PR #1, opened 2019-01-14
Adding license information

perpetualKid:
Adding project-wide license


twpol:
Thanks! To be honest I thought we had this in here somewhere already, but apparently not, well spotted. :)

PR #9, opened 2019-01-20
Blueprint https://blueprints.launchpad.net/or/+spec/station-level-crossing-trainspotter-camera Improved

Csantucci:
Pull request for improved version.

### DIFF
--- a/Source/RunActivity/Viewer3D/Cameras.cs
+++ b/Source/RunActivity/Viewer3D/Cameras.cs
@@ -1783,7 +1783,6 @@ namespace Orts.Viewer3D
         const float CameraAltitude = 2;
         // Height above the coordinate center of target.
         const float TargetAltitude = TerrainAltitudeMargin;
-        const float MaxHalfPlatformLength = 600;
         const int MaximumSpecialPointDistance = 300;
         const float PlatformOffsetM = 3.3f;
         public bool SpecialPoints = false;
@@ -1975,13 +1974,13 @@ namespace Orts.Viewer3D
                                 {
                                     PlatformDetails thisPlatform = train.signalRef.PlatformDetailsList[platformIndex];
                                     if (thisPlatform.TCOffset[0, thisRoute[routeIndex].Direction] + incrDistance < MaximumSpecialPointDistance * 0.7f
-                                        && thisPlatform.TCOffset[0, thisRoute[routeIndex].Direction] + incrDistance > 0)
+                                        && (thisPlatform.TCOffset[0, thisRoute[routeIndex].Direction] + incrDistance > 0 || FirstUpdateLoop))
                                     {
                                         // platform found, compute distance to viewing point
                                         distanceToViewingPoint = Math.Min(MaximumSpecialPointDistance * 0.7f,
-                                            incrDistance + thisPlatform.TCOffset[0, thisRoute[routeIndex].Direction] + thisPlatform.Length / 2.0f);
-                                        platformFound = true;
-                                        SpecialPointFound = true;
+                                            incrDistance + thisPlatform.TCOffset[0, thisRoute[routeIndex].Direction] + thisPlatform.Length * 0.7f);
+                                        if (FirstUpdateLoop && Math.Abs(train.SpeedMpS) <= 0.2f) distanceToViewingPoint = 
+                                                Math.Min(distanceToViewingPoint, train.Length * 0.95f);
                                         tdb.Move(distanceToViewingPoint);
                                         newLocation = tdb.WorldLocation;
                                         Traveller shortTrav;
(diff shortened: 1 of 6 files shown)


Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1778,6 +1783,13 @@ public class TracksideCamera : LookAtCamera
         const float CameraAltitude = 2;
         // Height above the coordinate center of target.
         const float TargetAltitude = TerrainAltitudeMargin;
+        const int MaximumSpecialPointDistance = 300;
+        const float PlatformOffsetM = 3.3f;
+        public bool SpecialPoints = false;
+        protected bool SpecialPointFound = false;
+        const float CheckIntervalM = 50f; // every 50 meters it is checked wheter there is a near special point
+        protected float DistanceRunM = 0f; // distance run since last check interval
+        protected bool FirstUpdateLoop = true; // first update loop
```
twpol: I don't think putting all of the special trackside camera fields into the base trackside camera class is a great way to organise this. There should be only tiny, or no, changes to `TracksideCamera` for this feature.

One possible way of doing that is to override `Update(elapsedTime)` in `SpecialTracksideCamera` to select the special locations if they're found, and only call `base.Update(elapsedTime)` if it can't find any special locations.

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1798,6 +1810,8 @@ public override bool IsUnderground
                     return true;
                 if (TrackCameraLocation == WorldLocation.None) return false;
                 var elevationAtCameraTarget = Viewer.Tiles.GetElevation(TrackCameraLocation);
+//                if (TrackCameraLocation.Location.Y + TerrainAltitudeMargin < elevationAtCameraTarget)
+//                    Trace.TraceInformation("CameraY {0}, elevationAtCamera {1}", TrackCameraLocation.Location.Y, elevationAtCameraTarget);
```
twpol: If you could remove all the commented out code, that'd be great. If you'd like to keep code like this for debugging in future, I'd recommend the `#if DEBUG_...` style that can be found in some other places.

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1885,48 +1899,202 @@ public override void Update(ElapsedTime elapsedTime)
...
-            if (!trainClose || (TrackCameraLocation == WorldLocation.None))
-            {
-                var tdb = trainForwards ? new Traveller(train.FrontTDBTraveller) : new Traveller(train.RearTDBTraveller, Traveller.TravellerDirection.Backward);
-                tdb.Move(MaximumDistance * 0.75f);
-                var newLocation = tdb.WorldLocation;
-                TrackCameraLocation = new WorldLocation(newLocation);
-                var directionForward = WorldLocation.GetDistance((trainForwards ? train.FirstCar : train.LastCar).WorldPosition.WorldLocation, newLocation);
-                if (Viewer.Random.Next(2) == 0)
+            if (!trainClose || (TrackCameraLocation == WorldLocation.None) || trySpecial)
+            {
```
twpol: The code in this block that locates the special locations is quite long and it isn't immediately clear why it is so complicated. Could you add some comments explaining what each small chunk is doing, e.g. what is `shortTrav` doing/for?

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1885,48 +1899,202 @@ public override void Update(ElapsedTime elapsedTime)
...
+                SpecialPointFound = false;
+                bool platformFound = false;
+                Traveller tdb;
+                if (FirstUpdateLoop && SpecialPoints) 
+                    tdb = trainForwards ? new Traveller(train.RearTDBTraveller) : new Traveller(train.FrontTDBTraveller, Traveller.TravellerDirection.Backward);
+                 else
+                    tdb = trainForwards ? new Traveller(train.FrontTDBTraveller) : new Traveller(train.RearTDBTraveller, Traveller.TravellerDirection.Backward);
+                WorldLocation newLocation = WorldLocation.None;
```
twpol: Where possible, it's preferable to use `var` for local variables.

Source/RunActivity/Viewer3D/Cameras.cs
Csantucci: Hi James, may I make a proposal? I have further added the roadcar based camera to this special points camera (see video mentioned in the ET forum), so I'd like to add it also within this pull request and then cope with your points. Else I have to make twice the adaptations accordingly to your requests. Can you agree?

twpol: That's absolutely fine!

Source/RunActivity/Viewer3D/Cameras.cs
Csantucci: Done.

Source/RunActivity/Viewer3D/Cameras.cs
Csantucci: Done. shortTrav is used to determine the direction of movement (left/right) of the camera from center of track to platform. I assume there is also another way to do this, but I couldn't find it.

Source/RunActivity/Viewer3D/Cameras.cs
Csantucci: Done

Source/RunActivity/Viewer3D/Cameras.cs
Csantucci: I have checked again if this is possible, but comparing the two versions of Cameras.cs I see that the changes are in many points, not only in the central part trying to find a special point. This was the reason why I did it that way (which is not the best one, I know). The other possibility is to have two separate Update(elapsedTime) methods, which is paid with the duplication of some code lines.

Csantucci: So I'm waiting for your comment about what solution you prefer.

twpol: Sorry, but we should separate this code. I don't think you need to duplicate much/anything, because the `SpecialTracksideCamera` can call code in the `TracksideCamera`. If you need some code in both classes even with the `base.Update(elapsedTime)` call, put it into a `protected` method in `TracksideCamera` and you can call it from both classes as needed.

Csantucci: I've generated an updated commit. I expect it is OK now for you.

PR #13, opened 2019-02-01
Improve cameras 2 3 at reversal points

Csantucci:
Bug fix for https://bugs.launchpad.net/or/+bug/1813790 Incoherent behaviour of cameras 2 and 3 at reversal points.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -414,6 +414,8 @@ namespace Orts.Simulation.Physics
         public double RunningTime;              // Total running time, used to check whether a locomotive is partly or totally unpowered due to a fault
         public int UnpoweredLoco = -1;          // car index of unpowered loco
 
+        public bool FormationReversed;          // flags the execution of the ReverseFormation method (executed at reversal points)
+
         public enum END_AUTHORITY
         {
             END_OF_TRACK,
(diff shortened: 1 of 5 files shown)


Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -414,6 +414,8 @@ public enum OUTOFCONTROL
         public double RunningTime;              // Total running time, used to check whether a locomotive is partly or totally unpowered due to a fault
         public int UnpoweredLoco = -1;          // car index of unpowered loco
 
+        public bool FormationReversed;          // flags the execution of the ReverseFormation method (executed at reversal points)
```
twpol:
```suggestion
        // TODO: Replace this with an event
        public bool FormationReversed;          // flags the execution of the ReverseFormation method (executed at reversal points)
```

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -1596,8 +1599,10 @@ public void TraincarsInitializeMoving()
         /// Update train 
         /// <\summary>
 
-        public virtual void Update(float elapsedClockSeconds)
+        public virtual void Update(float elapsedClockSeconds, bool auxiliaryUpdate = true)
         {
+            if (!auxiliaryUpdate)
+               FormationReversed = false;
```
twpol:
```suggestion
                FormationReversed = false;
```

twpol:
I am not sure if I've done the right thing in GitHub, but I want to say this is approved *with* those suggested changes. I don't know what GitHub will do when you accept them, and I don't seem to be able to force them in the web interface (although I could do them via Git if I had the time). If accepting them removes my approval I'll just re-add it.

PR #14, opened 2019-02-11
Removed confusing mentions of Demo Model 1

cjakeman:
As suggested by Peter Newall


cjakeman:
Also changed the banner "Now at Version 1.3" back to "Totally free software"

PR #18, opened 2019-02-17
Re-directed website Contact Us email

cjakeman:


cjakeman:
James, the Contact Us form has always been sending its action off-site because I failed to get PHP mail() working on the uktrainsim host.

It was sending its contents to a form handler on the host ccgi.jakeman.plus.com provided by PlusNet. PlusNet are doing their best to give up on this service and it's now died without warning, so I've transferred it to a project I'm paying for at wepp.co.uk. The host is One.com which has provided me with quite a reasonable service for a couple of years.

cjakeman:
James,
I was expecting GitHub to create 2 Pull Requests for me as these 2 commits are different topics (both website). The second change is in response to a suggestion about Blender received via the Contact Us page.

twpol:
> I was expecting GitHub to create 2 Pull Requests for me as these 2 commits are different topics (both website). The second change is in response to a suggestion about Blender received via the Contact Us page.

A PR is tied to the branch, so if you do more work on the same branch while the PR is open, you're updating the PR. Common practise is to start a new branch for each feature/bug that you want to PR, which will keep things separate in future.

PR #21, opened 2019-02-28
Timetable Update, including starting in paused state.

Roeterdink:
See :
bug report https://bugs.launchpad.net/or/+bug/1817878
blueprint https://blueprints.launchpad.net/or/+spec/timetable-trigger-function
thread http://www.elvastower.com/forums/index.php?/topic/32758-start-timetable-in-paused-mode

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -929,6 +929,9 @@ namespace Orts.Simulation.AIs
                         car.CarID = "AI" + train.Number.ToString() + " - " + (train.Cars.Count - 1).ToString();
                     }
 
+                    // associate location events
+                    Simulator.ActivityRun.AssociateEvents(train);
+
                 }
                 catch (Exception error)
                 {
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -537,6 +537,8 @@ namespace Orts.Simulation.AIs
                         Simulator.Signals.Update(true);
                         clockTime = runTime;
                         runTime += deltaTime;
+                        if (cancellation.IsCancellationRequested) // ping loader watchdog
+                            return;
 
                         int fullsec = Convert.ToInt32(runTime);
                         if (fullsec % 3600 == 0) Trace.Write(" " + (fullsec / 3600).ToString("00") + ":00 ");
@@ -571,6 +573,11 @@ namespace Orts.Simulation.AIs
                             PlayTrain = Simulator.Trains.GetAITrainByNumber(0) as TTTrain;
                             // train exists - set as player train
                             playerTrainStarted = (PlayTrain != null);
+
+                            if (playerTrain.LeadLocomotive == null)
+                            {
+                                playerTrainStarted = false;
+                            }
                         }
                         else
                         {
@@ -578,13 +585,9 @@ namespace Orts.Simulation.AIs
                         }
                     }
 
-                    if (playerTrain.LeadLocomotive == null)
-                    {
-                        playerTrainStarted = false;
-                    }
-
                     TimeSpan delayedStart = new TimeSpan((long)(Math.Pow(10, 7) * (clockTime - Simulator.ClockTime)));
                     Trace.TraceInformation("Start delayed by : {0}", delayedStart.ToString());
+                    StartList.RemovePlayerTrain();
                     TTTrain playerTTTrain = playerTrain as TTTrain;
                     playerTTTrain.InitalizePlayerTrain();
                     Simulator.ClockTime = runTime;
@@ -925,8 +928,7 @@ namespace Orts.Simulation.AIs
                         car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
                         car.CarID = "AI" + train.Number.ToString() + " - " + (train.Cars.Count - 1).ToString();
                     }
-                    // associate location events
-                    Simulator.ActivityRun.AssociateEvents(train);
+
                 }
                 catch (Exception error)
                 {
--- a/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
@@ -654,7 +654,7 @@ namespace Orts.Simulation.AIs
 
             if (MovementState == AI_MOVEMENT_STATE.AI_STATIC)
             {
-                physicsUpdate(elapsedClockSeconds);   //required to make train visible 
+                physicsUpdate(0);   //required to make train visible ; set elapsed time to zero to avoid actual movement
             }
             else
             {
@@ -1742,7 +1742,7 @@ namespace Orts.Simulation.AIs
                 else if (nextAspect == MstsSignalAspect.STOP)
                 {
                     // if stop but train is well away from signal allow to close; also if at end of path.
-                    if (distanceToSignal > 5 * signalApproachDistanceM ||
+                    if (DistanceToSignal.HasValue && DistanceToSignal.Value > 5 * signalApproachDistanceM ||
                         (TCRoute.TCRouteSubpaths[TCRoute.activeSubpath].Count - 1 == PresentPosition[0].RouteListIndex))
                     {
                         MovementState = AI_MOVEMENT_STATE.ACCELERATING;
@@ -1836,7 +1836,8 @@ namespace Orts.Simulation.AIs
                     }
                 }
             }
-            if (AuxActionnextActionInfo != null && MovementState == AI_MOVEMENT_STATE.STOPPED && tryBraking && distanceToSignal > clearingDistanceM
+            float distanceToNextSignal = DistanceToSignal.HasValue ? DistanceToSignal.Value : 0.1f;
+            if (AuxActionnextActionInfo != null && MovementState == AI_MOVEMENT_STATE.STOPPED && tryBraking && distanceToNextSignal > clearingDistanceM
                 && EndAuthorityType[0] != END_AUTHORITY.RESERVED_SWITCH && DistanceToEndNodeAuthorityM[0] <= 2.0f * junctionOverlapM)   // && ControlMode == TRAIN_CONTROL.AUTO_NODE)
             {
                 MovementState = AI_MOVEMENT_STATE.BRAKING;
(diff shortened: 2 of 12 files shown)


twpol:
There are an awful lot of changes in here, and you reference a bug and blueprint. Is this a single bug fix or new feature, or is it some mixture/multitude of things? With Git, there's no technical reason to mix things up any more, so I hope we can get away from these messy updates.

Roeterdink:
This update includes all changes I made in my 'private' version over the last two years or so. For various reasons I never got to commit these to the official version. I did not use GIT yet for my own version. I have now made an effort to bring the official version back in line with my private version, but it is impossible to separate these changes as they all affect each other, and there also changes on changes.
It is a once-off bulk update to get all back in line again, I am now using GIT with separate branches for separate issues.

twpol:
Okay, I'll let this slide but please:

1. Update this PR's description will a complete list of bugs fixed and features added with *links*
2. Keep each bug fix and new feature on a separate branch going forward - please ask if you have any problems with this, as I'm not likely to accept a PR like this one again

Roeterdink:
- bug report has been extended with furhter information
   I'm not aware of any other bug reports which are solved through this update
- set link to bug report in description

twpol:
@Roeterdink Thanks, I'll do a review of the changes now.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
```diff
@@ -6774,7 +7057,8 @@ public void ProcessTimetableStopCommands(TTTrainCommands thisCommand, int subrou
                 case "wait":
                     if (thisCommand.CommandValues == null)
                     {
-                        Trace.TraceError("Train : {0} : invalid wait command at {1}", Name, thisStationStop == null ? "no station" : thisStationStop.PlatformItem.Name);
+                        Trace.TraceInformation("Train : {0} : invalid wait command at {1}", Name, thisStationStop == null ? "note line" : thisStationStop.PlatformItem.Name);
```
twpol: This sounds like it should be a warning - something that is wrong in the data, but which we handled (possibly by ignoring it).

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
```diff
@@ -7367,6 +7651,39 @@ public void ProcessTimetableStopCommands(TTTrainCommands thisCommand, int subrou
...
+
+                    if (plattformReferenceID >= 0)
+                    {
+                        thisTrigger.platformId = plattformReferenceID;
+                        thisTrigger.activationType = TriggerActivationType.StationStop;
+
+                        if (thisCommand.CommandQualifiers != null && thisCommand.CommandQualifiers.Count > 0)
+                        {
+                            TTTrainCommands.TTTrainComQualifiers thisQualifier = thisCommand.CommandQualifiers[0]; // takes only 1 qualifier
+                            if (String.Equals(thisQualifier.QualifierName.Trim().ToLower(), "depart"))
```
twpol:
```suggestion
                            if (thisQualifier.QualifierName.Trim().ToLower() == "depart")
```

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
```diff
@@ -9566,7 +10014,8 @@ public override void CheckStationTask()
                             PresentPosition[1].Reverse(ValidRoute[0][PresentPosition[1].RouteListIndex].Direction, tempRoute, 0.0f, signalRef);
 
                             // reverse formation
-                            ReverseFormation(false);
+//                            ReverseFormation(false);
```
twpol: Please remove commented out code, especially when it's been replaced by something new.

Source/RunActivity/Viewer3D/Processes/GameStateViewer3D.cs
```diff
@@ -30,10 +30,11 @@ public class GameStateViewer3D : GameState
         bool FirstFrame = true;
         int ProfileFrames = 0;
 
-        public GameStateViewer3D(Viewer viewer)
+        public GameStateViewer3D(Viewer viewer, bool paused)
         {
             Viewer = viewer;
-            Viewer.Simulator.Paused = false;
+            Viewer.Simulator.Paused = paused;
```
twpol: As far as I can tell, all the places that call this do so with `paused = true`, so I'd prefer just coding that in here instead of passing a `bool` in.

Source/Orts.Simulation/Simulation/AIs/AI.cs
```diff
@@ -537,6 +537,8 @@ private void PrerunAI(int playerTrainOriginalTrain, TTTrain.FormCommand playerTr
                         Simulator.Signals.Update(true);
                         clockTime = runTime;
                         runTime += deltaTime;
+                        if (cancellation.IsCancellationRequested) // ping loader watchdog
```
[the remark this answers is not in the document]
Roeterdink: The first part is a for loop, as it is known when it should terminate (at the start time of the player train).
The second part (this loop) however is 'open ended'. The train can not start, it is delayed, but for how much is not clear. Setting up a 'for' loop would be misleading, as that creates the impression it is to run upto a certain point - but it is not. The check on midnight is an 'emergency exit' and terminates the program immediately.
Actually this is not quite correct for a train may be started after midnight and if this train is running late, the program would abort and not allow it to start.
This is on my list to review but I have not yet puzzled out what the proper 'emergency exit' criterium should be.

Source/Orts.Simulation/Simulation/AIs/AITrain.cs
```diff
@@ -3706,14 +3707,20 @@ public void AdjustControlsAccelMore(float reqAccelMpSS, float timeS, int stepSiz
...
                 {
                     //TODO: next code line has been modified to flip trainset physics in order to get viewing direction coincident with loco direction when using rear cab.
                     // To achieve the same result with other means, without flipping trainset physics, the line should be changed as follows:
                     //  car.SpeedMpS = car.Flipped ? -SpeedMpS : SpeedMpS;
                     car.SpeedMpS = car.Flipped ^ (car.IsDriveable && car.Train.IsActualPlayerTrain && ((MSTSLocomotive)car).UsingRearCab) ? -SpeedMpS : SpeedMpS;
                 }
+
+                if (CheckTrain)
+                {
+                    File.AppendAllText(@"C:\temp\checktrain.txt", "Forced speed increase : was " + LastSpeedMpS + " - now " + SpeedMpS + "\n");
```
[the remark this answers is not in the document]
Roeterdink: Agree - but at present there is no alternative.
There are only two output paths available at present - the location where 'saves' are stored, and the desktop.
In my view, neither are suitable for these test outputs.
These outputs don't belong in the 'saves' directory as one should not have to remove files manually from this location. And I certainly don't want these files on my desktop (I never place any files on my desktop at all, only a few links).
What is required is a user option (through menu or through register) defining a path for test outputs. If that is provided, I would be happy to change all related code.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
Roeterdink: All items found when processing timetables are consistently set as information.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
Roeterdink: Oh dear - habbit of a lifetime. I've worked for over 25 years with languages where one could not just compare strings. You found this one but there are many more like this, and I fear there will be more to come because it is just something I will do automatically.

Source/RunActivity/Viewer3D/Processes/GameStateViewer3D.cs
Roeterdink: This change was made before it was agreed to apply to all start/resume moments. So at first it was only appled for timetable mode and therefor required it to be passed in.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
Roeterdink: Done

Roeterdink: Well, I would have if the system wasn't hiding the 'commit' button?

Roeterdink:
I hope I did everything right. It does seem that the issue with the commented line has not been properly resolved. Somehow it did not show a 'commit' button for this change.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
twpol: That's okay, there's plenty of similar types of not-really-C# coding all through the code that we'll be trying to clean up, when and if we have time. 🙂

Source/Orts.Simulation/Simulation/AIs/AI.cs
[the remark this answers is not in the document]
twpol: I disagree it would be misleading; instead, it would clearly indicate this loop terminates. The problem with non-terminating loops like `while` is that now you've added a loader ping, if at any point in the future the code inside the loop changes and it - intentionally or, more likely, accidentally - becomes possible for it to not terminate, you've broken the ability for Open Rails to detect that and report a bug.

This one I'll let go but please remember this for future code.

Source/Orts.Simulation/Simulation/Timetables/TTTrain.cs
twpol: I guess that's something to review later.

twpol:
@Roeterdink The "Commit" button only appears when I have specifically given a suggest change; for other things that need changing, you need to go to this branch on your local repository, pull the changes made by the "Commit" button, make the other changes, commit, and push.

Let me know if you're not entirely clear about this, but this is useful for improving pull requests even without having a review, so it's good to know how to do it.

Roeterdink:
I've found a problem which is likely caused by update overlaps.
Comparing with the working version shows that a line is deleted in AI.cs.
But this line is not in my working version. It looks like this was added either while I was preparing the pull or after I made the pull.
Should I change my pull request to overcome this problem?

Is there anything else you're still waiting for?

Roeterdink:
Checked on the update from which the change to AI was missing, but all other changes are correct.
Seems it was not an update overlap but a genuine error while changing AI.
Have re-pushed with required changes (removed commented lines in TTTrain and removed bool parameter from Gatestateviewer).

PR #23, opened 2019-03-03
https://blueprints.launchpad.net/or/+spec/water-trough-filling - base…

peternewell:
… deployment of water trough dynamic filling

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
@@ -263,12 +263,24 @@ namespace Orts.Simulation.RollingStocks
         SmoothedData FuelRate = new SmoothedData(45); // Automatic fireman takes x seconds to fully react to changing needs.
         SmoothedData BurnRateSmoothKGpS = new SmoothedData(150); // Changes in BurnRate take x seconds to fully react to changing needs - models increase and decrease in heat.
         float FuelRateSmoothed = 0.0f;     // Smoothed Fuel Rate
+
+        // Water trough filling
         public bool HasWaterScoop = false; // indicates whether loco + tender have a water scoop or not
-        public float ScoopMinPickupSpeedMpS = 0.0f; // Minimum scoop pickup speed
-        public float ScoopMaxPickupSpeedMpS = 200.0f; // Maximum scoop pickup speed
-        public float ScoopResistanceN = 0.0f; // Scoop resistance
+//        public float ScoopMinPickupSpeedMpS = 0.0f; // Minimum scoop pickup speed
+        public float ScoopMaxPickupSpeedMpS = 200.0f; // Maximum scoop pickup speed - used in steam locomotive viewer
+//        public float ScoopResistanceN = 0.0f; // Scoop resistance
         public bool ScoopIsBroken = false; // becomes broken if activated where there is no trough
         public bool RefillingFromTrough = false; // refilling from through is ongoing
+        public float WaterScoopFillElevationM; // height water has to be raised to fill tender
+        public float WaterScoopDepthM; // depth that water scoop goes into trough (pan)
+        public float WaterScoopWidthM; // width of water scoop
+        public float WaterScoopVelocityMpS; // Velocity of water entering water scoop
+        public float WaterScoopDragForceN; // drag force due to scoop being in water trough
+        public float WaterScoopedQuantityGalukpS; // Amount of water scooped up by water scoop per second
+        public float WaterScoopInputAmountGals; // Water scooped in elapsed time
+        public float WaterScoopMinSpeedMpS; // Minimum speed for water pickup
+        public bool IsWaterScoopDown = false;
+
 
         // steam performance reporting
         public float SteamPerformanceTimeS = 0.0f; // Records the time since starting movement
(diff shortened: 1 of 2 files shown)


peternewell:
Is this pull request now valid?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -904,9 +919,12 @@ public override void Copy(MSTSWagon copy)
             IsFixGeared = locoCopy.IsFixGeared;
             IsSelectGeared = locoCopy.IsSelectGeared;
             HasWaterScoop = locoCopy.HasWaterScoop;
-            ScoopMinPickupSpeedMpS = locoCopy.ScoopMinPickupSpeedMpS;
-            ScoopMaxPickupSpeedMpS = locoCopy.ScoopMaxPickupSpeedMpS;
-            ScoopResistanceN = locoCopy.ScoopResistanceN;
+//            ScoopMinPickupSpeedMpS = locoCopy.ScoopMinPickupSpeedMpS;
+//            ScoopMaxPickupSpeedMpS = locoCopy.ScoopMaxPickupSpeedMpS;
+//            ScoopResistanceN = locoCopy.ScoopResistanceN;
```
twpol: There's a lot of commented-out code in this PR. It most cases, it should be removed, as we can always find it again with Git, and there's nothing useful to gain from the code as comments.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -1707,6 +1725,25 @@ public override void Initialize()
...
+            
+            if (WaterScoopDepthM == 0)
+            {
+               WaterScoopDepthM = 0.0889f; // Set to default of 3.5 ins
+            }
+            
+            if (WaterScoopWidthM == 0)
+            {
+               WaterScoopWidthM = 0.3048f; // Set to default of 1 ft
+            }
```
twpol: Can these defaults not be done the same way as everything else, in the field initialiser itself? Why must we check for 0 and set it instead of just starting with the default?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -1707,6 +1725,25 @@ public override void Initialize()
...
+
+            // Calculate minimum speed to pickup water
+               WaterScoopMinSpeedMpS = Me.FromFt((float)Math.Sqrt(2 * 32.2 * Me.ToFt(WaterScoopFillElevationM)));
```
twpol: This maths expression is inscrutable; please either use constants for the numbers (so I and others know what they mean/represent) or at least wrap the calculation inside a function with a clear name.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -2349,6 +2386,49 @@ private void UpdateTender(float elapsedClockSeconds)
...
+               {
+                  WaterScoopVelocityMpS = Me.FromFt((float)Math.Sqrt(Avalue - Bvalue));
+               }
+               else
+               {
+                  WaterScoopVelocityMpS = 0;
+               }
+               
+               // calculate volume of water scooped per period
+               float CuFttoGalUK = 6.22884f; // imperial gallons of water in a cubic foot of water
```
twpol: We should be using metric units everywhere internally to avoid confusion.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: This is one of a number of changes that will be made to implement additional functionality with filling from water troughs.

Previously my preference has been to leave some code that I am not 100% certain about until it has been through a test cycle in the community. Typically when a feature is completed, all commented code will be removed.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I have done it this way for parameters that can be overwritten by users through the ENG or WAG files.

In some instances (not currently in this case) an information message is provided in the log to confirm that the user has successfully overwritten the value.

Given that the code is only executed once when OR is initialised, what issues does it create?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I don't quite understand the issue with this statement, can you expand on the issue for me.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I agree ideally.

However given the age of steam locomotives, they are typically engineered in imperial units. Hence a lot of the relevant formula are in imperial units, and I have not found many metric formulas. I have chosen to keep formulas in their original forms so that they can be easily verified against source reference material. Also saves me from introducing errors by trying to convert an imperial formula to its metric equivalent.

Wherever possible I have taken the outputs of these formula and converted them to metric units for usage elsewhere in OR.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
twpol: We should be able to the AppVeyor automatic builds to test features while they are in PR ("Show all checks" at the bottom of the PR, then "Details", then go to "Artifacts"). Please try this - you should be able to link directly to the zip from a forum post - and let me know how it goes. This is my preferred approach going forward.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
twpol: The main issue is pulling the default values away from where they are for many other things, where people would typically expect to find the defaults, and simply unnecessary code. For example, if the field `WaterScoopWidthM` was initialised with `0.3048f` where it was defined, these four lines would be removed.

I don't understand the point of an informational message to tell the user they have set something, so I'm glad there's none in here.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
twpol: The numeric constants `2` and `32.2` mean nothing to me or anyone else looking at this code. If you put them into `const` fields, you can give them readable names *and* units - both of which are highly useful when reading the code. It is probably also worth commenting what the maths is here - why are we doing a square-root on a * b * c? What real-world thing is this?

I am separately sad that this converts metres into feet, does the maths, and then puts it back into metres - that probably helps nobody in this particular case, but definitely will make the calculation less accurate (every conversion makes things less accurate because computers can't represent floating point numbers exactly).

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
twpol: As long as we're keeping the use of non-metric units and conversions to a minimum, that's okay. Every conversion introduces a small error to the calculations, so we need to avoid repeatedly converting the same value, as well.

PR #25, opened 2019-03-05
Added comments to justify calculation of train speed

cjakeman:


cjakeman:
Branch title is correct-train-speed but found it wasn't wrong - the code just looked to be poor engineering obscured by clumsy code. Turned out to be sophisticated engineering (obscured by clumsy code).

twpol (approved):
I removed the VS files and extra logging, but otherwise this is good - thanks!

PR #26, opened 2019-03-06
Updates to OR manual for special viewpoint camera

Csantucci:


Source/Documentation/Manual/conf.py
```diff
@@ -54,15 +54,15 @@
...
+copyright = '2019, Open Rails'
 author = 'Open Rails Team'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = 'x.4329'
+version = 'X1.3.1-30'
```
twpol: I think we need to find a better solution to this, but don't worry about it in this PR. :)

PR #30, opened 2019-03-18
Upgrading GetText/GetOpt code to VS2017 .NET 4.7.2 and AnyCPU platform target

perpetualKid:
- Updating the GetText codebase to  target latest framework version, and AnyCPU configuration for x86 and x64 systems.
- Adding appropriate NUnit nuget packages to make the solution buildable
- Changes are primarily in auto-generated code and configs


twpol:
Do you know that we don't build the 3rd party libraries from source during the build/automatically? The source code in Source\3rdPartyLibs should align with the binaries in the same folder, so if we're updating the source I'd expect new binaries as well.

perpetualKid:
This PR is on purpose, to keep 3rd party code (changes) separate from OR, instead of having one large PR. 
While the binaries are a dependency to OR code, the GetText code is not a dependency... hence the new binaries will be checked in when OR is upgraded to NetFX 4.7.2, else the build of OR would fail because of version mismatch currently.

twpol:
It'll be a little confusing (if anyone notices ;) ) in the short-term but that's okay.

PR #31, opened 2019-03-18
  Update OR Launcher to check for NetFX 4.7.2 dependency

perpetualKid:
Preparing for NetFX upgrade


twpol:
I presume the intent here is to start getting people to upgrade .NET now, before the rest of the code depends on it?

perpetualKid:
correct, that way folks can ensure compatibility beforehand. 
Also I intent to keep the discussed upgrade(s) divided into multiple smaller chunks to allow for easier review and integration.

PR #33, opened 2019-03-19
Upgrade all projects to target NetFX 4.7.2

perpetualKid:
Includes updated binary dependencies (GetText/GetOpt), and code changes necessary to compile and run the programs.
Changes included are only addressing the .Net FX version upgrade, beside of following list of updates, all other changes are generated code.

Following changes where made for legacy compatibility reason and will be resolved over time:
- explicitly naming ORTS.Common.CancellationToken(Source) to avoid ambiguity with CancellationToken(Source) from System.Threading namespace
- adding useLegacyV2RuntimeActivationPolicy="true" to all app.config in configuration->startup node to allow mixedMode runtime as long there are still dependencies on .Net 2.0 (primarily XNA, will be resolved once we move to MG)
- adding Switch DoNotLoadLatestRichEditControl="true" in configuration->runtime->AppContextSwitchOverrides in ContentManager project to enforce RTF compatibility which has changed in .NET >= 4.6.2. The most appropriate documentation about is on SO https://stackoverflow.com/questions/48371211/richtextbox-removes-escape-character-from-the-text 

It is important to understand that this PR starts breaking compatibly with older OS which do not support .NET 4.7.2.
 
Approved Blueprint here https://blueprints.launchpad.net/or/+spec/dotnet-4


perpetualKid:
@twpol not sure what appveyor is complaining about

Csantucci:
Hi Steffen, this PR is too IT-oriented for my knowledge, so I'm not able to perform a serious review. You should select another reviewer.

perpetualKid:
> All looks good, sorry it took so long! Once people have had a day or two to check the unstable version, I'm happy to do a special edition of the testing version to get it to a wider audience before the standard  time on Friday.

no rush, give it the week, Friday should be fine

Csantucci:
I see that release 1.3.1-55, that should come out of this merge, is not yet seen by the updater and its .zip does not appear in James' site. I on the other side am not able to build with success starting from this merge. Is everything OK?

twpol:
(I don't know where Carlo's message has gone but I got an email about it from this PR...)

The automated build for unstable version failed because of the tests:

```
xUnit.net console test runner (32-bit .NET 2.0.50727.8838)
Copyright (C) 2013 Outercurve Foundation.

Could not load file or assembly 'E:\Documents\Coding\Jenkins\jobs\Open Rails Unstable (git)\workspace\Program\Tests.dll' or one of its dependencies. This assembly is built by a runtime newer than the currently loaded runtime and cannot be loaded.
```

I believe this is something for me to fix and not a problem with this PR - but this does highlight a known limitation of the AppVeyor builds - they don't run the tests/care about the output. IIRC this is because we have a bunch of failing tests and AppVeyor just fails everything for >0 failed tests. 😕

perpetualKid:
Interesting AppVeyor didn’t complain on the same when the PR was created or updated.

Carlo mentioned issues with the current Head, but I can’t see anything odd when pulling a clean version.

Csantucci:
Hi,
I wrote the message when I saw that release 55 didn't automatically build, and at the same time I had problems building it locally. By trying further some times something changed (don't know what) and I could build locally. So i deleted the message, because I supposed that the automatical build failure wasn't a problem of the PR itself (as in effect James is suspecting).

perpetualKid:
@James

 Are you looking into this? Seems also with 1.3.1-56 this is not working, even strangely AppVeyor reports all fine when building on PR

twpol:
@perpetualKid Yes, I'm just about to start looking now. The reason AppVeyor (AV) didn't notice is because it does not check for test results but my build does - although the real reason the AV build passed incorrectly is that the `Build.cmd` code is not stopping on the error running xunit. Annoyingly it is impossible to fix this in a PR (because it'll cause AV to fail) without also fixing the problem with the test runner. ☹

PR #34, opened 2019-03-20
Upgrade all projects to C# 7.3

perpetualKid:
This is the discussed upgrade of all projects in the solution to allow C# 7.3. 
No other changes are include.

It should be noted that this PR will break compatibility with VS 2015 (and before) as C# 7+ is only support in VS2017 and newer.

### DIFF
--- a/Source/ContentChecker/ContentChecker.csproj
+++ b/Source/ContentChecker/ContentChecker.csproj
@@ -31,7 +31,7 @@
     <PlatformTarget>x86</PlatformTarget>
     <ErrorReport>prompt</ErrorReport>
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
-    <LangVersion>6</LangVersion>
+    <LangVersion>7.3</LangVersion>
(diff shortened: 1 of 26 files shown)


perpetualKid:
Build engine (Appveyor) seems to have an issue with language versions beyond 6:

 CSC : error CS1617: Invalid option '7.3' for /langversion; must be ISO-1, ISO-2, Default or an integer in range 1 to 6.

perpetualKid:
build engine (AppVeyor) seems to have issues with C# language versions beyond 6:

 CSC : error CS1617: Invalid option '7.3' for /langversion; must be ISO-1, ISO-2, Default or an integer in range 1 to 6.

twpol:
> @twpol  Build engine (Appveyor) seems to have an issue with language versions beyond 6:
> CSC : error CS1617: Invalid option '7.3' for /langversion; must be ISO-1, ISO-2, Default or an integer in range 1 to 6.

Ah, AppVeyor defaults to Visual Studio 2015 still. I've updated the configuration to 2017 and it's passed now.

PR #37, opened 2019-03-23
Updating Copyright information

perpetualKid:


Source/Contrib/ActivityEditor/ActivityEditor/Properties/AssemblyInfo.cs
```diff
@@ -9,8 +9,8 @@
 [assembly: AssemblyDescription("Open Rails Transport Simulator")]
 [assembly: AssemblyConfiguration("")]
 [assembly: AssemblyCompany("Open Rails")]
-[assembly: AssemblyProduct("Open Rails (disabled)")]
```
twpol: We actually should leave this alone, as this is how this project is hidden from the main menu's dropdown (which is due to it not working properly IIRC).

twpol (changes requested):
Just put the product name back for the activity editor and we're good here 🙂

Source/Contrib/ActivityEditor/ActivityEditor/Properties/AssemblyInfo.cs
perpetualKid: not the most obvious way...

twpol: You are not wrong there! It actually works by comparing the executable's product name with that of the main menu, and if it is different it is excluded. This was originally to make sure oddities like Visual Studio's "foo.vshost.exe" didn't appear but was co-opted here. Obviously the real solution would be to fix this contributed app or remove it. 🙂

twpol (approved):
Thanks for all the housekeeping you're doing! 🎉

PR #39, opened 2019-03-31
Checking for new dependencies required for upgrade and MG

perpetualKid:
- Checking for NetFX 4.7.2, DirectX 9c (43) and OpenAL, and provide download links if necessary
- Remove OAL binary distribution


Source/Launcher/Program.cs
```diff
@@ -37,59 +36,116 @@ namespace ORTS
...
 	static class Program
 	{
-		/// <summary>
-		/// The main entry point for the application.
-		/// </summary>
-		[STAThread]
-		static void Main()
+        static string downloadUrl;
+        static string downloadText;
+        static List<string>missing = new List<string>();
```
twpol: This seems like an odd way to approach this. It would be clearer to have a struct with three fields (name, downloadText, downloadUrl) and make the list contain those.

Source/Launcher/Program.cs
```diff
@@ -37,59 +36,116 @@ namespace ORTS
...
-			var missing = new List<string>();
-			CheckNetFx(missing);
-			CheckXNA(missing);
-			CheckOR(missing, path);
-			if (missing.Count > 0)
+
+            CheckNetFx();
+			CheckXNA();
+            CheckDXRuntime();
+            CheckOpenAL();
```
twpol: Likewise, not passing in the `missing` list here is strange to me. Functions with side effects like this are harder to understand.

Source/Launcher/Program.cs
```diff
@@ -37,59 +36,116 @@ namespace ORTS
...
-				return;
-			}
+				if (MessageBox.Show(Application.ProductName + " requires the following:\n\n" + 
+                    string.Join("\n", missing.ToArray()) + 
+                    "\n\nWhen you click OK, we will guide you to download the required software.\n" +
+                    (missing.Count > 1 ? "If there are multiple items missing, you need to repeat this process until all dependencies are resolved.\n" : string.Empty) +
+                    "Click Cancel to quit.", 
+                    Application.ProductName, MessageBoxButtons.OKCancel) == DialogResult.OK && !string.IsNullOrEmpty(downloadUrl))
+                {
+                    DownloadDependency(missing[missing.Count - 1]);
```
twpol: Here we should be starting with the first dependency IMO.

Source/Launcher/Program.cs
perpetualKid: I don't agree, from semantics passing as variables within a static class does come at a cost, and if there would be any side effects, something else is inherently broken. 
But given the small code size of the artifact and for code clarity, I've updated the code.

twpol: The side effect here is the method modifying the static field (it's not mentioned on these lines so why would you expect it to change or even know it exists?), which is unnecessary and confusing when reading the code. Thank you for changing it!

PR #40, opened 2019-03-31
Monogame upgrade

perpetualKid:
Initial Monogame check-in, based on the experimental ORMG version from @Csantucci, with project cleanup to MG 3.7.1 binaries, and PR #33 and #34. This is a fully working version.

Blueprint https://blueprints.launchpad.net/or/+spec/monogame


nfsprodriver:
Here's a Build with latest upstream + Monogame upgrade and Net 4.8 working pretty good for me https://github.com/nfsprodriver/openrails/releases/download/1.3.1/Program.zip

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
```diff
@@ -97,10 +99,19 @@ internal RenderProcess(Game game)
...
             GraphicsDeviceManager.PreferredBackBufferFormat = SurfaceFormat.Color;
-            GraphicsDeviceManager.PreferredDepthStencilFormat = DepthFormat.Depth32;
+            GraphicsDeviceManager.PreferredDepthStencilFormat = DepthFormat.Depth24Stencil8;
             GraphicsDeviceManager.IsFullScreen = false;
             GraphicsDeviceManager.PreferMultiSampling = true;
             GraphicsDeviceManager.PreparingDeviceSettings += new EventHandler<PreparingDeviceSettingsEventArgs>(GDM_PreparingDeviceSettings);
+            var screen = Game.Settings.FastFullScreenAltTab ? Screen.FromControl(GameForm) : Screen.PrimaryScreen;
+            if (screen.Primary)
+            {
+                var wa = Screen.PrimaryScreen.WorkingArea;
```
twpol: Why are we only checking the working area on the primary screen? E.g. Windows 10 has taskbars on every monitor so this seems incorrect.

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
```diff
@@ -133,7 +146,12 @@ internal void Start()
             DisplaySize = new Point(GraphicsDevice.Viewport.Width, GraphicsDevice.Viewport.Height);
 
             if (Game.Settings.ShaderModel == 0)
-                Game.Settings.ShaderModel = GraphicsDevice.GraphicsDeviceCapabilities.PixelShaderVersion.Major;
+            {
+                if (GraphicsDevice.GraphicsProfile == GraphicsProfile.HiDef)
+                    Game.Settings.ShaderModel = 3;
+                else if (GraphicsDevice.GraphicsProfile == GraphicsProfile.Reach)
+                    Game.Settings.ShaderModel = 2;
```
twpol: I'm confused by this; the shader model is set to 4.0 in the effects files so this is not right.

Source/RunActivity/Viewer3D/RenderFrame.cs
```diff
@@ -888,64 +890,64 @@ void DrawSequencesDistantMountains(GraphicsDevice graphicsDevice, bool logging)
 #if DEBUG_RENDER_STATE
         static void DebugRenderState(RenderState renderState, string location)
         {
-            if (renderState.AlphaBlendEnable != false) throw new InvalidOperationException(String.Format("RenderState.AlphaBlendEnable is {0}; expected {1} in {2}.", renderState.AlphaBlendEnable, false, location));
-            if (renderState.AlphaBlendOperation != BlendFunction.Add) throw new InvalidOperationException(String.Format("RenderState.AlphaBlendOperation is {0}; expected {1} in {2}.", renderState.AlphaBlendOperation, BlendFunction.Add, location));
+//            if (renderState.AlphaBlendEnable != false) throw new InvalidOperationException(String.Format("RenderState.AlphaBlendEnable is {0}; expected {1} in {2}.", renderState.AlphaBlendEnable, false, location));
```
twpol: Kill off all these commented out lines please.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -1008,11 +1008,13 @@ public static void Mark(Viewer viewer)
 
     public class CabRenderer : RenderPrimitive
     {
-        private SpriteBatchMaterial _Sprite2DCabView;
+        //private SpriteBatchMaterial _Sprite2DCabView;
```
twpol: Remove if it's not needed.

Source/RunActivity/Viewer3D/Shaders.cs
```diff
@@ -48,18 +61,57 @@ static byte[] GetEffectCode(string filename)
...
-                return compiledEffect.GetEffectCode();
-            }
+            //using (var effectFileStream = File.OpenRead(effectFileName))
+            //{
+            //    // NOTE: We may need to implement a CompilerIncludeHandler here if we ever use #include in our shaders.
+            //    var compiledEffect = Effect.CompileEffectFromFile(effectFileStream, null, null, CompilerOptions.None, TargetPlatform.Windows);
+            //    if (!compiledEffect.Success)
+            //        throw new InvalidOperationException(compiledEffect.ErrorsAndWarnings);
+            //    return compiledEffect.GetEffectCode();
+            //}
```
twpol: Remove it?

Source/RunActivity/Viewer3D/Shaders.cs
```diff
@@ -255,7 +318,7 @@ public void SetBlurData(ref Matrix wvp)
 
         public void SetBlurData(Texture2D texture)
         {
-            imageTexture.SetValue(texture);
+            blurTexture.SetValue(texture);
```
twpol: Why do these need to be split out?

Source/RunActivity/Viewer3D/Shapes.cs
```diff
@@ -1002,7 +1002,7 @@ public override void PrepareFrame(RenderFrame frame, ElapsedTime elapsedTime)
     public class RoadCarShape : AnimatedShape
     {
         public RoadCarShape(Viewer viewer, string path)
-            : base(viewer, path, new WorldPosition(), ShapeFlags.ShadowCaster)
+            : base(viewer, path, new WorldPosition())
```
twpol: This is wrong; the default shape flags is `None`.

Source/RunActivity/Viewer3D/UserInput.cs
```diff
@@ -49,7 +49,7 @@ public static class UserInput
 
         public static RailDriverState RDState;
 
-        static InputSettings InputSettings;
+        public static InputSettings InputSettings;
```
twpol: Nope.

Source/RunActivity/Viewer3D/UserInput.cs
```diff
@@ -63,7 +63,9 @@ public static void Update(Game game)
             // Make sure we have an "idle" (everything released) keyboard and mouse state if the window isn't active.
             KeyboardState = game.IsActive ? new KeyboardState(GetKeysWithPrintScreenFix(Keyboard.GetState())) : new KeyboardState();
             MouseState = game.IsActive ? Mouse.GetState() : new MouseState(0, 0, LastMouseState.ScrollWheelValue, ButtonState.Released, ButtonState.Released, ButtonState.Released, ButtonState.Released, ButtonState.Released);
-            MouseButtonsSwapped = System.Windows.Forms.SystemInformation.MouseButtonsSwapped;
+            // Monogame returns swapped button status anyway.
```
twpol: So let's remove the `MouseButtonsSwapped` field entirely.

Source/Menu/Options.Designer.cs
```diff
@@ -61,6 +61,7 @@ private void InitializeComponent()
             this.numericSoundDetailLevel = new System.Windows.Forms.NumericUpDown();
             this.checkMSTSBINSound = new System.Windows.Forms.CheckBox();
             this.tabPageVideo = new System.Windows.Forms.TabPage();
+            this.checkEnableMultisampling = new System.Windows.Forms.CheckBox();
```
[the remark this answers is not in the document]
Csantucci: The problem is that with multisampling the MG version generally is quite slower than the XNA version.

Source/Contrib/TrackViewer/TrackViewer.cs
```diff
@@ -156,8 +157,8 @@ public TrackViewer(string[] args)
             SetAliasing();
             graphics.IsFullScreen = false;
             Window.AllowUserResizing = true;
-            Window.ClientSizeChanged += new System.EventHandler(Window_ClientSizeChanged);
-
+            Window.ClientSizeChanged += new System.EventHandler<EventArgs>(Window_ClientSizeChanged);
```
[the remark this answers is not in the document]
perpetualKid: the event signature changed to in XNA4/MG
public event EventHandler<EventArgs> ClientSizeChanged

Source/ORTS.Common/SystemInfo.cs
```diff
@@ -200,7 +200,7 @@ static string WriteInstalledRuntimes(TextWriter output, string versionKeyName, R
 
         static void WriteGraphicsAdapter(TextWriter output)
         {
-            foreach (var adapter in GraphicsAdapter.Adapters)
+ /*           foreach (var adapter in GraphicsAdapter.Adapters)
```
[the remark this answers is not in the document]
perpetualKid: removed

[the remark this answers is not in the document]
perpetualKid: Removed

Source/ORTS.sln
```diff
@@ -55,115 +55,184 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "ContentChecker", "ContentCh
 EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
 		Debug|Mixed Platforms = Debug|Mixed Platforms
+		Release|Any CPU = Release|Any CPU
```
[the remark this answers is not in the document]
perpetualKid: this is generated code/config, and will be changed/cleaned up in subsequent PR anyway

Source/Orts.Formats.Msts/LightCollection.cs
```diff
@@ -52,6 +52,14 @@ public LightState(STFReader stf)
                 new STFReader.TokenProcessor("transition", ()=>{ Transition = 1 <= stf.ReadFloatBlock(STFReader.UNITS.None, 0); }),
                 new STFReader.TokenProcessor("angle", ()=>{ Angle = stf.ReadFloatBlock(STFReader.UNITS.None, null); }),
             });
+            // Color byte order changed in XNA 4 from BGRA to RGBA
```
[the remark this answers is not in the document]
perpetualKid: http://www.shawnhargreaves.com/blog/breaking-changes-in-xna-game-studio-4-0.html
- Changed the Color type from BGRA to RGBA byte ordering

Source/Orts.Formats.Msts/AceFile.cs
```diff
@@ -181,7 +181,7 @@ static Texture2D Texture2DFromReader(GraphicsDevice graphicsDevice, BinaryReader
                         }
                         for (var x = 0; x < imageWidth; x++)
                         {
-                            buffer[imageWidth * y + x] = (channelBuffers[(int)SimisAceChannelId.Red][x] << 16) + (channelBuffers[(int)SimisAceChannelId.Green][x] << 8) + channelBuffers[(int)SimisAceChannelId.Blue][x];
+                            buffer[imageWidth * y + x] = channelBuffers[(int)SimisAceChannelId.Red][x] + (channelBuffers[(int)SimisAceChannelId.Green][x] << 8) + (channelBuffers[(int)SimisAceChannelId.Blue][x] << 16);
```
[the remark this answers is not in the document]
perpetualKid: so what would the docs suggest? Afaik this is due to the format change from BGRA to RGBA in XNA4

http://www.shawnhargreaves.com/blog/breaking-changes-in-xna-game-studio-4-0.html
- Changed the Color type from BGRA to RGBA byte ordering

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
perpetualKid: I don't think this is the same. OR used the ShaderModel in a way like GraphicsProfile HiDef or Reach, but not as an exact mapping to a specific shader version. Only HiDef is used anymore with MG, and run one shader version, so either we could remove  Game.Settings.ShaderModel at all, or would need to implement correct versioning throughout the code base.

Source/Menu/Options.Designer.cs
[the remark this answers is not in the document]
twpol: My preference in that case would be to configure MonoGame with a more-similar option to XNA to avoid a drastic change in performance.

A multisampling option hasn't been put on the roadmap/blueprints, nor discussed as part of the migration, so it is a surprise to me. If you don't wish to remove it, please create a forum discussion to ensure the community is okay with adding the feature (I assume they are, but this is the process we've got).

Source/Contrib/TrackViewer/TrackViewer.cs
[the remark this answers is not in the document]
twpol: Ah, your comment was trashed by GitHub so it looks like you said it changed to... exactly what it currently is in XNA 3.1. That's okay.

Source/Orts.Formats.Msts/AceFile.cs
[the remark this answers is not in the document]
twpol: I have to assume this is right but the docs say:

- XNA 3.0 (there's no 3.1 copy) at https://docs.microsoft.com/en-us/previous-versions/windows/xna/bb199274(v%3dxnagamestudio.30) says `Color` is "(Unsigned format) 32-bit ARGB pixel format with alpha, using 8 bits per channel."
- XNA 4.0 at https://docs.microsoft.com/en-us/previous-versions/windows/xna/bb199274(v%3dxnagamestudio.40) says `Color` is "(Unsigned format) 32-bit ARGB pixel format with alpha, using 8 bits per channel."

I trust Shawn's breaking changes page more than these docs, so I was just highlighting that the documentation contradicts this change.

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
twpol: Currently, we **are** using it to switch shader model, which is unrelated to XNA's HiDef/Reach definitions AFAIK.

This is because not everyone had shader model 3 supporting hardware/drivers when we initially introduced the newer scenery shaders. We choose based on whether version 3 is supported, and have separate version 2 and version 3 copies of the appropriate shaders (and the rest are all version 2).

For MonoGame we should remove the setting/code as everything is now shader model 4.

Source/Menu/Options.Designer.cs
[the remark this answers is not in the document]
Csantucci: Hi James,
yes, I agree on the public discussion. Having this on the forum allows also for the insertion of pictures and for tests by other people. I started discussion here http://www.elvastower.com/forums/index.php?/topic/33142-or-mg-and-multisampling/ and hope that there will be comments by you and other people.

Source/RunActivity/Viewer3D/Shaders.cs
Csantucci: This change was already applied by gpz, so I'm not able to provide an answer.

Source/RunActivity/Content/ShadowMap.fx
```diff
@@ -110,30 +124,18 @@ VERTEX_OUTPUT VSShadowMapForest(in VERTEX_INPUT In)
...
-	float2 offsetTexCoord = In.TexCoord + float2(0.5, 0.5);
-
-	Out.Position = mul(In.Position, WorldViewProjection);
-	Out.SampleCentre = offsetTexCoord * ImageBlurStep;
-	Out.Sample_01 = (offsetTexCoord - float2(0, 1.5)) * ImageBlurStep;
-	Out.Sample_23 = (offsetTexCoord + float2(0, 1.5)) * ImageBlurStep;
+	Out.Sample_03.xy = (offsetTexCoord - float2(1.5, 0)) * ImageBlurStep;
+	Out.Sample_03.zw = (offsetTexCoord + float2(1.5, 0)) * ImageBlurStep;
+	Out.Sample_47.xy = (offsetTexCoord - float2(0, 1.5)) * ImageBlurStep;
+	Out.Sample_47.zw = (offsetTexCoord + float2(0, 1.5)) * ImageBlurStep;
```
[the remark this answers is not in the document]
Csantucci: This has already been changed by gpz, so I'm not able to answer your question.

Source/RunActivity/Viewer3D/Common/Helpers.cs
```diff
@@ -131,7 +131,6 @@ public static bool IsSnow(Simulator simulator)
             { "Wrap", SceneryMaterialOptions.TextureAddressModeWrap },
             { "Mirror", SceneryMaterialOptions.TextureAddressModeMirror },
             { "Clamp", SceneryMaterialOptions.TextureAddressModeClamp },
-            { "Border", SceneryMaterialOptions.TextureAddressModeBorder },
```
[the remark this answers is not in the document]
Csantucci: This has already been changed by gpz, so I'm not able to comment here.

Source/RunActivity/Viewer3D/DDSLib.cs
```diff
@@ -1708,7 +1693,8 @@ private static void WriteTexture(BinaryWriter writer, CubeMapFace face, Texture
 
                 if (texture is TextureCube)
                 {
-                    (texture as TextureCube).GetData<byte>(face, i, null, data, 0, size);
+                    // FIXME: MonoGame fails with the following:
+                    //(texture as TextureCube).GetData<byte>(face, i, null, data, 0, size);
```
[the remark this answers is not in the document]
Csantucci: This has already been changed by gpz, so I'm not able to provide an answer.

Source/RunActivity/Viewer3D/DDSLib.cs
```diff
@@ -1752,7 +1738,8 @@ private static void WriteTexture(BinaryWriter writer, CubeMapFace face, Texture
 
             if (texture is TextureCube)
             {
-                (texture as TextureCube).GetData<byte>(face, mipLevel, null, data, 0, size);
+                // FIXME: MonoGame fails with the following:
+                //(texture as TextureCube).GetData<byte>(face, mipLevel, null, data, 0, size);
```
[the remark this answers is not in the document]
Csantucci: As above, this line has already been changed by gpz, so I'm not able to comment.

Source/RunActivity/Viewer3D/MSTSSky.cs
```diff
@@ -105,9 +101,13 @@ public MSTSSkyDrawer(Viewer viewer)
...
-            // Set starting value
-            mstsskyseasonType = -1;
-
+            // Set default values
+            mstsskyseasonType = (int)MSTSSkyViewer.Simulator.Season;
+            date.ordinalDate = 82 + mstsskyseasonType * 91;
+            // TODO: Set the following three externally from ORTS route files (future)
+            date.month = 1 + date.ordinalDate / 30;
+            date.day = 21;
+            date.year = 2010;
```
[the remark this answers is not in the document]
Csantucci: The change has already been done by gpz, so I'm not able to provide an answer.

Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -771,45 +807,30 @@ public override void SetState(GraphicsDevice graphicsDevice, Material previousMa
...
                 }
             }
             else
             {
-                // Enable alpha blending for everything: this allows distance scenery to appear smoothly.
-                rs.AlphaBlendEnable = true;
-                rs.DestinationBlend = Blend.InverseSourceAlpha;
-                rs.SourceBlend = Blend.SourceAlpha;
-
+                graphicsDevice.BlendState = BlendState.Opaque;
```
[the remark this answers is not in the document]
Csantucci: Already modified by gpz.

Source/RunActivity/Viewer3D/Popups/NoticeWindow.cs
```diff
@@ -143,7 +143,7 @@ public override void Draw(SpriteBatch spriteBatch)
                 else if (currentLength > AnimationLength - AnimationFade)
                     fade = (float)((currentLength - AnimationLength) / -AnimationFade);
 
-                var color = new Color(Color.White, fade);
+                var color = new Color(Color.White.R / 255f, Color.White.G / 255f, Color.White.B / 255f, fade);
```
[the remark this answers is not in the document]
Csantucci: This change has already been made by gpz, so I'm not able to comment here.

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
perpetualKid: wondering if we should assume DirectX11 hardware (which came out with Windows 7, 2009) and unify the shader version to 5.0 (DX11), as we wouldn't need to support pre-DX11 OS anymore

Source/Contrib/TrackViewer/TrackViewer.cs
```diff
@@ -1134,7 +1136,7 @@ static void InitLogging()
 
         void CalculateFPS(GameTime gameTime)
         {
-            float elapsedRealTime = (float)gameTime.ElapsedRealTime.TotalSeconds;
+            float elapsedRealTime = (float)gameTime.TotalGameTime.TotalSeconds;
```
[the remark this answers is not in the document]
perpetualKid: Changed, but I can't confirm or verify correctness

Source/RunActivity/Viewer3D/Shapes.cs
perpetualKid: why woulnd't a roadcar be allowed to cast shadows?

twpol: Road cars should have shadows. This change is **removing** shadows from road cars, by removing the `ShadowCaster` flag. 🙂

perpetualKid: ah right, just ignore above comment, the code was already updated

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
perpetualKid: I'd assume this code nor the original code OR/XNA code was built to handle multiple screens, and only handles the primary screen correct. I.e. it fails if screens have different resolution, or taskbars are located differently on each monitor. It may require a rework of the whole functionality.

twpol: That does not excuse this code. The code is making all kinds of incorrect assumptions and I would like it fixed. For example, the use of working area is only correct when we're in a window (not full screen) and has nothing to do with whether we are on the primary monitor, like the code is checking. We don't need to get multi-monitor support right here but we do need to not screw up the basics like that.

perpetualKid:
temporarily removed from Unstable builds until we figure the reason for graphics stuttering in newer builds (http://www.elvastower.com/forums/index.php?/topic/33263-monogame-now-in-or-unstable-official-version/page__view__findpost__p__250102)
This is not occuring in this PR branch, but only in combination with other PRs.

Source/RunActivity/Content/SceneryShader.fx
```diff
@@ -641,31 +581,17 @@ float4 PSSignalLight(in VERTEX_OUTPUT In) : COLOR0
...
-	}
-}
-
-technique TransferPS2 {
+technique ImagePS {
 	pass Pass_0 {
-		VertexShader = compile vs_4_0_level_9_1 VSTransfer2();
-		PixelShader = compile ps_4_0_level_9_1 PSImage4();
+		VertexShader = compile vs_4_0_level_9_3 VSGeneral();
+		PixelShader = compile ps_4_0_level_9_3 PSImage();
```
twpol: Does bumping the level here affect what hardware is supported?

Source/ORTS.Common/SystemInfo.cs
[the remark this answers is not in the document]
twpol: It looks like we can still list the device name, description, plus the device and vendor IDs. This may still be useful when debugging from someone's log file.

Source/ORTS.sln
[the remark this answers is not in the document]
twpol: I've read through this more carefully, and there's only one difference between the existing configurations and the new ones: RunActivity is built as Any CPU. Since this won't actually work at present, due to RailDriver, please remove this new configuration.

Source/RunActivity/Content/SceneryShader.fx
perpetualKid: https://en.wikipedia.org/wiki/Feature_levels_in_Direct3D
in short: yes. 
In reality I doubt there is much hardware from the ~before 2004 area around.

twpol: Probably true 🙂

And since "multiple render targets" is going to be needed 9_3 seems like a good minimum anyway. Do you know if it matters that they're currently a mixture of 9_1 and 9_3? Would it be useful to make every shader 9_3?

perpetualKid: probably better to avoid different version, I've just updated that.

PR #41, opened 2019-04-01
Correct vacuum brake issues

peternewell:
This patch is to correct an issue with correctly sensing the presence of the small steam ejector - I do not know how to separate other changes from this PR, and just associate the necessary file segments.


twpol:
@peternewell If you'd like, I can try doing a rebase so that this PR only includes the bug fix for https://bugs.launchpad.net/or/+bug/1822556?

peternewell:
That would be fine if it allows the patch to be committed.

twpol:
@peternewell I've reverted the extra commit included here - do the changes ("Files changed" tab) look good to you?

twpol:
Unfortunately I could not do a rebase myself, as that would need a force-push into your repository (which GitHub denies, understandably). Do you know how you got into this situation and how to avoid it in future?

peternewell:
> Do you know how you got into this situation and how to avoid it in future?
No, the only thing I could think of was that I had another outstanding PR, and some how this code fragments got mixed.

twpol:
> No, the only thing I could think of was that I had another outstanding PR, and some how this code fragments got mixed.

It looks like you made the changes for PR #23 on the master branch, instead of on its own branch, hence this situation. You should always create a new branch from master for each set of changes.

PR #42, opened 2019-04-05
Monogame AnyCPU Prefer32Bit

perpetualKid:
Upgrade all project build config to AnyCPU, but setting Prefer32Bit on all executables. This PR is an interim step towards x64 compatibility (which will require some code adoptions, coming next step)

@twpol As AnyCPU Pref32Bit sets the IMAGE_FILE_LARGE_ADDRESS_AWARE bit by default, there is no more need for separate executables and preference setting. The setting has been removed, please update the build config as necessary.

#40 is already pre-merged to ensure consistency.


YoRyan:
Temporarily removing from unstable because this PR's changes to Tests.csproj conflict with those of #234.

PR #43, opened 2019-04-05
Monogame AnyCpu x64

perpetualKid:
changes in this PR finally enable full support for both, 32bit and 64bit environments
- OpenAL and RailDriver have been updated to support both mode
- RD binary libraries added for both 32 and 64bit
- Removed 32bit preference bit in build configuration
- Updated code to show correct virtual memory size

https://trello.com/c/Ql02djix/390-64bit-version-4gb-ram-usage


perpetualKid:
Carlo and I found that the current way RD is implemented, does not work for both 64 and 32 bit in the same assembly, so we can't enable an x64 version without breaking 32bit.
This would either require to load the corresponding PIE assemblies dynamically at runtime, or different way accessing PIE RD (not using PIE Callback Handler interfaces but pulling data from RD.

twpol:
> Carlo and I found that the current way RD is implemented, does not work for both 64 and 32 bit in the same assembly, so we can't enable an x64 version without breaking 32bit.
> This would either require to load the corresponding PIE assemblies dynamically at runtime, or different way accessing PIE RD (not using PIE Callback Handler interfaces but pulling data from RD.

There seem to be a fair few suggestions for doing the dynamic loading:

- https://stackoverflow.com/a/10852827
- https://stackoverflow.com/a/30646096
- https://stackoverflow.com/a/31115656
- https://stackoverflow.com/a/156024

Is the plan to come back to this later after the RD loading is appropriately adjusted?

perpetualKid:
> https://stackoverflow.com/a/10852827
> https://stackoverflow.com/a/30646096
> https://stackoverflow.com/a/31115656
> https://stackoverflow.com/a/156024
> 
> Is the plan to come back to this later after the RD loading is appropriately adjusted?

If someone is interested in implementing such kind loading mechanism, it could be added here and the PR can be reopened and continued.

While technically feasible, I do not believe a dynamic load mechanism to be the cleanest way, and would only fall back if no other way possible. 
Given there's proof already in separate whole new RD implementation that a different implementation approach of PIE handler seems to work both for 32 and 64bit without such efforts, its a question of priorization also how important that is for OR:
- 64bit enablement
- RD for both 32 and 64bit
- new RD implementation
Here I've been trying to bring an intermediary fix to enable 1) and 2) without 3), but eventually we found this is not possible.

PR #48, opened 2019-04-12
Moving UserCommands from InputSettings into Common/Input

perpetualKid:
correcting enum name and move out from InputSettings to Orts.Common.Input
as per http://www.elvastower.com/forums/index.php?/topic/13336-raildriver-sdk/page__view__findpost__p__244639 
No functional changes, renaming only.
This is also prep for RD and other input.


perpetualKid:
> Only a single, minor, issue: please put the using directives that you have added in alphabetic order; e.g. in Weather.cs you didn't do that (using ORTS.Common.Input has been added at the end of the using list) .
I've run VS standard "Remove and Sort Usings" on all files affected by this PR. This sorts the namespaces alphabetically, but puts System namespaces first.

perpetualKid:
@twpol - are we good to merge here? We've been trying to follow the workflow, but wanted to double check with you

twpol:
@perpetualKid Yup, all good to merge here!

I'm about to write a comment in the forum but GitHub should be configured with the rules already.

perpetualKid:
> @perpetualKid Yup, all good to merge here!
> I'm about to write a comment in the forum but GitHub should be configured with the rules already.

it is indeed, we just wanted to check for the one

PR #50, opened 2019-04-12
OpenAL binary re-integration with updated default settings

perpetualKid:
Bug report: https://www.elvastower.com/forums/index.php?/topic/32923-something-wrong-with-bala/

- Going back to distribute specific binary version of OpenAL as part of OR (openal-soft v 1.16.0)
- implementing code to handle both x86 and x64 versions of oal binaries at runtime
- setting/updating configuration to allow max 1024 sound sources (default 256, which would be exceeded at some dense routes)
- remove old, stale oal-soft code with hardcoded config changes
- since OAL binaries are distributed again with OR, remove corresponding installation check in launcher


Csantucci (approved):
Rationale of the commit OK.
Code check OK,
Testing performed by dennisat on his test bed OK.
A final review by twpol would be welcome, but not absolutely necessary.

twpol (approved):

PR #52, opened 2019-04-17
Removing check and empty file creation

perpetualKid:
In some cases there seem a race condition to happen when the file is new created, setting the initial value would not be happen and honored until second start of the game.

Since WritePrivateProfileString will also just create the file if not existing, the prior check and creation is not necessary.

### DIFF
--- a/Source/RunActivity/Viewer3D/OpenAL.cs
+++ b/Source/RunActivity/Viewer3D/OpenAL.cs
@@ -439,10 +439,6 @@ namespace Orts.Viewer3D
             string configFile = Path.Combine(Environment.GetFolderPath(Environment.SpecialFolder.ApplicationData), "alsoft.ini");
             try
             {
-                if (!File.Exists(configFile))
-                {
-                    File.CreateText(configFile);
-                }
                 StringBuilder result = new StringBuilder(255);
                 if (ORTS.Common.NativeMethods.GetPrivateProfileString("General", "sources", string.Empty, result, 255, configFile) > 0)
                 {


perpetualKid:
> I'm willing to approve merging of this PR. However before this, accordingly to rules that James stated, you should file a bug report here https://bugs.launchpad.net/or/?orderby=-id&start=0 and insert the bug link in your PR.

@twpol it's counterproductive if we interpret the rules that way that I'd have to create a bug report for own code, and wait until someone else triages the bug before one could submit a change request (fix) - particularly at our current pace this slows down things further

Csantucci:
Steffen, I can set that bug as confirmed, and so we can proceed with the review. Creating the bug has also a statistic value.

Csantucci:
Creating a bug report requires less than 5 minutes.

twpol:
Okay, so two things here:

- Developers should be able to report a bug and move it into the appropriate state for a pull request (I need to look into the Launchpad settings as that might not be working right now)
- We used to allow commits that were small (max ~10 lines IIRC) and obvious fixes to go in without a bug report

I would be happy to migrate the small and obvious rule into the new contributing document, if people are happy to continue to allow such changes.

I understand that creating a bug report just to fix your own code seems excessive, but right now we're using Launchpad to track what's going on and we need to keep a record in there for that reason at least. Once this is all officially documented, people can officially make proposals to change it, and we can figure out the costs to doing so.

Csantucci:
I was using slight different rules about bug creation, but I'm OK with the rules James is suggesting. Until I'll get triaging rights, I'll go temporarily on creating pull requests without triaging the bug, but putting it into the "In progress" state. In force of the suggested rules, I'll now positively review Steffen's PR.

PR #56, opened 2019-05-08
CopyCreate LAA-exe .config file

perpetualKid:
the LAA-exe is missing a .exe.config file when creating, and does not seem to run without the config. Extending the build script to create an .exe.config for the LAA-exe by copying the one from non-LAA version. Thanks to mbm_OR@ET for reporting and providing a fix

https://bugs.launchpad.net/or/+bug/1828232


twpol:
FYI I'd strongly suggest not approving any PRs if the most recently merged PR has not been built automatically - doing so causes lost builds. I can't easily recreate X1.3.1-55 now.

Csantucci:
Sorry for this, James, I thought I did a good thing in removing in short time this problem.

twpol:
You did great getting it done quickly, and I hope that'll continue once I've fixed this build problem.

PR #57, opened 2019-05-08
Ensure that build fails when tests fail to run

twpol:
This is a fix for the issue where the build in AppVeyor succeeds incorrectly when the test suite fails to run.

Unfortunately, this cannot be merged without either:

1. Bypassing the AppVeyor build result
1. Fixing the problem with the test runner not loading the .NET 4 version of the test suite

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -129,8 +129,14 @@ REM Build locales.
 PUSHD Source\Locales && CALL Update.bat non-interactive && POPD || GOTO :error
 
 REM Run unit tests (9009 means XUnit itself wasn't found, which is an error).
+CALL :delete "xunit.xml" || GOTO :error
 xunit.console.x86 Program\Tests.dll /nunit xunit.xml
 IF "%ERRORLEVEL%" == "9009" GOTO :error
+CALL :file-size xunit.xml
+IF "%FileSize%" LEQ 100 (
+	>&2 ECHO ERROR: Test results file "xunit.xml" is %FileSize% bytes; expected more than 100 bytes.
+	GOTO :error
+)
 
 CALL :copy "Program\RunActivity.exe" "Program\RunActivityLAA.exe" || GOTO :error
 editbin /NOLOGO /LARGEADDRESSAWARE "Program\RunActivityLAA.exe" || GOTO :error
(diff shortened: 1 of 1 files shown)


twpol:
It does not look like there is an easy solution to this, so I am going to stop for the night and try again tomorrow or Friday.

Until this is resolve, I am not allowing any PRs to merge because we do not have working tests or builds.

perpetualKid:
> It does not look like there is an easy solution to this, so I am going to stop for the night and try again tomorrow or Friday.
> Until this is resolved, I am not allowing any PRs to merge because we do not have working tests or builds.

I suspect the xunit.dll which we have bundled with Tests.dll is version 1.9.2, and xunit only seems to support .NET  4.5+ from version 2+
However there seem to be a few breaking changes from xunit 1 to v2 and the unit tests would have to be migrated. Given there are only two unit tests implemented and both in areas not currently worked on, I'd put this on backlog and bypass unit tests for the time being if this requires a bigger effort to get builds working again.

twpol:
There's 192 tests, not two, although I get your point that we aren't testing much of the codebase. I've disabled the requirement for test results in my Jenkins and forcibly rebuilt the two missing versions, which are now available.

As for this issue, I managed to get the tests running by using the "clr4.x86" version of xUnit but only after I added the `useLegacyV2RuntimeActivationPolicy="true"` attribute. This is needed because some of the tests pull in the XNA framework (for the Vector classes, at least). The problem now is, how do we make this work in AppVeyor - I am somewhat hesitant to go modifying their installation of xUnit, even if they probably do reset everything before anyone else uses it.

perpetualKid:
> useLegacyV2RuntimeActivationPolicy="true"

we had to set the same flag at the recent PR where we upgraded NetFX, for the same reason to get it XNA working with new framework. Given that XNA is going away at the next step when we enable MG; might be worth to test if the MG PR (it's actually removed again in #42 ) would solve this?

YoRyan:
#234 upgrades the unit tests to xUnit 2.4.1. Perhaps this will allow you to resolve this PR.

cjakeman (commented):
HI James,

Is this PR still incompatible with the current code?

Chris

twpol:
Looks like we've got all tests passing which makes things easier – I can simply make any error running the xUnit tests fail the build now :)

twpol:
> Is this PR still incompatible with the current code?

The updated version I've just done is compatible

PR #58, opened 2019-05-12
Bug fix for  "Scripts don't work any more"

Csantucci:
https://bugs.launchpad.net/or/+bug/1828710
http://www.elvastower.com/forums/index.php?/topic/33041-problem-with-13156-g58c69ee6/#entry248022

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
+++ b/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
@@ -34,7 +34,7 @@ namespace Orts.Common.Scripting
     {
         readonly Simulator Simulator;
         readonly Dictionary<string, Assembly> Scripts = new Dictionary<string, Assembly>();
-        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider(new Dictionary<string, string>() { { "CompilerVersion", "v4.0" } });
+        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider();
 
         static CompilerParameters GetCompilerParameters()
         {
--- a/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
+++ b/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
@@ -34,7 +34,7 @@ namespace Orts.Common.Scripting
     {
         readonly Simulator Simulator;
         readonly Dictionary<string, Assembly> Scripts = new Dictionary<string, Assembly>();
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -34,7 +34,7 @@ public class ScriptManager
     {
         readonly Simulator Simulator;
         readonly Dictionary<string, Assembly> Scripts = new Dictionary<string, Assembly>();
-        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider(new Dictionary<string, string>() { { "CompilerVersion", "v3.5" } });
```
perpetualKid: Technically the provided change fixes the issue for now.
The ultimate fix would be to remove the version parameter at all, so the code compiles in the same version on default as the program is running on. Else we may run into the same issue again at some point when compiler version changes.

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -34,7 +34,7 @@ public class ScriptManager
...
+        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider(new Dictionary<string, string>() { { "CompilerVersion", "v4.0" } });
```
perpetualKid: Technically the provided change fixes the issue for now.
The ultimate fix would be to remove the version parameter at all, so the code compiles in the same version on default as the program is running on. Else we may run into the same issue again at some point when compiler version changes.
```suggestion
        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider();
```

Csantucci: Hi Steffen,
yes, what you say is correct. Take into account anyhow that MS has already changed its policy, because if you write "4.0" it's valid also for all 4.x versions. In fact if you write "4.5" or "4.7.2" you get an error.

Csantucci:
Seems I'm not authorized to merge, even if the changes are approved.

perpetualKid:
> Seems I'm not authorized to merge, even if the changes are approved.

same here, I think this is still because of the issue with the build server. Adding @twpol to comment/merge

Sharpe49:
FYI, I have done tests with the latest version of the scripts. It works correctly with the latest patch.

twpol:
@Csantucci Sorry, when I temporarily removed the development team permission while we resolved the test issue, GitHub made additional changes that I did not know about, so did not undo when I put the development team permission back. I hope it is sorted now.

Csantucci:
James, thank you that you sorted the test issue out!

PR #59, opened 2019-05-18
Update POT files and French locales. Fix for https://bugs.launchpad.net/or/+bug/1830120

Sharpe49:
Since UserCommand.cs is now in ORTS.Common, I updated the InputSettings.cs file too.


Sharpe49:
Perhaps I should also move the UserCommand locales in the other languages.

Sharpe49:
For your information, many of the locale files needs updating.

Sharpe49:
Alright, here's the link to the bug report :
https://bugs.launchpad.net/or/+bug/1830120

PR #60, opened 2019-05-20
Manually re-route AI trains through dispatcher window

Csantucci:
https://blueprints.launchpad.net/or/+spec/re-route-ai-dispatcher

### DIFF
--- a/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
@@ -3861,6 +3861,7 @@ namespace Orts.Simulation.Signalling
         ///   - if valid path only is requested and unreserved section is found (variable thisTrain required)
         ///   - end of track
         ///   - looped track
+        ///   - re-enter in original route (for manual re-routing)
         ///
         /// Returned is list of sections, with positive no. indicating direction 0 and negative no. indicating direction 1
         /// If signal or speedpost is required, list will contain index of required item (>0 facing direction, <0 backing direction)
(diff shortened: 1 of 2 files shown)


Source/Orts.Simulation/Simulation/Signalling/Signals.cs
```diff
@@ -8308,7 +8338,7 @@ public enum HoldState                // signal is locked in hold
 
         public Train.TCSubpathRoute signalRoute = new Train.TCSubpathRoute();  // train route from signal
         public int trainRouteDirectionIndex;    // direction index in train route array (usually 0, value 1 valid for Manual only)
-        private int thisTrainRouteIndex;        // index of section after signal in train route list
+        public int thisTrainRouteIndex;        // index of section after signal in train route list
```
perpetualKid: not a good practice to have public fields (and this is wide spread throughout the whole solution).
If we make this a public field, should ast least follow naming conventions (camel casing, start with upper case letter). 
More generally I'd suggest some refactoring separately to cleanup code.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
```diff
@@ -3869,7 +3870,7 @@ public void BreakDownRouteList(Train.TCSubpathRoute reqRoute, int firstRouteInde
         public List<int> ScanRoute(Train thisTrain, int firstSectionIndex, float firstOffset, int firstDirection, bool forward,
                 float routeLength, bool honourManualSwitch, bool autoAlign, bool stopAtFacingSignal, bool reservedOnly, bool returnSections,
                 bool searchFacingSignal, bool searchBackwardSignal, bool searchFacingSpeedpost, bool searchBackwardSpeedpost,
-                bool isFreight, bool considerSpeedReset = false)
+                bool isFreight, bool considerSpeedReset = false, bool checkReenterOriginalRoute = false)
```
perpetualKid: this is an awful list of parameters and should be candidate for refactoring (separately)

PR #62, opened 2019-05-22
New SignalScript Parser

perpetualKid:
separate files and namespace
https://blueprints.launchpad.net/or/+spec/separate-sigscr-parser

new Parser engine is build on tokenizer, and removes the regex and string-copy/index based approach, which is not only more stable to identify erronous scripts, but also much faster and less memory consuming:
http://www.elvastower.com/forums/index.php?/topic/32200-refactoring-signalscript-parser/

### DIFF
--- a/Source/Orts.Formats.Msts/Orts.Formats.Msts.csproj
+++ b/Source/Orts.Formats.Msts/Orts.Formats.Msts.csproj
@@ -50,6 +50,10 @@
     <Compile Include="CabViewFile.cs" />
     <Compile Include="EngineFile.cs" />
     <Compile Include="EnvironmentFile.cs" />
+    <Compile Include="Signalling\Enums.cs" />
+    <Compile Include="Signalling\Parser.cs" />
+    <Compile Include="Signalling\Tokenizer.cs" />
+    <Compile Include="Signalling\Tokens.cs" />
     <Compile Include="SignalScripts.cs" />
     <Compile Include="TerrainFlagsFile.cs" />
     <Compile Include="HazardFile.cs" />
--- a/Source/Orts.Formats.Msts/SignalScripts.cs
+++ b/Source/Orts.Formats.Msts/SignalScripts.cs
@@ -930,6 +930,12 @@ namespace Orts.Formats.Msts
                         }
                         else
                         {
+                            //substitute a trailing + or - operator token to become part of the (numeric) parameter 
+                            if (block.Tokens.Count > 1 && ((block.Tokens[0] as OperatorToken)?.Token == "-" || (block.Tokens[0] as OperatorToken)?.Token == "+"))
+                            {
+                                block.Tokens[1].Token = block.Tokens[0].Token + block.Tokens[1].Token;
+                                block.Tokens.RemoveAt(0);
+                            }
                             SCRParameterType parameter = ParameterFromToken(block.Tokens[0], block.LineNumber, localFloats, orSignalTypes, orNormalSubtypes);
                             result.Add(parameter);
                         }
@@ -1037,7 +1043,12 @@ namespace Orts.Formats.Msts
                         statement.Tokens.RemoveAt(0);
                         negated = true;
                     }
-
+                    //substitute a trailing + or - operator token to become part of the (numeric) term 
+                    if (statement.Tokens.Count > 1 && ((statement.Tokens[0] as OperatorToken)?.Token == "-" || (statement.Tokens[0] as OperatorToken)?.Token == "+"))
+                    {
+                        statement.Tokens[1].Token = statement.Tokens[0].Token + statement.Tokens[1].Token;
+                        statement.Tokens.RemoveAt(0);
+                    }
                     if (statement.Tokens.Count > 1 && Enum.TryParse(statement.Tokens[0].Token, out SCRExternalFunctions externalFunctionsResult) && statement.Tokens[1] is Enclosure)   //check if it is a Sub Function ()
                     {
                         Term1 = new SCRStatTerm(externalFunctionsResult, statement.Tokens[1] as Enclosure, 0, string.Empty, negated, localFloats, orSignalTypes, orNormalSubtypes);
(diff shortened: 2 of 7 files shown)


Sharpe49:
Oh my... I just looked at this code. It seems great ! 👍

Roeterdink:
Get an error on this line (line 7848) :
       else if (!Train_Requires_Next_Signal(nextNormalSignalId), 1)

Attached are sigcfg.dat and sigscr.dat, as well as debug print out of old version and new version (as far as it got)
[newsignalscript.zip](https://github.com/openrails/openrails/files/3213643/newsignalscript.zip)

perpetualKid:
> Get an error on this line (line 7848) :
> else if (!Train_Requires_Next_Signal(nextNormalSignalId), 1)

I'm trying to understand what kind of condition statement this is, maybe I'm missing something:
there is no valid operator between the function call and 1, and 1 is not a paramter for something else. 
C language (the base for MSTS signal scripts) does not provide such syntax, nor is such mentioned in the signal script specification. Is this an error in the script, or is there any other spec availabe describing this case?

Same for operator mismatch in line 5329 of this signal script (=#) - there is no such operator defined, the rounding assignment operator would be #= instead (although I see my code didn't caught that operator either, I'll commit a fix shortly).

Roeterdink:
Line : else if (!Train_Requires_Next_Signal(nextNormalSignalId), 1)
is fully valid.
The function is a boolean function and its outcome is negated.
This kind of test is used often in signalscripts, especially is statements for !enabled or !routeset().
Signal script is NOT C or C# - it's 'like' C or C# but has its own peculiarities, and this is one of them.

Roeterdink:
Oh wait - I see what you mean - it looks as if the '1' is out of place.
Could that be the problem? Still, the program should not crash.

As for "else if" and "elseif" : both are allowed.

perpetualKid:
> Signal script is NOT C or C# - it's 'like' C or C# but has its own peculiarities, and this is one of them.

the signal script spec states: 
"The scripting language is like a cut down version of ‘C’.  The layout of code within a script is the same as ‘C’."

> As for "else if" and "elseif" : both are allowed.

Spec defines only "if, else, else if	" for Conditional program control.

The next issues in the provided script is i.e. in SCRIPT RSL_SVLA where states are combined (ie line 10884) 
"nextNormalState == nextNormalState + 16;"

this again is not valid statement, since == i equality operator and not assignment.

We could discuss on the elseif as enhancement for OR, but else I'm sticking to the originial spec in lieu of an OR specific one.

adding @twpol for insights

Csantucci:
If MSTS accepts elseif, also OR should.

perpetualKid:
> If MSTS accepts elseif (which - as shown in this case - means that there are existing scripts that use it), also OR should.

the script is a good test case to capture issues and provide output, however I doubt it will run in MSTS as all in it's current state. There are quite a bunch of errors in the script, and I'm adding some more output to better find those - most are just swallowed or not even found in the current script parser implementation. As for the elseif, I have not found any MSTS script which uses elseif, and as said the MSTS signal script spec does not define elseif either.

Csantucci:
OK, I thought the elseif had been found in a script file of a public route. I did a search and in all routes I have I havent' found elseifs.

Roeterdink:
The situation here is that the software was not made according to the spec, but the spec was made according to what the software allowed and what we wanted to promote as the best options.
MSTS was very lenient it what it allowed for syntax and many 'signalmasters', in particular those who created scripts for payware routes, exploited this to some very weird limits.
Because of compatibility most of this had to be handled; even the few extreme issues which I did not include caused a lot if argument.
Many of those 'signalmasters' were very protective of their scripts and also very reluctant to adapt to OR.
The program had to and can indeed handle a lot of these non-specified constructions, but clearly we were not going to promote these by including them in the specification.
The result is that the specification does not cover all the program can and must be able to process, and therefor this specification can not be used as guideline to what should and should not be allowed.
Compatibility must be maintained with most of the existing scripts, at least those which OR can handle with the present code. That will take a lot of testing and will require the program to handle a lot more syntax constructions than specified. And those additional contructions are not detailed anywhere, whenever there was a situation with a route where the script was not processed correctly due to some non-specified construction the program was altered and that was it.
Only the following constructions were never included : =+, +=, ++, -=, --, =-
Also, the setup used by MLT to define a signalhead as two types was not implemented.

Csantucci:
A way to find out what these "custom" adaptations are about could be to get a log with all versions of OR (starting from which version?) where SIGSCRfile.cs was modified and to analyse the rationale of such versions. Could this be a viable way?

Roeterdink:
Signal script processing was the first software I wrote for OR - it dates back to 2011.
Any idea how many changes will have been made since then, and how are you going to separate real additions from changes required due to existing scripts?
And - there were no blueprints or bug reports in those early days.

perpetualKid:
> Signal script processing was the first software I wrote for OR - it dates back to 2011.
> Any idea how many changes will have been made since then, and how are you going to separate real additions from changes required due to existing scripts?
> And - there were no blueprints or bug reports in those early days.

At least with the given script, I see more issues with the script than with the Parser

- == and ==# are mis-used in assignment scenarios - I've added a warning message, and even an adhoc correction to treat as = or #=
- variable names misspelled, so they do not match the local or external floats declarations - I've added further trace messages
- in few places elseif is used, while mostly it's correctly spelled else if. I'm hesitant to implement an fix for elseif, but will have a look

There is one issue I've found with negative number which I'm looking into right now. Other than that, unless we find any scenarios which can't be covered with the current implementation, I'd favour to stick with a standardized approach, unless someone documents a spec

twpol:
The current signal script parser/code has a number of language features we know are used in content (or have tried to be used in content) which it does not support, so I would be hesitant to use the existing code as a guide for what should work or not.

My preferred approach seems to be what's already going on here: start with a very cut-down version of C, which we know forms the basis of the language MSTS uses, add additions to the language documented by MSTS, and finally figure out what is missing by using real-world signal scripts to guide us.

Compatibility is always hard, and in many cases it is challenging (or even impossible) to distinguish between the following two cases:

- Something that works because it is supported
- Something that works because it is ignored in just the right way

I would say that, for this signal script work, if we make sure that all unexpected constructs/code are reported as warnings and where possible we continue parsing the code (I am not sure how often this will actually be), that should make it possible for people to test out the new code on pre-existing content.

We'll have to decide on a case-by-case basis which weird and wonderful constructs we support and which we ignore, but we are at least in the situation where we can log everything unusual for content creators to look at - if they want - unlike MSTS.

twpol:
@perpetualKid Was this PR closed in error?

perpetualKid:
> @perpetualKid Was this PR closed in error?

on purpose, I think there is some broader incompatibility to be solved first.

Roeterdink:
Perhaps in view of this, we should reconsider this whole situation.
As is now clear, there is always a risk in replacing existing software that the new software does not fully behave similar as the software it replaces.
The question is if in this case, the effort is justified by the possible gain.
From a user point of view, and from operational point of view, there is no need to replace this function. There have been no issues, bugs or whatever for years now, and the code itself has also remained unchanged for quite a few years except for the occasional addition of new functions.
It may not be the best of code, it's not the worst either for at least it does what it is intended for - and without compiler warnings as well. That can not be said for all parts of OR.
And let's be honest about it - operational gain is absolutely minimal. A once-off saving of 200 msec during startup is not something which will set the whole community singing and dancing. In fact, without special tooling to measure it, one will never notice.
There are many parts of OR where restructuring is far more urgent. For starters, I would greatly appreciate if we could once more get back to the situation where OR can be compiled without warnings.
On a more complex and larger scale, the whole chain of Simulation - AI - Train/AITrain/TTTrain has been  altered, extended, redrawn and realtered so often that it threatens to collapse under its own weight. A full reconstruction of this part of the code, splitting it over functional classes, is long overdue.
Furthermore, many aspects of these functions, and for signalling as well, were all completely new, build from scratch when introduced. They have proven their worth, but over time it has become clear there are certain shortcomings, and usage has also extended our knowledge of the intended functionality. Real gain is to be had from reconstructing this logic in line with this new knowledge. I do already have some ideas on this, but it is clear that replacing the existing logic is going to be a very large project.
I suggest we see this as a leason learned, but now turn our attention to parts of the program which require our attention far more urgently. I am willing to start this discussion and work out some details of the main changes to various logic which I have in mind.
And let's start by sorting out those compiler warnings.

PR #63, opened 2019-05-25
When .s file requires missing .dds file, the equivalent .ace is picked if existing

Csantucci:
Bug fix for https://bugs.launchpad.net/or/+bug/1830498 .
See also http://www.elvastower.com/forums/index.php?/topic/33071-track-textures/

### DIFF
--- a/Source/RunActivity/Viewer3D/Materials.cs
+++ b/Source/RunActivity/Viewer3D/Materials.cs
@@ -65,9 +65,21 @@ namespace Orts.Viewer3D
                 try
                 {
                     Texture2D texture;
-                    if (Path.GetExtension(path) == ".dds" && File.Exists(path))
+                    if (Path.GetExtension(path) == ".dds")
                     {
-                        DDSLib.DDSFromFile(path, GraphicsDevice, true, out texture);
+                        if (File.Exists(path))
+                        {
+                            DDSLib.DDSFromFile(path, GraphicsDevice, true, out texture);
+                        }
+                        else
+                        {
+                            var aceTexture = Path.ChangeExtension(path, ".ace");
+                            if (File.Exists(aceTexture))
+                            {
+                                texture = Orts.Formats.Msts.AceFile.Texture2DFromFile(GraphicsDevice, aceTexture);
+                            }
+                            else texture = defaultTexture;
+                        }
                     }
                     else if (Path.GetExtension(path) == ".ace")
                     {


Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -65,9 +65,21 @@ public Texture2D Get(string path, Texture2D defaultTexture, bool required = fals
...
+                    if (Path.GetExtension(path) == ".dds")
                     {
-                        DDSLib.DDSFromFile(path, GraphicsDevice, true, out texture);
+                        if (File.Exists(path))
+                        {
+                            DDSLib.DDSFromFile(path, GraphicsDevice, true, out texture);
+                        }
+                        else
+                        {
+                            var aceTexture = Path.ChangeExtension(path, ".ace");
```
twpol: Why would we have a filename here of "foo.dds" when there is only a "foo.ace" on disk? That's a content creator mistake, surely?

Csantucci: If you read here http://www.elvastower.com/forums/index.php?/topic/33071-track-textures/ , and in particular here http://www.elvastower.com/forums/index.php?/topic/33071-track-textures/page__view__findpost__p__248467 , you see that there are commercial providers that are providing routes with Global folders where the standard track shapes refer to .dds files instead of .ace files. If you use a standard route (which hasn't .dds files) with such global folder it doesn't show the track shapes. And apparently it's not clear whether such special routes are recommended to be installed as mini-routes or not. So, this feature lets OR become a bit more "tolerant".

twpol: Okay, so the real problem here is that the shapes are in global and the textures per-route, and they've overwritten the global shapes, right?

If we're going to allow this mess to work, we should at least log a warning that something fishy is going on.

twpol: And probably add a comment explaining why this code exists, otherwise someone will break it in 6 months time when they have no idea why we need it.

Csantucci: Yes, that's what is occurring. I could add a log warning: this would result in a log line for every missing .ace, so let's hope that this does not create a burst of log lines.

Csantucci: PR with log warning and comment created. Waiting for review.

PR #68, opened 2019-05-31
Correction in OR manual.

Csantucci:
Corrected info about HASHEAD signal script keyword


perpetualKid:
> Approved, but perhaps Eugen's text is better.

as part of the approval process, I think it's expected the approver to Merge the PR also, unless there is some ask or request to the initiator or another party for a change (like fixing a typo)

Csantucci:
I'll modify the text next time I'll have changes for the manual.

PR #74, opened 2019-07-10
Removing on screen notifications of camera change

Csantucci:
Feature request: https://www.elvastower.com/forums/index.php?/topic/33200-removing-on-screen-notification-of-camera-change/page__mode__show


perpetualKid (approved):
would be better though to have this configurable

PR #81, opened 2019-07-29
20190714 add closeup to station stop

Roeterdink:
Add closeup command to station stops, see bug report #1838236


Roeterdink:
Link to bug report : https://bugs.launchpad.net/or/+bug/1838236

PR #82, opened 2019-07-29
20190714 correct next station window

Roeterdink:
Display of text "Train is to attach to <other train> at next stop" will only be displayed for correct stop. See bug report #1838235.


Roeterdink:
Link to bug report : https://bugs.launchpad.net/or/+bug/1838235

PR #87, opened 2019-07-29
20190714 correct station stops

Roeterdink:
Correct problems with station stop handling, see bug report #1838234


Roeterdink:
Link to bug report : https://bugs.launchpad.net/or/+bug/1838234

PR #90, opened 2019-07-29
20190724 weather files

Roeterdink:
Add processing of weather files (for timetable mode) : see blueprint https://blueprints.launchpad.net/or/+spec/weatherfile

### DIFF
--- a/Source/Menu/MainForm.Designer.cs
+++ b/Source/Menu/MainForm.Designer.cs
@@ -84,7 +84,7 @@ namespace ORTS
             this.comboBoxDifficulty = new System.Windows.Forms.ComboBox();
             this.label8 = new System.Windows.Forms.Label();
             this.panelModeTimetable = new System.Windows.Forms.Panel();
-            this.lableTimetableWeatherFile = new System.Windows.Forms.Label();
+            this.labelTimetableWeatherFile = new System.Windows.Forms.Label();
             this.comboBoxTimetableWeatherFile = new System.Windows.Forms.ComboBox();
             this.label24 = new System.Windows.Forms.Label();
             this.comboBoxTimetableTrain = new System.Windows.Forms.ComboBox();
--- a/Source/Menu/MainForm.Designer.cs
+++ b/Source/Menu/MainForm.Designer.cs
@@ -84,6 +84,8 @@ namespace ORTS
             this.comboBoxDifficulty = new System.Windows.Forms.ComboBox();
             this.label8 = new System.Windows.Forms.Label();
             this.panelModeTimetable = new System.Windows.Forms.Panel();
+            this.lableTimetableWeatherFile = new System.Windows.Forms.Label();
+            this.comboBoxTimetableWeatherFile = new System.Windows.Forms.ComboBox();
             this.label24 = new System.Windows.Forms.Label();
             this.comboBoxTimetableTrain = new System.Windows.Forms.ComboBox();
             this.label23 = new System.Windows.Forms.Label();
@@ -702,6 +704,8 @@ namespace ORTS
             // 
             // panelModeTimetable
             // 
+            this.panelModeTimetable.Controls.Add(this.lableTimetableWeatherFile);
+            this.panelModeTimetable.Controls.Add(this.comboBoxTimetableWeatherFile);
             this.panelModeTimetable.Controls.Add(this.label24);
             this.panelModeTimetable.Controls.Add(this.comboBoxTimetableTrain);
             this.panelModeTimetable.Controls.Add(this.label23);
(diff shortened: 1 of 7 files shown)


Source/Menu/MainForm.Designer.cs
```diff
@@ -84,6 +84,8 @@ private void InitializeComponent()
             this.comboBoxDifficulty = new System.Windows.Forms.ComboBox();
             this.label8 = new System.Windows.Forms.Label();
             this.panelModeTimetable = new System.Windows.Forms.Panel();
+            this.lableTimetableWeatherFile = new System.Windows.Forms.Label();
```
twpol:
```suggestion
            this.labelTimetableWeatherFile = new System.Windows.Forms.Label();
```
(Typo in name of the label)

Source/Orts.Formats.OR/WeatherFile.cs
```diff
@@ -253,9 +262,9 @@ public void Save(BinaryWriter outf)
     // fog
     public class WeatherSettingFog : WeatherSetting
     {
-        public float FogVisibilityM = 20000;                           // required fog density - range 0 - 20000
-        public float FogSetTimeS = 300;                                // required rate for fog setting - range 300 - 3600
-        public float FogLiftTimeS = 300;                               // required rate for fog lifting - range 300 - 3600 - required visibility is taken from next weather
+        public float FogVisibility = 1000;                             // required fog density - range 0 - 1000
+        public float FogSetTime = 300;                                 // required rate for fog setting - range 300 - 3600
+        public float FogLiftTime = 300;                                // required rate for fog lifting - range 300 - 3600 - required visibility is taken from next weather
```
twpol: What units are these values, if not meters and seconds?

Source/Orts.Formats.OR/WeatherFile.cs
```diff
@@ -101,15 +102,18 @@ public WeatherSettingOvercast(JsonReader json)
...
         {
+
+            // get values
             if (base.TryParse(item)) return true;
             switch (item.Path)
             {
-                case "Overcast": Overcast = item.AsFloat(Overcast); break;
-                case "OvercastVariation": OvercastVariation = item.AsFloat(OvercastVariation); break;
-                case "OvercastRateOfChange": OvercastRateOfChange = item.AsFloat(OvercastRateOfChange); break;
+                case "Overcast": Overcast = item.AsFloat(Overcast) >= 0 ? item.AsFloat(Overcast) : -1; break;
```
twpol: What's the purpose of this ternary? Why would you force all negative values to be -1?

Source/RunActivity/Viewer3D/Weather.cs
```diff
@@ -914,6 +923,94 @@ public AutomaticWeather(Viewer viewer)
...
+                // check if time is in sequence
+                if (weatherSet.Time < prevTime)
+                {
+                    Trace.TraceInformation("Invalid time value : time out of sequence : {0}", acttime.ToString());
+                    weatherSet.Time = prevTime + 1;
+                }
+                prevTime = weatherSet.Time;
+
+                // check settings
+                if (weatherSet.GetType() == typeof(WeatherSettingOvercast))
```
twpol: A much better way of doing this is:
```suggestion
                if (weatherSet is WeatherSettingOvercast)
```

Source/RunActivity/Viewer3D/Weather.cs
```diff
@@ -914,6 +923,94 @@ public AutomaticWeather(Viewer viewer)
...
+                    CheckValue(ref thisFog.FogVisibility, false, 10, 20000, acttime, "Fog Visibility");
+                }
+            }
+        }
+
+        // check value, set random value if allowed and value not set
+        void CheckValue(ref float setValue, bool randomize, float minValue, float maxValue, TimeSpan acttime, string description)
+        {
+            // overcast
+            if (setValue < 0 && randomize)
```
twpol: Since you're checking for negative here, the ternary that explicitly sets them to -1 when loading is unnecessary.

Source/ORTS.Menu/ORTimetables.cs
```diff
@@ -137,6 +139,35 @@ public static List<TimetableInfo> GetTimetableInfo(Folder folder, Route route)
...
+            if (route != null)
+            {
+                var directory = System.IO.Path.Combine(route.Path, "WeatherFiles");
+
+                if (Directory.Exists(directory))
+                {
+                    foreach (var weatherFile in Directory.GetFiles(directory, "*.weather-or"))
+                    {
+                        FileInfo weatherFileInfo = new FileInfo(weatherFile);
+                        weatherInfo.Add(weatherFileInfo.Name);
```
twpol: It might be better to add the full paths (i.e. `weatherFile`) to the list here, and only shorten to a filename when displaying in the dropdown.

twpol: The way we do that on all the other dropdowns is by using a class to store the entry, and the `ToString()` method is overridden to provide the nice display value. It would be good to try that here too.

Source/Orts.Formats.OR/WeatherFile.cs
Roeterdink: These are meters and seconds, but the fog variables were the only ones which had this reflected in the names.
Out of consistency, it should either be all or nothing, and 'nothing' led to fewer changes.

Source/ORTS.Menu/ORTimetables.cs
Roeterdink: I was struggling here to just show the name, and admittedly the idea of creating a special class with dedicated ToString() never crossed my mind.
Will look into this.

Roeterdink: Updated code with new class 'WeatherFilesInfo', which holds information on available files.

Source/Orts.Formats.OR/WeatherFile.cs
Roeterdink: All checks on negative values have been removed.

Source/RunActivity/Viewer3D/Weather.cs
Roeterdink: Replaced all "typeof" checks with "is" statement.

Source/Orts.Formats.OR/WeatherFile.cs
twpol: We have a requirement that [all fields, etc. containing real world measurements must be suffixed with the unit](https://github.com/openrails/openrails/blob/master/CONTRIBUTING.md#physics-requirements).

I would like to set a good example here, and renaming each field is only an F2 away in Visual Studio, so please could you restore these unit suffixes. If consistency dictates adding them to other, related fields, so be it - I would rather we ended up with more correct unit suffixes than less. 🙂

Roeterdink: Set the suffixes for all relevant variables in the main class definitions.

twpol: Thanks!

PR #100, opened 2019-09-10
Add warning if train has no locomotive but is required

WallyCZ:
This trace message helps little bit when finding cause of NullReferenceException...


Csantucci (approved):
What increases readability of the logfile and clarity of events is welcome.

PR #101, opened 2019-09-15
https://blueprints.launchpad.net/or/+spec/stock-visual-effects - chan…

peternewell:
…ges to allow for steam heat boiler to be located in non-locomotive context


peternewell:
Close to fix merge issues

PR #104, opened 2019-10-04
Diesel performance

peternewell:
https://blueprints.launchpad.net/or/+spec/diesel-performance - correct some issues with diesels not starting


peternewell:
It appears that two PRs have become merged into one. This PR to be closed and rebuilt as a separate version.

PR #105, opened 2019-10-05
refactor: Reformat all code with dotnet format

twpol:
This is the result of running `dotnet format` (with version 3.1.37601+256f37159dc60b29f76999749f57f4338324f102) on the entire `Source` directory.

The [supported formatting options](https://github.com/dotnet/format/wiki/Supported-.editorconfig-options) are all about whitespace, for now, so this diff is entirely adding and removing spaces.


YoRyan:
Closing this PR as too much has changed since it was approved. I would support a dotnet format run, though.

PR #110, opened 2019-10-12
Fix bug of stop signal skipped when inside platform

Csantucci:
https://bugs.launchpad.net/or/+bug/1847844


Csantucci:
OK, I've corrected in the test output. Pls. review the PR again.

PR #114, opened 2019-11-07
Runactivity/dispatch window on top

cjakeman3:
Added trick to bring window to the front, even though invisible, so that it appears immediately when Ctrl+9 is pressed.

Manual updated to match.

See blueprint https://blueprints.launchpad.net/or/+spec/dispatch-window-on-top


cjakeman3:
Hi Carlo,

"Changes to ORTS.sln, Runactivity.csproj and Sound.cs, and deletion of Readme file within the Program folder don't seem related. Maybe you could revert such unrelated changes."

Thanks for reviewing and spotting that. Those changes were not intended. I'll try to undo them tomorrow.

PR #119, opened 2019-11-28
Bug fix for https://bugs.launchpad.net/or/+bug/1854381 Bell plays forever in AI trains after automatic horn blow

Csantucci:


Csantucci:
The fix lets the bell play for 30 seconds after horn blow, instead for an infinite time.
Also the remaining time when a save is performed during the bell play is recomputed.

PR #120, opened 2019-12-10
Bug fix for https://bugs.launchpad.net/or/+bug/1208960 Wrong warning about missing animations in Milestones and Speedposts

Csantucci:


Csantucci:
The patch is within class SpeedPostShape, therefore it affects only speedposts.

Csantucci:
Thank you for cooperation!

PR #130, opened 2020-01-01
Bug fix for https://bugs.launchpad.net/or/+bug/1857203 Sound thread safety crash

Csantucci:


Csantucci:
With the contribution of dennisat.

Csantucci:
Thank you Steffen. Such event handlers are only related to sound, that is they e.g. start or stop a sound. From the experience of players the occurrence of a catch{} is quite rare, and the fact that in such case there is a sound which is not managed as it should is IMHO a price that can be paid to avoid heavier CPU load with Interlocked mechanisms. And of course what is implemented here, which avoids the crash, is better than the actual situation where a crash can occur.

PR #137, opened 2020-01-21
Harmonize initialization of autopilotable player train with that of standard player trains

Csantucci:
See bug https://bugs.launchpad.net/or/+bug/1858323 Unjustified player train initialization differences.
As of now autopilotable player trains have an initialization procedure that differs from that of standard player trains. Some differences are justified, and some not. This patch modifies the initialization procedure of autopilotable player trains to in the points where differences with standard player trains is not justified.

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
@@ -403,9 +403,10 @@ namespace Orts.Simulation.AIs
                 }
             }
 #endif
-            // check deadlocks
+            // check deadlocks; do it after placing for player train, like done for it when autopilot option unchecked
 
-            CheckDeadlock(ValidRoute[0], Number);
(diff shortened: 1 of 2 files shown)


Csantucci:
Hi Rob,
Thanks for approval. I don't see changes about display of elevation in Trac.k Monitor.
The changes in AITrain.cs are to PostInit(), while the changes in Train.cs are related to Datalogging, which crashed when the train was autopiloted because LeadLocomotiveIndex is = -1 in that case. I don't see other changes in the "Files changed" tab here.

Csantucci:
Those parameters are written in the Datalog file. That part of code wasn't executed before the changes, because the override PostInit used for the autopiloted train didn't start the Datalogging. Now I have added the possibility of having datalogging also for the autopiloted train, so that part of code is now executed and would generate a crash. So these changes must be done together with the harmonization.

PR #152, opened 2020-04-10
Improved window shading, add letterboxing mode for legacy 2D cabs

YoRyan:
This pull request was born out of my personal frustration with 2D cabs in Open Rails. Most of them are built for the 4:3 world of MSTS, whereas Open Rails takes full advantage of your widescreen monitor by stretching or clipping the cabs to fit. Personally, I would much rather maintain full visibility of the cab, and I suspect this is the reason others have [requested](https://trello.com/c/f3sfTLYR/312-user-selectable-resolution-for-full-screen-mode) the ability to specify a custom fullscreen resolution - it's not the custom resolution that is important _per se_, but rather the ability to force the graphics card into a 4:3 aspect ratio so that 2D cabs will render in their entirety.

I propose that Open Rails do the letterboxing by itself. To that end, this PR:

* Replaces the popup window shader with PNG textures drawn with SpriteBatch. I had to do this because using viewports breaks the ResolveBackBuffer call, which OR currently uses to render glass window backgrounds. SpriteBatch should be a significantly faster (and IMHO, simpler) than the shader, and ResolveBackBuffer is one of the blockers for a complete Monogame port (@perpetualKid), so if nothing else, I suggest taking a close look at this commit.

* Adds a new flag UserSetting.Letterbox2DCab, disabled by default, to activate cab letterboxing. And adds a new checkbox to the Video settings tab to expose this flag.

* Scales OR's viewport to match the 2D cab, with black letterboxing, when letterboxing is requested and the player is in cab view. In all other views and conditions, the sim will revert to the full viewport. It's quite elegant.

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
@@ -6323,7 +6323,7 @@ namespace Orts.Simulation.AIs
             ResetActions(true, true);
             if (SpeedMpS != 0) MovementState = AI_MOVEMENT_STATE.BRAKING;
             else if (this == Simulator.OriginalPlayerTrain && Simulator.ActivityRun != null && Simulator.ActivityRun.Current is ActivityTaskPassengerStopAt && ((ActivityTaskPassengerStopAt)Simulator.ActivityRun.Current).IsAtStation(this) &&
-                ((ActivityTaskPassengerStopAt)Simulator.ActivityRun.Current).BoardingEndS > Simulator.ClockTime)
+                ((ActivityTaskPassengerStopAt)Simulator.ActivityRun.Current).BoardingS > 0)
             {
                 StationStops[0].ActualDepart = (int)((ActivityTaskPassengerStopAt)Simulator.ActivityRun.Current).BoardingEndS;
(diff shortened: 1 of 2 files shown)


Csantucci:
Hi YoRyan,
and thank you for your PR. To have it merged into the official OR release, the feature has to be approved by the OR management team (I'm not inside it). To have the approval you should create a Trello card as explained here http://www.elvastower.com/forums/index.php?/topic/28517-open-rails-12-planning/page__view__findpost__p__201634 . When the Trello card will be approved, any developer, including myself, will have the authorization to merge your PR. In the meantime I ask you if I can insert your feature within OR NewYear MG, which is the unofficial release described in this thread http://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/

YoRyan:
Thank you for pointing that out, Csantucci. I should have read the contributing guidelines before I submitted this PR.

I have created a [discussion thread](http://www.elvastower.com/forums/index.php?/topic/33908-letterboxing-for-2d-cabs/) on Elvas Tower for this new feature. In the meantime, the code is GPL and you are absolutely free to include it in your fork. Please let me know if I can be of any assistance.

YoRyan:
Withdrawn in favor of #163 and #164.

PR #155, opened 2020-04-15
Allowing manual CallOn from dispatcher window

cesarBLG:
In activity mode, TrainHasCallOn() and TrainHasCallOn_Restricted() sigscr functions will always return the same value for the same signal. I add the possibility of manually allowing a train to call on, selecting an option in signal's menu of dispatcher window. This option is only available in signals whose scripts make use of call on functions.
Discussion: http://www.elvastower.com/forums/index.php?/topic/33899-manual-shunting-using-dispatcher-window/
Roadmap: https://trello.com/c/I19VVKlz

### DIFF
--- a/Source/Orts.Simulation/MultiPlayer/Message.cs
+++ b/Source/Orts.Simulation/MultiPlayer/Message.cs
@@ -3535,10 +3535,7 @@ namespace Orts.MultiPlayer
                     }
                     break;
                 case 4:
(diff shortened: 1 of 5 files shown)


cjakeman:
Rob Roeterdink thinks this is a good idea.
Trello card created and I've moved it to Release 1.4

Thanks, César

PR #161, opened 2020-04-21
Bug fix for  https://bugs.launchpad.net/or/+bug/1463537 Flashing MSTSwater 

Csantucci:
See also http://www.elvastower.com/forums/index.php?/topic/33897-flashing-water/


Csantucci:
Comment added.

PR #163, opened 2020-04-22
Fix glass windows in the Monogame port

YoRyan:
Currently, Monogame cannot render transparent glass windows because of the use of `ResolveBackBuffer()` in the current code, which was [deprecated](https://www.shawnhargreaves.com/blog/resolvebackbuffer-and-resolvetexture2d-in-xna-game-studio-4-0.html) in XNA 4.0 and has no obvious replacement. This PR would replace PopupWindowShader with a few SpriteBatch calls to achieve Monogame-ready glass windows.

[Related Trello card.](https://trello.com/c/5Aj0EM4w/457-draw-glass-windows-with-spritebatch-instead-of-resolvebackbuffer) [Elvas Tower discussion.](http://www.elvastower.com/forums/index.php/topic/33908-letterboxing-for-2d-cabs/)


YoRyan:
Withdrawn due to poor performance.

PR #165, opened 2020-04-22
Bug fix for https://bugs.launchpad.net/or/+bug/1874209 Waiting point does not expire

Csantucci:


Csantucci:
Replaced values with a constant.

PR #166, opened 2020-04-22
Bug fix for https://bugs.launchpad.net/or/+bug/1874240 route_cleared_to_signal not working for non NORMAL signals

cesarBLG:
See http://www.elvastower.com/forums/index.php?/topic/33924-using-route-cleared-to-signal-in-non-normal-signals/


Csantucci:
Hi César,  sorry for being bureaucratic, but being this a bug fix, you should create a bug report entry here https://bugs.launchpad.net/or and reference it here. After that the bug fix can be merged.

PR #168, opened 2020-04-23
Bug fix for https://bugs.launchpad.net/or/+bug/1874383 Remote player trains at standstill not updated in MP

cesarBLG:
When a remote train is at standstill, the server won't update its route, so signals aren't properly cleared. If a signal is at danger, the train has to move a bit to clear it. With this fix player trains controlled remotely are updated if stopped (only in server), as is done with local player train.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Simulator.cs
+++ b/Source/Orts.Simulation/Simulation/Simulator.cs
@@ -674,7 +674,7 @@ namespace Orts.Simulation
 
             foreach (Train train in Trains)
             {
-                if (train.SpeedMpS != 0 &&
+                if ((train.SpeedMpS != 0 || (train.ControlMode == Train.TRAIN_CONTROL.EXPLORER && train.TrainType == Train.TRAINTYPE.REMOTE && MPManager.IsServer())) &&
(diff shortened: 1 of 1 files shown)


Csantucci:
This has been generated so by the author (not me), to reduce amount of data to be transmitted. At least in my country, it is known that you have to move a bit to clear the signal. So I am a bit uncertain if this should be implemented.

cesarBLG:
This data is not transmitted, but only calculated in the server by its own. It might increase the load in server. Maybe adding remote trains to a list which is updated every second, or updating one remote train per loop could be implemented. With long freight trains, if you have to release the brakes in order to update the signal, and then apply them again, it can be a pain.

cesarBLG:
Anyway, a low percentage of trains will be stopped, so CPU usage in the server won't be reduced too much avoiding the update (as far as I know updating the train doesn't imply broadcasting any data, but I might be wrong).

Csantucci:
CPU usage is not an issue I think. I checked in MPManager.cs, and in fact it seems that only moving trains are broadcast to the clients. So I think that your fix can be accepted.

PR #171, opened 2020-04-29
Document C# scripting in the manual

YoRyan:
Open Rails' four scripting interfaces are considerably powerful and should be documented. This manual update includes a high-level overview of each interface and some tips on developing scripts in Visual Studio.


Csantucci:
Ryan, what you are doing here is a long overdue work, thank you!
Only one remark: you write that the script must reside in a "Script" subfolder of the locomotive's folder. That's not exact. The script may reside everywhere, provided it is correctly referred to. Simply change the path in line 
ORTSTrainControlSystem ( "YourTCS.cs" ).
for this.
This is an example:
ORTSTrainControlSystem( "..\\..\\common.Script\\IT\\TCS_SCMT_IT" )
 It is even advisable in some cases to have the scripts residing in a subfolder of a Common.script folder within TRAINSET. This way, analogously to Common.cab, Common.snd and so on, more .eng files residing in different folders can cleanly share the same copy of the script.

YoRyan:
That trick isn't documented for cabview and sound files, so this would be the first time it appears in the manual. Nevertheless, I have added it.

PR #173, opened 2020-04-29
GetValidWindowSize

cjakeman:
Ensures user's entry in Menu > Options > Video > Window Size is valid or ignores it.

A user entered "1024X780", RunActivity.exe did not run and nothing in log file, so was stumped.

Bug Report at [https://bugs.launchpad.net/or/+bug/1875887](https://bugs.launchpad.net/or/+bug/1875887)

### DIFF
--- a/Source/Menu/Options.cs
+++ b/Source/Menu/Options.cs
@@ -461,7 +461,8 @@ namespace ORTS
             Settings.DistantMountainsViewingDistance = (int)numericDistantMountainsViewingDistance.Value * 1000;
             Settings.ViewingFOV = (int)numericViewingFOV.Value;
             Settings.WorldObjectDensity = (int)numericWorldObjectDensity.Value;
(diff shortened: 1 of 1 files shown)


Source/Menu/Options.cs
```diff
@@ -538,6 +539,15 @@ void buttonOK_Click(object sender, EventArgs e)
             Settings.Save();
         }
 
+        private string GetValidWindowSize()
+        {
+            // "1024X780" instead of "1024x780" then "Start" resulted in an immediate return to the Menu with no OpenRailsLog.txt and a baffled user.
+            var windowSizeArray = comboWindowSize.Text.ToLower().Replace(" ", "").Split('x');
+            return (int.TryParse(windowSizeArray[0], out int width) && int.TryParse(windowSizeArray[1], out int height))
```
perpetualKid, line 546: this will crash if the string.Split does not return an array of expect length, and should be checked also!

PR #175, opened 2020-04-30
Web server documentation and bug fixes

YoRyan:
- Add a description of the web server to the manual.
- Fix web content not getting copied to the Program folder.
- Fix other small bugs.
- Use case-insensitivity where appropriate for more intuitive behavior.
- Decode URI strings ("index%20-%20ApiSample.html" -> "index - ApiSample.html").

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -210,6 +210,9 @@ When activated, the server can be accessed (on the local machine only) at
            | data types.
          - Orts.Viewer3D.WebServices.WebServer.ApiSampleData
 
+    There is no rate-limiting or caching, so mind your API calls' effects on the
+    simulator's performance.
+
 Audio Options
 =============
 
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -168,6 +168,48 @@ Disable TCS scripts
 This option disables the train control system scripts for locomotives where
 these have been implemented.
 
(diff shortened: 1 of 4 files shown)


Csantucci:
Ryan, the AppVeyor test has failed due to an internal error. To resolve this, either delete and present again the PR, or do an even simple change in one of the modified files, so that Appveyor starts again. (I don't know whether there is a more elegant way).

YoRyan:
It looks like a rebase did the trick!

Csantucci:
It would also be nice if Mauricio's (mbm_OR) improvements, which are present in the OR NewYear MG version, could be taken into consideration. If you agree with that, Chris, you could contact Mauricio about this.

YoRyan:
Hi Chris,

Yes, I am also happy to collaborate. Here is my personal web server wish list:

- It would be preferable to use an existing HTTP server framework, like ASP.NET Core, which - unlike legacy ASP.NET - is cross-platform and open-source. Rolling our own implementation inevitably misses some of the nuances of the protocol, like case-insensitive headers.
- I would like to create an API for manipulating the controls of the train. I am working on a touch-enabled web app to emulate a Raildriver on a tablet and I would need this to port the app to Open Rails.

If you don't mind my taking ownership of the web server code for a little while, I can take a look at Mauricio's improvements and then report back with a new patch against the unstable branch.

Finally, feel free to direct me to a more suitable place for discussion than this GitHub issue.

PR #177, opened 2020-05-04
added more checks to GetValidWindowSize()

cjakeman:
Thanks for your catch, PerpetualKid,

_this will crash if the string.Split does not return an array of expect length, and should be checked also!_

THis new version avoids that fatal error and also does some sanity checking.


peternewell:
Seems ok.

peternewell (approved):
Seems ok

PR #178, opened 2020-05-05
Fading signal lamps

YoRyan:
Change signal lamps to fade on and off. Effect is customizable by route builders.

[(Elvas Tower)](http://www.elvastower.com/forums/index.php?/topic/33951-smooth-signal-lamp-transitions/) [(Trello)](https://trello.com/c/P9rYgv8V/461-fading-signal-lamps)


YoRyan:
Done! I have changed the `OnOffTime` property. The flash ones will have to stay as they are as they are Kuju names.

PR #186, opened 2020-05-09
Fix 3D cab stuttering in Monogame

YoRyan:
With Monogame, there is an annoying stuttering effect in the 3D cab view; this bug was most recently reported on Elvas Tower against the NewYear version of OR, but it also occurs with the official Monogame branch.

Bug report: https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__255667

The cause seems to lie within the 3D digit and gauge rendering code, in which OR recreates a new ShapePrimitive to draw the digit/gauge every single frame. Apparently, this is not a very efficient operation under Monogame. So, I propose introducing a new MutableShapePrimitive class that permits the direct manipulation of the vertex and index buffers. This seems to remove the performance bottleneck, but the drawback is that we have to hardcode a maximum number of cab display digits (I chose 6, as used by the constructor).

This is a very technically involved fix, so I am all ears to any suggestions and criticism.

### DIFF
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2639,6 +2639,7 @@ namespace Orts.Viewer3D.RollingStock
 
     public class ThreeDimCabDigit
     {
+        const int MaxDigits = 6;
         PoseableShape TrainCarShape;
         VertexPositionNormalTexture[] VertexList;
         int NumVertices;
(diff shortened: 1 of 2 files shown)


cjakeman (commented):
Hi Ryan,

Good to see some new ideas, especially MutableShapePrimitive. Also nice to see foreach in use too.

Please would you add a Bug Report for this issue so that it can be included in the records.

In the meantime, I have added a "for unstable" label which should get it included into the Unstable version.

Thanks,
Chris

YoRyan:
[Done.](https://bugs.launchpad.net/or/+bug/1878975) Unfortunately, this bug is back, because my changes were overwritten by one of the auto-merged commits.

cjakeman:
Thanks, Ryan.

Earlier today I spotted that your PR had dropped out of the list waiting to be approved and asked James for his advice. I'll let you know what he says.

If you can re-submit the PR, I will approve it.

Thanks,

YoRyan:
Succeeded by #191.

PR #194, opened 2020-05-20
Update Diesel Code

peternewell:
https://blueprints.launchpad.net/or/+spec/diesel-performance

Correct an issue with one prime mover not shutting down, and added first attempt at unloading power

http://www.elvastower.com/forums/index.php?/topic/33990-locomotive-unloading-speed/


peternewell:
Hi Chris,

Thanks for that.

I am breaking out of lockdown for a few days, so will update the manual when I return.

Cheers

Peter

peternewell:
Manual now updated, can you approve if you are happy with it?

PR #198, opened 2020-05-24
https://trello.com/c/8RK0kuCt/465-enable-rotation-for-digital-2d-cabview-controls

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34052-rotation-of-digital-instrument-in-cvf-files/page__pid__259073#entry259073


Csantucci:
Ok, modified code and updated OR manual.

PR #199, opened 2020-05-25
Rewrite Steam Heating Code

peternewell:
https://blueprints.launchpad.net/or/+spec/rewrite-steam-heat - Add steam heating boiler to electric locomotive

### DIFF
--- a/Source/Orts.Simulation/Simulation/Confirmer.cs
+++ b/Source/Orts.Simulation/Simulation/Confirmer.cs
@@ -50,6 +50,7 @@ namespace Orts.Simulation
       , PlayerDiesel
       , HelperDiesel
       , DieselFuel
+      , SteamHeatBoilerWater
       // Steam power
       , SteamLocomotiveReverser
       , Regulator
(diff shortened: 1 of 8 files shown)


peternewell:
Hi Chris,

Thanks for reviewing.

This was based on the approach that I needed to use with the steam locomotive for the watercontroller, where the tender water level wasn't restored correctly unless I saved to a "dummy" value and then restored from it.

The only time that "RestoredCurrentLocomotiveSteamHeatBoilerWaterCapacityL" is non negative is after a restore, so I used this logic to determine when a restore was done.

Thanks

Peter

cjakeman:
Hi Peter,

> This was based on the approach that I needed to use with the steam locomotive for the watercontroller, where the tender water level wasn't restored correctly unless I saved to a "dummy" value and then restored from it.

This sounds like a work-around for a problem which would be fine for the Unstable version, but probably not for the Testing version.

Shouldn't we look for a solution to restoring both tender water level and CurrentLocomotiveSteamHeatBoilerWaterCapacityL correctly?

Is there anything I can do to help?

Chris

peternewell:
Hi Chris,

> Shouldn't we look for a solution to restoring both tender water level and CurrentLocomotiveSteamHeatBoilerWaterCapacityL correctly?

What do you mean by "correctly", as this approach does work, and appears a "valid" way of doing it?

Thanks

Peter

cjakeman:
Hi Peter,

> What do you mean by "correctly", as this approach does work, and appears a "valid" way of doing it?

I was thinking we should be able to save and restore these values in the usual way. If there's a problem in doing that, it would be better to fix the problem rather than continue with a workaround.

What issues did you have with saving and restoring the tender water level?

Best wishes,

Chris

peternewell:
Hi Chris,

> I was thinking we should be able to save and restore these values in the usual way. If there's a  problem in doing that, it would be better to fix the problem rather than continue with a workaround.

Ok, I will have a look at it and see what I can do.

peternewell:
Hi Chris,

> I was thinking we should be able to save and restore these values in the usual way. If there's a problem in doing that, it would be better to fix the problem rather than continue with a workaround.

I have updated it now so that it restores directly.

Thanks

Peter

PR #201, opened 2020-06-03
Fix cab letterboxing memory leak

YoRyan:
There is a small memory leak in the cab letterboxing code that was reported [at Elvas Tower](http://www.elvastower.com/forums/index.php?/topic/34069-potential-memory-leak/). This is a fix. Beware: It won't auto-merge cleanly into the unstable branch.


YoRyan:
Will withdraw this until the Monogame port is stable.

PR #203, opened 2020-06-05
Extended starting friction parameters

YoRyan:
Blueprint [Parameters to adjust friction at low or no speed](https://blueprints.launchpad.net/or/+spec/stationary-friction)

Additional .wag parameters for starting friction. Based off of code [shared](http://www.elvastower.com/forums/index.php?/topic/33919-orts-extended-friction-test/) by NickOnWheels of Elvas Tower.

I also refactored the super-sized `UpdateTrainBaseResistance` method. This results in a rather large diff, but my hope is that it's now much clearer which physics model applies during which states.

> README for new starting friction equation - by NickonWheels
> 
> First and foremost, the three Davis lines
> 
> ORTSDavis_A
> ORTSDavis_B
> ORTSDavis_C
> 
> work as before. However now it is POSSIBLE to replace ORTSBearingType with
> 
> ORTSStandstillFriction - the amount of static friction / defaults to newton but can use pounds, just like ORTSDavis_A
> ORTSMergeSpeed - the speed at which low speed calculations and Davis equations meet (while accelerating and decelerating) / uses meters per second, other units are untested
> 
> The older code section uses 5mph as merge speed throughout.
> 
> If you don´t want to bother with this new section ORTSBearingType (or lack thereof in regards to plain bearings) still works as before.
> Of course it´s hard to know what values seem to fit as FCalc won´t tell that.

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -61,10 +61,11 @@ number should be used instead of zero.
 
 When a car is pulled from steady state, an additional force is needed due
 to higher bearing forces. The situation is simplified by using a different
-calculation at low speed (5 mph and lower). Empirical static friction
-forces are used for different classes of mass (under 10 tons, 10 to 100
+calculation at low speed (``ORTSMergeSpeed`` and lower). Empirical static
+friction forces are used for different classes of mass (under 10 tons, 10 to 100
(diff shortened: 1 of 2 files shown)


YoRyan:
I'll see if I can adapt his provided readme text.

YoRyan:
Manual's updated!

PR #209, opened 2020-06-08
Enable rotation of digital clock cab control; by Paolo

Csantucci:
Completion of https://trello.com/c/8RK0kuCt/465-enable-rotation-for-digital-2d-cabview-controls


Csantucci:
Correct extended HUD, by mbm_OR

Csantucci:
Hi Chris,
this is only an extension of the already merged PR that allowed rotation of digital cab controls; it uses the same method used by the already merged PR. I can add a line to the manual for this.
Paolo wasn't yet sure to be able to generate PRs when the feature was ready, so we agreed that I presented it in his name, but now it's not a problem to me to withdraw it and to let him present it, if you prefer.

PR #210, opened 2020-06-08
Add ORTS_Battery and ORTS_Powerkey commands

Paoloserpi78:
New command ORTS_BATTERY it's a simple Two state command with 2 dedicated sound triggers (177 and 178) that reproduces the switch located on the back of the cab capable of enabling all 24v voltage instruments.
The combination of keys to turn the switch on and off is "Ctrl + B", activation is also possible via mouse if the switch is visible on the front of the cab.
New command ORTS_POWERKEY it's a simple Two state command with 2 dedicated sound triggers (179 and 180) that reproduces the key of the cab desk, used also to turn on diagnostic monitors.
The combination of keys to turn the switch on and off is "Ctrl + B", activation is also possible via mouse if the switch is visible on the front of the cab.


Sharpe49:
Hello,

You have to send a new blueprint on launchpad.net/or for new functionnalities.

By the way, the blueprint for the master key (better term for "power key") is already available since I was already working on it.

I do not agree with this implementation mainly because you don't integrate this functionnality with the different scripts (these informations are useful for all of the subsystems of a train).
Also, for me, the battery should be integrated inside the PowerSupply classes and the master key inside a new class.

Paoloserpi78:
> Hi Paolo,
> 
> I've resolved the conflict between your CabViewFile.cs and another and merged the result.

Thanks so much! i try a lot of time also with the help of Carlo but without success

cjakeman:
Following discussions with Paolo, Cedric expects the code to change a lot.
I think we should close the pull request for the time being.

PR #220, opened 2020-06-20
Further TCS extensions

Sharpe49:
Added ability to change the customized TCS cabview control name at a specific identifier
Added support for switches for TCS cabview controls
Added a blinker helper class for scripts

Blueprint : https://blueprints.launchpad.net/or/+spec/expand-potential-of-tcs-scripts
Forum thread : http://www.elvastower.com/forums/index.php?/topic/33693-expand-potential-of-tcs-scripting/page__st__60


Sharpe49:
Alright, I updated the PR comment and the manual.
I also added the constants and an InitializeMoving function.

The next changes for the TCS will be in another PR, so the extensions are not yet completed.

PR #224, opened 2020-07-01
Additional Sound Events

peternewell:
https://blueprints.launchpad.net/or/+spec/addtional-steam-locomotive-sound-triggers - Add sound event for gear positions, and adjust injector in AI mode so that sound event only turns on of off with the injector.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
@@ -5453,11 +5453,7 @@ namespace Orts.Simulation.RollingStocks
                     Injector2IsOn = false;
                     Injector2Fraction = 0.0f;
                     InjectorLockedOut = true;
-                    if (!Injector1SoundIsOn) // turn injector sound on only when injector starts
-                    {
-                        Injector1SoundIsOn = true;
-                        SignalEvent(Event.WaterInjector1On);
-                    }
+                    PlayInjector1SoundIfStarting();
                 }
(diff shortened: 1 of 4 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -5447,6 +5453,11 @@ private void UpdateFiring(float absSpeedMpS)
                     Injector2IsOn = false;
                     Injector2Fraction = 0.0f;
                     InjectorLockedOut = true;
+                    if (!Injector1SoundIsOn) // turn injector sound on only when injector starts
+                    {
+                        Injector1SoundIsOn = true;
+                        SignalEvent(Event.WaterInjector1On);
+                    }
```
YoRyan, line 5460: I notice you repeat this logic many times. Consider factoring it out into a helper method?

YoRyan:
Hi Peter,

Code looks good. Please also add documentation of the new triggers to the manual.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: Unfortunately I am a "self taught" coder, so always happy to have examples of better ways to code stuff and achieve the same outcome.

Thanks

YoRyan: See 7952b1944047a2197044f4b9707af1579a0431e5.

YoRyan:
Approved with a manual update (courtesy of Peter) and some minor tweaks. Thanks again for your contributions.

PR #225, opened 2020-07-04
Fix advanced adhesion wheelslip upon resuming from save: https://bugs.launchpad.net/or/+bug/1294410

YoRyan:
On steep grades and at slow speeds, advanced adhesion physics can cause a wheelslip upon resuming from a save. This is because the state of the wheelslip model is not being properly restored.

This PR also includes a fix for the projected speed display on the Track Monitor, which displays "NaN" for a short time after resuming due to an initial acceleration calculation that comes out to "negative infinity." I have chosen to fix this by fixing the acceleration to 0 m/s/s^2 upon resume, but @peternewell has also [proposed](https://www.trainsim.com/vbts/showthread.php?333695-Resuming-Activity-From-Save-Locomotive-Wheelslip-and-Power-Reduction-Occurs!&p=1970175#post1970175) saving the state of the Track Monitor's acceleration projection. Perhaps another developer could chime in on which alternative seems more proper for Open Rails.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -647,8 +647,8 @@ namespace Orts.Simulation.Physics
             Number = inf.ReadInt32();
             TotalNumber = Math.Max(Number + 1, TotalNumber);
             Name = inf.ReadString();
-            SpeedMpS = inf.ReadSingle();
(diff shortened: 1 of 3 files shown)


YoRyan:
Done and done.

Ryan

YoRyan:
Hi Chris,

I have incorporated Peter's track monitor saving change. Since this has been sitting in unstable for quite some time, I was wondering if you could take a second look.

Thanks,
Ryan

PR #229, opened 2020-07-07
Fix MSTS-compatible physics reverting to Davis physics: https://bugs.launchpad.net/or/+bug/1886566

YoRyan:
We can't rely on the state of the Davis instance variables as an indicator of the current physics model, since they are updated by other parts of the physics code. Therefore, determine the current physics model at initialization time.


cjakeman:
https://bugs.launchpad.net/or/+bug/1886566

PR #230, opened 2020-07-11
Monogame migration

twpol:
### Monogame migration

**Discussion:** http://www.elvastower.com/forums/index.php?/topic/33263-monogame-now-in-or-unstable-official-version/
**Blueprint:** https://blueprints.launchpad.net/or/+spec/monogame
**Roadmap:** https://trello.com/c/IXfkjwSb/408-switch-from-xna-to-monogame

I have created this PR from Peter's original migration and merged the latest Testing Version code; the following checklist is the remaining issues I wish to at least investigate, if not address, with the code included here. Until then, it is not suitable for review or merging.

- [x] Review all of Peter's changes
  - [x] `TrackViewer.CalculateFPS()` - incorrect FPS calculation
  - [x] `AceFile.Texture2DFromReader()` - duplicate code
- [x] Merge Testing Version
- [x] Small issues
  - [x] `UserInput.InputSettings`
  - [x] `RenderProcess.ShadowMapMipCount`
  - [x] `DDSLib.GenerateNewCubeTexture()` with hasMipMaps
  - [x] `DDSLib.WriteTexture()`
  - [x] `RenderFrame.DebugRenderState()`
- [x] Reinstate border texture address mode
  - [x] `SceneryMaterialOptions.TextureAddressModeBorder`
  - [x] `SceneryMaterial.GetShadowTextureAddressMode()`
  - [x] `Helpers.TextureAddressingModeNames`
  - [x] `SharedShape.SubObject.UVTextureAddressModeMap`
- [x] Upgrade MonoGame to latest version
  - [x] Upgrade to MonoGame 3.7.1.189
  - [x] `SystemInfo.WriteGraphicsAdapter()`
  - [x] `Viewer.UpdateAdapterInformation()`
  - [x] `RenderRrocess.GDM_PreparingDeviceSettings()` with NVIDIA PerfHUD
- [x] Instancing investigation
  - [x] RenderFrame.cs - DummyVertexBuffer, DummyVertexDeclaration, DummyVertexData
  - [x] Shapes.cs - use if DummyVertexBuffer
  - [x] Terrain.cs - use of DummyVertexBuffer
  - [x] Water.cs - use of DummyVertexBuffer
- [x] Reinstate window glass
  - [x] `PSPopupWindowGlass()`
  - [x] `WindowManager.Draw()`
  - [x] `PopupWindowMaterial.SetState()`
- [x] Fix obvious bugs
  - [x] Windowed -> full-screen window with "Fast full-screen alt-tab" is in the wrong position and the wrong size
  - [x] Full-screen window with "Fast full-screen alt-tab" is in the wrong position
- [x] `Shader.GetEffectCode()`
- [x] `Viewer.SaveScreenshotToFile()`
- [x] Review all changes to `Color`
- [x] Review all `GraphicsProfile` references
- [x] ~Review Carlo/Steffen's changes~
- [x] Known issues
  - [x] Flickering distant scenery
    - https://github.com/openrails/openrails/pull/230#issuecomment-706375176
  - [x] Smoke effects are wrong colour
    - http://www.elvastower.com/forums/index.php?/topic/34479-u2020-10-16-2036-and-smoke-colour/page__view__findpost__p__263764
    - https://bugs.launchpad.net/bugs/1900215
  - [x] Giant snowballs from the sky
    - http://www.elvastower.com/forums/index.php?/topic/34479-u2020-10-16-2036-and-smoke-colour/page__view__findpost__p__263793
  - [x] Pixelated display
    - http://www.elvastower.com/forums/index.php?/topic/34489-graphical-problem-with-unstable-version/page__view__findpost__p__263887
  - [x] 2D cab doesn't darken at night
    - http://www.elvastower.com/forums/index.php?/topic/34511-2d-cab-doesnt-darken-at-night/page__view__findpost__p__264114
  - [x] Many implicit truncation warnings from shaders
    - http://www.elvastower.com/forums/index.php?/topic/34512-implicit-truncation-warnings/page__view__findpost__p__264115
  - [x] Multisample anti-aliasing (MSAA) defaults to more than XNA, leading to (sometimes significant) performance regressions
    - http://www.elvastower.com/forums/index.php?/topic/34518-unstable-monogame-fps-issue/page__view__findpost__p__264177
  - [x] Cannot open a route in Track Viewer
    - http://www.elvastower.com/forums/index.php?/topic/34485-cant-start-a-route-in-trackviewer/page__view__findpost__p__263837


pzgulyas:
Some years have passed since I wrote the code, but regarding the DummyVertexBuffer in RenderFrame.cs and Shapes.cs I can say, IIRC that it is there for substituting the InstanceBuffer in cases the SceneryShader is called without instancing. When the current shader code is called with unbound instance buffer, it leads to some crash, IIRC. An alternative to the dummy second binding could be to double the shader code so that one variation could be called in case of instanced objects with two vertex buffers bound, another variation for the uninstanced ones with one vertex buffer only.

twpol:
> Some years have passed since I wrote the code, but regarding the DummyVertexBuffer in RenderFrame.cs and Shapes.cs I can say, IIRC that it is there for substituting the InstanceBuffer in cases the SceneryShader is called without instancing. When the current shader code is called with unbound instance buffer, it leads to some crash, IIRC. An alternative to the dummy second binding could be to double the shader code so that one variation could be called in case of instanced objects with two vertex buffers bound, another variation for the uninstanced ones with one vertex buffer only.

Thanks, it does appear that it no longer lets us get away with calling a shader without some of its data (the instances in this case). It looks like a lot of effort to split the shader code (particularly with the shadow shader too) so I've kept the dummy vertex buffer.

pzgulyas:
Regarding the GraphicsProfile: I haven't migrated the vs_2_0 and ps_2_0 branch of the shaders in a hope that it might not be needed when MonoGame will become official. (And later forgot about that entirely.) But now I see, that there might still be a need for the corresponding vs_4_0_level_9_1 and ps_4_0_level_9_1 functions, see [this post](http://www.elvastower.com/forums/index.php?/topic/34344-immediate-benefit-from-monogame/page__view__findpost__p__262044). (Currently I don't have the development environment set up due to technical problems, sorry for that.)

twpol:
MonoGame provides access to profiles (HiDef and Reach) but does not provide the underlying DirectX feature level (FL) being used, which means we can't automatically support some scenarios.

We have two rendering quality levels currently:

* Low quality rendering (e.g. 1 shadow map) needing FL 9_1+
* High quality rendering (e.g. 4 shadow maps) needing FL 9_3+

However, the MonoGame profiles correspond to FL 9_1 (Reach) and FL 10_0 (HiDef), which means that any device which supports FL 9_3 but not FL 10_0 will not get the high quality rendering, even though it is supported.

I have chosen to address this with a hidden setting `DirectXFeatureLevel` (replacing the similar hidden setting `ShaderModel`) which defaults to `9_1` or `10_0` depending on MonoGame profile, but also supports being manually set to `9_3` if you are unfortunately on FL 9_3 hardware.

Hardware FL | MonoGame profile | DirectXFeatureLevel | Rendering
---|---|---|---
9_1 | Reach | _(not set)_ or `"9_1"` | Low
9_3 | Reach | _(not set)_ | Low
9_3 | Reach | `"9_3"` | High
10_0 | HiDef | _(not set)_ or `"10_0"` | High

Source/RunActivity/Viewer3D/Scenery.cs
```diff
@@ -524,8 +524,8 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
                 }
             }
 
-            // Model instancing requires shader model 3 or higher.
-            if (Viewer.Settings.ModelInstancing && Viewer.Settings.ShaderModel >= 3)
+            // Model instancing requires feature level 9_3 or higher.
+            if (Viewer.Settings.ModelInstancing && Viewer.Settings.IsDirectXFeatureLevelIncluded(ORTS.Settings.UserSettings.DirectXFeature.Level9_3))
```
pzgulyas, line 528: I am just heckling here :-) , but there graphics chips exist supporting 9_1 and 10_0, but not supporting 9_3. So while on these chips the 4 shadow maps rendering wouldn't work because of the inadequate number of registers, model instancing probably still would. But this is just a guess from my side, and since probably neither of us have a test computer for this specific case, it is not worth dealing with this. Cheers :-)

YoRyan:
James, first, thanks for all of your hard work on this PR. Sifting through all of those changes is no small feat! If this port is close to being ready, I was wondering if we might release it to the unstable branch so that we can review it, and players can test it? Then, we can merge in @Csantucci and @perpetualKid's changes piecemeal.

It's looking pretty good on my system except for distant scenery objects, which display some kind of flickering effect. Unfortunately, it's not captured well by screenshots.
MG flicker
[OpenRailsLog.txt](https://github.com/openrails/openrails/files/5357188/OpenRailsLog.txt)

Source/RunActivity/Viewer3D/Scenery.cs
twpol: If the hardware supports 10_0, we'll be using the HiDef profile — which requires 10_0+ — and this check will be true.

The check will be true because the function `IsDirectXFeatureLevelIncluded` is exploiting this quote from https://docs.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-devices-downlevel-intro:

> A feature level always includes the functionality of previous or lower feature levels.

So, if we know 10_0 is supported, we'll return `true` for 9_3 and 9_1 no matter what the hardware is.

If there is hardware that supports 10_0 and 9_1 but not 9_3, that's pretty confusing, but then information on DirectX feature levels in the 9_* range is spread thinly and sometimes confusingly. 😕

MonoGame is limiting us here, by only allowing us to check for the Reach and HiDef profiles – feature levels 9_1 and 10_0, respectively. If you are unfortunate enough to have 9_3 hardware, MonoGame will tell us it only supports Reach, which we must interpret as 9_1, so the `DirectXFeatureLevel` setting exists specifically for this case – by setting it to "9_3" you can gain access to our high-quality rendering (which only needs 9_3 hardware).

twpol:
> James, first, thanks for all of your hard work on this PR. Sifting through all of those changes is no small feat! If this port is close to being ready, I was wondering if we might release it to the unstable branch so that we can review it, and players can test it? Then, we can merge in @Csantucci and @perpetualKid's changes piecemeal.

Cheers, and I was coming to the same conclusion.

Having worked my way through the code from Peter's original version from December 2015 with MonoGame 3.5, a bunch of issues I encountered (such as window position, full-screen behaviour) were resolved with no work other than upgrading to MonoGame 3.7. So I feel it's better to let people re-find any remaining bugs rather than copying code unnecessarily from other forks.

Of course, for any issue we **do** find, those forks will be a vital source for fixes. 🙂

Source/RunActivity/Viewer3D/Scenery.cs
pzgulyas: Yeah, it is confusing indeed. There was a request about the Intel GMA 4500 MHD, which is exactly this type of chip, but we cannot test it anyway, so it is not worth making guesses.

YoRyan:
The flickering effect I mentioned is no longer displaying as of the latest unstable. That's another blocker out of the way.

twpol:
> The flickering effect I mentioned is no longer displaying as of the latest unstable. That's another blocker out of the way.

Ah, cool. It might have been a different artifact from the uninitialised shader variable that others saw as pixelation.

YoRyan:
With the Track Viewer fixes applied, I believe there are no more outstanding reported bugs. I vote we take this out of draft and merge.

cjakeman (commented):
James,

Just some thoughts:

I'm puzzled to see
using Microsoft.Xna.Framework;
is still present in a number of files and apparently the Color class is used from that.
(I don't know anything about graphics, but it seems odd.)

I also see lots of references to XNAViewMatrix and XNAProjectionMatrix. Presumably this is just a naming convention which is now a bit misleading.

Chris

YoRyan:
Nothing to worry about, Chris, since Monogame intentionally retains as many module names as possible. Microsoft.Xna.Framework is [part](https://docs.monogame.net/api/Microsoft.Xna.Framework.Color.html) of the official API.

Source/RunActivity/Viewer3D/MSTSSky.cs
```diff
@@ -652,58 +644,50 @@ public override void Render(GraphicsDevice graphicsDevice, IEnumerable<RenderIte
             MSTSSkyShader.WindSpeed = Viewer.World.MSTSSky.mstsskywindSpeed;
             MSTSSkyShader.WindDirection = Viewer.World.MSTSSky.mstsskywindDirection; // Keep setting this after Time and Windspeed. Calculating displacement here.
 
+            for (var i = 0; i < 5; i++)
+                graphicsDevice.SamplerStates[i] = SamplerState.LinearWrap;
```
YoRyan, line 648: How do we know there will always be 5 sampler states to set? (Not a graphics guy.)

Source/RunActivity/Viewer3D/Processes/RenderProcess.cs
```diff
@@ -132,20 +140,32 @@ internal void Start()
...
-                Game.Settings.ShaderModel = 3;
+            try
+            {
+                // Validate that the DirectX feature level is one we understand
+                Enum.Parse(typeof(ORTS.Settings.UserSettings.DirectXFeature), "Level" + Game.Settings.DirectXFeatureLevel);
+            }
+            catch (Exception)
+            {
+                Game.Settings.DirectXFeatureLevel = "";
+            }
```
YoRyan, line 151:
```suggestion
            // Validate that the DirectX feature level is one we understand
            if (!Enum.IsDefined(typeof(ORTS.Settings.UserSettings.DirectXFeature), "Level" + Game.Settings.DirectXFeatureLevel))
                Game.Settings.DirectXFeatureLevel = "";
```
Prefer Enum.IsDefined over an exception-swallowing try-catch.

Source/RunActivity/Viewer3D/MSTSSky.cs
pzgulyas: There are 5 sampler_states defined in the SkyShader.fx effect file, this is where 5 comes from.

YoRyan: Got it, thanks. Maybe consider annotating it `SKYSHADER_SAMPLER_STATES` for mortals like me?

Source/RunActivity/Viewer3D/DDSLib.cs
```diff
@@ -516,15 +507,15 @@ private static int MipMapSizeInBytes(int map, int width, int height, bool isComp
         }
 
         //Get the byte data from a mip-map level.
-        private static void GetMipMaps(int offsetInStream, int map, bool hasMipMaps, int width, int height, bool isCompressed, FourCC compressionFormat, int rgbBitCount, bool partOfCubeMap, BinaryReader reader, LoadSurfaceFormat loadSurfaceFormat, ref byte[] data, out int numBytes)
+        private static void GetMipMaps(int offsetInStream, int map, bool hasMipMaps,int width, int height, bool isCompressed, FourCC compressionFormat, int rgbBitCount, bool partOfCubeMap, BinaryReader reader,LoadSurfaceFormat loadSurfaceFormat, ref byte[] data, out int numBytes)
```
YoRyan, line 510:
```suggestion
        private static void GetMipMaps(int offsetInStream, int map, bool hasMipMaps, int width, int height, bool isCompressed, FourCC compressionFormat, int rgbBitCount, bool partOfCubeMap, BinaryReader reader, LoadSurfaceFormat loadSurfaceFormat, ref byte[] data, out int numBytes)
```
There are some strange whitespace changes in this file--like this one, where some spaces between arguments were removed.

Source/RunActivity/Viewer3D/DDSLib.cs
```diff
@@ -593,7 +584,7 @@ private static void GetMipMaps(int offsetInStream, int map, bool hasMipMaps, int
         }
 
         //Xna only supporst mip-map on textures with full chains == last-mip is 1x1
-        private static bool CheckFullMipChain(int width, int height, int numMip)
+        private static bool CheckFullMipChain(int width,int height,int numMip)
```
YoRyan, line 587:
```suggestion
        private static bool CheckFullMipChain(int width, int height, int numMip)
```
As above.

Source/RunActivity/Viewer3D/DDSLib.cs
```diff
@@ -915,11 +900,11 @@ private static TextureCube GenerateNewCubeTexture(LoadSurfaceFormat loadSurfaceF
         }
 
         //new 2d-map texture
-        private static Texture2D GenerateNewTexture2D(LoadSurfaceFormat loadSurfaceFormat, FourCC compressionFormat, GraphicsDevice device, int width, int height, int numMips, uint pixelFlags, int rgbBitCount)
+        private static Texture2D GenerateNewTexture2D(LoadSurfaceFormat loadSurfaceFormat, FourCC compressionFormat, GraphicsDevice device, int width,int height, bool hasMipMaps, uint pixelFlags, int rgbBitCount)
         {
             SurfaceFormat surfaceFormat = SurfaceFormatFromLoadFormat(loadSurfaceFormat, compressionFormat, pixelFlags, rgbBitCount);
 
-            Texture2D tx = new Texture2D(device, width, height, numMips, TextureUsage.None, surfaceFormat);
+			Texture2D tx = new Texture2D(device, width, height, hasMipMaps, surfaceFormat);
```
YoRyan, line 907:
```suggestion
             Texture2D tx = new Texture2D(device, width, height, numMips, TextureUsage.None, surfaceFormat);
```
And here.

YoRyan (commented):
I mean, for the most part, James, the code quality of this port looks very good. But I'm not sure how thoroughly you expect us to review such a technically complex PR? Graphics knowledge is in short supply on our team.

If you just want a sanity check on your C#, then any of us would do, but if you want somebody to make sure you've crossed your sampler states and dotted your DirectX feature levels, then my hunch is that only @pzgulyas is in a position to do so.

Source/RunActivity/Viewer3D/MSTSSky.cs
pzgulyas: It's a fair request. :-) In long term however a single sampler object would be enough, given that all of them are the same, but as of now the shaders syntax is kept in the now-deprecated "effect" style, where the sampler object is yet bound to the texture.

cjakeman:
> I mean, for the most part, James, the code quality of this port looks very good. But I'm not sure how thoroughly you expect us to review such a technically complex PR? Graphics knowledge is in short supply on our team.

I reckon that this has been in Unstable for long enough now to be reliable.

James, if you can accept Ryan's suggestions, then I'll approve this PR.

Best wishes,

Chris

Source/RunActivity/Viewer3D/DDSLib.cs
twpol: These were from the original changes, so may have been automatically applied by Visual Studio when Peter originally did the migration.

I've applied [dotnet-format](https://github.com/dotnet/format) to the whole file (which we should adopt as our standard code style) to fix all these cases.

twpol:
> I mean, for the most part, James, the code quality of this port looks very good. But I'm not sure how thoroughly you expect us to review such a technically complex PR? Graphics knowledge is in short supply on our team.
> 
> If you just want a sanity check on your C#, then any of us would do, but if you want somebody to make sure you've crossed your sampler states and dotted your DirectX feature levels, then my hunch is that only @pzgulyas is in a position to do so.

If @pzgulyas is willing to check any of the code, that would be fantastic, but having it be part of Unstable Version for weeks is as good a check as we need right now for graphical functionality.

The area we had trouble with the original PR was inclusion of not-really-migration changes, so that's worth checking — have I included anything completely extraneously?

YoRyan (approved):

YoRyan:
> The area we had trouble with the original PR was inclusion of not-really-migration changes, so that's worth checking — have I included anything completely extraneously?

None that I can see. This is an exceptionally clean port.

Chris, this PR has my seal of approval.

cjakeman:
> Chris, this PR has my seal of approval.

Thanks, Ryan.

Merging now . . .

YoRyan:
And with that, Monogame has landed! 🥳

🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

PR #231, opened 2020-07-12
Add each gear positions sound events

peternewell:
https://blueprints.launchpad.net/or/+spec/addtional-steam-locomotive-sound-triggers -  Add each gear positions sound events for diesel with multiple gear positions. Fix steam injector so that it only turns off when injector initially stopped.


YoRyan:
Oops! That looked like code at first glance.

PR #234, opened 2020-07-22
Rehabilitate unit testing suite, migrate to xUnit 2

YoRyan:
As discussed on [Elvas Tower](http://www.elvastower.com/forums/index.php?/topic/34219-xunit-test-issues/),

- Add the xunit.runner.visualstudio NuGet package so test discovery will work in modern versions of Visual Studio
- Delete the in-repo copy of xUnit, switch to NuGet for distribution
- Migrate from xUnit 1.9.2 to 2.4.1
- Resolve STFReader bugs spotlighted by the unit tests
- Update some other tests to reflect the current behavior of STFReader


twpol (dismissed):
I'm not 100% certain the changes to the STF tests are correct (in particular, the skip underscores - that **is** a thing that happens in specific cases in STF) but otherwise I'm happy.

YoRyan:
Thanks. For a given test, I wasn't sure whether the commented version, the un-commented version, or the current behavior of the STF reader was correct.

If you wish, I can investigate the STF reader changes that would be required to make the tests pass as written (warn on missing block after comment, warn on missing block after skip, skip underscored tokens).

cjakeman (commented):
Hi Ryan,

Did you discuss this with Rob Roeterdink? I was wondering why he didn't allow Tab as a delimiter from the start and he probably had a reason.

Best wishes,

Chris

YoRyan:
In [May 2017](http://www.elvastower.com/forums/index.php?/topic/30326-update-timetable-mode-signalling/), Rob added Tab as a valid separator character - this is reflected in our code, but not yet in the manual, nor in that unit test. I'm not sure why he did not allow Tab at first.

YoRyan:
It looks like one of the changes in this PR [broke](http://www.elvastower.com/forums/index.php?/topic/34187-or-consist-format/page__pid__261013__st__80#entry261013) certain Polish locomotives. I am investigating...

YoRyan:
Fixed by bc61523 and bbdede3. I had to remove the test that checks for an EOF before the document-ending whitespace.

twpol:
> If you wish, I can investigate the STF reader changes that would be required to make the tests pass as written (warn on missing block after comment, warn on missing block after skip, skip underscored tokens).

I wouldn't worry too much about it right now; changing the tests to pass (rather than the parser) is more likely to match reality, given the parser is battle-tested over many years.

We can come back later and improve these tests, especially as I've had the thought we may need to replace the parser with something closer to the JSON one to make supporting MSTS and OR formats for train cars...

YoRyan:
I have added a short tutorial on running the tests to the Developing Code webpage.

cjakeman (approved):
Hi Ryan,

Thanks for seeing this through - it's been quite a long haul but the system is better for it.

I've also pushed your change to www.openrails.org up to the site.

Approved with thanks,

Chris

YoRyan:
Thank you, Chris!

PR #240, opened 2020-07-26
Need to remove this request: Update ru.po

Weter-ORTS:
Couple of spell-corrections;
Pair of missed string-translations added.


Weter-ORTS:
This work was merged in last request, so this one is no longer needed.

Sharpe49:
Ok, done.

PR #246, opened 2020-07-31
Spanish manual

cjakeman:
https://blueprints.launchpad.net/or/+spec/non-english-documents
Adds a Spanish version of the Manual.pdf provided by Javier (Xavivilla) to the distribution.
Also extends the Menu.exe to automatically add any non-English documents into the Documents drop-down if that language has been selected.


YoRyan:
Hi Chris,

It looks like this PR was based off of the unstable branch, which means it includes the changes of every other PR in our queue. Could you try rebasing against master to get it down to a manageable size?

cjakeman:
Hi Ryan,

Thanks for catching this. Guess I was getting confused.

It was just a few files, so I'll try again.

Chris

PR #247, opened 2020-07-31
adds Manual.pdf in Spanish

cjakeman:
https://blueprints.launchpad.net/or/+spec/non-english-documents
Adds a Spanish version of the Manual.pdf provided by Javier (Xavivilla) to the distribution.
Also extends the Menu.exe to automatically add any non-English documents into the Documents drop-down if that language has been selected.

### DIFF
--- a/Source/Menu/MainForm.cs
+++ b/Source/Menu/MainForm.cs
@@ -231,21 +231,19 @@ namespace ORTS
                 var path = dir + @"\Documentation\";
                 if (Directory.Exists(path))
                 {
-                    foreach (string entry in Directory.GetFiles(path))
+                    // Load English documents
+                    LoadDocuments(docs, path, null);
+
+                    // Find any non-English documents by looking in \Documentation\<language code>\, e.g. \Documentation\es\
+                    foreach (var codePath in Directory.GetDirectories(path))
                     {
-                        // These are the following formats that can be selected.
-                        if (entry.Contains(".pdf") || entry.Contains(".doc") || entry.Contains(".docx") || entry.Contains(".pptx") || entry.Contains(".txt"))
-                        {
-                            var i = entry.LastIndexOf("\\");
-                            docs.Add(new ToolStripMenuItem(entry.Substring(i + 1), null, (Object sender2, EventArgs e2) =>
-                            {
-                                var docPath = (sender2 as ToolStripItem).Tag as string;
-                                Process.Start(docPath);
-                             }) { Tag = entry });
-                        }
-                        contextMenuStripDocuments.Items.AddRange((from tool in docs
-                                                                  orderby tool.Text
-                                                                  select tool).ToArray());
+                        // Extract the last folder in the path - the language code, e.g. "es"
+                        var i = codePath.LastIndexOf(@"\");
+                        var code = codePath.Substring(i + 1);
+
+                        // include any non-English documents that match the chosen language
+                        if (code == Settings.Language)
+                            LoadDocuments(docs, codePath, code);
                     }
                 }
                 else
@@ -261,6 +259,31 @@ namespace ORTS
             {
                 LoadFolderList();
                 Initialized = true;
+            }
+        }
+
+        private void LoadDocuments(List<ToolStripItem> docs, string folderPath, string code)
+        {
+            foreach (var filePath in Directory.GetFiles(folderPath))
+            {
+                var ext = System.IO.Path.GetExtension(filePath);
+                var name = System.IO.Path.GetFileNameWithoutExtension(filePath);
+                // These are the following formats that can be selected.
+                if (".pdf .doc .docx .pptx .txt".Contains(ext))
+                {
+                    var codeLabel = "";
+                    if (code != null)
+                        codeLabel = $" [{code}]";
+                    docs.Add(new ToolStripMenuItem($"{name}{ext}{codeLabel}", null, (Object sender2, EventArgs e2) =>
+                    {
+                        var docPath = (sender2 as ToolStripItem).Tag as string;
+                        Process.Start(docPath);
+                    })
+                    { Tag = filePath });
+                }
+                contextMenuStripDocuments.Items.AddRange((from tool in docs
+                                                          orderby tool.Text
+                                                          select tool).ToArray());
             }
         }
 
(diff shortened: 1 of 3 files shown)


pzgulyas:
Just a question: Why the pdf was added here? Shouldn't the editable source be added to the repository, with the pdf compilation done automatically by the build server?

cjakeman:
Hi Peter,

Good to hear from you.

The code changes are worth making as they apply to any non-English documents not just the manual.

My intention is to help Javier (Xavivilla) move his rst files into the solution on GitHub and to amend the build process to compile these into the Spanish PDF just like the English one. I don't have a timetable for that yet, so adding the PDF is just a stop-gap.

I included a comment about this in a README.TXT in the PR.

Best wishes,

Chris

Source/Menu/MainForm.cs
```diff
@@ -231,21 +231,19 @@ orderby tool.Text
                 var path = dir + @"\Documentation\";
                 if (Directory.Exists(path))
                 {
-                    foreach (string entry in Directory.GetFiles(path))
+                    // Load English documents
+                    LoadDocuments(docs, path, null);
```
YoRyan, line 235:
```suggestion
                    LoadDocuments(docs, path);
```
Simplify by making the `code` parameter optional.

Source/Menu/MainForm.cs
```diff
@@ -231,21 +231,19 @@ orderby tool.Text
...
-                                var docPath = (sender2 as ToolStripItem).Tag as string;
-                                Process.Start(docPath);
-                             }) { Tag = entry });
-                        }
-                        contextMenuStripDocuments.Items.AddRange((from tool in docs
-                                                                  orderby tool.Text
-                                                                  select tool).ToArray());
+                        // Extract the last folder in the path - the language code, e.g. "es"
+                        var i = codePath.LastIndexOf(@"\");
+                        var code = codePath.Substring(i + 1);
```
YoRyan, line 242:
```suggestion
                        var code = System.IO.Path.GetFileName(codePath);
```
Use System.IO.Path instead of manually substringing.

Source/Menu/MainForm.cs
```diff
@@ -264,6 +262,31 @@ orderby tool.Text
             }
         }
 
+        private void LoadDocuments(List<ToolStripItem> docs, string folderPath, string code)
```
YoRyan, line 265:
```suggestion
        private void LoadDocuments(List<ToolStripItem> docs, string folderPath, string code = "")
```
Simplify by making the `code` parameter optional.

Source/Menu/MainForm.cs
```diff
@@ -264,6 +262,31 @@ orderby tool.Text
...
+            foreach (var filePath in Directory.GetFiles(folderPath))
+            {
+                var ext = System.IO.Path.GetExtension(filePath);
+                var name = System.IO.Path.GetFileNameWithoutExtension(filePath);
+                // These are the following formats that can be selected.
+                if (".pdf .doc .docx .pptx .txt".Contains(ext))
+                {
+                    var codeLabel = "";
+                    if (code != null)
+                        codeLabel = $" [{code}]";
```
YoRyan, line 276:
```suggestion
                    var codeLabel = string.IsNullOrEmpty(code) ? "" : $" [{code}]";
```

Source/Menu/MainForm.cs
```diff
@@ -264,6 +262,31 @@ orderby tool.Text
...
+                    docs.Add(new ToolStripMenuItem($"{name}{ext}{codeLabel}", null, (Object sender2, EventArgs e2) =>
```
YoRyan, line 277:
```suggestion
                    docs.Add(new ToolStripMenuItem($"{name}{ext}{codeLabel}", null, (object sender2, EventArgs e2) =>
```
Visual Studio suggests using the lower-case name.

Source/Menu/MainForm.cs
```diff
@@ -264,6 +262,31 @@ orderby tool.Text
...
         }
 
+        private void LoadDocuments(List<ToolStripItem> docs, string folderPath, string code)
+        {
+            foreach (var filePath in Directory.GetFiles(folderPath))
+            {
+                var ext = System.IO.Path.GetExtension(filePath);
+                var name = System.IO.Path.GetFileNameWithoutExtension(filePath);
+                // These are the following formats that can be selected.
+                if (".pdf .doc .docx .pptx .txt".Contains(ext))
```
YoRyan, line 272:
```suggestion
                if (new[] { ".pdf", ".doc", ".docx", ".pptx", ".txt" }.Contains(ext.ToLowerInvariant()))
```
You need to test for the presence of the entire extension here; your current check passes partial matches such as ".p"  (.pdf) and ".tx" (.txt).

YoRyan (changes requested):
Hi Chris,

Thanks for the code. Just have a few refactoring suggestions and a bug to fix.

Source/Menu/MainForm.cs
cjakeman: 3 lines collapse to 1 but is it easier to read?

Source/Menu/MainForm.cs
cjakeman: Good idea

Source/Menu/MainForm.cs
cjakeman: Good catch; definitely a bug.

Source/Menu/MainForm.cs
cjakeman: Didn't realise GetFileName could also get you the final folder from a sequence of folders. Neat.

cjakeman:
Hi Ryan,

Ready for your approval.

Chris

YoRyan (approved):
Right, then. Approved and merged.

twpol:
I'm afraid this has resulted in two problems:
* Unstable Version is now more than 4x its original size:
  image
* In the Unstable Version, the "Documents" menu is enabled but empty for everyone but Spanish users

In both cases the fix is to move the copying of the PDF into `Build.cmd` along with the other documentation work, so that it is not part of the Unstable Version.

We also should be avoiding adding large binary files to the repository, because they make the `.git` folder bigger for everyone, forever.

cjakeman:
Hi James,

Sorry about that. I did it that way because I had the PDF and not the source files and wanted to try out the code change to access the PDF in the Unstable version. I should have used a dummy PDF to avoid bloating the repo.

I've just contacted Xavavilla to get the Spanish text and images so that I can generate a PDF as part of the build.

Thanks for sorting this,

Chris

PR #252, opened 2020-08-08
Script compilation with Roslyn in order to support C# v7

Sharpe49:
https://blueprints.launchpad.net/or/+spec/script-csharp7

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
+++ b/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
@@ -15,7 +15,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
 
-using Microsoft.CSharp;
+using Microsoft.CodeDom.Providers.DotNetCompilerPlatform;
 using Orts.Simulation;
 using ORTS.Common;
 using System;
@@ -34,7 +34,8 @@ namespace Orts.Common.Scripting
     {
         readonly Simulator Simulator;
         readonly Dictionary<string, Assembly> Scripts = new Dictionary<string, Assembly>();
-        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider();
+        static readonly ProviderOptions ProviderOptions = new ProviderOptions(Path.Combine(new Uri(Path.GetDirectoryName(Assembly.GetExecutingAssembly().CodeBase)).LocalPath, "roslyn", "csc.exe"), 10);
+        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider(ProviderOptions);
 
         static CompilerParameters GetCompilerParameters()
         {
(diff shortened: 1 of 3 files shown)


Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -34,7 +34,8 @@ public class ScriptManager
     {
         readonly Simulator Simulator;
         readonly Dictionary<string, Assembly> Scripts = new Dictionary<string, Assembly>();
-        static readonly CSharpCodeProvider Compiler = new CSharpCodeProvider();
+        static readonly ProviderOptions ProviderOptions = new ProviderOptions("roslyn/csc.exe", 10);
```
YoRyan, line 37: This, to me, is a blocker - you're assuming that the working directory will always be the directory with RunActivity.exe, but this will not always be the case, particularly if you're running OR from the command line or through a debugger.

You should explicitly request the full path to the Open Rails executable. For reference, [this](https://github.com/openrails/openrails/blob/4d61887252746f5adc2714d9f7950d7a4fe8a097/Source/RunActivity/Viewer3D/Processes/Game.cs#L95) is how it's done for the Content\ directory. Since the Orts.Simulation project lacks a reference to Windows.Forms, I suggest storing this in that very same `Game()` constructor and passing it to the `Simulator` class, but it's up to you.

Source/Orts.Simulation/Orts.Simulation.csproj
```diff
@@ -42,6 +44,9 @@
       <HintPath>..\3rdPartyLibs\GNU.Gettext.dll</HintPath>
     </Reference>
     <Reference Include="Microsoft.Xna.Framework, Version=3.1.0.0, Culture=neutral, PublicKeyToken=6d5c3888ef60e27d, processorArchitecture=x86" />
+    <Reference Include="Microsoft.CodeDom.Providers.DotNetCompilerPlatform, Version=3.5.0.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
+      <HintPath>..\packages\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.3.5.0-preview1\lib\net45\Microsoft.CodeDom.Providers.DotNetCompilerPlatform.dll</HintPath>
```
YoRyan, line 48: CodeDom 3.6.0 is out and it's a non-preview version, so can we bump this package up to there?

YoRyan:
Thanks for looking into this, by the way! As a fellow TCS script writer, I would very much appreciate an upgrade to C# 7.

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
Sharpe49: Alright, this is fixed with the new commit.

Source/Orts.Simulation/Orts.Simulation.csproj
Sharpe49: Alright, it has been updated.

By the way, my title is wrong : we are in fact supporting C# v8. ;)

Sharpe49:
@YoRyan For your information, when a pull request is linked to a blueprint, you have to wait for the blueprint approval.

@cjakeman Can you take a look at this blueprint?
https://blueprints.launchpad.net/or/+spec/script-csharp7
Thank you.

YoRyan:
My apologies; I'm still getting used to our development workflow. Who has the authority to approve blueprints? The management team?

Sharpe49:
Exactly.

PR #248, opened 2020-07-31
Russian translation

Weter-ORTS:
Translated almost all, that I knew.

PR #253, opened 2020-08-10
Fix translations not displaying for the HUD Dispatcher Information tab: https://bugs.launchpad.net/or/+bug/1891066

YoRyan:
The Dispatcher Information header, which displays the number of active trains, is not being properly translated into other languages. Discovered by @Weter-ORTS during his work on the Russian translation (#248).


YoRyan:
I don't like to implicitly cast things to string, but that's just me.

Sharpe49:
Alright, I understand now.

PR #257, opened 2020-08-19
Missing link restored

cjakeman:
Extended HUD - Manual Mode
http://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/


cjakeman:
Hi Ryan,

Yes, please. This is a very small correction but no more in the pipeline just now.

Thanks,

PR #259, opened 2020-08-22
Fix duplicate couplers upon resuming from save: https://bugs.launchpad.net/or/+bug/1892403

YoRyan:
The `Couplers` list is already populated when reading the save stream, so the `.Add()` calls in the current code add duplicate couplers each time the game is restored from a save. Fix by reassigning the entire list instead.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
@@ -1527,6 +1527,14 @@ namespace Orts.Simulation.RollingStocks
             AuxPowerOn = true;
         }
 
+        /// <summary>
+        /// Read the coupler state(s) from a save stream.
+        /// </summary>
+        /// <remarks>
+        /// Has no side effects besides advancing the save stream, thus avoiding any shared-state pitfalls.
+        /// </remarks>
+        /// <param name="inf">The save stream.</param>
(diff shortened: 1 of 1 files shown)


cjakeman (commented):
Hi Ryan,

I can see this starts a new list and does the job just fine, but I can't see how the yield return construction is any better than the original. It doesn't look any faster (not actually timed it) and it's not so easy to read.

How about a comment in the code to explain why it's done that way?

Anyway, I'll approve after 7 days as usual.

Chris

YoRyan:
Hi Chris,

I have added my rationale in the form of XML documentation.

```
        /// <summary>
        /// Read the coupler state(s) from a save stream.
        /// </summary>
        /// <remarks>
        /// Has no side effects besides advancing the save stream, thus avoiding any shared-state pitfalls.
        /// </remarks>
        /// <param name="inf">The save stream.</param>
        /// <returns>A list of newly restored <see cref="MSTSCoupling"/> instances.</returns>
```

By moving that logic to a separate, static method, we minimize the scope involved and any potential side effects. In my opinion, this makes the code safer to use and easier to reuse.

I haven't tested performance either, but I wouldn't expect any improvement over the current code. On the other hand, I would expect it to slightly outperform Rob Roeterlink's proposal to use `Couplers.Clear()`, since that tends to be a linear-time operation.

cjakeman (approved):
Hi Ryan,

Thanks for responding. Approved.

Chris

PR #262, opened 2020-08-25
Manual Braking

peternewell:
Initial release of Manual braking - https://blueprints.launchpad.net/or/+spec/manual-braking

### DIFF
--- a/HUDWindow.cs
+++ b/HUDWindow.cs
@@ -1,0 +1,1400 @@
+﻿// COPYRIGHT 2011, 2012, 2013 by the Open Rails project.
+//
+// This file is part of Open Rails.
+//
+// Open Rails is free software: you can redistribute it and/or modify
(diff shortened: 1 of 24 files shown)


cjakeman (commented):
Hi Peter,

Seems a good start and a useful extension. Lots of support on the forum too.

Would it help to use the Microsoft.Xna.Framework includes a MathHelper class which provides:

  public static float Clamp(float value, float min, float max);


Apart from the Manual, what else is needed for this to become operational?

Chris

peternewell:
Hi Chris,

Thanks for the feedback.

I am familiar with the Clamp function and will have a look at it.

This patch is operational as it is. I will probably be doing some tweaking over time on it as well, based on testing feedback.

As you suggest some notes will be required for the Manual. I will try and add some stuff early next week for this.

peternewell:
PR closed to fix some commit errors.

PR #264, opened 2020-08-29
add temporary fix

cjakeman:
Copy of Carlo's fix for OpenRails_MG_NoWD
In both cases this is temporary as the root cause is not yet known.
Forum thread at http://www.elvastower.com/forums/index.php?/topic/34297-mg71-crash-index-was-out-of-range-must-be-non-negative-and-less-than-the-size-of-the-collection/
Carlo's fix at https://github.com/Csantucci/openrails/commit/540fab09fb6be37ec490461b5974606b13165e3e

PLEASE DO NOT APPROVE THIS PR - it's fine for the Unstable version, but we need a solution to the root cause before making a change to the Testing version.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -281,10 +281,20 @@ namespace Orts.Simulation.Physics
                 float sectionStart = -offset;
 
                 // Get all track circuit sections in front of the train
-                List<int> sectionIndexes = signalRef.ScanRoute(this, position.TCSectionIndex, position.TCOffset,
-                        position.TCDirection, true, -1, true, false,
-                        false, false, true, false, false, false, false, IsFreight);
-                List<TrackCircuitSection> sections = new List<TrackCircuitSection>();
+                //List<int> sectionIndexes = signalRef.ScanRoute(this, position.TCSectionIndex, position.TCOffset,
+                //        position.TCDirection, true, -1, true, false,
+                //        false, false, true, false, false, false, false, IsFreight);
+                //List<TrackCircuitSection> sections = new List<TrackCircuitSection>();
+
+                var sectionIndexes = new List<int>();
+                var sections = new List<TrackCircuitSection>();
+                if (position.TCSectionIndex >= 0) // Temporary fix for TCSectionIndex == -1 but don't know why that occurs.
+                {
+                    // Get all track circuit sections in front of the train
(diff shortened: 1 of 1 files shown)


YoRyan:
Hi Chris,

For myself (and anyone else curious about this issue), could you link to the relevant bug report or forum thread?

Thanks.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -281,10 +281,20 @@ public float CheckDistanceM
...
-                List<int> sectionIndexes = signalRef.ScanRoute(this, position.TCSectionIndex, position.TCOffset,
-                        position.TCDirection, true, -1, true, false,
-                        false, false, true, false, false, false, false, IsFreight);
-                List<TrackCircuitSection> sections = new List<TrackCircuitSection>();
+                //List<int> sectionIndexes = signalRef.ScanRoute(this, position.TCSectionIndex, position.TCOffset,
+                //        position.TCDirection, true, -1, true, false,
+                //        false, false, true, false, false, false, false, IsFreight);
+                //List<TrackCircuitSection> sections = new List<TrackCircuitSection>();
+
+                var sectionIndexes = new List<int>();
```
pzgulyas, line 289: In this line one additional new list is created even if the error doesn't occure. The proper method would be just to declare here and not assign.

cjakeman: Hi Peter,

Thanks for looking at this. You are quite correct - that an empty list is created and is then not used.

It's a practice that my team has adopted so that we never have to check for lists which are null. Instead they are initially empty.

The developers do less work in coding and debugging and the computer does a bit more.

Best wishes,

Chris

pzgulyas:
Chris, the new empty list can be created in the "else" section to avoid the null-check. But creating 1 additional new list when the error doesn't occur (probably 99% of the cases) just spins the garbage collector, which in a video game causes micro-stuttering. Creating new non-elementary objects in the Update() and Render() loops must be avoided as much as possible.

Source/Orts.Simulation/Simulation/Physics/Train.cs
pzgulyas: Chris, the new empty list can be created in the "else" section to avoid the null-check. But creating 1 additional new list when the error doesn't occur (probably 99% of the cases) just spins the garbage collector, which in a video game causes micro-stuttering. Creating new non-elementary objects in the Update() and Render() loops must be avoided as much as possible.

Sharpe49:
This pull request is no longer needed with PR #270.

PR #265, opened 2020-08-29
Corrections & additions to Russian translation

Weter-ORTS:
Reworked conception of  key inputs arrangement in "Common" module;
Some more strings translated, some-corrected.


Weter-ORTS:
Thanks, @YoRyan I've visited Coal to Newcastle website and found some answers to my questions about unclear HUD strings there.

PR #268, opened 2020-09-02
Added Manual Brakes

peternewell:
https://blueprints.launchpad.net/or/+spec/manual-braking - added initial functionality


cjakeman (approved):
Hi Peter,

Thanks for updating the manual and looking at Math.Clamp().

I'm not sure why you weren't able to add changes to #262, but this looks fine.

Approved with thanks,

Chris

PR #270, opened 2020-09-03
Revert and new fix for https://bugs.launchpad.net/or/+bug/1613304

Sharpe49:


Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -244,82 +244,9 @@ public enum TRAINTYPE
         public float allowedAbsoluteMaxSpeedSignalMpS;   // Max speed as set by signal independently from train features
         public float allowedAbsoluteMaxSpeedLimitMpS;    // Max speed as set by limit independently from train features
         public float allowedAbsoluteMaxTempSpeedLimitMpS;    // Max speed as set by temp speed limit independently from train features
+        public float maxTimeS = 120;                     // check ahead for distance covered in 2 mins.
+        public float minCheckDistanceM = 5000;           // minimum distance to check ahead
```
YoRyan, line 248:
```suggestion
        public float maxTimeS { get; } = 120;       // check ahead for distance covered in 2 mins.
        public float minCheckDistanceM { get; } = 5000;  // minimum distance to check ahead
```
Use pure getters to ensure these constants don't get changed.

I *also* suggest renaming "maxTimeS" to "MaxCheckTimeS" and "minCheckDistanceM" to "MinCheckDistanceM". If you agree, use VIsual Studio's rename tool to make the change.

YoRyan (changes requested):
I cannot speak to the correctness of the signal logic, but code-wise, these changes look fine to me. I just have one style suggestion.

Roeterdink (approved):

Source/Orts.Simulation/Simulation/Physics/Train.cs
Sharpe49: This is in fact the same orthograph as before the first fix. You should really look only at the second commit to see what's been changed.

YoRyan: I see. Forget about it, then.

YoRyan (approved):

PR #272, opened 2020-09-05
build: Add index to PDFs

twpol:
It turns out that we aren't generating the index for PDF files, so I've fixed that.

I've also removed the explicit index/search entries in the TOC from PDFs, as neither of them was correct for PDFs (they're good for HTML though).


YoRyan (approved):
Thanks, James. This change looks good to me, although it seems to introduce a Perl dependency.

PR #274, opened 2020-09-07
Bug fix for https://bugs.launchpad.net/or/+bug/1893565 Pantograph control sequence does not work correctly

Csantucci:
See also http://www.elvastower.com/forums/index.php?/topic/33763-pantograph-control-sequence-does-not-work-correctly/page__view__findpost__p__262364 and following posts


YoRyan (approved):
Thanks, Carlo. This looks good to me.

PR #278, opened 2020-09-15
Bug fix for https://bugs.launchpad.net/or/+bug/1895692 False done indication on drop cars at location; by edwardk

Csantucci:
See also http://www.elvastower.com/forums/index.php?/topic/34402-issue-with-dropoffwagonsatlocation-event/page__pid__262879#entry262879

### DIFF
--- a/Source/Orts.Simulation/Simulation/Activity.cs
+++ b/Source/Orts.Simulation/Simulation/Activity.cs
@@ -1388,7 +1388,6 @@ namespace Orts.Simulation
         {
             foreach (var trainItem in Simulator.Trains)
             {
-                //bool lEngine = false;
                 int nCars = 0;//all cars other than WagonIdList.
                 int nWagonListCars = 0;//individual wagon drop.
                 foreach (var item in trainItem.Cars)
--- a/Source/Orts.Simulation/Simulation/Activity.cs
+++ b/Source/Orts.Simulation/Simulation/Activity.cs
@@ -1321,9 +1321,8 @@ namespace Orts.Simulation
                 case EventType.AssembleTrainAtLocation:
                     if (atSiding(OriginalPlayerTrain.FrontTDBTraveller, OriginalPlayerTrain.RearTDBTraveller, this.SidingEnd1, this.SidingEnd2))
                     {
-                        consistTrain = null;
                         consistTrain = matchesConsist(ChangeWagonIdList);
-                        triggered = (consistTrain != null ? true : false);
+                        triggered = consistTrain != null;
                     }
                     break;
                 case EventType.DropOffWagonsAtLocation:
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1334,7 +1334,7 @@ override public Boolean Triggered(Activity activity)
                     {
                         consistTrain = null;
```
YoRyan, line 1335: Side note, can't we delete this line? It's a redundant assignment.

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1334,7 +1334,7 @@ override public Boolean Triggered(Activity activity)
...
                         consistTrain = matchesConsistNoOrder(ChangeWagonIdList);
-                        triggered = (consistTrain != null ? false : true);
+                        triggered = (consistTrain != null ? true : false);
```
YoRyan, line 1337: Can be simplified to:
```suggestion
                        triggered = consistTrain != null;
```

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1388,12 +1388,11 @@ private Train matchesConsistNoOrder(List<string> wagonIdList)
         {
             foreach (var trainItem in Simulator.Trains)
             {
-                bool lEngine = false;
+                //bool lEngine = false;
```
YoRyan, line 1391: It's bad form to litter the source code with commented code sections. Just delete it.
```suggestion
```

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1438,7 +1435,19 @@ static bool includesWagons(Train train, List<string> wagonIdList)
...
+            {
+                var car = train.Cars[i];
+                if (car.Train.TrainType == Train.TRAINTYPE.STATIC)
+                {
+                    lNotFound = true;
+                    return lNotFound;
+                }
+
+            }
+
```
YoRyan, line 1450: Correct me if I'm wrong, but couldn't this simplify to:
```suggestion
            // The Cars list is a global list that includes STATIC cars.  We need to make sure that the active train/car is processed only.
            if (train.TrainType == Train.TRAINTYPE.STATIC)
                return true;

            bool lNotFound = false;
```

Source/Orts.Simulation/Simulation/Activity.cs
Csantucci: OK, line removed together with a similar nearby one.

YoRyan (approved):
Thanks, Carlo. This change now looks good to me.

PR #281, opened 2020-09-26
fix: Allow building the manual without the rest of Open Rails

twpol:
This is a slight change which allows you to compile the manual (HTML or PDF) without compiling the rest of Open Rails, which saves on a lot of time and dependencies (or messing with configuration files).


YoRyan (approved):
Thanks, James. This change is quite helpful.

PR #283, opened 2020-09-28
Implement static features-only mode for the Track Monitor: https://blueprints.launchpad.net/or/+spec/static-track-monitor

YoRyan:
https://blueprints.launchpad.net/or/+spec/static-track-monitor
http://www.elvastower.com/forums/index.php?/topic/34424-removing-signal-aspects-from-the-track-monitor/

This mode is toggleable with Shift+F4. It hides all diverging switch icons and conceals the signal aspects with generic signal images. I am not entirely sold on the grey blob; perhaps a traffic light image, such as Coolhand101 suggested, would look more appealing.


YoRyan:
Manual documentation has been added. Taking this out of draft.

Sharpe49 (commented):
I think you should add an option in the menu in order to set a default state of the track monitor.

YoRyan:
What I'd like to do is allow RunActivity.exe to save and restore certain options. That way, once you've turned this on, you won't have to do so again for the next session. I can think of a few other settings that should be saved and restored in this way, such as cab pillarboxing and the siding and platform label toggles.

But I'd like to get this approved and merged first.

cjakeman (approved):
Hi Ryan,

All looks very good. Well done.

Approved with thanks.

PR #282, opened 2020-09-28
Timetable trains shown on a map

cjakeman:
https://blueprints.launchpad.net/or/+spec/timetable-window

Adds a tab to the map window which shows the timetabled trains with their HUD values and signals info. Useful for players and timetable creators. Prototype tested by Rick Loader on a very large timetable.

Manual updated (7.5.2 Timetable Tab)

The code piggy-backs off the dispatch window. I've tried hard to leave as much of that code unchanged as possible since it has lots of functions including multi-player that I am not competent to test thoroughly.

The graphics look a bit better because:
- anti-aliasing is used
- platforms are visible
- locos are shown in a different colour
- the lead vehicle is a different colour so the direction of a train is obvious
- objects and their labels can be shown or hidden

PR #284, opened 2020-09-30
Correction to "Timetable trains shown on a map #282"

cjakeman:
See bug reported at http://www.elvastower.com/forums/index.php?/topic/34431-timetable-map-window/


cjakeman:
These changes were based on the Unstable branch, which is not recommended.
They are being replaced by a commit to the previous Pull Request 282

PR #285, opened 2020-10-10
changes for new web-server

cjakeman:
Had to bypass file dates as now too new.
Made session folder path generic.
Also disabled tracking of visits.


YoRyan:
Since you've already made these changes to get the site moved onto the new server, I suppose we have little choice but to approve. 😅

Note that if we want to make the move to GitHub Pages, our end goal should be to replace PHP with Jekyll.

YoRyan (approved):

PR #291, opened 2020-10-16
Expand to max 48 TCS commands

Csantucci:
https://trello.com/c/Q6m2nqhn/481-expand-amount-of-tcs-commands
Needed for complex TCSs.


YoRyan (approved):
Simple enough. Any reason for settling on 48 commands?

Csantucci:
Thanks Ryan.
I have set the Command number to the same value of the Control (Command and display) number, as they partially use the same code, and 48 at the moment should be enough for the needs. However it may be that this could be extended later on 64, but that would again be a small job to do.

PR #293, opened 2020-10-20
Further TCS handles to get signal data and to set sound events

Csantucci:
https://trello.com/c/S38d2RoV/482-further-tcs-handles-related-to-signals-and-sound-events
These handles allow to get in a single struct some significant data for any next signal; they allow also to set any sound event and to get the actual direction of the train, that could not coincide with the direction of the player locomotive.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -128,6 +128,11 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<int, float> NextPostDistanceM;
         /// <summary>
+        /// Distance to end of authority.
+        /// int: direction; 0: forwards; 1: backwards
+        /// </summary>
+        public Func<int, float> EOADistanceM;
+        /// <summary>
         /// Train's length
         /// </summary>
         public Func<float> TrainLengthM;
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -587,6 +587,6 @@ namespace ORTS.Scripting.Api
         public string MainHeadSignalTypeName;
         public Aspect Aspect;
         public float DistanceM;
-        public float SpeedLimit;
+        public float SpeedLimitM;
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -105,7 +105,7 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<string, float> NextGenericSignalDistanceM;
         /// <summary>
-        /// Features of next generic signal. Not for NORMAL signals
+        /// Features of next generic signal. 
         /// string: signal type (DISTANCE etc.)
         /// int: position of signal in the signal sequence along the train route, starting from train front; 0 for first signal;
         /// float: max testing distance
@@ -587,5 +587,6 @@ namespace ORTS.Scripting.Api
         public string MainHeadSignalTypeName;
         public Aspect Aspect;
         public float DistanceM;
+        public float SpeedLimit;
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -18,6 +18,7 @@
 using System;
 using System.IO;
 using ORTS.Common;
+using Orts.Common;
 using Orts.Simulation.RollingStocks;
 
 namespace ORTS.Scripting.Api
@@ -79,6 +80,11 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<bool> DoesNextNormalSignalHaveTwoAspects;
         /// <summary>
+        /// Name of Head 0 of next normal signal.
+        /// int: position of signal in the signal sequence along the train route, starting from train front; 0 for first signal;
+        /// </summary>
+        public Func<int, string> NextNormalSignalMainHeadSignalType;
+        /// <summary>
         /// Aspect of the next DISTANCE signal.
         /// </summary>
         public Func<Aspect> NextDistanceSignalAspect;
@@ -87,17 +93,24 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> NextDistanceSignalDistanceM;
         /// <summary>
-        /// Signal type of main head of hext generic signal.
+        /// Signal type of main head of hext generic signal. Not for NORMAL signals
         /// </summary>
         public Func<string, string> NextGenericSignalMainHeadSignalType;
         /// <summary>
-        /// Aspect of the next generic signal.
+        /// Aspect of the next generic signal. Not for NORMAL signals
         /// </summary>
         public Func<string, Aspect> NextGenericSignalAspect;
         /// <summary>
-        /// Distance to next generic signal.
+        /// Distance to next generic signal. Not for NORMAL signals
         /// </summary>
         public Func<string, float> NextGenericSignalDistanceM;
+        /// <summary>
+        /// Features of next generic signal. Not for NORMAL signals
+        /// string: signal type (DISTANCE etc.)
+        /// int: position of signal in the signal sequence along the train route, starting from train front; 0 for first signal;
+        /// float: max testing distance
+        /// </summary>
+        public Func<string, int, float, SignalFeatures> NextGenericSignalFeatures;
         /// <summary>
         /// Next normal signal has a repeater head
         /// </summary>
@@ -123,21 +136,25 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> SpeedMpS;
         /// <summary>
-        /// Train's direction.
+        /// Locomotive direction.
         /// </summary>
         public Func<Direction> CurrentDirection;
         /// <summary>
-        /// True if train direction is forward.
+        /// True if locomotive direction is forward.
         /// </summary>
         public Func<bool> IsDirectionForward;
         /// <summary>
-        /// True if train direction is neutral.
+        /// True if locomotive direction is neutral.
         /// </summary>
         public Func<bool> IsDirectionNeutral;
         /// <summary>
-        /// True if train direction is reverse.
+        /// True if locomotive direction is reverse.
         /// </summary>
         public Func<bool> IsDirectionReverse;
+        /// <summary>
+        /// Train direction.
+        /// </summary>
+        public Func<Direction> CurrentTrainMUDirection;
         /// <summary>
         /// True if train brake controller is in emergency position, otherwise false.
         /// </summary>
@@ -322,6 +339,10 @@ namespace ORTS.Scripting.Api
         /// Trigger Deactivate sound event
         /// </summary>
         public Action TriggerSoundSystemDeactivate;
+        /// <summary>
+        /// Trigger generic sound event
+        /// </summary>
+        public Action<Event> TriggerGenericSound;
         /// <summary>
         /// Set ALERTER_DISPLAY cabcontrol display's alarm state on or off.
         /// </summary>
@@ -559,5 +580,12 @@ namespace ORTS.Scripting.Api
         /// Red color. Train control system intervention speed. Computer has to apply full service or emergency brake to maintain speed restriction.
         /// </summary>
         Intervention,
+    }
+
+    public struct SignalFeatures
+    {
+        public string MainHeadSignalTypeName;
+        public Aspect Aspect;
+        public float DistanceM;
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -587,6 +587,6 @@ namespace ORTS.Scripting.Api
         public string MainHeadSignalTypeName;
         public Aspect Aspect;
         public float DistanceM;
-        public float SpeedLimitM;
+        public float SpeedLimitMpS;
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -80,11 +80,6 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<bool> DoesNextNormalSignalHaveTwoAspects;
         /// <summary>
-        /// Name of Head 0 of next normal signal.
-        /// int: position of signal in the signal sequence along the train route, starting from train front; 0 for first signal;
-        /// </summary>
-        public Func<int, string> NextNormalSignalMainHeadSignalType;
-        /// <summary>
         /// Aspect of the next DISTANCE signal.
         /// </summary>
         public Func<Aspect> NextDistanceSignalAspect;
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -593,13 +593,15 @@ namespace ORTS.Scripting.Api
         public readonly Aspect Aspect;
         public readonly float DistanceM;
         public readonly float SpeedLimitMpS;
+        public readonly float AltitudeOrLengthM;
 
-        public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distanceM, float speedLimitMpS)
+        public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distanceM, float speedLimitMpS, float altitudeOrLengthM)
         {
             MainHeadSignalTypeName = mainHeadSignalTypeName;
             Aspect = aspect;
             DistanceM = distanceM;
             SpeedLimitMpS = speedLimitMpS;
+            AltitudeOrLengthM = altitudeOrLengthM;
         }
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -593,15 +593,15 @@ namespace ORTS.Scripting.Api
         public readonly Aspect Aspect;
         public readonly float DistanceM;
         public readonly float SpeedLimitMpS;
-        public readonly float AltitudeOrLengthM;
+        public readonly float AltitudeM;
 
-        public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distanceM, float speedLimitMpS, float altitudeOrLengthM)
+        public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distanceM, float speedLimitMpS, float altitudeM)
         {
             MainHeadSignalTypeName = mainHeadSignalTypeName;
             Aspect = aspect;
             DistanceM = distanceM;
             SpeedLimitMpS = speedLimitMpS;
-            AltitudeOrLengthM = altitudeOrLengthM;
+            AltitudeM = altitudeM;
         }
     }
 }
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -584,9 +584,17 @@ namespace ORTS.Scripting.Api
 
     public struct SignalFeatures
     {
-        public string MainHeadSignalTypeName;
-        public Aspect Aspect;
-        public float DistanceM;
-        public float SpeedLimitMpS;
+        public readonly string MainHeadSignalTypeName;
+        public readonly Aspect Aspect;
+        public readonly float DistanceM;
+        public readonly float SpeedLimitMpS;
+
+        public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distanceM, float speedLimitMpS)
+        {
+            MainHeadSignalTypeName = mainHeadSignalTypeName;
+            Aspect = aspect;
+            DistanceM = distanceM;
+            SpeedLimitMpS = speedLimitMpS;
+        }
     }
 }
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -1652,7 +1652,7 @@ namespace Orts.Simulation.Physics
 
             // prepare train data for Train Control System
             if (IsActualPlayerTrain)
-                UpdateTrainData();
+                UpdatePlayerTrainData();
 
             // log train details
 
@@ -13990,51 +13990,78 @@ namespace Orts.Simulation.Physics
         /// Generation of Train data useful for TCS and TrackMonitor; player train only
         /// <\summary>
 
-        public class TrainSignal
-        {
-            public float DistanceToTrainM;
-            public SignalObject ObjectDetails;
+        // Contains data about all types of signals
+        public List<TrainObjectItem>[,] PlayerTrainSignals; // first index 0 forward, 1 backward; second index signal type (NORMAL etc.)
 
-            public TrainSignal(float distanceToTrainM, SignalObject objectDetails)
+        public List<TrainObjectItem>[] PlayerTrainSpeedposts; // 0 forward, 1 backward
+        public List<TrainObjectItem>[,] PlayerTrainDivergingSwitches; // 0 forward, 1 backward; second index 0 facing, 1 trailing
+        public List<TrainObjectItem>[] PlayerTrainMileposts; // 0 forward, 1 backward
+        public List<TrainObjectItem>[] PlayerTrainTunnels; // 0 forward, 1 backward
+
+
+        //================================================================================================//
+        /// <summary>
+        /// Initializes train data for TCS and TrackMonitor
+        /// </summary>
+        /// 
+        public void InitializePlayerTrainData()
+        {
+            if (PlayerTrainSignals == null)
             {
-                DistanceToTrainM = distanceToTrainM;
-                ObjectDetails = objectDetails;
+                PlayerTrainSignals = new List<TrainObjectItem>[2, signalRef.ORTSSignalTypeCount];
+                PlayerTrainSpeedposts = new List<TrainObjectItem>[2];
+                PlayerTrainDivergingSwitches = new List<TrainObjectItem>[2, 2];
+                PlayerTrainMileposts = new List<TrainObjectItem>[2];
+                PlayerTrainTunnels = new List<TrainObjectItem>[2];
+                for (int dir = 0; dir < 2; dir++)
+                {
+                    for (int fn_type = 0; fn_type < signalRef.ORTSSignalTypeCount; fn_type++)
+                        PlayerTrainSignals[dir, fn_type] = new List<TrainObjectItem>();
+                    PlayerTrainSpeedposts[dir] = new List<TrainObjectItem>();
+                    for (int i = 0; i < 2; i++)
+                        PlayerTrainDivergingSwitches[dir, i] = new List<TrainObjectItem>();
+                    PlayerTrainMileposts[dir] = new List<TrainObjectItem>();
+                    PlayerTrainTunnels[dir] = new List<TrainObjectItem>();
+                }
+            }
+            else
+            {
+                foreach (var playerTrainSignalList in PlayerTrainSignals)
+                    playerTrainSignalList?.Clear();
+                foreach (var playerTrainSpeedpostList in PlayerTrainSpeedposts)
+                    playerTrainSpeedpostList?.Clear();
+                foreach (var playerTrainDivergingSwitchList in PlayerTrainDivergingSwitches)
+                    playerTrainDivergingSwitchList?.Clear();
+                foreach (var playerTrainMilepostList in PlayerTrainMileposts)
+                    playerTrainMilepostList?.Clear();
+                foreach (var playerTrainTunnelList in PlayerTrainTunnels)
+                    playerTrainTunnelList?.Clear();
             }
         }
-
-        public List<TrainSignal>[,] TrainSignalLists; // first index 0 forward, 1 backward; second index signal type (NORMAL etc.)
 
         //================================================================================================//
         /// <summary>
         /// Updates the train data for TCS and TrackMonitor
         /// </summary>
         /// 
-        public void UpdateTrainData()
+        public void UpdatePlayerTrainData()
         {
-            UpdateSignalData(10000.0f);
+            UpdatePlayerTrainData(10000.0f);
             //TODO add generation of other train data
         }
 
         //================================================================================================//
         /// <summary>
-        /// Updates the signal data;
+        /// Updates the Player train data;
+        /// For every section it adds the TrainObjectItems to the various lists;
+        /// this first in forward direction and then in reverse direction
         /// </summary>
 
-        public void UpdateSignalData(float maxDistanceM)
+        public void UpdatePlayerTrainData(float maxDistanceM)
         {
-            if (!((MSTSLocomotive)Simulator.PlayerLocomotive).TrainControlSystem.CustomTCSScript)
-                return;
-            if (TrainSignalLists == null)
-            {
-                TrainSignalLists = new List<TrainSignal>[2, signalRef.ORTSSignalTypeCount];
-                for (int dir = 0; dir < 2; dir++)
-                    for (int fn_type = 0; fn_type < signalRef.ORTSSignalTypeCount; fn_type++)
-                        TrainSignalLists[dir, fn_type] = new List<TrainSignal>();
-            }
-            else
-                foreach (var trainSignalList in TrainSignalLists)
-                    trainSignalList?.Clear();
-            // fill up the lists
+            InitializePlayerTrainData();
+            // fill in the lists
+            TrainObjectItem thisItem;
             for (int dir = 0; dir < 2; dir++)
             {
                 if (ValidRoute[dir] == null || dir == 1 && PresentPosition[dir].TCSectionIndex < 0)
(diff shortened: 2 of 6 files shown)


pzgulyas:
I don't really like the way this function is implemented. It is actually a partial copy of the code residing in the Train.cs from a function building the TrainInfo database. This function has one of the most complex logics in OR, and I think it is not a good idea copying this into multiple functions, each with just a little modification. Later noone will ever be able to sort out how this whole thing works. In TrainControlSystem.cs, together with this, among the new functions there are multiple copies of this same logic, all with little modifications, and all of them are partial copies of the GetTrainInfoAuto() or the AddSwitch_MilepostInfo() functions. Rather the original functions should be extended to gather the appropriate objects with a few lines of code, leaving the searching logic there, and only there.

For this the enum TRAINOBJECTTYPE should be extended with e.g. a GENERIC_SIGNAL member, and just adding them to the ObjectInfoForward and ObjectInfoBackward lists of TrainObjectItem-s. And by extending the TrainObjectItem to have the thisSignal.SignalRef as a member, all info could be easily extracted about heads, aspects, etc., whatever is needed.

The same is the case with the other new functions, the enum TRAINOBJECTTYPE could also be extended with a FACING_SWITCH, TRAILING_SWITCH, TUNNEL, whatever. This would leave the whole logic residing only in one single place.

Csantucci:
Peter, what you suggest is a way to do it. I didn't want to touch internal structures (TrainObjectItem) only to simplify access to the TCS script management; while it can be straightforward to add thisSignal.SignalRef to TrainObjectItem to get the missing info (like signal head names), I wonder if adding also GENERIC_SIGNAL to TRAINOBJECTTYPE doesn't generate too heavy  ObjectInfoForward and Backward lists. But I'm open to what you propose. However I'd like to have at least also the opinion of Rob about this (I think he originally generated that part of OR code) before to proceed.

YoRyan:
A thought: It would be really spiffy if, rather than walking the track circuits every time some part of Open Rails calls `NextSuperSpecialTrackFeature...()`, we had some kind of data structure that could retrieve upcoming features in constant-time.

pzgulyas:
Ryan, I absolutely like this idea. The TrainInfo database is not needed to be rebuilt every time end for every caller separately. It could be built in a central location only once per frame, or even just 10 times a second, or so. And everything could just query this data structure for the needed info.

pzgulyas:
> I didn't want to touch internal structures (TrainObjectItem) only to simplify access to the TCS script management;

I think it is better to do it at one place correctly, than making multiple copies of complex queries.

Csantucci:
If we are talking about a re-done work, I don't think it's wise to be based on the TrainInfo structure, which mixes up signals with switches with speedposts. When a TCS script wants to access such data, it doesn't ask for a generic item, but for a Normal signal, or a switch (leading or trailing), or a speedpost, or a not-Normal signal. So best would be having separate lists for each of these items. And I'd build such lists directly from the internal structures and not from TrainInfo, and only if TCS scripts are present. Messing up additional data in TrainInfo, which aren't used for the original use of it (Track Monitor) is only an unnecessary intermediate step (internal structures --> TrainInfo --> Data structures where to pick data for script).

pzgulyas:
There's no need to build separate lists, it is only a matter of TrainInfoForward.Select(x => x.type == TRAINOBJECTTYPE.GENERIC_SIGNAL)[0] for example. The whole point is __not to multiply__ the searching logic in the track circuit lists, because that is super complex, and in case of rewriting the track circuit lists, all searches would also have to be rewritten. So have the code for searching only in one centralized place. If that is the TrainInfo for the Track Monitor, then that is the TrainInfo for the Track Monitor. Doesn't matter where it is, just be it only in one place.

Csantucci:
Why having everything together if you have info separately in the internal structures and need such info separately for the script? Regrouping and then selecting require unneeded CPU time. Having separate structures isn't against centralizing the generation of them (I agree about centralizing). E.g. generic signals are already separated per type and per track circuit section in the internal structures; so generating structures separated only per type is quite fast. Normal signals are present in the NextSignalObject() structure (I'm using the term structure improperly), so generating the script related data requires only a copy of the relevant data. Same is true for speedposts, same is true for switches, which are a subset of the track circuit sections.

pzgulyas:
I have to revise/refine my preceeding comments. I'm not actually against having separate structures (although I find convenient and simple to have everything in the single TrainInfo and selcting from that). I am only against multiplying the code of searching in the track circuit sections database. IMO the functions GetTrainInfoAuto() and AddSwitch_MilepostInfo() should be reused everywhere this is needed, instead of copying part of it to many locations. They can also be modified to fulfill all needs where such searching is needed. E.g. if one doesn't want to add the found items to the TrainInfo, but rather to some own list, probably some if()-s or similar need to be added there to the relevant places. (But again, I find the TrainInfo convenient to have every such thing, but accept if you find it unsympathic for some reason.)

pzgulyas:
P.S.: Don't care about the CPU time, when a script is running that time is getting used anyway, there is no difference. And if the TrainInfo building is optimized by Rayan's idea, it is actually getting much-much lower.

pzgulyas:
Just to be clear, I meant something like this (untested):
[TrainInfo.zip](https://github.com/openrails/openrails/files/5417335/TrainInfo.zip)

(I'm sorry, not sure how to attach further code suggestions to a GitHub PR.)

YoRyan:
Well, don't concern yourself with the implementation details too much. Some of the performance optimizations we've discussed might, upon actual testing, well be illusory and not worth it.

Point is, there ought to be a common set of functions attached to the Train class that search for track features in either direction in a well-defined and centralized manner:

```
public class Train {
    ...
    IEnumerable<TrackItem> ForwardTrackItems(float maxDistanceM)
    IEnumerable<TrackItem> BackwardTrackItems(float maxDistanceM)
    IEnumerable<Signal> ForwardTrackSignals(float maxDistanceM)
    IEnumerable<Signal> ForwardTrackSpeedposts(float maxDistanceM)
    IEnumerable<Milepost> ForwardTrackMileposts(float maxDistanceM)
    IEnumerable<Milepost> BackwardTrackSignals(float maxDistanceM)
    IEnumerable<Speedpost> BackwardTrackSpeedposts(float maxDistanceM)
    IEnumerable<Speedpost> BackwardTrackMileposts(float maxDistanceM)
}
```

The Track Monitor and TCS code should then be reworked to use the new functions.

For new API calls, I would also advise dropping the exclusion of `NORMAL` signals, which is a holdover from the origins of the code in the Track Monitor. This "feature" causes quite the headache for TCS scripts that need to locate those signals.

Roeterdink:
I had no intention to participate in this discussion but as my opninion is explicitly asked for, I will make one exception.
I am no longer actively involved with code development for OR, so obviously I also no longer have any plans to restructure the track circuit logic.
Also, I have not really been following this discussion, certainly not in any detail, so I don't really have an opinion on this. So please feel free to continue as you seem fit.

Regards,
Rob Roeterdink

pzgulyas:
Ryan, this would be perfect, probably this would be the best way to solve my code maintenance concerns too.

Csantucci:
I agree with Ryan's solution too.
I wonder now whether this reworking should be done within this PR, which in its origin has a more limited scope, or in another PR.

YoRyan:
The generic TrackItem search ought to return every feature along the path, regardless of whether or not it is shown on the Track Monitor. We should lean toward more information, not less.

I'm not sure if the signal lists are sorted by type but to allow for these kinds of optimizations, there could be calls to request signals by type:

```
IEnumerable<Signal> ForwardTrackSignals(type, float maxDistanceM)
IEnumerable<Signal> BackwardTrackSignals(type, float maxDistanceM)
```

I am unavailable for dev work this weekend, but I would be happy to collaborate with you on this.

As for restructuring the PR, I suppose that's up to you. The sound event call looks like a good addition. If you add manual documentation for that and strip the signal search stuff, I would approve right away.

Csantucci:
Thanks for being available to help, Ryan.
I am generating the part related to the generic (not NORMAL) signals, and intend to start inserting that in this PR. Before going further I'll of course wait about comments, suggestions and change proposals by you.. I won't handle the sound event call alone, that requires further time and I need all handles operating for my script.

Csantucci:
Updated PR with centralized management of not NORMAL signals.

cesarBLG:
Hi Carlo,
With this update, my TCS script crashes. I used GetNextGenericSignalAspect() since the very first Update() call, when TrainSignalLists is not defined, causing a null pointer exception. I have already solved this in the script, but it might be worth fixing it in OR, as script writers might not expect this behaviour.
I suggest one of this fixes:
a) Call UpdateSignalData before updating TCS scripts
b) Initialize TrainSignalLists with empty lists before first update of TCS scripts
c) Add a null-pointer check in NextGenericSignalFeatures(), ScriptedTrainControlSystem.cs line 580

Csantucci:
César,
thank you for pointing that out. I'm just completing the work by extending it to NORMAL signals and to speedposts, so I'll also add one of your fixes.

Csantucci:
The management of relevant info passed from OR core to script interface has been quite reworked, in the direction suggested by Ryan. Waiting for further comments or approval.

cesarBLG:
A maximum search distance of 10 km might not be enough in high speed lines, at least when searching for NORMAL signals and speedposts. Therefore, maximum distance should be extended to cover all Train.ValidRoute when requesting information about these items, if not found in the required distance. This problem only arises in MANUAL and EXPLORER modes, as auto modes are handled separately.
My TCSs aren't affected by this at the moment (they might be in the future when I finish ETCS), but I just wanted to note this.
I'll think a way to fix this.
@Sharpe49 does your TCS script need a higher distance? If not, maybe I'm overthinking the problem.

Just another thought: With this implementation, scripts won't be able to locate ends of track. Previously they were manually added from TrainInfo as danger signals. I might be wrong though.

Csantucci:
César,
that can be modified for NORMAL signals and speedposts. What do you think about linking the maxDistance with the allowed speed in the point of calculation?

cesarBLG:
Setting maxDistance depending on allowed speed could solve the issue in HSLs. However, consider two stations, A and B, 20 km away, without signals between them. A train departing from station A with a beacon-based ATP system on board should know all speed limits in advance, up to station B.
I have been thinking, and I believe this is not the right time to fix this. There are many problems regarding signal behaviour in EXPLORER mode when long distances are involved, so anyway scripts would misbehave.

YoRyan:
Carlo, Peter and I were suggesting that the track monitor code should also be refactored to share the lookahead code with the TCS interface. That way, such a complex piece of the software will exist in only one place and be easier to maintain. Your new patches don't really go far enough; they only affect your own TCS calls.

I am working on an implementation that aims to achieve what we suggested. Hopefully, I'll have something to present soon.

Csantucci:
Ryan,
sorry, but I don't understand your point. The track monitor code has been simplified because it is getting the data from the new lists, as the TCS interface does, and such lists are built in a single module, and this is what you suggested.
Take also into account that the title of this PR is "Further TCS handles to get signal data and to set signal events". So I think that what I have done has already quite widened the scope of the title. Further improving the track monitor code (how ?) is in my opinion another story, that I don't see why it should be part of this PR.

Csantucci:
César, do you have a suggestion for the formula of the link between maxDist and train max speed?

cesarBLG:
The formula seems to fit well. However, I wonder if there is a CPU overhead if no max distance is checked at all. Search will anyway stop at the end of ValidRoute, whose length is calculated via SNCA. I will do some tests, but I would say that ValidRoute length is similar or lower than Simulator.TRK.Tr_RouteFile.SpeedLimit * 250.0f, so it could be OK to remove the maxDistance check.

Csantucci:
César,
what you say might be true, however consider that some time ago there was an idea of Serana about increasing the max distance to check if ahead signals might be cleared in Explore mode, and that was especially devoted to HST. So I think that increasing max distance here is not bad, and that maybe max distance should be increased to clear signals in Explore mode. Therefore I think I will insert the check.

cesarBLG:
I have found a 200 km ValidRoute distance in Northeast Corridor, so it's a sensible decision to keep the max distance check, just in case something weird happens with route calculation.
Prior to Serana's patch this summer, signals were cleared only 5 km ahead of the train, which is not enough in several situations. Now, the route is extended up to SignalNumClearAhead, which is the right way to go. However, I have found some issues (including a crash when trying to reset a signal twice) which need further investigation.

There is another breaking change with this PR, concerning End of Authority objects. They were considered as signals when using NextSignalXXX() family of functions, but now they are excluded. This new behaviour is what I would expect from this functions, but then there is no way for the TCS to know if the train is approaching an end of track.

Csantucci:
So probably a specific script call to know the distance of the EOA could be generated.
I have also a question on tunnels: they are now collected in a list, but the list isn't accessed. Tunnel data are needed e.g. for ETCS DMIs. What would be the best script request for tunnel data? Asking for data about n-th tunnel within a specified distance, or asking for all tunnels within a specified distance in a single request?

cesarBLG:
In my opinion, having information of all tunnels at a time is more practical, but it requires allocating data in the heap, which is expensive. In ETCS, this information is usually collected only whenever a beacon is passed or MA is updated from RBC, so garbage collection is not a problem, but it might not be the same for other TCSs.
Another thing to take into account is that both speedposts and signals are accessed one by one, so it would be better to do the same for tunnels to keep consistency.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -79,6 +80,11 @@ public abstract class TrainControlSystem : AbstractScriptClass
         /// </summary>
         public Func<bool> DoesNextNormalSignalHaveTwoAspects;
         /// <summary>
+        /// Name of Head 0 of next normal signal.
+        /// int: position of signal in the signal sequence along the train route, starting from train front; 0 for first signal;
+        /// </summary>
+        public Func<int, string> NextNormalSignalMainHeadSignalType;
```
YoRyan, line 86: Any reason for this call when `NextGenericSignalFeatures` returns the same information, and more?

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -123,22 +136,26 @@ public abstract class TrainControlSystem : AbstractScriptClass
...
         public Func<bool> IsDirectionNeutral;
         /// <summary>
-        /// True if train direction is reverse.
+        /// True if locomotive direction is reverse.
         /// </summary>
         public Func<bool> IsDirectionReverse;
         /// <summary>
+        /// Train direction.
+        /// </summary>
+        public Func<Direction> CurrentTrainMUDirection;
```
YoRyan, line 157: How does "MU Direction" differ from "locomotive direction"? We should include a blurb about this.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -560,4 +581,12 @@ public enum MonitoringStatus
         /// </summary>
         Intervention,
     }
+
+    public struct SignalFeatures
+    {
+        public string MainHeadSignalTypeName;
+        public Aspect Aspect;
+        public float DistanceM;
+        public float SpeedLimit;
```
YoRyan, line 590: `SpeedLimit` needs SI units here.

YoRyan (changes requested):
The code for the track monitor and the _existing_ TCS calls could be cleaned up and improved. I'm making good progress on that right now. Maybe that's out of the scope of this PR, and that's fair.

So to move forward, let's not concern ourselves with the implementation details too much. Instead, let's talk about your new TCS calls, since the API is permanent, while we can always modify the internal implementation later. Please take a look at my feedback, and then I'll approve.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: Also, script writers might be interested in information on the signal heads. Might want to consider including that into this struct.

cesarBLG: I suggested a similar struct with information about signal heads [http://www.elvastower.com/forums/index.php?/topic/33693-expand-potential-of-tcs-scripting/page__view__findpost__p__262321](here), but Peter thought this would be bad for performance. Using a fixed-size array or something for signal heads might force passing SignalFeatures by the stack, but I don't know if that will happen in C#.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: The only reason is that there is already a script out using this call. But I agree deleting it from the official OR release, and will do.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: Will add the SI unit on SpeedLimit.
I am of the same opinion of Peter about not generating a heavier Struct (which by the way wouldn't be optimized for speedposts and so on), considering also that my small experience says that most of the calls don't need such info. Maybe a dedicated call like GetExtendedSignalInfo could be created, passing info about all heads, but that could be added later with no  problems.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: That's a good question. For sure they are two different variables :), Looking at the code I expect that they couldn't coincide for flipped locos or for locos run from rear cab (maybe backwards, or in explore mode, or during shunting). So I think it's good to have that info.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: Until new calls get mainlined, we make no guarantees about their availability.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: Right, but then the call is named `TrainMUDirection`, which at first glance might apply to both the train and the locomotive. :-)

Also consider adding a call for the locomotive's direction, but returning a `Direction` enum instead of a boolean. That would make it even more clear that they are two separate pieces of information.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: Yes, but these calls replace the Locomotive() call, which IS mainlined and which someone wants to remove...

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: I disagree with Peter on the performance consideration; I would rather have an API that's easier to use. If building and destroying these structs takes a noticeable amount of time, then there are ways to avoid those operations by reusing the same struct across multiple calls.

That said, the way I would accomplish that would be use a class with public getters and internal setters instead of a struct, but we can handle these micro-optimizations later once we've agreed on the API.

YoRyan, line 590: Another thought: Not all signals impose speed limits. Use a `float?` type so you don't have to ambiguously code this condition with `0` or `float.MaxValue`.

Also, is it possible for signals not to have aspects?

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
cesarBLG: There's already a CurrentDirection() call available returning an enum. IsDirectionReverse() is a legacy call which was created a few years ago.

YoRyan: Totally missed that, thanks for pointing it out. I'll be satisfied if we add documentation clarifying the definitions of locomotive direction, train direction, and MU direction.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
pzgulyas: I don't think it is desirable to expect from the script writers to do null-checks, so I would not change it to "float?". The aspects enum has a "none" member.

pzgulyas: To use a stuct vs. a class must be decided now, because altering that is a breaking change, which is not allowed later. It is not the time of destroying that matters here, rather it is specifically the Garbage Collector that causes problems. A struct is actually not affected by the GC, because that is a value type, in contrast with a class, which is a reference type. A struct is a good approach here, because that is automatically destroyed when it goes out of scope, but a class is just registered for later destoying by the GC. If it is decided to use a class instead, your approach of reusing the same class across all calls must be used.

There were extended efforts in OpenRails to hunt down class instance creations in the game loop in the past, to minimize the triggering of the GC. The problem with the GC is that when it runs, it hangs the game loop for some time, thus causes micro-stuttering. Ideally there should be no new class instance creations in the game loop at all, all needed classes must be created at the game initialization time, and never destroyed/disposed. I learned this from James @twpol, he might jump in here either to deny or confirm this.

YoRyan: The type checker will actually throw an error if you use a `float?` in a location that requires a plain old `float` and remind you to downcast with `.Value` or `??`. Optionals are much more sophisticated than `if (x != null)` and it would be a shame not to take advantage of them.

YoRyan: Technically some scripts could continue to work if the names and syntax remained identical - as in `NextGenericSignalInfo(...).Aspect` - but yes, I realize now that if you do anything involving the type declaration, that would introduce the possibility of breakage.

Okay, so stick with structs, but I suggest adding the `readonly` keyword to keep their fields immutable.

YoRyan: Er, that's ["nullable reference type"](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types) in C# speak. It's a potential null pointer that forces you to deal with the null pointer condition.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: I will add that to the script .cs file.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: OK

Csantucci: I will leave things as they are.

Csantucci:
I have got a problem: the new signal data management has dropped the "trick" of Peter, that returned the data of the last signal in the list, if the data of a further signal were required. This created a problem in the first Italian high speed route (Firenze-Roma) with max speed 250 km/h. The SCMT TCS script foresees, as in reality, that for HS routes the aspect of the next signal is conditioned by the aspect of the 2 subsequent signals. Unfortunately the actual formula within the core OR code to determine where to stop generation of the next signal list (when in auto mode) is as follows:
`            float maxDistance = Math.Max(AllowedMaxSpeedMpS * maxTimeS, minCheckDistanceM);`
that is, it is depending also from the train max speed. Unfortunately in this high speed route also regional trains with max speed of 160 km/h run on some sections, so that maxDistance is too short to get data of 3 signals.
Therefore I suggest to change this formula into something like that
`float maxDistance = Math.Max((IsActualPlayerTrain ? (float)Simulator.TRK.Tr_RouteFile.SpeedLimit : AllowedMaxSpeedMpS) * maxTimeS, minCheckDistanceM);`
that works. This should be changed also in other places in the code where this maxDistance is computed. I wouldn't change it for all trains. AI trains don't need that, and having longer searches increases computing times. Can this be OK?

pzgulyas:
Just for reference: Is this the same function that PR#270 and [this ET post](http://www.elvastower.com/forums/index.php?/topic/34297-mg71-crash-index-was-out-of-range-must-be-non-negative-and-less-than-the-size-of-the-collection/page__view__findpost__p__262475) was about, the public void InitializeSignals(bool existingSpeedLimits) in Train.cs?

If I understand that correcty, this affected only the Explorer (Multiplayer) mode, and only those trains that are inserted into the game running by a speed?

pzgulyas:
If so, then the AllowedMaxSpeedMpS initialized like this:
`AllowedMaxSpeedMpS = TrainMaxSpeedMpS;   // set default`

and TrainMaxSpeedMpS is by the comment:
`public float TrainMaxSpeedMpS;                   // Max speed as set by route (default value)`

Where the comment is in contrast how it is initialized in the InitializeSpeeds() function:
`TrainMaxSpeedMpS = Math.Min((float)Simulator.TRK.Tr_RouteFile.SpeedLimit, ((MSTSLocomotive)Simulator.PlayerLocomotive).MaxSpeedMpS);`

Also there is a conditional overwrite in the above mentioned InitializeSignals() function:
`TrainMaxSpeedMpS = (this.LeadLocomotive as MSTSLocomotive).MaxSpeedMpS;`

and there is also an overwrite in the UpdateCarSpeeds() function:
`                if (TrainMaxSpeedMpS <= 0f)
                {
                    if (car is MSTSLocomotive)
                        TrainMaxSpeedMpS = (car as MSTSLocomotive).MaxSpeedMpS;
                    if (car is MSTSElectricLocomotive)
                        TrainMaxSpeedMpS = (car as MSTSElectricLocomotive).MaxSpeedMpS;
                    if (car is MSTSDieselLocomotive)
                        TrainMaxSpeedMpS = (car as MSTSDieselLocomotive).MaxSpeedMpS;
                    if (car is MSTSSteamLocomotive)
                        TrainMaxSpeedMpS = (car as MSTSSteamLocomotive).MaxSpeedMpS;
                }`

pzgulyas:
But all in all, to me it looks like if the game is started with a high-speed locomoitive+cars, the TrainMaxSpeedMpS variable should be set correctly. Where is goes wrong then?

cesarBLG:
The problem with explorer mode was similar to this one, but not the same. In explorer mode, what happened was that CheckExplorerPath() constructed a route with a fixed size of 5 km, causing further signals not to clear correctly. As a side problem, signals in HSLs weren't accessible by TCSs. This was solved in PR #270 by clearing signals up to SignalNumClearAhead and extending ValidRoute[], so with this both TCSs and signalling logic work well in explorer mode. While reviewing Carlo's changes yesterday, I noticed some bugs, so I have filed PR #301 refining the logic in CheckExplorerPath.

If I'm not wrong, the problem with auto mode is in UpdateSignalState(), where SignalObjectItems is built to cover a short distance, not enough for HSLs. Ideally, SignalObjectItems should have a number of signals up to SNCA if possible, but the logic there is too complex and something could be broken.
Carlo, can the code used for explorer mode in UpdatePlayerTrainData() be used for auto mode too?

pzgulyas:
Just a note: PR #270 was submitted by @Sharpe49, not by Carlo.

Csantucci:
Peter, if the player train is a HST the signals are collected up to a bit above 10000 meters, which is OK at least in my case. But if the player train is a regional train with max speed = 160 Km/h, the signals are collected up to about 5300 meters, which is not enough for the SCMT TCS.
I wouldn't at all touch the allowed speed logic calculation, which is used in many places and is quite delicate. I don't see a reason to do that, and moreover I don't see the reason to do this within this PR. What I propose is related to max signal distance calculation, and not to max allowed speed for trains. What I propose is not a hack, because only player trains use TCS, so only they need to get signal data further away.
César, I'll see your proposed patch, thank you. However for the purposes of the problem in AUTO mode I think that my patch (extended to all places where this max distance is computed) could be enough.

cesarBLG:
> Just a note: PR #270 was submitted by @Sharpe49, not by Carlo.

I know, although it leads to confusion in my previous comment. I was doing some tests about _this_ PR because I found that maximum search distance of 10 km wasn't enough for HSLs, and I came up against the bug related to PR #270.

I'll try to clarify things a bit. It's confusing because we are speaking about different max distance checks:
- Maximum distance to check for signals and speedposts for use by TCS, as defined in UpdatePlayerTrainData(), defined in this PR. Initially defined as fixed 10 km, now extended using TRK speed. No further actions should be required for this.
- Maximum distance to construct ValidRoute[] in explorer mode, which affects both signalling logic and PlayerTrainData. Previously set to a fixed value of 5 km, PR #270 set no distance based limit, but uses SNCA. My PR #301 refines the logic. No further actions required either.
- Maximum distance to check when building SignalObjectItems in Auto mode. Currently calculated using TrainMaxSpeed, not convenient as Carlo has said. Something has to be done in order to get HSLs TCS working in auto_signal.

pzgulyas:
Carlo, I'm just thinking loud, don't take it wrong. Just the comment at 
`public float TrainMaxSpeedMpS; // Max speed as set by route (default value)`
popped into my eyes, and I started thinking of why it is not set to the value of the original intention. Probably the variable got reused to the speed logic calculation, and that became its primary use by the time...

Looking more at the code, your proposed change looks valid to me. However if you intend to change that line at multiple places, probably it would look nicer if you defined a variable near TrainMaxSpeedMpS like
`public float RouteMaxSpeedMpS; // Max speed as set by route`
and initialize that to (float)Simulator.TRK.Tr_RouteFile.SpeedLimit, and use that, just for showing the contrast with TrainMaxSpeedMpS and AllowedMaxSpeedMpS. But it is just nicer to my eyes, it is actually up to you.

pzgulyas:
César, is your PR is in line with [this ET post](http://www.elvastower.com/forums/index.php?/topic/34297-mg71-crash-index-was-out-of-range-must-be-non-negative-and-less-than-the-size-of-the-collection/page__view__findpost__p__262475) I mentioned before? Your PR seem to reintroduce the minCheckDistanceM constant for in use in the explorer mode, however Rob writes the following in his post:

_"Then to explorer mode. This same constant (minCheckDistanceM) was used - or abused, if you like - to set up the limits for the checks in explorer mode as explained above. But this certainly is not the main function of this constant.
So, if a change is needed in explorer mode to extend the search, the first thing to do is to drop the use of this constant. This use is not what this constant is intended for, so leave it to its intended function and set up another limit check for explorer."_

pzgulyas:
Also:
_"The best option, however, is simply to remove the scan restriction alltogether. By removing the constant minCheckDistanceM from the call to ScanRoute and replacing it by the value -1, the scan is performed for the full (remaining) route of the train, which is similar as to its use in other modes."_

It is in contrast with your PR which specifically does the opposite: remove -1 from the ScanRoute function, and adds minCheckDistanceM.

Csantucci:
Peter,
in effect the description of the variable TrainMaxSpeedMpS is not correct. I don't like very much creating a new variable in Train.cs, because there would be a variable for each train, and that variable would hold the same value for all trains. Maybe it could belong to Simulator.cs.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: Setting the Struct as readonly generates a compile error in the TrainControlSystem.cs statements where I set their values.

YoRyan: Ok, so how are you coding signals without speed limits? This needs to be documented, too.

YoRyan:
[use-immutable-structs.patch.txt](https://github.com/openrails/openrails/files/5502929/use-immutable-structs.patch.txt)

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: It's possible if you give the struct a constructor. See my [attached patch](https://github.com/openrails/openrails/files/5502929/use-immutable-structs.patch.txt).

YoRyan: Correct me if I'm wrong, but I believe signal objects without speed limits have `.actual_speed = -1 m/s`.

Csantucci: OK, I'll apply it.

Csantucci: Yes, it's the rule used in the signalling code.

Csantucci: Ryan, your patch does work for direct calls to NextGenericSignalFeatures(), but doesn't work for calls to NextGenericSignalItem().
Apparently from NextGenericSignalItem() it enters NextGenericSignalFeatures(), returns correctly to NextGenericSignalItem(), but instead of returning from it it enters again in the Return statement of NextGenericSignalFeatures(), and exiting to it into NextGenericSignalItem the parameter values are lost.

Csantucci:
Inserted changes as agreed, with the exception of the readonly keyword for the struct.

YoRyan:
[Uploading immutable-structs-2.patch.txt…]()

YoRyan:
[immutable-structs-2.patch.txt](https://github.com/openrails/openrails/files/5503635/immutable-structs-2.patch.txt)

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: I didn't expect you to store the result of a function call in an instance variable. Try this [revised patch](https://github.com/openrails/openrails/files/5503635/immutable-structs-2.patch.txt), please.

Csantucci: Patch OK, thanks. I've just pushed the new PR release including it.

Sharpe49: I would have preferred nullable type too. It's cleaner than checking for negative values.

Csantucci:
I have just added the script hook EOADistanceM(int) returning the distance to end of authority in direction M, as discussed here with César. I would be the right moment to add also the hook to get distance and length of the n-th tunne in the required direction up to a certain distance from train, as these data are now already collected in a list. The handling metod could be NextTunnelFeatures(int itemSequenceIndex, float maxDistanceM, int direction); the question is where to return distance and length: in a struct also in this case?
Re the issue about returning or not the nulls, also the original code (before my changes) had following lines in SearchTrainInfo
`         if (searchFor == Train.TrainObjectItem.TRAINOBJECTTYPE.SIGNAL && signalsFound == 0)
            {
                SignalSpeedLimits.Add(-1);
                SignalAspects.Add(Aspect.None);
                SignalDistances.Add(float.MaxValue);
            }
            if (searchFor == Train.TrainObjectItem.TRAINOBJECTTYPE.SPEEDPOST && postsFound == 0)
            {
                PostSpeedLimits.Add(-1);
                PostDistances.Add(float.MaxValue);
            }`
and changing this would break compatibility with existing scripts.

cesarBLG:
What do you think about passing a Direction argument to EOADistanceM() instead of an integer?

For tunnel information, there's no advantage in having two separate calls for tunnel length and distance, so I would prefer a single call returning a struct. Something which has to be discussed is the return value if nth tunnel is not found in required distance. If the function returned a list of tunnels, it would simply return an empty list, but in that case it has to be specified that the function must be called only occasionally to keep performance.

I find cleaner the use of a nullable for signal speed limits. Compatibility with old scripts can be maintained with something like this:
```csharp
T NextGenericSignalItem<T>(int itemSequenceIndex, ref T retval, float maxDistanceM, Train.TrainObjectItem.TRAINOBJECTTYPE type, string signalTypeName = "UNKNOWN")
{
    var item = NextGenericSignalFeatures(signalTypeName, itemSequenceIndex, maxDistanceM, type);
    MainHeadSignalTypeName = item.MainHeadSignalTypeName;
    ItemAspect = item.Aspect;
    ItemDistance = item.DistanceM;
    ItemSpeedLimit = item.SpeedLimitMpS ?? -1;
    return retval;
}
```

Csantucci:
Again a fail running Appveyor, because a package couldn't be loaded. How can I repeat the build withouth modifying the PR and without retiring and re-presenting it?

By the way the modification adds the field AltitudeOrLengthM to the SignalFeatures struct, which contains the altitude of the related NORMAL signal.

YoRyan:
Do a `git merge master` to merge the new appveyor.yml file into this PR.

I'm confused about this new property. It's derived from the value of `trainSignal.SignalObject.tdbtraveller.Y`, but it's named "AltitudeOrLength"? Under what circumstances would the Y coordinate be the "length" of the signal?

Csantucci:
Thank you Ryan, I was sure I had done that, but I didn't.

I have called that field "AltitudeOrLength", because I was considering using the same struct also to host tunnel data (distance from train and length are needed). What do you think about that?

YoRyan:
Very bad idea. I would strongly suggest using separate structs with unambiguously named fields.

pzgulyas:
+1 for Ryan

Csantucci:
OK, I've done it. There might be code duplication though, at the moment I'll add tunnel support.

Peter, thanks for your words. In fact I need just now TCS access to panto operations and wondered if I should add calls for that or if you would intervene with your general solution, that includes such access. I must also confess that it's not clear to me whether with your solution the TCS script should have the burden of getting all locomotive commands and passing them back.

pzgulyas:
Just proceed with your solution.

I would like to make a note about the keyboard "Commands" (e.g. PantographCommand, or how it is called), that when implementing additional locomotive commands for the scripting interface, please make sure you never ever call the corresponding "Command" that is attached to the LocomotiveViewer (e.g. the PantographCommand, or how it is called). The problem with them is that no matter what locomotive script is calling them, they always operate on the locomotive attached to the Viewer, not on the one attached to the script. So such command need to be kind of "reimplemented", the function backing them has to be called directly, not via a kind of "new PantoghraphCommand".

Probably you were aware of this, just I wanted to make a note to be sure.

Csantucci:
Thanks Peter; all calls pass through Locomotive or through Locomotive.Train, and therefore should be addressed to the correct plae.

Sharpe49:
Could you add the name of the current draw state of the signal to the SignalFeatures struct? I need this information to be more accurate for the Crocodile system.
Also, for the same system, can we do something similar for speed posts? I would like to be able to get information about SpeedWarning speedposts.

cesarBLG:
I'm going to add milepost and tunnel hooks for the TCS. Carlo, is there any reason to reuse the SignalFeatures for that? I find cleaner using separate structs for these items.

We have to decide what to do about exposing deeper signal features. For example, I thought of using the signalRef id as balise group NID_BG in ETCS (the ID is needed in order to link one balise group with the next one), but this can only be done with the Locomotive() call, unless we add the signal id to the SignalFeatures struct. If added to the struct, it's a waste of resources for the rest of TCSs. Something similar happens with the draw_state names.

I think all this accesses to the internal code could be avoided if we add a proper interface between signals (or beacons) and trains. What about creating a handler (either as an extension in the sigscr or using C# scripts as suggested by @Sharpe49) which is called when the train passes a signal, and it sends some information for the train?

Csantucci:
César,
the mayority of you prefer to have separate structs for mileposts and tunnels, so please do so.
Take into account that tunnel data are already collected in train.cs within this PR; what is missing are only the TCS hooks. So, I'd suggest that you either wait until this PR is merged into master or you generate a PR to my branch or you prepare a patch that I apply to the PR.
Cédric,
Re the draw state there is the problem that one signal can have multiple heads, each with its own draw state, so this can become complicated.
Re the speed warnings, as far as I know they are not part of the tdb, so collecting them is not trivial. Speed reductions can already be accessed.

Sharpe49:
For the draw state, you can give a Dictionary<string, string> where the first string is the name of the head and the second string is the name of the draw state.

This is a SpeedWarning speed post.
image
In TSRE, when TrackDB items are shown, there is a TDB item for this object.

Csantucci:
Merged César's contribution adding hooks for tunnel and milepost info.

cesarBLG:
At the moment I am using Locomotive().Train.signalRef.ORTSSignalTypes.IndexOf("ASFA") to check if beacons are present in a route. If not, I do some suppositions and simulate their presence in some places.
If I use NextGenericSignalFeatures("ASFA", 0, 20) in routes without that signal type defined, I will get an IndexOutOfRangeException with the current implementation. If I catch that exception, I will know that beacons don't exist and I have to simulate them. However, I don't think any script call should raise an exception, so this has to be fixed. In that case, I need a way to check if the fn_type is defined. The solutions I've come up with are:
- Make NextGenericSignalFeatures return a nullable SignalFeatures struct
- Return a negative value for DistanceM if the required signal type does not exist
- Create a specific call like RouteHasSignalType(type)
- Add a specific parameter in the .trk route file

Which do you like the most?

Csantucci:
In coherence to what has been implemented up to now, I'd leave DistanceM to
the default initial value which is float.MaxValue.

Il giorno dom 29 nov 2020 alle ore 11:12 César Benito Lamata <
notifications@github.com> ha scritto:

cesarBLG:
If DistanceM is float.MaxValue there's no way to differentiate a) signal type not being defined in sigcfg from b) signal type defined but not found withinin maxDistanceM. I was suggesting a special value in SignalFeatures struct in order to not having to create a specific script call, but maybe it's more clear to have it as a separate function.

Csantucci:
If you need to have two distinct return values, then -1 is OK for me, but before spending time modifying it let's see if it is OK also for the reviewer(s).

YoRyan:
Strong recommendation for a nullable SignalFeatures return. From the POV of a script writer, it's the least worst option versus:

- Remembering to catch a `NullReferenceException`
- Remembering to check for the special "not found" values of `DistanceM`, `SpeedLimitMpS`, `AltitudeM`, and the other fields
- Remembering to call a function to check for the existence of a signal first, which actually introduces a potential race condition, too

In addition, I stress that if you return a `SignalFeatures?` type, Intellisense will nag script writers to deal with the "not found" condition. This is not true with any of the above approaches. That's why I don't accept the argument that TCS functions should never return nullable types.

pzgulyas:
You might be right about the nullables. When it is explicitly "nullable", it has a value indeed. (I still keep the argument that when the return type is not explicitly nullable, e.g. a class, in that case it should not return a null. But the current functions iirc don't return a class, which is also good.)

Csantucci:
If I understand César's need correctly, making the struct nullable doesn't solve his problem. He needs two different "not existing" return values for a) signal type not being defined in sigcfg and b) signal type defined but not found withinin maxDistanceM.

cesarBLG:
Carlo, you're right, I would like to differentiate both situations. What about having a property defining the result of the last search?
``` csharp
public Func<SearchResult> GetSearchResult;
public enum SearchResult
{
    FOUND,
    NOT_FOUND_MAX_DISTANCE,
    NOT_FOUND_NOT_EXISTING,
    NOT_FOUND_NO_VALID_PATH
}
```
This property would be set whenever a script calls any signalling function.

Peter, the nullable types are defined as `struct Nullable<T>` so I don't think using them will result in additional heap allocations.

Csantucci:
So I see I'm in big minority about the nullable issue; I'll accept what the majority prefers, although this will create some annoyance to people using one of my scripts. If someone provides a patch or a PR to my PR branch, I'll merge it.
Cédric, I looked about the speed warning posts. Such speed warning posts are flagged with the IsWarning flag within TrakDatabaseFile.cs. If you look in private void ScanSection() within Signals.cs, around line 976, you will see that speed warnings are not inserted neither in the speedpost lists nor in the milepost lists. So they aren't immediately retrievable. Therefore I consider adding them a non trivial, separate task, because the problem would subsist also using the Locomotive() hook, which this PR aims at least partiallly to replace.

pzgulyas:
If the list of these items are needed, then they might be either added to the existing list by flagging them similarly how the different element types are flagged in the traininfo list, or they could be added to a distinct list, where the scripting interface function is able to search in. I still think it would be a trivial solution to add everything to the traininfo list, together with generic signals, trailing switches, tunnels, etc.

YoRyan:
This PR has been sitting on the queue for quite some time. I suggest merging first, then refactoring later.

Csantucci:
Thanks Ryan. Pls. wait a moment, I have found a bug that I am fixing now.

Csantucci:
Wait request terminated: fixed bug acquiring switch distance and added check about not existing signal type.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -14097,13 +14099,315 @@ public String[] AddRestartTime(String[] stateString)
...
+                            {
+                                if (tunnelStartOffset > lengthOffset)
+                                {
+                                    thisItem = new TrainObjectItem(tunnelStartOffset + sectionDistanceToTrainM, (int)thisTunnel[sectionDirection].TotalLength, TrainObjectItem.TRAINOBJECTTYPE.TUNNEL);
+                                    PlayerTrainTunnels[dir].Add(thisItem);
+                                }
+                                else if (PlayerTrainTunnels[dir].Count == 0 && (thisTunnel[sectionDirection].TunnelEnd < 0 || thisTunnel[sectionDirection].TunnelEnd > lengthOffset))
+                                {
+                                    // Train is in tunnel, compute remaining length
+                                    var remainingLength = thisTunnel[sectionDirection].TotalLength - lengthOffset - (tunnelStartOffset < 0 ? thisTunnel[sectionDirection].TCSStartOffset : tunnelStartOffset);
```
cesarBLG, line 14379: I made a mistake in the calculation of the remaining length of the tunnel. It should be:
``` csharp
var remainingLength = thisTunnel[sectionDirection].TotalLength - lengthOffset + (tunnelStartOffset < 0 ? -thisTunnel[sectionDirection].TCSStartOffset : tunnelStartOffset);
```
(There was a missing minus sign)

Could you please insert the modification?

Csantucci: OK, corrected.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: Bump! This is the last blocker before I can approve.

Csantucci: OK, I had forgotten this. Hook now deleted.

YoRyan (approved):

YoRyan:
Right, this has waited long enough. Thanks for your patience, Carlo.

Csantucci:
Thanks, I'm relieved.

PR #294, opened 2020-10-20
Adjust hotbox code

peternewell:
Set hotboxes so that they only activate on player trains. - https://bugs.launchpad.net/or/+bug/1900762


YoRyan (approved):
Looks good, thanks for the patch.

PR #298, opened 2020-11-02
Describes improved F6 labels

cjakeman:
Function of F6 and Shift+F6 has been improved.
Manual now describes these changes.
See http://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/page__view__findpost__p__262726


YoRyan (approved):
Looks good. Thanks for taking this on, Chris.

PR #300, opened 2020-11-04
Brake pipe overcharge

cesarBLG:
https://trello.com/c/nHJrSDA9/486-brake-pipe-overcharge

Adds "overcharge" brake controller position, which increases brake pipe pressure above the maximum to fully release all brakes.

See http://www.elvastower.com/forums/index.php?/topic/34536-brake-pipe-overcharge/

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -752,8 +752,10 @@ namespace Orts.Simulation.RollingStocks
                 case "engine(enginecontrollers(brake_dynamic": DynamicBrakeController.Parse(stf); break;
 
                 case "engine(trainbrakescontrollermaxsystempressure":
+                case "engine(trainbrakescontrollermaxoverchargepressure":
                 case "engine(trainbrakescontrollermaxreleaserate":
                 case "engine(trainbrakescontrollermaxquickreleaserate":
+                case "engine(trainbrakescontrolleroverchargeeliminationrate":
                 case "engine(trainbrakescontrollermaxapplicationrate":
                 case "engine(trainbrakescontrolleremergencyapplicationrate":
                 case "engine(trainbrakescontrollerfullservicepressuredrop":
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -752,10 +752,10 @@ namespace Orts.Simulation.RollingStocks
                 case "engine(enginecontrollers(brake_dynamic": DynamicBrakeController.Parse(stf); break;
 
                 case "engine(trainbrakescontrollermaxsystempressure":
-                case "engine(trainbrakescontrollermaxoverchargepressure":
+                case "engine(ortstrainbrakescontrollermaxoverchargepressure":
                 case "engine(trainbrakescontrollermaxreleaserate":
                 case "engine(trainbrakescontrollermaxquickreleaserate":
-                case "engine(trainbrakescontrolleroverchargeeliminationrate":
+                case "engine(ortstrainbrakescontrolleroverchargeeliminationrate":
                 case "engine(trainbrakescontrollermaxapplicationrate":
                 case "engine(trainbrakescontrolleremergencyapplicationrate":
                 case "engine(trainbrakescontrollerfullservicepressuredrop":
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -709,19 +709,22 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                             if (car.BrakeSystem.AngleCockAOpen) //  AND Front brake cock opened
                             {
                                 car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipeTimeVariationS * p1 / brakePipeTimeFactorS;
-
+                                if (car.BrakeSystem.BrakeLine1PressurePSI < 0)
+                                    car.BrakeSystem.BrakeLine1PressurePSI = 0;
                             }
 
                             if (car0.BrakeSystem.AngleCockBOpen && car != car0) //  AND Rear cock of wagon opened, and car is not the first wagon
                             {
                                 car0.BrakeSystem.BrakeLine1PressurePSI -= TrainPipeTimeVariationS * p0 / brakePipeTimeFactorS;
-
+                                if (car.BrakeSystem.BrakeLine1PressurePSI < 0)
+                                    car.BrakeSystem.BrakeLine1PressurePSI = 0;
                             }
                         }
                         if (car == train.Cars[train.Cars.Count - 1] && car.BrakeSystem.AngleCockBOpen) // Last car in train and rear cock of wagon open
                         {
                             car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipeTimeVariationS * p1 / brakePipeTimeFactorS;
-
+                            if (car.BrakeSystem.BrakeLine1PressurePSI < 0)
+                                car.BrakeSystem.BrakeLine1PressurePSI = 0;
                         }
                         p0 = car.BrakeSystem.BrakeLine1PressurePSI;
                         car0 = car;
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -695,7 +695,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                             // Based on the principle of pressure equualization between adjacent cars
                             // First, we define a variable storing the pressure diff between cars, but limited to a maximum flow rate depending on pipe characteristics
                             // The sign in the equation determines the direction of air flow.
-                            float TrainPipePressureDiffPropogationPSI = (p0>p1 ? -1 : 1) * Math.Min (TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));
+                            float TrainPipePressureDiffPropogationPSI = (p0>p1 ? -1 : 1) * Math.Min(TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));
 
                             // Air flows from high pressure to low pressure, until pressure is equal in both cars.
                             // Brake pipe volumes of both cars are taken into account, so pressure increase/decrease is proportional to relative volumes.
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -48,7 +48,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
         protected float MaxReleaseRatePSIpS = 1.86f;
         protected float MaxApplicationRatePSIpS = .9f;
         protected float MaxAuxilaryChargingRatePSIpS = 1.684f;
-        protected float MaxAuxilaryOverchargeEliminationRatePSIpS = 0.05f; // Allows overcharge elimination of 0.4 bar in 2 minutes
+        protected float BrakeInsensitivityPSIpS = 0;
         protected float EmergResChargingRatePSIpS = 1.684f;
         protected float EmergAuxVolumeRatio = 1.4f;
         protected string DebugType = string.Empty;
@@ -112,6 +112,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
             MaxReleaseRatePSIpS = thiscopy.MaxReleaseRatePSIpS;
             MaxApplicationRatePSIpS = thiscopy.MaxApplicationRatePSIpS;
             MaxAuxilaryChargingRatePSIpS = thiscopy.MaxAuxilaryChargingRatePSIpS;
+            BrakeInsensitivityPSIpS = thiscopy.BrakeInsensitivityPSIpS;
             EmergResChargingRatePSIpS = thiscopy.EmergResChargingRatePSIpS;
             EmergAuxVolumeRatio = thiscopy.EmergAuxVolumeRatio;
             TwoPipes = thiscopy.TwoPipes;
@@ -234,6 +235,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                 
                 // OpenRails specific parameters
                 case "wagon(brakepipevolume": BrakePipeVolumeM3 = Me3.FromFt3(stf.ReadFloatBlock(STFReader.UNITS.VolumeDefaultFT3, null)); break;
+                case "wagon(ortsbrakeinsensitivity": BrakeInsensitivityPSIpS = stf.ReadFloatBlock(STFReader.UNITS.PressureRateDefaultPSIpS, null); break;
             }
         }
 
@@ -449,9 +451,9 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                     AuxResPressurePSI += dp;
                     BrakeLine1PressurePSI -= dp * AuxBrakeLineVolumeRatio;  // Adjust the train brake pipe pressure
                 }
-                if (AuxResPressurePSI > BrakeLine1PressurePSI) // Allow small flow from auxiliary reservoir to brake pipe to restore normal pressure after overcharge
+                if (AuxResPressurePSI > BrakeLine1PressurePSI) // Allow small flow from auxiliary reservoir to brake pipe so the triple valve is not sensible to small pressure variations when in release position
                 {
-                    float dp = elapsedClockSeconds * MaxAuxilaryOverchargeEliminationRatePSIpS;
+                    float dp = elapsedClockSeconds * BrakeInsensitivityPSIpS;
                     if (AuxResPressurePSI - dp < BrakeLine1PressurePSI + dp * AuxBrakeLineVolumeRatio)
                         dp = (AuxResPressurePSI - BrakeLine1PressurePSI) / (1 + AuxBrakeLineVolumeRatio);
                     AuxResPressurePSI -= dp;
@@ -688,13 +690,16 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                         train.TotalTrainBrakePipeVolumeM3 += car.BrakeSystem.BrakePipeVolumeM3; // Calculate total brake pipe volume of train
 
                         float p1 = car.BrakeSystem.BrakeLine1PressurePSI;
-                        if (car == train.Cars[0] || car.BrakeSystem.FrontBrakeHoseConnected && car.BrakeSystem.AngleCockAOpen && car0.BrakeSystem.AngleCockBOpen)
+                        if (car != train.Cars[0] && car.BrakeSystem.FrontBrakeHoseConnected && car.BrakeSystem.AngleCockAOpen && car0.BrakeSystem.AngleCockBOpen)
                         {
+                            // Based on the principle of pressure equualization between adjacent cars
+                            // First, we define a variable storing the pressure diff between cars, but limited to a maximum flow rate depending on pipe characteristics
+                            // The sign in the equation determines the direction of air flow.
+                            float TrainPipePressureDiffPropogationPSI = (p0>p1 ? -1 : 1) * Math.Min (TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));
 
-                            float TrainPipePressureDiffPropogationPSI = Math.Min (TrainPipeTimeVariationS * (p1 - p0) / brakePipeTimeFactorS, p1 - p0);
-                            // Appears to work on the principle of pressure equalisation, ie trailing car pressure will drop by fraction of total volume determined by lead car.
-
-                            // TODO - Confirm logic - it appears to allow for air flow to coupled car by reducing (increasing) preceeding car brake pipe - this allows time for the train to "recharge".
+                            // Air flows from high pressure to low pressure, until pressure is equal in both cars.
+                            // Brake pipe volumes of both cars are taken into account, so pressure increase/decrease is proportional to relative volumes.
+                            // If TrainPipePressureDiffPropagationPSI equals to p1-p0 the equalization is achieved in one step.
                             car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipePressureDiffPropogationPSI * brakePipeVolumeM30 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
                             car0.BrakeSystem.BrakeLine1PressurePSI += TrainPipePressureDiffPropogationPSI * car.BrakeSystem.BrakePipeVolumeM3 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
                         }
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -448,6 +448,14 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                     AuxResPressurePSI += dp;
                     BrakeLine1PressurePSI -= dp * AuxBrakeLineVolumeRatio;  // Adjust the train brake pipe pressure
                 }
+                if (AuxResPressurePSI > BrakeLine1PressurePSI) // Allow small flow from auxiliary reservoir to brake pipe to restore normal pressure after overcharge
+                {
+                    float dp = elapsedClockSeconds * 0.05f;
+                    if (AuxResPressurePSI - dp < BrakeLine1PressurePSI + dp * AuxBrakeLineVolumeRatio)
+                        dp = (AuxResPressurePSI - BrakeLine1PressurePSI) / (1 + AuxBrakeLineVolumeRatio);
+                    AuxResPressurePSI -= dp;
+                    BrakeLine1PressurePSI += dp * AuxBrakeLineVolumeRatio;
+                }
             }
             if (TwoPipes
                 && !NoMRPAuxResCharging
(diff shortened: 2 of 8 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -752,8 +752,10 @@ public override void Parse(string lowercasetoken, STFReader stf)
                 case "engine(enginecontrollers(brake_dynamic": DynamicBrakeController.Parse(stf); break;
 
                 case "engine(trainbrakescontrollermaxsystempressure":
+                case "engine(trainbrakescontrollermaxoverchargepressure":
```
YoRyan, line 755: New properties should be prefixed with `orts`.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -752,8 +752,10 @@ public override void Parse(string lowercasetoken, STFReader stf)
...
                 case "engine(trainbrakescontrollermaxreleaserate":
                 case "engine(trainbrakescontrollermaxquickreleaserate":
+                case "engine(trainbrakescontrolleroverchargeeliminationrate":
```
YoRyan, line 758: As above.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -448,6 +448,14 @@ public override void Update(float elapsedClockSeconds)
                     AuxResPressurePSI += dp;
                     BrakeLine1PressurePSI -= dp * AuxBrakeLineVolumeRatio;  // Adjust the train brake pipe pressure
                 }
+                if (AuxResPressurePSI > BrakeLine1PressurePSI) // Allow small flow from auxiliary reservoir to brake pipe to restore normal pressure after overcharge
+                {
+                    float dp = elapsedClockSeconds * 0.05f;
```
YoRyan, line 453: Please explain the justification for the number `0.05f` and move it to a constant with a descriptive name.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
```diff
@@ -195,6 +201,10 @@ public void Parse(string lowercasetoken, STFReader stf)
                     MaxPressurePSI = stf.ReadFloatBlock(STFReader.UNITS.PressureDefaultPSI, null);
                     break;
 
+                case "engine(trainbrakescontrollermaxoverchargepressure":
```
YoRyan, line 204: Needs the `orts` prefix.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
```diff
@@ -205,6 +215,10 @@ public void Parse(string lowercasetoken, STFReader stf)
                     QuickReleaseRatePSIpS = stf.ReadFloatBlock(STFReader.UNITS.PressureRateDefaultPSIpS, null);
                     break;
 
+                case "engine(trainbrakescontrolleroverchargeeliminationrate":
```
YoRyan, line 218: As above.

YoRyan (dismissed):
Thanks for your contribution. To be merged, this PR will also require blueprint approval and a manual update.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: While I know all new properties have to start with ORTS, I was in doubt with braking parameters, as many of them have been added without ORTS prefix (I mentioned it in ET, as I wasn't sure with the naming). Thanks for clarifying this.

YoRyan: I think the enums are fine without the prefix, but for anything under the `engine(` block I would expect it.

cesarBLG:
I think it's worth adding all brake controller positions and their meanings to the manual. I'll do that first, and then I'll include overcharge.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -448,6 +449,14 @@ public override void Update(float elapsedClockSeconds)
                     AuxResPressurePSI += dp;
                     BrakeLine1PressurePSI -= dp * AuxBrakeLineVolumeRatio;  // Adjust the train brake pipe pressure
                 }
+                if (AuxResPressurePSI > BrakeLine1PressurePSI) // Allow small flow from auxiliary reservoir to brake pipe to restore normal pressure after overcharge
```
peternewell, line 452: Given the fact that overcharging is not present in all air brake systems, I would suggest that an additional "switching factor" be used.

For example, something like "if (AuxResPressurePSI > BrakeLine1PressurePSI && OverChargingActive)"

where OverChargingActive = a parameter that is set when the user enters overcharging parameters in the ENG/WAG file.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -686,17 +695,8 @@ protected static void PropagateBrakeLinePressures(float elapsedClockSeconds, Tra
...
-                            }
-                           else
-                            {
-                                car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipePressureDiffPropogationPSI * brakePipeVolumeM30 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
-                                car0.BrakeSystem.BrakeLine1PressurePSI += TrainPipePressureDiffPropogationPSI * car.BrakeSystem.BrakePipeVolumeM3 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
-                            }                      
+                            car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipePressureDiffPropogationPSI * brakePipeVolumeM30 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
+                            car0.BrakeSystem.BrakeLine1PressurePSI += TrainPipePressureDiffPropogationPSI * car.BrakeSystem.BrakePipeVolumeM3 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
```
peternewell, line 699: From memory, as the pressure change is propagated along the train pipe a certain amount of change happens in each of the cars (as suggested by + and - statements). If the locomotive pressure is not clamped, then its brake pipe pressure can exceed the maximum pressure (this maybe what you are trying to achieve in your change, but in a normal system it may not be the case).

Changes to this section of code may impact on the propagation delay along the train pipe, so any changes should be tested as a before and after to make sure that timings have not varied.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: Is there any technical reason which justifies not having that reverse flow in all triple valves? It is so small that shouldn't affect brake application, and only occurs if valve is in release position.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: I haven't found any difference in brake pipe charging and discharging times. Note that when applying and releasing brakes, brake pipe pressure will be below maximum pressure, so the clamp will have no effect. Also, TrainPipePressureDiffPropagationPSI can't be greater than the difference of pressures, so the pressures won't increase magically.

However, I have seen something which might cause problems:
`float TrainPipePressureDiffPropogationPSI = Math.Min (TrainPipeTimeVariationS * (p1 - p0) / brakePipeTimeFactorS, p1 - p0);`
This is OK when p1 > p0, but when p1 < p0 I don't think this code is doing what the writer expected. What's your opinion about it?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: Without studying a wide range of triple valve types, I can't confirm that this leakage occurs. Certainly the documentation that I have looked at quickly does not appear to support overcharging, and hence talks about the Aux Res flow stopping when it is equal to the train pipe pressure.

The size of flow is immaterial, as philosophically I would prefer to apply functionality only to the features that are applicable to a particular equipment type. So as not all triple valves appear to have supported overcharging, I would still suggest that the feature only be enabled when overcharging is required rather then generically across all types of equipment.

As an aside, given the variation in operation of the different brake components used over the years, I would prefer to see documentation on how to set up a particular brake type, as this would assist OR users and eliminate confusion. For example, an A6-ET brake is different to a 26-L, etc. Hence some of the tokens, and "other variations" will need to be defined and coded differently. So it would be good if your brake system was described, and then coded as required.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: > "This is OK when p1 > p0, but when p1 < p0 I don't think this code is doing what the writer expected. What's your opinion about it?"

I am sorry but I don't understand the issue, as I have always interpreted it as creating either a -ve or +ve number depending upon whether the train pipe is charging or releasing. There appears to be no separate propagation code for releasing/charging, and hence the one code section needs to handle both. The Min statement is just taking minimum +ve or -ve number to use as the difference value.

> "Note that when applying and releasing brakes, brake pipe pressure will be below maximum pressure, so the clamp will have no effect."

Is this the case, the way that I have interpreted the code, is as follows.

When the brakes are fully released, the BP = MaxPressure, lets say 70psi. So for the locomotive (car0), OR process the following line of code,

`car0.BrakeSystem.BrakeLine1PressurePSI += TrainPipePressureDiffPropogationPSI * car.BrakeSystem.BrakePipeVolumeM3 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
`
So in this instance wouldn't we get the following result, BP = BP + Diff = 70+ psi? The clamp statement would then reduce the loco value to 70psi? Hence it is providing a form of feedback for the locomotive pressure.

Other cars in the train are treated differently, ie only the subtraction statement is used. (Either it goes up or down depending upon whether Diff is -ve or +ve.)

Unfortunately, at the moment I haven't got time to get my head fully back around this code, so I am not able to fully document any specific impacts, apart from the "vague" memory and comments already made for these proposed changes. Suffice to say that like a lot of other code in OR, it is quite complex, and often very hard to understand the logic used for its creation by the original developer. Typically I have found it better to leave code alone unless I can fully understand the logic for its original creation.

cesarBLG: Consider the situation of brake release. Brake pipe pressure in car0 (lead locomotive) will be 70 psi, and pressure in another car will be lower than 70, e. g., 60 psi. Then, p1-p0 = -10 psi. This illustrates the two points I'm trying to tell:
- First: `p1-p0` is -10, but `TrainPipeTimeVariationS * (p1 - p0) / brakePipeTimeFactorS` will be around -9 psi. The minimum of both magnitudes is -10. However, if both were positive, the minimum would be 9. I'm pretty sure that what the original author meant was: 
`float TrainPipePressureDiffPropogationPSI = (p1>p0 ? 1 : -1) * Math.Min(TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));`
The way it is written now it's not symmetric, but one would expect symmetry from a physical system (the brake pipe) which has no preferred direction.

- Second: p1-p0 < 0, so TrainPipePressureDiffPropagationPSI < 0. Therefore, pressure in car0 will be decreased, and increased in car1. The algorithm used here works by the principle of pressure equalization between adjacent cars, so air will flow from the car with higher pressure to the car with lower, never the other way round.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: This leakage is more related to brake insensitivity than to overcharge. I haven't been able to found any document for triple valves, but for distributors UIC regulation says that brakes shouldn't be applied if brake pipe pressure is decreased at a lower rate than 0.3 bar in 60 seconds. What do you think about creating a `Wagon(ORTSBrakeInsensitivity(` parameter to regulate this flow? Note that this brake insensitivity is also needed in order to not apply brakes due to a small brake pipe leakage.

peternewell: > "I haven't been able to found any document for triple valves"

I looked at a number of manuals when trying to add some features to the brake code. Here is but one example -

https://archive.org/details/westinghouseairb00inteuoft

This publication describes a number of different types of triple valves. Naturally there maybe more modern or non US types of valves which will have variations to those described.

By the time the different triple valves, brake control valves, other functions, such as load drop compensation, etc are considered, the number of variations can be reasonably large.

> "Note that this brake insensitivity is also needed in order to not apply brakes due to a small brake pipe leakage."

This is perhaps a more modern feature that may not have been present in older systems? In non self-lapping systems some level of leakage occurred in the brake pipe, and this gradually allowed the brakes to be applied.

Overall, my personal feeling is that the current "default" brake functionality should remain untouched as it provides a "reasonable" base foundation. Extra functionality can then be added on top of it to customise the operation to a particular equipment type or era.

My concern was not with the functionality that you were proposing (as I assumed that you had done sufficient research to ensure that it was correct), but more with the fact that it was also possibly changing the base functionality.

> "What do you think about creating a Wagon(ORTSBrakeInsensitivity( parameter to regulate this flow?"

I have no problems with this if you feel it is necessary for the functionality that you are adding, and it is an addon on top of the base functionality, rather then a change to the base code. 

If you wish to discuss anything other information in detail, how about sending me a PM?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: > "The way it is written now it's not symmetric, but one would expect symmetry from a physical system (the brake pipe) which has no preferred direction."

I would agree with your thoughts in regard to the symmetry, and would be happy to see the Diff value changed accordingly. It should have a developer comment describing the intent of the action for future developers. 

I don't think that the suggested change will work. It may also require an ABS function around the first factor as well.

> "Second: p1-p0 < 0, so TrainPipePressureDiffPropagationPSI < 0."

If p0 = 60, and p1 = 70, ie brake pipe needs to reduce pressure along the train. In this case p1-p0 > 0, so TrainPipePressureDiffPropagationPSI > 0. Therefore, pressure in car0 will be increased, and decreased in car1. Whilst the calculation of the Diff value and the Volume offsetting may mean it might not go above the MaxPressure, it will probably still be good to retain the Clamp statement as a safeguard.

cesarBLG: The clamp is what prevents the brake pressure from being overcharged, which is what I'm trying to achieve. I have studied the formula and I'm sure that pressure won't increase over the maximum in other situations. If that happens, it will be in another part of the code, and it should be identified and fixed there.
There are some vehicles where this clamp has sense, as they are overcharge-protected, but this should be set on a per car basis, and not just for the first car in the train.

peternewell (commented):
I am sorry but I hit the wrong button when responding, so I am not sure how to undo what I did.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: > "I have studied the formula and I'm sure that pressure won't increase over the maximum in other situations. If that happens, it will be in another part of the code, and it should be identified and fixed there."

    As long as the current overcharge-protected functionality is preserved, then it is fine for the code changes. However it would be good to validate the fact that it does/doesn't need to be corrected elsewhere in the code through some form of testing, etc, rather then "breaking" the code, and hoping that somebody notices it.

> "There are some vehicles where this clamp has sense, as they are overcharge-protected, but this should be set on a per car basis, and not just for the first car in the train."

I am sorry but I don't understand this.

Given the fact that we are talking about clamping the brake pipe at a max pressure, and that the locomotive is the only car in the train that controls changes in the brake pipe (apart from minor leakage), why would it be necessary to set it on a per car basis? The locomotive is the overall pressure reference for the brake pipe on the whole train, and the pressure in all other cars should follow accordingly.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: Thanks for the link! It's really useful.
I have added the parameter in order to keep things as they were by default.
If I have more doubts I'll write you.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
```diff
@@ -168,8 +172,6 @@ public override void UpdatePressure(ref float pressureBar, float elapsedClockSec
                 }
             }
 
-            if (pressureBar > MaxPressureBar())
```
peternewell, line 171: Sorry another couple of queries.

If these lines are removed, does that allow "overcharge-protected" tokens to exceed the MaxPressureBar?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
```diff
@@ -104,13 +105,16 @@ public override void UpdatePressure(ref float pressureBar, float elapsedClockSec
                     switch (notch.Type)
                     {
                         case ControllerState.Release:
-                            pressureBar += x * ReleaseRateBarpS() * elapsedClockSeconds;
-                            epState = -1;
-                            break;
                         case ControllerState.FullQuickRelease:
```
peternewell, line 108: Is there a benefit in merging the Release and FullQuickRelease tokens together. I would prefer them to remain separate as it make the code easier to "read", and doesn't imply that the have the same functionality.

Also does the addition of the DecreasePressure code imply that they are no longer overcharge-protected tokens?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
cesarBLG: I modified release and quick release positions to ensure that EQ pressure never goes over MaxPressureBar. Pressure only increases above maximum while overcharge is active.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
cesarBLG: I'll keep them separate then.
If overcharge is present (this will only happen if Overcharge is defined and selected), it has to be eliminated really slowly to avoid application of brakes, so that's the reason for using DecreasePressure. If no overcharge token is defined, pressure will be equal or lower than MaxPressure and DecreasePressure will have no effect.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: > As long as the current overcharge-protected functionality is preserved, then it is fine for the code changes. However it would be good to validate the fact that it does/doesn't need to be corrected elsewhere in the code through some form of testing, etc, rather then "breaking" the code, and hoping that somebody notices it.

I have checked all formulas and pressure will never increase over the maximum without overcharge feature. However pressure can become negative, so I have added some checks.

> Given the fact that we are talking about clamping the brake pipe at a max pressure, and that the locomotive is the only car in the train that controls changes in the brake pipe (apart from minor leakage), why would it be necessary to set it on a per car basis? The locomotive is the overall pressure reference for the brake pipe on the whole train, and the pressure in all other cars should follow accordingly.

In my opinion, if overcharge protection (i. e. pressure clamp) were implemented it should be done depending on each car. There may be the case where the locomotive tries to overcharge the brake pipe but some of the wagons are overcharge protected. The pressure should be clamped when at least one vehicle has this protection. Vehicles with overcharge protection are not a common situation, though.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
peternewell: > "I'll keep them separate then."

Thanks for that.

Have you seen the document in this post? - 

http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__264462

On pg 14 for the Knorr brake it is proposing an alternate token to follow the overcharging one. 

I am not sure whether anything needs to be done to the Neutral token, as it is supposed to isolate the pipe from the brake valve. However perhaps it is worth you posing a question in the thread, and thus any necessary changes could be done as part of the changes that you are adding.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: My understanding of a non-overcharge triple valve is it just follows the train brake pipe pressure. In railway operating instructions I have seen, it is suggested that the train brake pipe pressure can be varied by the crew depending upon the type of train that it is. Hence the triple valve works at these different pressures without any individual alterations, and in this instance the  brake pipe pressure sets the maximum operating pressure of the triple valve, and thus would only need to be clamped on the locomotive. 

At the moment I can't speak for the (overcharge) triple valves that you have described. Do you have any documents that you can point me to?

cesarBLG: I don't think there exist such 'overcharge triple valves'. Overcharge is a situation that might happen, and all triple valves will react to it in a way or another. Some of them will have overcharge protections, so as soon as the BP pressure is greater (e. g. 5.3 bar) than aux pressure (say 5.1 bar) the valve will release the brakes but the auxiliary pressure won't be increased. Other valves (generally older ones) will simply equalize both pressures. In both situations, if the auxiliary pressure is higher than BP's, it should slowly decrease while the valve is in release position, so in the end auxiliary pressure is equal to normal system pressure (this is what I have introduced with brake insensitivity parameter, see https://rail21.pagesperso-orange.fr/Freinage_Distributeur_EN.pdf for the timings). If auxiliary reservoir pressure weren't decreased, then brakes would be applied even if the controller was in release position. This is what happens if overcharge elimination process is interrupted, leading to brakes being wrongly applied. Some systems (https://patents.google.com/patent/US6375275B1/en) will act if overcharge is not correctly eliminated.

There are two types of overcharge:
 - Undesired overcharge: usually happens when the brake pipe is connected for a long time to the main reservoir in order to achieve a fast brake pipe charging.
 - Intentional overcharge: the driver can overcharge the brake pipe on purpose to release stuck brakes, especially in twin pipe systems. For example, if a locomotive with max pressure of 5.2 bar is replaced with a 4.9 bar one, the control reservoirs in brake distributors remain at 5.2 bar, so the new locomotive can't fully release all brakes. The driver momentarily increases pressure in BP over 5.3 bar, so the brakes are released. Then, the pressure starts decreasing very slowly, and the control reservoir pressure is finally 4.9 bar, so the new locomotive can operate the brakes normally. Something similar happens in single pipe systems.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
cesarBLG: Currently Neutral position does not isolate EQ from BP, so this has to be fixed. It's pretty straightforward to do it, but it's off-topic here I think.
Which is the other token you mention?
I have posted a message in that thread.

peternewell: That was the only token.

It would be good to tidy it all up in the one go.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
peternewell: Thanks for that.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
peternewell: Thanks for the comments.

If I understand what you are saying, the triple valve needs to be treated differently depending upon which type it is:

i) Older Style - will only respond to the train brake pipe. In this type the pressure should decrease at the "current" release rate, as it is merely following the train brake pipe pressure (as it has always done).

ii) "Overcharge" style - will behave as you have described, and decrease at the rates that you have indicated.

So perhaps the only difference between the two will be the decrease rate value.

cesarBLG: Current implementation causes auxiliary reservoir pressure to not decrease at all if the valve is in Release position. If it decreased following BP pressure, brakes would never be applied. Therefore, the flow from auxiliary reservoir to brake pipe, if any, has to be slow.

I don't think it's necessary to have two different types of triple valves. If you have never found any triple valve with overcharge protections, they must be really uncommon, if they actually exist.

The thing here is, how do _normal_ triple valves react to an overcharge? They aren't able to differentiate overcharge from normal releases, as they have no control reservoir to check against. Therefore, they will charge auxiliary reservoir with a pressure equal to brake pipe's. Unless I'm missing some weird exception, this happens with all triple valves.

Another different question is: how do triple valves react to a really slow decrease of BP pressure, while in Release position? This situation happens both when overcharge is eliminated and due to small leakages. When the valve is in Lap position, the answer is simply that brakes are applied until BP and aux have the same pressure again. However, if the valve is in Release position, the UIC allowing a 'brake pipe insensitivity' suggests that brakes aren't applied if the pressure reduction is small.

Distributors know when the brake pipe is overcharged, as BP pressure will be higher than control reservoir pressure. Some of them might react sealing the reservoirs so only the BP is overcharged (this is the only example of overcharge-protected equipment). Currently OR uses emergency reservoir as a fixed pressure control reservoir, so it won't be affected by overcharge. However, the behaviour of the control reservoir in OR seems to have some differences with what I have read in manuals.

peternewell:
I have a couple of trains with air brakes that work ok in version Stable version 1.3.1 and don't appear to work in the latest unstable version. The issue does not appear to be present in my development environment which currently doesn't have your recently introduced overcharge code changes.

It appears that the EQ pressure goes above the maximum system pressure set in the ENG file of 70psi.

To see the issue, you can download the following route (stock and activities included). 

http://www.zigzag.coalstonewcastle.com.au/route/downloads/

Run Tutorial #3, and try releasing and applying the brakes. It appears that the EQ pressure is overcharging.

Can you have a quick look at this in relation to your changes, and confirm that the issue with the brakes not releasing correctly is not associated with your changes because the brake tokens used in the ENG files should not be able to overcharge.

Have you done any testing on other "legacy" setups to ensure that the changes haven't impacted them?

Thanks

cesarBLG:
It was affecting controllers initially set at Running, Lap or Hold positions. The equalizing reservoir pressure was initialized to a fixed value of 90 psi instead of setting it to a system dependent max pressure.

peternewell:
Thanks for that.

I am unable to confirm that all is ok now, as the latest unstable version is still at 25/11/2020.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -679,44 +690,41 @@ protected static void PropagateBrakeLinePressures(float elapsedClockSeconds, Tra
...
-                            }
-                           else
-                            {
-                                car.BrakeSystem.BrakeLine1PressurePSI -= TrainPipePressureDiffPropogationPSI * brakePipeVolumeM30 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
-                                car0.BrakeSystem.BrakeLine1PressurePSI += TrainPipePressureDiffPropogationPSI * car.BrakeSystem.BrakePipeVolumeM3 / (brakePipeVolumeM30 + car.BrakeSystem.BrakePipeVolumeM3);
-                            }                      
+                            // Based on the principle of pressure equualization between adjacent cars
+                            // First, we define a variable storing the pressure diff between cars, but limited to a maximum flow rate depending on pipe characteristics
+                            // The sign in the equation determines the direction of air flow.
+                            float TrainPipePressureDiffPropogationPSI = (p0>p1 ? -1 : 1) * Math.Min (TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));
```
YoRyan, line 698:
```suggestion
                            float TrainPipePressureDiffPropogationPSI = (p0>p1 ? -1 : 1) * Math.Min(TrainPipeTimeVariationS * Math.Abs(p1 - p0) / brakePipeTimeFactorS, Math.Abs(p1 - p0));
```

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
```diff
@@ -142,8 +144,10 @@ public ScriptedBrakeController(MSTSLocomotive locomotive)
             Locomotive = locomotive;
 
             MaxPressurePSI = 90;
+            MaxOverchargePressurePSI = 95;
```
YoRyan, line 147: Sane defaults, I presume, or...?

YoRyan (commented):
I cannot speak to the correctness of the brake simulation, but I can say that the style looks good and this is almost ready to go.

Barring any sudden discoveries of bugs in the physics model, I think this is ready to merge.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
cesarBLG: A typical overcharge application increases pressure in 5 psi above max pressure. It is intended that content creators use the related .eng parameters to set appropriate values for those variables. For existing stock the brake controller doesn't have an "Overcharge" position, so it will be unused.

YoRyan (approved):
Approving and merging if there are no more unintended behavioral changes.

cjakeman:
Hi Ryan,
 
> Approving and merging if there are no more unintended behavioral changes.

It looks as though you approved but forgot to merge, so I'm doing it for you.

Best wishes,

Chris

PR #301, opened 2020-11-06
Extend path up to 5 km or first signal, and clear depending on SNCA

cesarBLG:
https://bugs.launchpad.net/or/+bug/1903196
There are some issues with the calculation of ValidRoute in explorer mode. In some cases, path is too short and there is no information in TrackMonitor and TCS scripts. In other situations, path is extended up to hundreds of kilometers, which will cause signals 100 km away to be cleared, making difficult multiplayer operations.
I mentioned it some time ago here: http://www.elvastower.com/forums/index.php?/topic/33930-improving-calculation-of-train-path-in-explorer-mode/

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -247,6 +247,7 @@ namespace Orts.Simulation.Physics
         public float maxTimeS = 120;                     // check ahead for distance covered in 2 mins.
         public float minCheckDistanceM = 5000;           // minimum distance to check ahead
         public float minCheckDistanceManualM = 3000;     // minimum distance to check ahead in manual mode
+        public float minCheckDistanceExplorerM => Math.Max(AllowedMaxSpeedMpS * 240, 5000);      // minimum distance to check ahead in explorer mode
 
         public float standardOverlapM = 15.0f;           // standard overlap on clearing sections
         public float junctionOverlapM = 75.0f;           // standard overlap on clearing sections
@@ -9012,7 +9013,7 @@ namespace Orts.Simulation.Physics
                 List<int> tempSections = new List<int>();
 
                 tempSections = signalRef.ScanRoute(this, requiredPosition.TCSectionIndex, requiredPosition.TCOffset,
-                        requiredPosition.TCDirection, forward, minCheckDistanceM, true, false,
+                        requiredPosition.TCDirection, forward, minCheckDistanceExplorerM, true, false,
                         true, false, true, false, false, false, false, IsFreight);
 
                 if (tempSections.Count > 0)
@@ -9098,10 +9099,10 @@ namespace Orts.Simulation.Physics
 
             // if route does not end with signal and is too short, extend
 
-            if (!endWithSignal && totalLengthM < minCheckDistanceM)
+            if (!endWithSignal && totalLengthM < minCheckDistanceExplorerM)
             {
 
-                float extendedDistanceM = minCheckDistanceM - totalLengthM;
+                float extendedDistanceM = minCheckDistanceExplorerM - totalLengthM;
                 TCRouteElement lastElement = newRoute[newRoute.Count - 1];
 
                 int lastSectionIndex = lastElement.TCSectionIndex;
@@ -9327,7 +9328,7 @@ namespace Orts.Simulation.Physics
                         int numCleared = 0;
                         totalLengthM = 0;
                         offsetM = direction == 0 ? requiredPosition.TCOffset : thisSection.Length - requiredPosition.TCOffset;
-                        for (int iindex = 0; iindex < newRoute.Count && (firstSignalPassed || totalLengthM < minCheckDistanceM) && (!firstSignalPassed || numCleared != 0); iindex++)
+                        for (int iindex = 0; iindex < newRoute.Count && (firstSignalPassed || totalLengthM < minCheckDistanceExplorerM) && (!firstSignalPassed || numCleared != 0); iindex++)
                         {
 
                             thisSection = signalRef.TrackCircuitList[newRoute[iindex].TCSectionIndex];
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -247,7 +247,17 @@ namespace Orts.Simulation.Physics
         public float maxTimeS = 120;                     // check ahead for distance covered in 2 mins.
         public float minCheckDistanceM = 5000;           // minimum distance to check ahead
         public float minCheckDistanceManualM = 3000;     // minimum distance to check ahead in manual mode
-        public float minCheckDistanceExplorerM => Math.Max(AllowedMaxSpeedMpS * 240, 5000);      // minimum distance to check ahead in explorer mode
+        /// <summary>
+        /// Minimum distance to check ahead in explorer mode
+        /// </summary>
+        private float MinCheckDistanceExplorerM
+        {
+            get
+            {
+                int checkTimeS = 120, checkDistanceM = 5000;
+                return Math.Max(AllowedMaxSpeedMpS * checkTimeS, checkDistanceM);
+            }
+        }
 
         public float standardOverlapM = 15.0f;           // standard overlap on clearing sections
         public float junctionOverlapM = 75.0f;           // standard overlap on clearing sections
@@ -7894,8 +7904,7 @@ namespace Orts.Simulation.Physics
                 MisalignedSwitch[direction, 0] = -1;
                 MisalignedSwitch[direction, 1] = -1;
 
-                List<int> tempSections = new List<int>();
-                tempSections = signalRef.ScanRoute(this, requiredPosition.TCSectionIndex, requiredPosition.TCOffset,
+                List<int> tempSections = signalRef.ScanRoute(this, requiredPosition.TCSectionIndex, requiredPosition.TCOffset,
                         requiredPosition.TCDirection, forward, minCheckDistanceManualM, true, false,
                         true, false, true, false, false, false, false, IsFreight);
 
@@ -9010,10 +9019,8 @@ namespace Orts.Simulation.Physics
 
                 // build new route
 
-                List<int> tempSections = new List<int>();
-
-                tempSections = signalRef.ScanRoute(this, requiredPosition.TCSectionIndex, requiredPosition.TCOffset,
-                        requiredPosition.TCDirection, forward, minCheckDistanceExplorerM, true, false,
+                List<int> tempSections = signalRef.ScanRoute(this, requiredPosition.TCSectionIndex, requiredPosition.TCOffset,
+                        requiredPosition.TCDirection, forward, MinCheckDistanceExplorerM, true, false,
                         true, false, true, false, false, false, false, IsFreight);
 
                 if (tempSections.Count > 0)
@@ -9099,10 +9106,10 @@ namespace Orts.Simulation.Physics
 
             // if route does not end with signal and is too short, extend
 
-            if (!endWithSignal && totalLengthM < minCheckDistanceExplorerM)
+            if (!endWithSignal && totalLengthM < MinCheckDistanceExplorerM)
             {
 
-                float extendedDistanceM = minCheckDistanceExplorerM - totalLengthM;
+                float extendedDistanceM = MinCheckDistanceExplorerM - totalLengthM;
                 TCRouteElement lastElement = newRoute[newRoute.Count - 1];
 
                 int lastSectionIndex = lastElement.TCSectionIndex;
@@ -9126,7 +9133,7 @@ namespace Orts.Simulation.Physics
                     }
                 }
 
-                List<int> tempSections = new List<int>();
+                List<int> tempSections = null;
 
                 if (nextSectionIndex >= 0 && MisalignedSwitch[direction, 0] < 0)
                 {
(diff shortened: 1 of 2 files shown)


pzgulyas:
I just repeat my posts in PR #293, because it is offtopic there, and belongs here.

César, is your PR in line with [this ET post](http://www.elvastower.com/forums/index.php?/topic/34297-mg71-crash-index-was-out-of-range-must-be-non-negative-and-less-than-the-size-of-the-collection/page__view__findpost__p__262475) I mentioned before? Your PR seem to reintroduce the minCheckDistanceM constant for use in the explorer mode, however Rob writes the following in his post:

_"Then to explorer mode. This same constant (minCheckDistanceM) was used - or abused, if you like - to set up the limits for the checks in explorer mode as explained above. But this certainly is not the main function of this constant.
So, if a change is needed in explorer mode to extend the search, the first thing to do is to drop the use of this constant. This use is not what this constant is intended for, so leave it to its intended function and set up another limit check for explorer."_

Also:
_"The best option, however, is simply to remove the scan restriction alltogether. By removing the constant minCheckDistanceM from the call to ScanRoute and replacing it by the value -1, the scan is performed for the full (remaining) route of the train, which is similar as to its use in other modes."_

It is in contrast with your PR which specifically does the opposite: removes -1 from the ScanRoute function, and adds minCheckDistanceM.

cesarBLG:
I restored maxDistanceM because that was used before #270, but I am not modifying it, which was Rob's concern. I will create a new variable. I used that because I had been testing explorer mode before #270 and Rob's comment, when it was still used.
I don't know if there is a point in having a long route which in fact will probably be modified by MP dispatcher. I would like to hear Rob's opinion about how long ValidRoute should be.
ScanRoute sometimes returns a short route, and it has to be extended in order to have enough information for TrackMonitor and TCSs, so a minimum distance has to be defined (so maxDistance might not be the best naming).
Another distance has to be checked: if the train is far away from the first signal, there is no point on clearing it before it has approached.

pzgulyas:
By Rob's comment, if you set that parameter to -1, then the searching distance will be unlimited. This is exactly what #270 did.

pzgulyas:
If you would like to possibly hear Rob's comments, you need to mention him like this: @Roeterdink. Otherwise he would never find this post. :-)

cesarBLG:
If the idea is to extend ValidRoute to the maximum possible distance, more changes have to be done, as this is not achieved with #270. In my opinion it would be better to have a ValidRoute which only covers the distance up to the signal whose aspect doesn't affect the train.

cesarBLG:
I have been thinking about it and now I'm not sure. Extending the path up to an end of track seems to have a huge overhead in long (over 1000 km routes), but I have found a case where my proposal isn't valid: ETCS overlaps and danger points, which require information further than the first uncleared signal.

Sharpe49:
I have tested this fix on conventional and high speed lines. Works perfectly.
I have not yet tested on lines with really long block sections (15 km for example).

Sharpe49:
Alright, tested the new commit on both high speed lines and conventional lines (both with normal block sections or long block sections), still works perfectly on my side.

cesarBLG:
@Roeterdink what's your opinion about this approach? If you feel this is OK, I think it's ready to merge. Having a ValidRoute extended up to the end of the path becomes quite difficult in explorer mode and needs several changes in the logic used.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -247,6 +247,7 @@ public enum TRAINTYPE
         public float maxTimeS = 120;                     // check ahead for distance covered in 2 mins.
         public float minCheckDistanceM = 5000;           // minimum distance to check ahead
         public float minCheckDistanceManualM = 3000;     // minimum distance to check ahead in manual mode
+        public float minCheckDistanceExplorerM => Math.Max(AllowedMaxSpeedMpS * 240, 5000);      // minimum distance to check ahead in explorer mode
```
YoRyan, line 250:
```suggestion
        /// <summary>
        /// Minimum distance to check ahead in explorer mode.
        /// </summary>
        private float minCheckDistanceExplorerM
        {
            get
            {
                const float checkTimeS = 240, checkDistanceM = 5000;
                return Math.Max(AllowedMaxSpeedMpS * checkTimeS, checkDistanceM);
            }
        }
```
A few things I'd like to see changed here:
- We want to avoid public instance variables if we can help it.
- Annotate the `240` and `5000` constants.
- Move the variable description to a proper XML comment. This is required to get help text working in Intellisense; it's not just a stylistic choice. (Yes, almost all of our current code violates this standard, but you've got to start somewhere.)

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
```diff
@@ -10384,6 +10361,39 @@ public TrackMonitorSignalAspect TranslateTMAspect(MstsSignalAspect SigState)
...
+                {
+                    ReqNumClearAhead = 0;
+                }
+                else
+                {
+                    ReqNumClearAhead = isPropagated ? signalNumClearAhead - 1 : SignalNumClearAheadActive - 1;
+                }
+            }
+
+            return ReqNumClearAhead;
```
YoRyan, line 10395: Can be simplified to
```suggestion
            if (SignalNumClearAhead_MSTS > -2)
                return propagated ? signalNumClearAhead - SignalNumNormalHeads : SignalNumClearAhead_MSTS - SignalNumNormalHeads;
            else if (SignalNumClearAheadActive == -1)
                return propagated ? signalNumClearAhead : 1;
            else if (SignalNumClearAheadActive == 0)
                return 0;
            else
                return isPropagated ? signalNumClearAhead - 1 : SignalNumClearAheadActive - 1;
```

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -9257,12 +9318,56 @@ public void UpdateExplorerMode(int signalObjectIndex)
...
+                    if (unclearedSignal && signalIndex < newRoute.Count)
                     {
                         SignalObject reqSignal = signalRef.SignalObjects[nextUnclearSignalIndex];
-                        newRoute = reqSignal.requestClearSignalExplorer(newRoute, forward ? routedForward : routedBackward, false, 0);
+                        bool firstSignalPassed = false;
+                        int numCleared = 0;
+                        totalLengthM = 0;
+                        offsetM = direction == 0 ? requiredPosition.TCOffset : thisSection.Length - requiredPosition.TCOffset;
+                        for (int iindex = 0; iindex < newRoute.Count && (firstSignalPassed || totalLengthM < minCheckDistanceExplorerM) && (!firstSignalPassed || numCleared != 0); iindex++)
+                        {
```
YoRyan, line 9332:
```suggestion
                        for (int iindex = 0; iindex < newRoute.Count; iindex++)
                        {
                            var continue1 = firstSignalPassed || totalLengthM < minCheckDistanceExplorerM;
                            var continue2 = !firstSignalPassed || numCleared != 0;
                            if (!continue1 || !continue2)
                                break;
```
That's a lot of booleans to keep track of, there; please consider this style instead, with more descriptive names for "continue1" and "continue2".

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -9092,6 +9097,62 @@ public void UpdateExplorerMode(int signalObjectIndex)
...
+                if (nextSection != null && nextSection.CircuitType == TrackCircuitSection.TrackCircuitType.Junction)
+                {
+                    if (nextSection.Pins[0, 0].Link != lastSectionIndex && nextSection.Pins[0, 1].Link != lastSectionIndex && nextSection.Pins[1, nextSection.JunctionLastRoute].Link != lastSectionIndex)
+                    {
+                        MisalignedSwitch[direction, 0] = nextSection.Index;
+                        MisalignedSwitch[direction, 1] = lastSectionIndex;
+                    }
+                }
+
+                List<int> tempSections = new List<int>();
```
YoRyan, line 9129:
```suggestion
                List<int> tempSections;
```
This unnecessary initialization incurs a slight performance penalty.

YoRyan (changes requested):
Thanks for tackling this. Again, I cannot speak to the correctness of the signal logic - any reason we have to use special logic for explore the route mode, versus manual mode? - so here are my suggestions regarding coding style.

Source/Orts.Simulation/Simulation/Physics/Train.cs
cesarBLG: Do I change the case of the property name?

YoRyan: The C# guideline (enforced by Visual Studio) is that instance variable names need to use PascalCase, but *technically* that only applies to public, not private, variable names.

Still, I think it would be a good idea to change the case. Do use the auto-rename tool - I can't make that change in a GitHub suggestion.

Source/Orts.Simulation/Simulation/Physics/Train.cs
cesarBLG: I didn't notice that, thanks for pointing. I have changed that in manual mode too.

cesarBLG:
I have found this piece of code in ProcessExplorerSwitch():
```csharp
//<CSComment> Probably also in singleplayer the logic of multiplayer should be used, but it's unwise to modify it just before a release
TrainRouted thisRouted = direction == Direction.Reverse ^ !MPManager.IsMultiPlayer() ? routedBackward : routedForward;
```
Maybe it's time to use the same logic in singleplayer. I don't see any reason to use routedBackward when the direction is forward.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
Roeterdink: This suggestion shows the difference between software developers and software testers.
Developers like compressed code, but testers prefer singular flows. I've spend most of my professional life as software test engineer, and in that time I build up a firm dislike of multiple-exit methods as it often makes debugging more complex.

Roeterdink:
> 
> 
> I have found this piece of code in ProcessExplorerSwitch():
> 
> ```cs
> //<CSComment> Probably also in singleplayer the logic of multiplayer should be used, but it's unwise to modify it just before a release
> TrainRouted thisRouted = direction == Direction.Reverse ^ !MPManager.IsMultiPlayer() ? routedBackward : routedForward;
> ```
> 
> Maybe it's time to use the same logic in singleplayer. I don't see any reason to use routedBackward when the direction is forward.

In both Manual and Explorer mode, all actions regarding clearance were originally performed in both directions, allowing free movement of the trains in either direction. The 'TrainRouted' holds the information for the proper direction in which the train is presently moving. To remove this would restrict explorer mode to forward direction only, and I do not think this would go down well with many users who prefer 'pathless' running in explorer mode.

cesarBLG:
I'm one of the users of 'pathless' explorer mode, so I obviously want it working in both directions. What I meant was to change this:
```diff
- TrainRouted thisRouted = direction == Direction.Reverse ^ !MPManager.IsMultiPlayer() ? routedBackward : routedForward;
+ TrainRouted thisRouted = direction == Direction.Reverse ? routedBackward : routedForward;
```
As it's written now, trainRouted will be opposite to train's direction in single player mode.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
Roeterdink: Another point about this change.
This logic is now part of method TranslateTMAspect. This method was created to translate the signal enumeration to the cabview enumeration, and as such it has no function within the signal logic.
Admittedly I do not know if the code w.r.t. SNCA was introduced as part of this commit or if it was introduced earlier, for I have not kept up with the changes made to OR for quite some time now.
But this logic has no place in this method. As it has now been collected in a separate method, I would suggest to remove the call to this method from TranslateTMAspect and move is to the location where TranslateTMAspect is called, and keep TranslateTMAspect clean as translation between enums only.

Roeterdink:
> 
> 
> Alright, tested the new commit on both high speed lines and conventional lines (both with normal block sections or long block sections), still works perfectly on my side.

Have you tested only explorer mode, or have you also tested activity and timetable mode to ensure these changes do not in any way interfere with the signal logic in these modes?
It is important to know that the changes are indeed restricted to explorer mode and do not somehow alter the behaviour of the signalling in other modes.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
cesarBLG: I think it's just a 'git diff' error. This code was in function requestClearSignalExplorer(), but I moved it to a new function GetReqNumClearAheadExplorer(), in order to use this from CheckExplorerPath() too. I haven't added SNCA to TranslateTMAspect, of course.

cesarBLG:
I have only modified ProcessExplorerSwitch() and CheckExplorerPath(), so according to the name of this methods I would tend to think that Activity and Timetable modes remain unchanged.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
YoRyan: Starting the function with `int ReqNumClearAhead = 0;` actually suppresses the type checker. It will no longer warn you when you've forgotten to assign a value to it in one of the if branches.

YoRyan (approved):

YoRyan:
Side note: I wish we had some form of automated testing for the signaling code. It is much more efficient than spending hours testing the simulator across explorer, activity, and timetable modes. But that's a topic for another day.

Sharpe49:
I tested the latest commit for all three modes. No problems encountered.

Sharpe49 (approved):

Sharpe49:
Any last word before merge ?

PR #304, opened 2020-11-10
added signalling timing functions

cjakeman:
As requested at http://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/page__view__findpost__p__264472


Sharpe49 (dismissed):
You have added your VS Code settings. Is that intentional?

cjakeman:
Hi Cedric,

No. Thanks for catching that.

Best wishes,

Chris

Sharpe49:
Why is there a revert?

YoRyan (approved):

PR #306, opened 2020-11-13
Write back to the settings store on quit

YoRyan:
Some key commands write to the settings store while in-game. It would be convenient if these change persisted after close, so that players don't have to press the same keys every time to restore their preferences.

I've made sure that at least the cab letterboxing feature and the track monitor immersive mode are saved and restored in this manner. Any suggestions for other in-game modes that should be looked at?

Feature request: https://www.elvastower.com/forums/index.php?/topic/34562-in-game-toggles-that-should-stick-but-dont/

### DIFF
--- a/Source/ORTS.Settings/UserSettings.cs
+++ b/Source/ORTS.Settings/UserSettings.cs
@@ -407,6 +407,10 @@ namespace ORTS.Settings
         public string TrackMonitorDisplayMode { get; set; }
         [Default(true)]
         public bool Use3DCab { get; set; }
+        [Default(0x7)] // OSDLocations.DisplayState.Auto
+        public int OSDLocationsState { get; set; }
+        [Default(0x1)] // OSDCars.DisplayState.Trains
+        public int OSDCarsState { get; set; }
 
         #endregion
 
--- a/Source/ORTS.Settings/UserSettings.cs
+++ b/Source/ORTS.Settings/UserSettings.cs
@@ -352,8 +352,6 @@ namespace ORTS.Settings
         [Default(false)]
         public bool DataLogger { get; set; }
         [Default(false)]
-        public bool Letterbox2DCab { get; set; }
-        [Default(false)]
         public bool Profiling { get; set; }
         [Default(0)]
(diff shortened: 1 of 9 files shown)


YoRyan:
I have received no further suggestions, so I'm taking this out of draft.

cjakeman (dismissed):
Hi Ryan,

You've done well to find your way through this "jungle" of code.
Approved, with thanks,

Please would you resolve the conflict so that I can merge this PR.

Chris

YoRyan:
Thanks, Chris!

YoRyan:
Resolved! Thanks, Chris.

cjakeman (approved):
Hi Ryan,

Sorry I let this one go stale.

Approved, with thanks,

Chris

twpol:
Hmm, I think this was a bad idea.

I've lost my content settings and had some supposed-to-be-ephemeral settings persisted into the registry since this landed.

We should be saving **only the settings we modify** while running.

Some of the settings (like the profiling ones at https://github.com/openrails/openrails/blob/6f548846f687df55c470005fe30a0f6f63b4aa56/Source/ORTS.Settings/UserSettings.cs#L355) are completely internal and should never be saved, which we could also implement explicitly. (But this alone would not be enough because of e.g. https://github.com/openrails/openrails/blob/48ac1230ca73cb0344e83cdebf8bc7229de20e93/Source/ORTS.Common/SettingsBase.cs#L106)

YoRyan:
Hmm, I see. We have already have a `[DoNotSave]` attribute that will explicitly prevent settings from being saved; it just wasn't applied correctly to Profiling and other such internal settings.

We could add this attribute to the other settings that need to be blacklisted. Do you think this approach would be sufficient, or should we scrap it in favor of an explicit whitelist with something like a `[SaveAtExit]` attribute?

twpol:
The `[DoNotSave]` attribute will certainly help things like the profiling settings, which should never be saved.

But we'd still need some more changes to make sure **nothing** is saved when you use the `/skip-user-settings` option.

PR #308, opened 2020-11-17
added manual braking to News items

cjakeman:
As suggested by Peter Newell


peternewell (approved):
Seems good.

Thanks for that.

PR #311, opened 2020-11-22
Bug fix for https://bugs.launchpad.net/or/+bug/1905146 On restore panto 1 is always risen

Csantucci:
Launching event of a panto rise is needed when you meet an AI train or after restore of a moving player train, because it restarts the related sound event. Up to now always the event for panto 1 was launched. However launching the event not only triggers the related sound trigger, but also physically rises the panto if it it was down, and another one was up; so panto 1 was risen even if it hat to be down, because another panto was up. With the patch the first up panto is selected, and the event is launched for that one.

### DIFF
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
@@ -38,8 +38,19 @@ namespace Orts.Viewer3D.RollingStock
             if (ElectricLocomotive.Train != null && (car.Train.TrainType == Train.TRAINTYPE.AI ||
                 ((car.Train.TrainType == Train.TRAINTYPE.PLAYER || car.Train.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN || car.Train.TrainType == Train.TRAINTYPE.AI_PLAYERHOSTING) &&
                 (car.Train.MUDirection != Direction.N && ElectricLocomotive.PowerOn))))
+                // following reactivates the sound triggers related to certain states
+                // for pantos the sound trigger related to the raised panto must be reactivated, else SignalEvent() would raise also another panto
             {
-                ElectricLocomotive.SignalEvent(Event.Pantograph1Up);
+                var iPanto = 0;
+                foreach (var panto in ElectricLocomotive.Pantographs.List)
+                {
+                    if (panto.State == ORTS.Scripting.Api.PantographState.Up)
+                    {
+                        ElectricLocomotive.SignalEvent((Event)((int)Event.Pantograph1Up + 2 * iPanto));
+                        break;
+                    }
+                    iPanto++;
+                }
                 ElectricLocomotive.SignalEvent(Event.EnginePowerOn);
                 ElectricLocomotive.SignalEvent(Event.ReverserToForwardBackward);
                 ElectricLocomotive.SignalEvent(Event.ReverserChange);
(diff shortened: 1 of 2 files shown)


YoRyan:
It's possible to raise multiple pantographs at the same time. Do we have to activate the raise event for all of them when resuming from a save? If so, then that `break` statement shouldn't be there.

Csantucci:
This needs a more complete explanation, also beyond the title, which is incomplete. The aim of these lines is not rising the pantos, which are risen within the restore code itself, but to restart the pantograph rise sound after resume and to start it on an AI train which is met.
In most of the .sms files of electric locos the pantograph triggers are used not only to get the short sound of the rising pantograph, but also to start the long-term sounds for ventilation, power electronics and so on. Without these lines of code such sounds wouldn't play within the player train after game resume neither on AI trains met at any moment during the game. 
The problem before the bug correction was that it rose pantograph 1 even if it hadn't to be risen (because that call both rises the panto and triggers the sound). So the fix triggers only the event of the first pantograph that is already up. Why only one? Because these long-term sounds are usually the same for any pantograph risen, so it's enough to trigger such sounds only for one panto. But if you prefer, I can remove the break. It doesn't change much.

YoRyan:
I see. To my mind, we should activate the sound events for all raised pantographs. You never know when some content creator might depend on all of the sounds playing together.

As far as the code goes, I also dislike the fragile int<->Event casting. I would suggest a lookup table instead, something like

```
Event evt;
switch (iPanto)
{
    case 0: evt = Event.Pantograph1Up;
    case 1: evt = Event.Pantograph2Up;
    case 2: evt = Event.Pantograph3Up;
    case 3: evt = Event.Pantograph4Up;
}
```

Csantucci:
There are more than one place in the OR code where the enum sequence order matters. But again, if you prefer, I can add code lines to satisfy what you prefer.

YoRyan:
If the integer values are significant, then they ought to be explicitly specified. Relying on the order is a practice we should try to avoid in any new code, IMO.

Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
```diff
@@ -38,19 +38,41 @@ public MSTSElectricLocomotiveViewer(Viewer viewer, MSTSElectricLocomotive car)
...
                 ElectricLocomotive.SignalEvent(Event.ReverserChange);
             }
         }
 
-        /// <summary>
-        /// A keyboard or mouse click has occured. Read the UserInput
-        /// structure to determine what was pressed.
-        /// </summary>
-        public override void HandleUserInput(ElapsedTime elapsedTime)
+        Event evt;
```
YoRyan, line 68: Delete unused variable

Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
```diff
@@ -38,19 +38,41 @@ public MSTSElectricLocomotiveViewer(Viewer viewer, MSTSElectricLocomotive car)
...
+
+
+    /// <summary>
+    /// A keyboard or mouse click has occured. Read the UserInput
+    /// structure to determine what was pressed.
+    /// </summary>
+    public override void HandleUserInput(ElapsedTime elapsedTime)
```
YoRyan, line 75: Restore the correct indentation

YoRyan (changes requested):
Thanks, Carlo. Just a couple of blockers and I'll approve.

Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
Csantucci: I left some mess there. It's corrected now.

Source/RunActivity/Viewer3D/RollingStock/MSTSElectricLocomotiveViewer.cs
Csantucci: Remainder of introduction of correction. Deleted now.

YoRyan (approved):

PR #312, opened 2020-11-23
Some locomotive related TCS script hooks, see https://blueprints.launchpad.net/or/+spec/locomotive-related-tcs-hooks

Csantucci:
Hooks include:
- pantograph state (both read and write)
- throttle percent (read)
- current elevation percent.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -204,13 +204,13 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> AccelerationMpSS;
         /// <summary>
-        /// Locomotive elevation.
+        /// Locomotive altitude.
         /// </summary>
-        public Func<float> ElevationM;
+        public Func<float> AltitudeM;
         /// <summary>
         /// Track gradient percent at the locomotive's location.
         /// </summary>
-        public Func<float> CurrentGradientPercent;
+        public Func<float> CurrentElevationPercent;
         /// <summary>
         /// Line speed taken from .trk file.
         /// </summary>
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -292,11 +292,15 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Action SetPantographsDown;
         /// <summary>
-        /// Cut power by pull all pantographs down.
-        /// PowerSupplyEvent: may be LowerPantograph or RaisePantograph
+        /// Raise specified pantograph
         /// int: pantographID, from 1 to 4
         /// </summary>
-        public Action<PowerSupplyEvent, int> SetPantograph;
+        public Action<int> SetPantographUp;
+        /// <summary>
+        /// Lower specified pantograph
+        /// int: pantographID, from 1 to 4
+        /// </summary>
+        public Action<int> SetPantographDown;
         /// <summary>
         /// Set the circuit breaker or power contactor closing authorization.
         /// </summary>
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -172,6 +172,10 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> ThrottlePercent;
         /// <summary>
+        /// Returns dynamic brake percent
+        /// </summary>
+        public Func<float> DynamicBrakePercent;
+        /// <summary>
         /// True if traction is authorized.
         /// </summary>
         public Func<bool> TractionAuthorization;
@@ -192,9 +196,21 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> BrakeCutsPowerAtBrakeCylinderPressureBar;
         /// <summary>
-        /// Elevation percent at the locomotive's location.
+        /// Track slope percent at the locomotive's location.
         /// </summary>
-        public Func<float> CurrentElevationPercent;
+        public Func<ControllerState> TrainBrakeControllerState;
+        /// <summary>
+        /// Locomotive acceleration.
+        /// </summary>
+        public Func<float> AccelerationMpSS;
+        /// <summary>
+        /// Locomotive elevation.
+        /// </summary>
+        public Func<float> ElevationM;
+        /// <summary>
+        /// Track slope percent at the locomotive's location.
+        /// </summary>
+        public Func<float> CurrentSlopePercent;
         /// <summary>
         /// Line speed taken from .trk file.
         /// </summary>
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -162,7 +162,7 @@ namespace ORTS.Scripting.Api
         /// Checks the state of any pantograph
         /// int: pantograph ID (1 for first pantograph)
         /// </summary>
-        public Func<int, PantographState> PantographState;
+        public Func<int, PantographState> GetPantographState;
         /// <summary>
         /// True if all pantographs are down.
         /// </summary>
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -158,6 +158,10 @@ namespace ORTS.Scripting.Api
         /// True if circuit breaker or power contactor opening order is true.
         /// </summary>
         public Func<bool> CircuitBreakerOpeningOrder;
+         /// <summary>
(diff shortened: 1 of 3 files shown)


YoRyan:
Can I also ask for calls to retrieve the brake controller positions, in notches in addition to percentages? That information is the only reason I use the Locomotive handle in my scripts.

Csantucci:
If you write here the code line you are using, we are sure that I implement the right thing.

YoRyan:
Certainly. Here's an example:

```C#
ControllerState brake = Locomotive().TrainBrakeController.TrainBrakeControllerState;
bool suppressing = brake == ControllerState.Suppression || brake == ControllerState.ContServ || brake == ControllerState.FullServ;
```

I also need access to the `Locomotive().AccelerationMpSS` property.

cesarBLG:
Could you add access to the altitude of the locomotive? This is useful when calculating the gradient between the infill balise group and the main signal.

Sharpe49:
For the locomotive, I will need to be able to reduce the traction automatically before opening the circuit breaker. When the neutral section has been passed, I will close automatically the circuit breaker and I need to set the traction to its original value.
Since this is a bit more complicated, I don't know if it should be integrated now or when we have cruise control.

Csantucci:
All the requests here and on the other TCS hooks PR demonstrate one thing: that keeping the Locomotive() hook is the right thing, because more and more needs will follow.

That said, for Cédric's request I think it should be feasible: reading the throttle position is already available, so writing it should be added. Read/write access to circuit breaker is already available. 

Re Ryan's request for the TrainBrakeControllerState I ask him how to give to the script access to the ControllerState enum. 

Apart from the ones requested up to now, I won't add code for further requests if there isn't a related PR in my remote repository's branch,

Csantucci:
Following further hooks added (all of them getters):
DynamicBrakePercent
TrainBrakeControllerState
AccelerationMpSS
ElevationM

Renamed CurrentElevationPerCent to CurrentGradientPerCent

YoRyan:
> All the requests here and on the other TCS hooks PR demonstrate one thing to me: that keeping the Locomotive() hook is the right thing, because more and more needs will follow.

I still have to agree with @pzgulyas here; exposing the Locomotive class to the TCS interface means that we cannot make any changes to that class without breaking any scripts. I doubt it would be advisable to freeze such a massive portion of the internal codebase like that. What if we decide to break it up and refactor it, as some other developers have suggested?

pzgulyas:
Carlo, the more and more requests are actually exactly the same ones, that we were talking about originally. And once the requests are fulfilled, there will be no more mass requests. I would expect some more additional requests of the systems possible to control by keyboard (get, set), all portions of signalling data, and some more that probably already are available. The ones that I've collected in the list at the beginning of [this file](https://github.com/pzgulyas/openrails-1/blob/ScriptedControls/Source/Orts.Simulation/Common/Scripting/EngineScript.cs)

Having the Locomotive() hook in the scripting interface is just principally wrong. It must be removed not later than just before the next stable release. I understand, that it is an actual work to add the proper interface functions, but this needs to be done to maintain the freedom of modifying the codebase in long term. (We are planning on long term, right?)

Csantucci:
To me the PR is ready for final review.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -259,6 +292,12 @@ public abstract class TrainControlSystem : AbstractScriptClass
         /// </summary>
         public Action SetPantographsDown;
         /// <summary>
+        /// Cut power by pull all pantographs down.
+        /// PowerSupplyEvent: may be LowerPantograph or RaisePantograph
+        /// int: pantographID, from 1 to 4
+        /// </summary>
+        public Action<PowerSupplyEvent, int> SetPantograph;
```
Sharpe49, line 299: Currently, this is not right because you can use something inappropriate in the parameters (for exemple, StopEngine)

I would prefer something either more generic :
public Action<PowerSupplyEvent>
public Action<PowerSupplyEvent, int> SignalEvent;
Which is similar to the existing Action<Event> SignalEvent

Or more specific : 
public Action<int, bool> SetPantograph;
With the boolean indicating if we want the pantograph up or down.

Besides this, everything else looks fine to me.

Sharpe49 (changes requested):

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Csantucci: OK, I have created two specific hooks, one to raise the specified panto, and one to lower it.
I wasn't so happy with the boolean. I hope it is OK now.

YoRyan:
A few things that need to be addressed:

1. There ought to be a way to enumerate how many pantographs (0-4?) are present on the locomotive.
2. What happens if a script calls a pantograph function with an invalid pantograph ID? Currently you are indexing directly into `Locomotive.Pantographs`, so I think this would result in a rather nasty `KeyError` condition. Consider creating a [custom exception](https://docs.microsoft.com/en-us/dotnet/standard/exceptions/how-to-create-user-defined-exceptions) for this condition.
3. Manual documentation. 😁

Csantucci:
Ryan,
1) I don't understand your point.
2) What can be done is checking for lower than 1 and higher than 4, and generating a Trace.TraceError() in that case. Can this be OK?
3) single TCS script hooks don't have specific manual documentation, as you can easily check; the specific documentation is the summary in Common\Scripting\TrainControlSystem.cs, which script builders must access anyway, where I'm describing the parameters used, differently from older hooks that don't have such description.

YoRyan:
Okay, so documentation is good so long as your XML comments are present.

I'm saying not all locomotives have exactly 4 pantographs. Is that correct? My reading of Pantograph.cs is that locomotives can have anywhere from 0 to `int.MaxValue` pantographs. Thus, scripts need a way to 1) find out how many pantographs are present on the locomotive, and 2) be notified if they pass an invalid pantograph ID. (Let's not use `Trace.TraceError()` because that's completely invisible to the script.)

Csantucci:
Ryan,
you are overcomplicating things. If someone doesn't accept the PR as it is with the addition of the Trace.TraceError, I consider withdrawing the PR.

Sharpe49:
I think an exception is a bit excessive because the end user may not understand that this is not an Open Rails issue (and may report a bug on Launchpad) but a script issue. To me, TraceError is appropriate.

YoRyan:
I'm sorry. I should talk with code when I become so picky: [honor-pantograph-count.diff.txt](https://github.com/openrails/openrails/files/5614149/honor-pantograph-count.diff.txt)

(I suppose a custom exception is overkill, too; nevertheless, surely `const numPantographs = 4` is a bad assumption?)

Csantucci:
YoRyan,
remember that a loco might have panto1 and panto3, but not panto2... anyhow adding your simple hook is not a problem.
Please note too that pantoID runs from 1 to 4 and not from 0 to 3 (as stated also in the XML comment about the call). That said, I'll modify your proposed patch accordingly and will integrate it.

Csantucci:
Ryan,
rethinking at your patch, I'm not convinced about it. A locomotive might have only the first two pantographs, but there could be in the consist also a locomotive with 4 pantographs. In this case it would not be possible to move them with your patch. I think that the only check which must be done is that pantoID is within 1 and 4, and this is what I am pushing now.
I have also renamed CurrentElevationPercent to CurrentGradientPercent and changed the sign to it, because CurrentElevationPercent is negative going uphill,

YoRyan:
Actually, the `Pantographs` list is 0-indexed. I just set a breakpoint in `MSTSLocomotive.Update` with the AEM-7 and inspected it myself; pantograph 1 is index 0, and pantograph 2 is index 1. So the ID's are 0-3, even if they're labeled 1-4 in the engine file.

YoRyan:
Oops, didn't realize you were sending `PowerSupplyEvent.RaisePantograph`. I thought you were manipulating the `Pantographs` list directly.

pzgulyas:
> (I suppose a custom exception is overkill, too; nevertheless, surely `const numPantographs = 4` is a bad assumption?)

I think neither of the script interface functions should ever generate any exceptions.

Csantucci:
Peter, I'm not sure of having well understood your sentence, because "should" can be interpreted in two ways, but if you mean that by setting, let's say, pantoID to 100 you don't get crashes (withouth the boundary check Trace.TraceError) I agree with you. Anyhow setting a boundary check helps the script developer writing a correct script.

pzgulyas:
Just to make a note, although I think Carlo is aware of this: whether a Pantograph with a specific number is the front or the back one, depends on the locomotive "flip"-ping, and also depends on whether it has been travelled through a reversal point or not. The panto numbers in the script probably would need to be maintained regardless. (The same applies to doors left-right side.)

Csantucci:
Yes, you're right, which leads to the fact that the addition of two further hooks (IsFlipped and IsRearCab) is desirable.

Csantucci:
Added IsFlipped and IsRearCab hooks.

Csantucci:
Changes have been made, I expect this may be approved.

Sharpe49:
Seems good to me.

Sharpe49 (approved):

Csantucci:
Thanks, Cédric.

PR #313, opened 2020-11-24
build: Add xunit to AppVeyor configuration

twpol:
[As was pointed out](http://www.elvastower.com/forums/index.php?/topic/34274-ci-failure-due-to-lazarus-download/page__view__findpost__p__264908), because we don't have the same AppVeyor configuration nor Chocolatey packages on different branches, the cache feature is not really effective.

This PR adds XUnit to the configuration which, combined with https://github.com/openrails/openrails/pull/230/commits/88ce21f9a0a8e330fdb3ee056cdc265e3437dcd4, should allow the AppVeyor configuration to match on `master` and `unstable` thus allowing the use of the cache.

Any open PRs can, if they absolutely need it, merge `master` into their PR (after this one is merged) to also align with the configuration and cache.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -12,13 +12,13 @@ install:
 - ps: >-
     $ErrorActionPreference = "Stop"
 
-    choco install --yes --no-progress lazarus 7zip.portable xna31
+    choco install --yes --no-progress lazarus 7zip.portable xunit xna31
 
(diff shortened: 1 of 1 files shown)


YoRyan (approved):
Looks good, James. Hopefully this resolves the CI issue once and for all.

PR #314, opened 2020-11-25
Play a standardized, pleasing sound when taking a screenshot

YoRyan:
Currently, taking a screenshot with the PrintScreen key plays the "control error" sound of the current locomotive. I don't know about you, but I find the default MSTS buzzer sound quite jarring and obnoxious. Worse still, some locomotives have this sound event intentionally disabled, which creates a usability problem.

This PR changes the screenshot function to play a camera shutter sound instead, regardless of the current locomotive. You can listen to that sound [here](https://freesound.org/people/jdaniel1999/sounds/376205/). It is stored in the Program\Content\ folder.


Sharpe49 (approved):
Seems good to me.

YoRyan:
Thanks, Cedric. I was hoping for an opinion from the management team, but I doubt they would raise any objections, anyway. Merging.

PR #315, opened 2020-11-27
Bug fix for https://bugs.launchpad.net/or/+bug/1905957 - Locomotives with only a rear 3D cab aren't managed properly

cesarBLG:
See http://www.elvastower.com/forums/index.php?/topic/24040-3d-cabs/page__view__findpost__p__264525
When a locomotive has only a 3D cab in rear but not in front, the cab change procedure handles the rear cab as a frontal cab, not setting the variable UsingRearCab correctly, so direction dependent controls like reverser aren't inverted.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -714,7 +714,14 @@ namespace Orts.Simulation.RollingStocks
             if (!(this is MSTSSteamLocomotive))
                 InitializeFromORTSSpecific(cvfFilePath, extendedCVF);
 
-            return new CabView3D(cvfFile, CabViewpoints, extendedCVF, CabViewType.Front, noseAhead, shapeFilePath);
+            var cabViewType = CabViewType.Front;
+            if (CabViewpoints.Count == 1 && ((CabViewpoints[0].StartDirection.Y >= 90 && CabViewpoints[0].StartDirection.Y <= 270)
+                 || (CabViewpoints[0].StartDirection.Y <= -90 && CabViewpoints[0].StartDirection.Y >= -270)))
+            {
+                CabViewpoints.Insert(0, new PassengerViewPoint());
+                cabViewType = CabViewType.Rear;
+            }
+            return new CabView3D(cvfFile, CabViewpoints, extendedCVF, cabViewType, noseAhead, shapeFilePath);
         }
 
         /// <summary>
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -714,13 +714,13 @@ namespace Orts.Simulation.RollingStocks
             if (!(this is MSTSSteamLocomotive))
                 InitializeFromORTSSpecific(cvfFilePath, extendedCVF);
 
-            var cabViewType = CabViewType.Front;
-            if (CabViewpoints.Count == 1 && ((CabViewpoints[0].StartDirection.Y >= 90 && CabViewpoints[0].StartDirection.Y <= 270)
-                 || (CabViewpoints[0].StartDirection.Y <= -90 && CabViewpoints[0].StartDirection.Y >= -270)))
-            {
+            var cabViewAngle = CabViewpoints[0].StartDirection.Y;
+            var cabViewType = (cabViewAngle >= 90 && cabViewAngle <= 270) || (cabViewAngle <= -90 && cabViewAngle >= -270) ? CabViewType.Rear : CabViewType.Front;
+
+            // only one cabview, and it looks rear; insert a void one at first place to maintain fast indexing
+            if (CabViewpoints.Count == 1 && cabViewType == CabViewType.Rear)
                 CabViewpoints.Insert(0, new PassengerViewPoint());
-                cabViewType = CabViewType.Rear;
-            }
+
             return new CabView3D(cvfFile, CabViewpoints, extendedCVF, cabViewType, noseAhead, shapeFilePath);
         }
 
--- a/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
@@ -1639,7 +1639,7 @@ namespace Orts.Simulation.RollingStocks
             {
                 var loco = this as MSTSLocomotive;
                 var i = (int)CabViewType.Front;
-                if (loco == null || loco.CabView3D == null) return false;
+                if (loco == null || loco.CabView3D == null || loco.CabView3D.CabViewType != CabViewType.Front) return false;
                 return (loco.CabView3D.ViewPointList.Count > i);
             }
         }


Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -714,7 +714,14 @@ protected CabView3D BuildCab3DView()
...
                 InitializeFromORTSSpecific(cvfFilePath, extendedCVF);
 
-            return new CabView3D(cvfFile, CabViewpoints, extendedCVF, CabViewType.Front, noseAhead, shapeFilePath);
+            var cabViewType = CabViewType.Front;
+            if (CabViewpoints.Count == 1 && ((CabViewpoints[0].StartDirection.Y >= 90 && CabViewpoints[0].StartDirection.Y <= 270)
+                 || (CabViewpoints[0].StartDirection.Y <= -90 && CabViewpoints[0].StartDirection.Y >= -270)))
+            {
+                CabViewpoints.Insert(0, new PassengerViewPoint());
+                cabViewType = CabViewType.Rear;
+            }
```
YoRyan, line 723:
```suggestion
            var cabViewAngle = CabViewpoints[0].StartDirection.Y;
            var cabViewType = (cabViewAngle >= 90 && cabViewAngle <= 270) || (cabViewAngle <= -90 && cabViewAngle >= -270) ? CabViewType.Rear : CabViewType.Front;
            if (CabViewpoints.Count == 1 && cabViewType == CabViewType.Rear)
                CabViewpoints.Insert(0, new PassengerViewPoint());
```
I'd prefer this form. I think it makes the logic clearer and easier to follow.

YoRyan: Also, I understand the need to set `cabViewType` correctly, but what's with the `CabViewpoints.Insert(0, new PassengerViewPoint());` line? It seems to be inserting another front-facing viewpoint, but why should this be necessary?

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -1639,7 +1639,7 @@ public bool HasFront3DCab
             {
                 var loco = this as MSTSLocomotive;
                 var i = (int)CabViewType.Front;
-                if (loco == null || loco.CabView3D == null) return false;
+                if (loco == null || loco.CabView3D == null || loco.CabView3D.CabViewType != CabViewType.Front) return false;
```
YoRyan, line 1642: Nearby in this file, there is a counterpart `HasRear3DCab` getter that presumably needs to be changed, too.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: Some parts of the code (e. g. cab change) expect the rear cab to be indexed at 1. For 2D cabs a void cabview is inserted, so I followed a similar logic for 3D cabs.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
cesarBLG: Locomotives with only a front cab will have just one CabViewpoint, so HasRear3DCab will work as expected. The HasFrontCab and HasRearCab 2D counterparts follow the same logic.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
YoRyan: Sorry, could you point me to the line where this happens for 2D cabs? I don't see anything like that in the `CabView BuildCabView(string wagFilePath, string cvfFileName)` function.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
YoRyan: I see, thanks!

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: The cabview type is determined in `BuildCabView()`
https://github.com/openrails/openrails/blob/7abeb7d51b2af715c20742adcd7334d1d053aaca/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L647-L649
(there's an unnecessary initialization there)

And the void viewpoint is added in `LoadWagFromFile(string wagFilePath)`:
https://github.com/openrails/openrails/blob/7abeb7d51b2af715c20742adcd7334d1d053aaca/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L462-L468

YoRyan: So I'm not so sure about this change. That 2D cab code inserts the void object into the CabView list, which is not the same thing as the CabView ViewPoint list. Meanwhile, the code suggests we can have many 2D cabs view, but only one 3D cab view:
https://github.com/openrails/openrails/blob/7abeb7d51b2af715c20742adcd7334d1d053aaca/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L368-L369

And there's code to index into the CabView list - hence inserting a dummy entry for the missing front CabView:
https://github.com/openrails/openrails/blob/7abeb7d51b2af715c20742adcd7334d1d053aaca/Source/RunActivity/Viewer3D/Cameras.cs#L2071-L2073

...But I'm not seeing any equivalent lines for CabView.ViewPointList.

So, a few more questions:

- Is it possible that setting the proper CabViewType value (Front or Rear) for the 3D cab was enough to get the controls to behave correctly, and that we don't have to insert a dummy ViewPoint at all?
- In the original Elvas Tower thread, VAPOR3D reports that the second 3D cab takes on the same position as that of the first 3D cab, and that none of the digital controls are rendering. This, to me, seems consistent with my assumption that the current code was not written with multiple 3D cabs in mind. And I don't see anything in this patch that addresses those issues.

I'd like to help debug this, but I don't have any locomotives with multiple 3D cabs. I've asked VAPOR3D if he might be willing to provide a sample.

cesarBLG: The viewpoint for 3D cabs is switched in ThreeDimCamera class:
https://github.com/openrails/openrails/blob/7abeb7d51b2af715c20742adcd7334d1d053aaca/Source/RunActivity/Viewer3D/Cameras.cs#L1879-L1889
- The dummy cabview (for 2D cabs) and the dummy viewpoint (for 3D cabs) are used to index the cabs, so that Viewpoint[(int)CabViewType] returns the adecuate value. This dummy viewpoints are never seen because of the HasFrontCab and HasRearCab properties. I don't think it's a clean solution, but is there some benefit from refactoring it now?
- VAPOR3D was misusing the parameter StartDirection(), that's why he got wrong directions. After this patch he got everything working except the digital gauges. I modified two locomotives so one had a front cab and the other a rear one and both of them had all the controls working.
- Regarding the UsingRearCab, it might be OK to set it for rear-only 3D cabs too (unless there is a weird locomotive with 2D cabs at both ends but only a rear 3D cab, which I don't think). This parameter is only useful for the first player locomotive, after a cab change it is recalculated in `Viewer.ChangeCab()`

YoRyan: Thanks, that's what I wanted to know. I do think this system is unnecessarily obtuse and likely doesn't help performance at all, but we can save that refactor for another day.

YoRyan (approved):

PR #317, opened 2020-12-04
Bug fix for https://bugs.launchpad.net/or/+bug/1906786 - Approach Control doesn't work for non NORMAL signals

cesarBLG:
Approach control is only available for NORMAL signals at the moment. This caused some route developers to set level crossing signals (which have to be cleared a fixed distance before the train passes them) as NORMAL signals, but that interferes with signalling logic. With this proposed changes approach_control_position_forced() will work for all signals.

I haven't extended this to approach_control_position() and approach_control_speed(), but if there are use cases I can do it.

### DIFF
--- a/Source/Documentation/Manual/operation.rst
+++ b/Source/Documentation/Manual/operation.rst
@@ -1194,18 +1194,24 @@ has reduced its speed to a specific value. Such control is used for diverging
 routes, to ensure the speed of the train is reduced sufficiently to safely 
 negotiate the switches onto the diverging route.
 
-Two script functions for use in OR have been defined which can be used to 
+Three script functions for use in OR have been defined which can be used to
 control the signal until the train has reached a specific position or has 
 reduced its speed.
 
 These functions are::
 
     APPROACH_CONTROL_POSITION(position)
+    APPROACH_CONTROL_POSITION_FORCED(position)
     APPROACH_CONTROL_SPEED(position, speed)
 
 These functions are Boolean functions: the returned value is 'true' if a train 
-is approaching the signal and is within the required distance of the signal 
-and, for APPROACH_CONTROL_SPEED, has reduced its speed below the required values.
+is approaching the signal and is within the required distance of the signal and,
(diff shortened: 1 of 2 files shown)


YoRyan:
You have to be careful with a change like this, because some route developers might be _expecting_ the approach control function to disregard non-NORMAL signals.

So instead of changing the behavior of the existing function, I would like to see a new function defined for this use case. How about a function that allows the developer to select any arbitrary set of any signal types (think `NORMAL|SHUNTING|SPEED|...`) to activate approach control? I'm not sure how difficult that would be to define with the signal scripting syntax.

cesarBLG:
So you are saying that some route developers expect that if they call Approach_Position_Control_Forced(distance) from the script of a non NORMAL signal, it will return false always? Then I don't find any reason for why they are using approach control for that signal in the first time. Take into account that this changes only affect the _forced_ version, which wasn't working properly before #159 in April.

YoRyan:
Oh sorry, I had my frame of reference reversed. Then yes, if you request approach control from a signal script, there's no reason to expect it shouldn't work.

YoRyan:
In that case, rather than creating a separate `APPROACH_CONTROL_POSITION_FORCED` function, I suggest just changing `APPROACH_CONTROL_POSITION` to work with all signal types.

cesarBLG:
The `approach_position_control_forced` function already existed, but it wasn't documented.

I'm not sure about extending `approach_position_control` to work with non normal signals. It is supposed to clear only if the calling function is the next signal for the train. It is clear which is the next NORMAL signal for a train, but for non normal signals it's not trivial to determine it. For them, next signal might mean 'next signal without normal heads' or 'next signal having the same head as the calling signal'. Until this is clarified, I would prefer letting this function as it is now. I haven't found any cases where this function would be useful for non normal signals yet.

The forced version doesn't check if signal is the first for the train, it only clears if the train is within the required distance, so it's useful for all signal types.

YoRyan (approved):
Yes, that sounds good to me. Approving.

PR #318, opened 2020-12-05
Reintroduce "Allowing manual Call On from dispatcher window" - https://trello.com/c/I19VVKlz/450-manual-call-on-from-dispatcher-window

cesarBLG:
For some reason #155 has vanished, so I create this PR with the feature.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
@@ -8408,7 +8408,7 @@ namespace Orts.Simulation.Signalling
         public bool StationHold = false;        // Set if signal must be held at station - processed by signal script
         protected List<KeyValuePair<int, int>> LockedTrains;
 
-        public bool CallOnEnabled = false;
+        public bool CallOnEnabled = false;      // set if signal script file uses CallOn functionality
 
         public bool enabled
         {
@@ -12596,6 +12596,8 @@ namespace Orts.Simulation.Signalling
             bool[] returnValue = new bool[2] { false, false };
             MstsSignalAspect thisAspect = this_sig_lr(MstsSignalFunction.NORMAL);
 
+            SetManualCallOn(false);
+
             // signal not enabled - set lock, reset if cleared (auto signal can clear without enabling)
 
             if (enabledTrain == null || enabledTrain.Train == null)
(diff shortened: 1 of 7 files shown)


Source/Orts.Simulation/Simulation/Signalling/Signals.cs
```diff
@@ -12116,6 +12118,7 @@ public bool CheckTimingTrigger(int reqTiming, string dumpfile)
 
         public bool TrainHasCallOn(bool allowOnNonePlatform, bool allowAdvancedSignal, string dumpfile)
         {
+            CallOnEnabled = true;
```
YoRyan, line 12121: Here's a difference between this code and the one presented in #155 - in the previous PR, you set this property in SIGSCRfile.cs, but now, you've made it always true. Can you explain this change?

YoRyan:
Sorry you lost your PR, and thanks for creating a new one. For future reference, don't merge into the `unstable` branch. That one's automatically managed by the Code Bot. Always use `master`.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
cesarBLG: Rob created two additional callon functions between the previous PR and this one, and all of them call this function. Setting this variable within TrainHasCallOn() is better than having to set it separately for each function.

The aim of this variable is to show the "Enable Call On" option in the signal's context menu only for signals which support this feature.

cesarBLG:
I think it was my first PR, and I wasn't aware of the code bot. Now that I know about it, I always compare against the `master` branch.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
YoRyan: The purpose of the function is to "Test if train has call-on set," which strongly implies that there shouldn't be any side effects. Therefore, I think it's best to set this variable back in SIGSCRfile.cs, regardless of the code duplication.

cesarBLG: I moved it back to SIGSCRfile.cs

YoRyan (approved):
Thanks! Approving and merging.

PR #319, opened 2020-12-07
notrriggered => nottriggered x 2

cjakeman:
Corrected spelling errors in Parser for activities found by John (RR1)


YoRyan:
Looks good

YoRyan (dismissed):

YoRyan:
So, what's the history behind this property? Is it a custom ORTS one, or does the MSTS Activity Editor output `texttodisplayoncompletionifnotrriggered`?

If the latter is the case, then we need to retain support for the misspelling to retain backwards compatibility with MSTS.

Sharpe49:
I did a search inside my activity files, it is spelled TextToDisplayOnCompletionIfNotTriggered.
So the fix is correct.

Sharpe49 (approved):

YoRyan:
Perfect. Thanks for checking, Cedric.

PR #321, opened 2020-12-18
Improvements for EP and air braking systems

cesarBLG:
There are some issues with the behaviour of some of the brake controller notches, which have been corrected:
- Full service position decreases pressure to 0, instead of using the full service pressure drop parameter
- Neutral position should isolate the EQ reservoir from the brake pipe, so no leakage is compensated.

Additionally, two new notches have been added:
- EPOnly: this token applies EP brakes without reducing pressure in brake pipe.
- EPFullService: this is used for non self lapping EP systems. The brake controller should have three positions: Release, EPHold and EPFullService, to increase or reduce application of EP brakes.
 
Another feature added is graduated release of the brakes without the need of the "Graduated release on all cars" option. In order to not having to create more brake tokens, the existing ones have been modified, so there will be a slightly different behaviour for this tokens. If it's better to create new tokens rather than modifying existing ones, I'll do that. These are the modified notches:
- EPApply: applies both EP and air brakes with graduated release. The continuous service position does the same as EPApply but without graduated release.
- GraduatedSelfLap: applies air brakes with graduated release. There's another (already existing) function, GraduatedSelfLapLimitedHolding that doesn't provide graduated release. IIRC this was the behaviour in MSTS.

Still have to add a slow pressure reduction notch for AWS and include some documentation provided by darwins from Elvas Tower.

@peternewell what do you think about this changes?

Discussion: http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/
Related trello card: https://trello.com/c/0SvMICQt/485-air-brake-and-ep-brake-options

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -143,10 +143,6 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<float> TrainLengthM;
         /// <summary>
-        /// Train's actual absolute speed.
-        /// </summary>
-        public Func<float> SpeedMpS;
-        /// <summary>
         /// Locomotive direction.
         /// </summary>
         public Func<Direction> CurrentDirection;
--- a/Source/Documentation/Manual/cabs.rst
+++ b/Source/Documentation/Manual/cabs.rst
@@ -188,6 +188,39 @@ Here is an example of a cab light control block within the .cvf file::
 			      MouseControl ( 1 )
             )
 
+Dedicated buttons for brake controllers
+---------------------------------------
+
+.. index::
+    single: ORTS_BAILOFF
+
+In addition to the BailOff keyboard command, a cabview control named
+ORTS_BAILOFF is available. It is used to release the brakes of the engine
+while keeping the train brakes applied.
+
+.. index::
+    single: ORTS_QUICKRELEASE
+
+In some brake controllers, there is a button that provides a full and quick
+release of the train brake when pressed. OR supports this via the
+ORTS_QUICKRELEASE cabview control.
+
+.. index::
+    single: ORTS_OVERCHARGE
+
+Some brake controllers have a dedicated button to overcharge the brake pipe.
+The ORTS_OVERCHARGE cabview control can be used for this purpose.
+
+Here is an example of one of this controls within the .cvf file::
+
+			TwoState (
+            Type ( ORTS_BAILOFF TWO_STATE )
+			      Position ( 120 425 30 21 )
+			      Graphic ( BailOff.ace )
+			      NumFrames ( 2 2 1 )
+			      Style ( PRESSED )
+			      MouseControl ( 1 )
+            )
 
 Signed Traction Braking control
 -------------------------------
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -1922,32 +1922,31 @@ Open Rails software has implemented its own braking physics in the
 current release. It is based on the Westinghouse 26C and 26F air brake
 and controller system. Open Rails braking will parse the type of braking
 from the .eng file to determine if the braking physics uses passenger or
-freight standards, self-lapping or not. This is controlled within the
-Options menu as shown in :ref:`General Options <options-general>` above.
+freight standards, self-lapping or not. 
 
-Selecting :ref:`Graduated Release Air Brakes <options-general>` in *Menu >
-Options* allows partial release of the brakes. Some 26C brake valves have a
-cut-off valve that has three positions: passenger, freight and cut-out. Checked
-is equivalent to passenger standard and unchecked is equivalent to freight
-standard.
+There are two different features regarding graduated release of brakes.
+If the train brake controller has a self-lapping notch that provides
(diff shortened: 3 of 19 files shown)


peternewell:
> 
> @peternewell what do you think about this changes?
> 
In principal I agree with the changes, however I have a few concerns, and will raise them with you in a PM through ET.

YoRyan:
Not to be too nosy, Peter, but it would be helpful to have your comments in the public record, if only in summary form.

YoRyan:
As for myself, there are a couple of things I think we should keep in mind:

- We need to be careful about preserving the physics and handling behavior of existing content - yes, going all the way back to the MSTS Kuju moddels. Some minor realism-oriented changes are okay.
- We *do* have a scripted brake interface, so some of the more obscure cases can rely on C# scripting instead of one-off .eng tokens. (I think? Correct me if I'm wrong about the scope of the brake controller.)

cesarBLG:
The scripting interface is useful to model uncommon or very realistic controllers. For example, it allowed the creation of the PBL26 brake controller used in France and Spain which has a centered stable position, and springs for Apply and Release positions. It's also the way to go for modern locomotives with electronically controlled brakes. However, the tokens that I introduce here are used in a variety of trains, and it's nice to provide them for content creators without programming knowledge.

YoRyan:
Agreed, and it's great that you're trying to strike that balance.

peternewell:
> 
> 
> Not to be too nosy, Peter, but it would be helpful to have your comments in the public record, if only in summary form.
>

In summary, I want to ensure that we don't make global changes which impact in areas where they shouldn't. In principal I have no problems with the changes that Cesar has proposed, it is probably more about how they fit into a bigger picture view. 

For example, I agree that graduated release should not be a tick box menu item, but it is how it should be implemented that I want to work with Cesar on. 

{As an aside, changes to global menu tickbox raise the issue of how we depreciate the existing menu tick box item, as I believe it should be removed from the menu. This highlights what I feel is a bit of an issue in OR, as we seem to have an approach where we need to maintain all functionality, whereas in reality some functionality should be removed if it has become redundant or superfluous. }

In principal we only have a handful of different brake control valves (see Darwin's document in the relevant thread) and if we define the correct tokens for each one we should get an "accurate" representation of each different brake type. Ideally users could almost "cut and paste" the relevant brake type into their ENG file.

I have found that this PR comment interface is a bit cumbersome to work with if sharing detailed information including reference documents, so therefore I have started a PM with Cesar to share some thoughts with.

cesarBLG:
@YoRyan You were interested some time ago in air and dynamic brake blending based on controller position. I have been doing some tests with a modern EMU which has an speed dependent blending, with the addition of a SetDynamicBrake() hook to the C# brake controller. I could add this hook together with this PR if you like so.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -143,10 +143,6 @@ public abstract class TrainControlSystem : AbstractScriptClass
         /// </summary>
         public Func<float> TrainLengthM;
         /// <summary>
-        /// Train's actual absolute speed.
```
Sharpe49, line 146: Is that intentional ?

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -143,10 +143,6 @@ public abstract class TrainControlSystem : AbstractScriptClass
...
-        /// </summary>
-        public Func<float> SpeedMpS;
-        /// <summary>
```
Sharpe49, line 149: Is that intentional ?

cesarBLG: Yes, I moved it to AbstractScriptClass. Does this cause a problem?

YoRyan:
> ... with the addition of a SetDynamicBrake() hook to the C# brake controller. I could add this hook together with this PR if you like so.

Sure, I think such a hook is warranted.

The setup I discussed is common enough in North American locomotives that I'd also like to see some token support, but we can deal with that some other time.

cesarBLG:
Darwins and I have found several brake controllers where dynamic brake is used, so I agree with you that this should be supported without scripting. In fact, I've achieved a working version for the MSTSBrakeController using a brake controller to dynamic brake application interpolator. This would cover most situations, but I know some EMUs where the dynamic brake application depends on speed (due to different performance of dynamic braking at different speeds). Ryan, could you provide some information about how is blending managed in American locos?

cesarBLG:
In order to not have this PR open for too much time, pending improvements will be addressed later with subsequent pull requests, so this is ready now.
I added a table with the description of brake controller notches. Unfortunately, it seems that it is not shown correctly in the PDF version of the Manual. The table should expand to several pages, but it is forced to be in one page and it's partially hidden. Is there a way to solve this?

cjakeman (approved):
Hi Cesar,

I think this has been in Unstable for long enough and both Ryan and Cedric have inspected and commented on it.

Approved, with thanks,

Chris

cjakeman:
Discussion: http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/
Related trello card: https://trello.com/c/0SvMICQt/485-air-brake-and-ep-brake-options

PR #324, opened 2020-12-20
Mark signal glow material for not loosing it.

pzgulyas:
I wonder if this wasn't it missing for you guys. :-) (Hopefully the PR is now to the right branch.)

### DIFF
--- a/Source/RunActivity/Viewer3D/Signals.cs
+++ b/Source/RunActivity/Viewer3D/Signals.cs
@@ -413,6 +413,7 @@ namespace Orts.Viewer3D
             internal void Mark()
             {
                 SignalTypeData.Material.Mark();
+                SignalTypeData.GlowMaterial?.Mark();
             }
         }
 


Source/RunActivity/Viewer3D/Signals.cs
```diff
@@ -413,6 +413,7 @@ void renderEffect(Material material)
             internal void Mark()
             {
                 SignalTypeData.Material.Mark();
+                SignalTypeData.GlowMaterial?.Mark();
```
YoRyan, line 416: Is there any reason to guard against null here? The glow material isn't loaded from the route data.

pzgulyas: In the constructor below this line there is an if-else condition, and the GlowMaterial isn't set in both cases.

YoRyan: Gotcha

YoRyan (approved):

Sharpe49:
You forgot a part of the OR workflow: 
Is there a bug in Launchpad resolved by this? A blueprint? A Trello item?

YoRyan:
Definitely just a bug, but it hadn't filed on Launchpad. Oops. I guess I should have asked for a report, but I deferred to Peter's judgment.

PR #325, opened 2020-12-23
Fix invisible forest shadows: https://bugs.launchpad.net/or/+bug/1863217

YoRyan:
A revised, smaller forest shadows patch that merges cleanly into master and Monogame. With a little help from @pzgulyas...


Csantucci (approved):
The shadows show up now, the fix has been agreed with gpz, the fix is small even if conceptual, so I may approve the PR.

PR #328, opened 2020-12-24
3 new banner images replace redundant ones

cjakeman:
New banner images from VAPOR3D.
I've not extended the current list of 70 as I found some near duplicates that could be retired.


YoRyan:
If they're redundant, why not just overwrite the images instead of renaming the old ones to "was X.jpg"? We'll always have them in Git history.

cjakeman:
> If they're redundant, why not just overwrite the images instead of renaming the old ones to "was X.jpg"? We'll always have them in Git history.

Good point. I had forgotten to delete them.

Thanks,

Chris

YoRyan (approved):
Cheers! Approved and merged.

PR #329, opened 2020-12-26
Display shadow only behind text in on-screen messages

Csantucci:
As of now shadows of on-screen messages cover the whole width of the screen, even if the string is very short. This is not nice to see and can also partially hide the cabview controls. The patch displays the shadow only behind of the text. See http://www.elvastower.com/forums/index.php?/topic/34709-shadow-behind-on-screen-messages/
Trello box https://trello.com/c/FaitVFxc/489-shadow-only-behind-text-in-on-screen-messages


YoRyan (approved):
A small but much-needed change. Thanks for tackling this.

Csantucci:
Ryan, thanks for fast approval.

PR #330, opened 2020-12-30
Bug fix for https://bugs.launchpad.net/or/+bug/1909684 AI train speed  variable not initialized

Csantucci:
Please check comment in Launchpad


YoRyan (approved):
Fix looks good. Regarding your comment on Launchpad, I guess you could submit a new PR that renames the `AITrain.LastSpeedMpS` and `AITrain.SpeedMpS` variables, if you think it's worth your time to refactor.

Csantucci:
Thanks Ryan. OK, I'll leave the eventual refactoring for later. BTW, FYI only LastSpeedMpS is defined both in Train.cs and AITrain.cs; SpeedMpS is defined only in Train.cs and therefore doesn't lead to misunderstandings.

PR #331, opened 2021-01-03
Correct issues with brake cuts power in vacuum brakes

peternewell:
https://blueprints.launchpad.net/or/+spec/vacuum-brake-update - fixes some issues identified with vacuum braking.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -1337,11 +1337,16 @@ namespace Orts.Simulation.RollingStocks
             {
                 if (TrainBrakeController.MaxPressurePSI == 21 || TrainBrakeController.MaxPressurePSI == 25) // If 21 or 25 has been entered assume that it is 21InHg or 25InHg, and convert it to the correct psi equivalent
                 {
+                    float TempMaxPressure = TrainBrakeController.MaxPressurePSI;
+
+                    // Convert assumed inHg value to psi
+                    TrainBrakeController.MaxPressurePSI = Bar.ToPSI(Bar.FromInHg(TrainBrakeController.MaxPressurePSI));
+
                     if (Simulator.Settings.VerboseConfigurationMessages)
                     {
-                        Trace.TraceInformation("TrainBrakeController.MaxPressurePSI is assumed to be {0} Inhg, - confirmed as a value of {1} InHg", TrainBrakeController.MaxPressurePSI, Bar.ToInHg(Bar.FromPSI(Bar.ToPSI(Bar.FromInHg(TrainBrakeController.MaxPressurePSI)))));
+                        Trace.TraceInformation("TrainBrakeController.MaxPressurePSI is assumed to be {0} Inhg, - confirmed as a value of {1} InHg", TempMaxPressure, Bar.ToInHg(Bar.FromPSI(TrainBrakeController.MaxPressurePSI)));
                     }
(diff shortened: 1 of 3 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -1333,35 +1333,44 @@ public override void Initialize()
...
-            if ((BrakeSystem is VacuumSinglePipe) && TrainBrakeController.MaxPressurePSI > 10 && TrainBrakeController.MaxPressurePSI < 15)
+            if ((BrakeSystem is VacuumSinglePipe))
             {
-                Trace.TraceInformation("TrainBrakeController.MaxPressurePSI being incorrectly read as {0} Inhg, - set to value of {1} InHg", TrainBrakeController.MaxPressurePSI, Bar.ToInHg(Bar.FromPSI(TrainBrakeController.MaxPressurePSI)));
-                TrainBrakeController.MaxPressurePSI = Bar.ToInHg(Bar.FromPSI(TrainBrakeController.MaxPressurePSI));
+                if (TrainBrakeController.MaxPressurePSI == 21 || TrainBrakeController.MaxPressurePSI == 25) // If 21 or 25 has been entered assume that it is 21InHg or 25InHg, and convert it to the correct psi equivalent
+                {
+                    if (Simulator.Settings.VerboseConfigurationMessages)
+                    {
+                        Trace.TraceInformation("TrainBrakeController.MaxPressurePSI is assumed to be {0} Inhg, - confirmed as a value of {1} InHg", TrainBrakeController.MaxPressurePSI, Bar.ToInHg(Bar.FromPSI(Bar.ToPSI(Bar.FromInHg(TrainBrakeController.MaxPressurePSI)))));
```
cjakeman, line 1342: Hi Peter,

These conversions puzzle me. Why are we changing a value in InHg to bar and then to PSI and then to bar and finally to InHg? There must be something I'm missing.

Also, when units of measure are missing like this, how about adding a warning in the log that vacuum pressures are best defined with units (bar, psi, inHg) to avoid ambiguity.

Best wishes, 

Chris

cjakeman (commented):
Hi Peter,

I notice this PR is in Draft mode. Can you click on the "Ready for review" button so I can approve it?

Thanks,

Chris

Thanks,

Chris

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
peternewell: Hi Chris,

> These conversions puzzle me. Why are we changing a value in InHg to bar and then to PSI and then to bar and finally to InHg? There must be something I'm missing.

I agree it is a bit complicated.

I wanted to provide both the input value from the user, and the new default value created. So there are two ways to do this. The way that I have done it, and the other alternative is to temporarily save the user value, and this will reduce the number of conversion in one place.

In the interest of making the logic for the code clearer, I will change the commit to try and reflect the 2nd approach.
 
> Also, when units of measure are missing like this, how about adding a warning in the log that vacuum pressures are best defined with units (bar, psi, inHg) to avoid ambiguity.
The OR manual highlights the need to add UoM to parameters.

Indirectly the associated error messages are to highlight that there is a lack of clarity that OR has attempted to "fix". Most MSTS files would be configured without the UoM.

peternewell:
> I notice this PR is in Draft mode. Can you click on the "Ready for review" button so I can approve it?
Sorry about that. It should be ready for review now.

Thanks

Peter

cjakeman (approved):
Hi Peter,

All is good now.

Approved with thanks,

Chris

PR #332, opened 2021-01-04
Bug fix for https://bugs.launchpad.net/or/+bug/1910154. Log File generates the same header warning message.

mbm-OR:
This fix allows to differentiate the warning message between wagon and engine.

### DIFF
--- a/Source/Orts.Simulation/MultiPlayer/Message.cs
+++ b/Source/Orts.Simulation/MultiPlayer/Message.cs
@@ -576,7 +576,7 @@ namespace Orts.MultiPlayer
                                     string wagonFilePath = MPManager.Simulator.BasePath + @"\trains\trainset\" + cars[i];
                                     if (!File.Exists(wagonFilePath))
                                     {
-                                        Trace.TraceWarning("Ignored missing wagon {0}", wagonFilePath);
+                                        Trace.TraceWarning($"Ignored missing {(wagonFilePath.EndsWith(".eng")? "engine" : "wagon")} {wagonFilePath}");
                                         continue;
                                     }
 
--- a/Source/Orts.Simulation/MultiPlayer/Message.cs
+++ b/Source/Orts.Simulation/MultiPlayer/Message.cs
@@ -576,7 +576,7 @@ namespace Orts.MultiPlayer
                                     string wagonFilePath = MPManager.Simulator.BasePath + @"\trains\trainset\" + cars[i];
                                     if (!File.Exists(wagonFilePath))
                                     {
-                                        Trace.TraceWarning($"Ignored missing {(wagonFilePath.EndsWith(".eng")? "engine" : "wagon")} {wagonFilePath}");
+                                        Trace.TraceWarning($"Ignored missing rolling stock {wagonFilePath}");
                                         continue;
                                     }
 
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -901,7 +901,7 @@ namespace Orts.Simulation.AIs
 
                 if (!File.Exists(wagonFilePath))
                 {
-                    Trace.TraceWarning("Ignored missing wagon {0} in consist {1}", wagonFilePath, consistFileName);
+                    Trace.TraceWarning($"Ignored missing {(wagon.IsEngine ? "engine" : "wagon")} {wagonFilePath} in consist {consistFileName}");
                     continue;
                 }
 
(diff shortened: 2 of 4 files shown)


Csantucci (approved):
Everything improving readability of the logfile is welcome. Thanks, Mauricio.

Source/Orts.Simulation/MultiPlayer/Message.cs
```diff
@@ -576,7 +576,7 @@ public override void HandleMsg()
                                     string wagonFilePath = MPManager.Simulator.BasePath + @"\trains\trainset\" + cars[i];
                                     if (!File.Exists(wagonFilePath))
                                     {
-                                        Trace.TraceWarning("Ignored missing wagon {0}", wagonFilePath);
+                                        Trace.TraceWarning($"Ignored missing {(wagonFilePath.EndsWith(".eng")? "engine" : "wagon")} {wagonFilePath}");
```
YoRyan, line 579: I'd rather not hardcode a file extension like this. Instead, we should use a term like "vehicle" that covers both engines and wagons.

YoRyan (changes requested):

Source/Orts.Simulation/MultiPlayer/Message.cs
mbm-OR: This pull request allows to improve the readability of the log file.
The string interpolation doesn’t modify any file extensions. 
The following code, could be another option with the same output messages.
[Unstable-LogFile-Improve-Readability-Not-string-interpolation.txt](https://github.com/openrails/openrails/files/5788405/Unstable-LogFile-Improve-Readability-Not-string-interpolation.txt)

YoRyan: No, it's not the use of string interpolation that I'm concerned about. It's that you're assuming every file that OR ever loads will be a .eng or .wag. We cannot expect this to hold true forever, and we also cannot expect any future developer to go back and update this section of code. If we add a new ".eng-or" format, for example, your code is going to print "wagon."

Also, as written, your check is case-sensitive, so it also thinks ".ENG" is a "wagon."

mbm-OR: This pull request uses the extensions that OR code, is currently looking for.
In the future, if the extensions are modified, I’m sure that the developers will improve the code according to the new changes.
About case-sensitive, OR code is always working with “**.eng**” and “**.wag**” extensions, if you consider removing the case-sensitive from this pull request, not a problem here.

YoRyan: > This pull request uses the extensions that OR code, is currently looking for.

But it's important to keep in mind not just what OR can do now, but also what it may be able to do in the future.

> In the future, if the extensions are modified, I’m sure that the developers will improve the code according to the new changes.

This is not a given. If somebody adds support for a new format down the line, it's quite unlikely they're going to go out of their way to update your logic -- which is difficult to find, because it's in the logging section -- to distinguish between "engine" and "wagon." Therefore, for maintainability's sake, it's best that we not make this distinction at all, except in a few well-defined locations.

Csantucci: Rejecting changes because in an undefined future things could change doesn't seem the best policy to me.

mbm-OR: Let's try to strike a balance between different point of view about this pull request.

The Message.cs file is part of the Multiplayer mode, and the code is activated only when the "wagonFilePath" doesn’t exist.
To avoid misunderstandings, it would be reasonable to replace only the text “Ignored missing **wagon**” by “Ignored missing **rolling stock**”. 
[Unstable-LogFile-Warning-Message.txt](https://github.com/openrails/openrails/files/5791294/Unstable-LogFile-Warning-Message.txt)


The rest of this pull request code, using the “wagon.IsEngine” flag, is well-defined and easy to locate.

Otherwise, revert this pull request.

Csantucci: Mauricio, I can agree with this.

YoRyan: Thanks, that's what I was suggesting. I'm sorry if I didn't make clear that the rest of the PR was fine by me.

> Rejecting changes because in an undefined future things could change doesn't seem the best policy to me.

Well, we do refactor things every now and then, don't we? Wouldn't it be better if we avoided such issues in the first place? I believe I've explained why spreading file extensions around is something to avoid.

YoRyan (approved):
Resolved and approved!

PR #335, opened 2021-01-15
Fix broken web track monitor API call and displays

YoRyan:
This PR fixes the currently nonfunctional web Track Monitor, which was first noticed by @mbm-OR. It also fixes an index out of range exception generated by this code.

### DIFF
--- a/Source/RunActivity/RunActivity.csproj
+++ b/Source/RunActivity/RunActivity.csproj
@@ -185,6 +185,7 @@
     <Compile Include="Viewer3D\Weather.cs" />
     <Compile Include="Viewer3D\WebServices\TrackMonitorDisplay.cs" />
     <Compile Include="Viewer3D\WebServices\TrainDrivingDisplay.cs" />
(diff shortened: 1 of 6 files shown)


cjakeman (commented):
Hi Ryan,

Looks good. Should we leave it in Unstable for a week?

What was the issue that stopped the API call working?

Chris

YoRyan:
To be honest, few know that the web server exists in the first place, so I don't think there would be much of a benefit to waiting. But it's up to you.

As for the issue: As best as I can tell, at some point JSON serialization stopped working for the TrainInfo class. I've fixed this by creating a new struct exclusive to the web server code. (This is best practice anyway, since, as Peter pointed out, it's a bad idea to expose internal data structures to outside code, TCS or HTTP or otherwise.)

cjakeman (approved):
Hi Ryan,

Thanks for the info.

Approved.

Chris

cjakeman:
Fix for http://www.elvastower.com/forums/index.php?/topic/34740-monogame-v85v86-chronic-crashes-due-to-web-server/

PR #336, opened 2021-01-15
Add secret button to access launcher arguments for easier debugging

YoRyan:
When debugging RunActivity.exe, I often need to copy the launch parameters (`-start -explorer ...`) set by the Open Rails launcher into my Visual Studio project so that I can reproduce a specific test scenario. Now, *yes*, they can be extracted from OpenRailsLog.txt, but this format is not very convenient, because the log file prints the arguments on separate lines and does not escape the spaces with quotes.

Therefore, I propose adding a secret function that copies the exact launch arguments set by the launcher to the clipboard. This function would be activated by holding down the Alt key while clicking on the Start button. Hopefully, no player would do this by accident, but in the event that somebody does, an instructional message is also presented to explain what just happened.

### DIFF
--- a/Source/Menu/Program.cs
+++ b/Source/Menu/Program.cs
@@ -152,14 +152,27 @@ namespace ORTS
                             break;
                     }
 
-                    var processStartInfo = new System.Diagnostics.ProcessStartInfo();
-                    processStartInfo.FileName = MainForm.RunActivityProgram;
-                    processStartInfo.Arguments = String.Join(" ", parameters.ToArray());
-                    processStartInfo.WindowStyle = System.Diagnostics.ProcessWindowStyle.Normal;
-                    processStartInfo.WorkingDirectory = Application.StartupPath;
-
-                    var process = Process.Start(processStartInfo);
-                    process.WaitForExit();
+                    var joinedParameters = string.Join(" ", parameters);
+                    if ((Control.ModifierKeys & Keys.Alt) == Keys.Alt)
+                    {
+                        Clipboard.SetText(joinedParameters);
+                        MessageBox.Show(
+                            "RunActivity.exe arguments have been copied to the clipboard:" +
+                            $"\n\n{joinedParameters}\n\n" +
+                            "This is a debugging aid. If you wanted to start the simulator instead, select Start without holding down the Alt key.");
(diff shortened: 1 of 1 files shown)


cjakeman:
> When debugging RunActivity.exe, I often need to copy the launch parameters (`-start -explorer ...`) set by the Open Rails launcher into my Visual Studio project so that I can reproduce a specific test scenario.

Good idea; it's a pain for me too. 

Your clipboard idea will work fine, but wouldn't it convenient to have it also available in the log file? You'll likely have that open as part of the debugging process and it would then also be present in log files that users send in.

YoRyan:
They're already present in the log file. To take one example from Elvas Tower:

```
Version    = 1.3.1.4328
Build      = 1.3.1.4328 (2000-01-02 02:24:16Z)
Logfile    = C:\Users\rosar\OneDrive\Escritorio\OpenRailsLog.txt
Executable = RunActivityLAA.exe
Argument   = -start
Argument   = -explorer
Argument   = C:\Program Files (x86)\Microsoft Games\Train Simulator\ROUTES\FCGR Electrificado Finalizado\PATHS\9001 CEMEN V BARKER.pat
Argument   = C:\Program Files (x86)\Microsoft Games\Train Simulator\TRAINS\CONSISTS\hotbox.con
Argument   = 12:00
Argument   = 1
Argument   = 0
```

The problem is that this can't just be copied to the command line to run Open Rails; you need to strip the `Argument =` text at the beginning of each line, and to encapsulate each argument in double quotes. So, by hand, you would have to turn this into:

```
-start -explorer "C:\Program Files (x86)\Microsoft Games\Train Simulator\ROUTES\FCGR Electrificado Finalizado\PATHS\9001 CEMEN V BARKER.pat" "C:\Program Files (x86)\Microsoft Games\Train Simulator\TRAINS\CONSISTS\hotbox.con" 12:00 1 0
```

Hence having a secret function to do this for you. I went with the clipboard because it is easy to code and leaves no traces behind.

Sharpe49 (approved):
The code seems good to me.
@cjakeman Are the blueprints still used? I don't really see them being updated and this modification is not linked to a blueprint.

PR #338, opened 2021-01-15
Bug fix for https://bugs.launchpad.net/or/+bug/1911908 Variable1 goes negative when steam loco runs backwards

Csantucci:
Variable1 is required to be always positive for the .sms files.
See http://www.elvastower.com/forums/index.php?/topic/34789-volumecurve-variable1controlled-in-a-chuff-stream/#entry267315


YoRyan (approved):
Looks good to me.

PR #340, opened 2021-01-16
Create C# signalling scripting interface

cesarBLG:
When building advanced signalling systems, the SIGSCR language becomes too limited. With this simple addition, started by @Sharpe49 , C# scripts can be used in addition to the SIGSCR.
This includes two new features specifically designed for C# scripts:
- A new TextAspect variable, used both in signal and TCS scripts, to have a wider range of signal aspects.
- A function to send string messages to other signals.

Discussion: http://www.elvastower.com/forums/index.php?/topic/34792-c-signal-scripts/
Blueprint: https://blueprints.launchpad.net/or/+spec/csharp-signal-scripts


Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -47,6 +47,7 @@ static CompilerParameters GetCompilerParameters()
             cp.ReferencedAssemblies.Add("System.dll");
             cp.ReferencedAssemblies.Add("System.Core.dll");
             cp.ReferencedAssemblies.Add("ORTS.Common.dll");
+            cp.ReferencedAssemblies.Add("Orts.Formats.Msts.dll");
```
cesarBLG, line 50: I'm not happy with this, I would prefer converting MstsSignalAspect and MstsBlockState to other enum types so this is not required. Good candidates are Aspect enum as used in TrainControlSystem and InternalBlockState used by signalling logic, or new enums can be created.

Source/Orts.Simulation/MultiPlayer/Message.cs
```diff
@@ -34,6 +34,7 @@
 using System.IO.Compression;
 using System.Linq;
 using Event = Orts.Common.Event;
+using System.Text;
```
YoRyan, line 37: Super minor nitpick, but keep the imports sorted in alphabetical order. :)

Source/Orts.Simulation/MultiPlayer/Message.cs
```diff
@@ -3018,7 +3019,11 @@ public class MSGSignalStatus : Message
         static byte[] preState;
         static SortedList<long, SignalHead> signals;
         public bool OKtoSend = false;
+        bool SendEverything;
         static byte[] signalsStates;
+        int changedTextStates = 0;
+        static string[] signalTextStates;
+        static string[] preTextState;
```
YoRyan, line 3026: So unfortunately, we have to maintain backwards compatibility in the multiplayer protocol. (Yes, this sucks.) The reason is because there is a [public relay server](http://www.tsimserver.com/ORFiles031205/ServerInfo.html), whose source code has not been made public and whose protocol is fixed, that is used by many Open Rails players. We will get complaints if Open Rails can no longer connect to this server.

I am not familiar with this section of the multiplayer code, so maybe there's no issue, but it's something to be aware of FYI.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
```diff
@@ -0,0 +1,170 @@
...
+    public abstract class CsSignalScript
+    {
+        // References and shortcuts. Must be private to not expose them through the API
+        private SignalHead SignalHead { get; set; }
+        private SignalObject SignalObject => SignalHead.mainSignal;
+        private int SigFnIndex(string sigFn) => SignalObject.signalRef.ORTSSignalTypes.IndexOf(sigFn);
+        private SignalObject SignalObjectById(int id) => SignalObject.signalRef.SignalObjects[id];
+
+        // Public interface
+        public string DebugFileName = String.Empty;
```
YoRyan, line 19: Let's document the public interface with docstrings (`/// <summary> ... </summary>`).

Source/Orts.Simulation/MultiPlayer/Message.cs
cesarBLG: Today I participated in a MP session using the public server and there were not more crashes than usual. We have to test it again, with more signals using text aspects, to see if it works well. I think the server just broadcasts the message, as it is sent as a base64 compressed string, and there is no point in decoding it there.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
cesarBLG: Sure! I have half of them already documented in my working copy, as well as the interface with TCS

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
Sharpe49: I think you can use Aspect from ORTS.Scripting.Api without any trouble. You just have to convert it back at the right locations inside OR.

cjakeman:
Hi Cesar,

Is this still in "Draft"? If not, then would you press "Ready for review" to change it over.

Thanks,

Chris

cesarBLG:
Hi Chris,
I finished the coding part, but I left this as a draft just in case there were new suggestions before final review.

I have a lot of work from university, but if I get some free time I'll add a note in the manual for documentation, and then I'll take the PR out of draft.

Source/RunActivity/Viewer3D/Debugging/DebugViewerForm.cs
```diff
@@ -1937,6 +1937,7 @@ private void boxSetSignalChosen(object sender, EventArgs e)
                       else if (drawstate2 > 0) { sigHead.state = MstsSignalAspect.APPROACH_2; }
                       else { sigHead.state = MstsSignalAspect.APPROACH_3; }
                       sigHead.draw_state = sigHead.def_draw_state(sigHead.state);
+					  sigHead.TextSignalAspect = "";
```
YoRyan, line 1940: Sorry, but what does this line too?

cesarBLG: The way the dispatcher window is implemented bypasses the signal logic and forces a fixed aspect. In order to not confuse other signals and TCS scripts when an (msts)aspect is forced, the text aspect has to be reset, otherwise it would be kept in an inconsistent state.

YoRyan: Thanks, that clears things up!

Source/Orts.Simulation/MultiPlayer/Message.cs
YoRyan: I take it all went well.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
```diff
@@ -0,0 +1,375 @@
...
+        /// </summary>
+        public Dictionary<int, int> SharedVariables => SignalObject.localStorage;
+        /// <summary>
+        /// Current time in seconds
+        /// </summary>
+        public float ClockTimeS => (float)SignalObject.signalRef.Simulator.GameTime;
+        /// <summary>
+        /// Represents a timer which can be started and stopped several times, with a configurable trigger value
+        /// </summary>
+        public class Timer
```
YoRyan, line 121: Ideally we wouldn't have to copy and paste this code for every new scriptable subsystem. I'll make a note to investigate whether we can share it among multiple classes.

YoRyan (dismissed):
Thanks for tackling this, Cesar. Everything looks good to me, after adding some suggestions of my own. Do test to see if any further problems crop up, and if not, then we can merge with @cjakeman's approval.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
cesarBLG: One possibility is that CsSignalScript inherits from AbstractScriptClass, where these are defined. Then we would have to create a new abstract class for train scripts, adding DistanceM(), SpeedMpS() and Odometer functionality there.

What do you think about using the Orts.Scripting.Script namespace instead of Orts.Signalling.Signals for the signal scripts?

cesarBLG:
Thanks for your code refactoring, Ryan. I sometimes forget that C# is a high level language and has useful built-in features, and I reinvent the wheel. The serialization trick is much cleaner than my approach, for sure.

Source/Orts.Simulation/MultiPlayer/Message.cs
```diff
@@ -3125,18 +3125,10 @@ public MSGSignalStatus(string m)
                 {
                     gZipStream.Read(signalsStates, 0, signalsStates.Length); // Read integer values for aspect and draw state
 
-                    byte[] intbuffer = new byte[4];
-                    gZipStream.Read(intbuffer, 0, 4);
-                    int textAspectCount = BitConverter.ToInt32(intbuffer, 0);
-                    for (int i = 0; i < textAspectCount; i++)
+                    if (new BinaryFormatter().Deserialize(gZipStream) is SignalTextState[] textAspects)
```
cesarBLG, line 3128: I found this article in Microsoft docs: https://aka.ms/binaryformatter
If binary deserialization allows remote code execution, probably it would be better to avoid it.
@YoRyan An intermediate approach could be using BinaryReader and BinaryWriter classes as we do for saves. What do you think?

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
YoRyan: That sounds like a good idea. I agree it would be more fitting to place a scriptable subsystem under the Script namespace.

Source/Orts.Simulation/MultiPlayer/Message.cs
YoRyan: Hmm, yes, that is a legitimate security concern. I agree that BinaryReader would be a good intermediate approach that would still be pretty clean. Feel free to revert my commit and implement this, or I can do so if you're busy.

On a related note, I'm not happy with the state of the rest of the multiplayer code, either. I feel like a maliciously formatted message could easily cause a crash. Ideally, the whole thing would use a safe serialization format like JSON.

cesarBLG: Everything about multiplayer needs some cleanup, including signalling, train position report and communication. The signalling code is spreaded with several MPManager.IsMultiPlayer() checks that cause signals to work different in MP.

It would be much cleaner to use JSON, but a lot of people use the public server, so it's a problem.

YoRyan: Yes, in the long run, we've got to ditch the public server, too, either by writing a replacement or implementing peer-to-peer session borwsing. It's unacceptable that a closed-source program is such a central part of our multiplayer infrastructure.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
cesarBLG: Now the signal scripts inherit from AbstractScriptClass, so the built-in timer can be used, They must reside in the ORTS.Scripting.Script namespace to be loaded.

YoRyan (approved):
Thanks, Cesar! This looks good to me. Pending final approval by @cjakeman.

cjakeman (approved):
Looks very good.

Well done, everyone.

Approved with thanks,

Chris

Sharpe49:
Great, I will update the scripts for the French lines.

PR #344, opened 2021-01-19
Bug fix for https://bugs.launchpad.net/or/+bug/1912388 Crash with $attach and $detach in timetable mode

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34801-tt-crash-when-player-attaches-train/


YoRyan (approved):
Thanks for the fix. This one looks good!

PR #345, opened 2021-01-22
Bug fix for https://bugs.launchpad.net/or/+bug/1912683. Tapping on the OR-icon, it didn’t change anything.

mbm-OR:
Tapping on the web TrainDrivingInfo OR-icon, it didn’t change anything.
More info here: http://www.elvastower.com/forums/index.php?/topic/34740-monogame-v85v86-chronic-crashes-due-to-web-server/page__view__findpost__p__267403

### DIFF
--- a/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
+++ b/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
@@ -57,7 +57,7 @@ async function ApiTrackMonitor() {
 	// GET to fetch data, POST to send it
 	// "/API/APISAMPLE" /API is a prefix hard-coded into the WebServer class
 	let [tnInfo, tmDisplay] = await Promise.all([ApiGet("TRAININFO"), ApiGet("TRACKMONITORDISPLAY")]);
-	
+
 	let Str = "<table>";
 	let endIndexFirst = 0,
 		endIndexTrackLeft = 0,
@@ -255,7 +255,7 @@ async function ApiTrainDriving() {//*** TrainDriving code ***
 		lastColor = false;
 		keyColor = false;
 		symbolColor = false;
-	
+
 		// FirstCol
 		if (data.FirstCol != null) {
 			endIndexFirst = data.FirstCol.length;
@@ -336,7 +336,7 @@ async function ApiTrainDriving() {//*** TrainDriving code ***
 	}
 
 	// space at bottom
-	Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
+	Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' ontouchstart='this.onclick()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
 	Str += "</table>";
 	TrainDriving.innerHTML = Str;
 }
@@ -368,7 +368,7 @@ async function DownloadImage(path) {
 			image.onerror = reject;
 			image.src = path;
 		});
-    }
+	}
 }
 
 function DisplayItem(alignment, colspanvalue, isColor, colorCode, item, small){
@@ -417,14 +417,7 @@ function dragMouseElement(tdDrag) {
 		initY = touch.clientY;
 		document.ontouchend = closeDrag;
 		document.ontouchmove = touchMove;
-
-        var currentsrc = touch.target.currentSrc;
-        if (currentsrc != undefined){
-            var png = 'or_logo.png';
-            if (currentsrc.slice(-png.length) === png){
-                changeNormalTextMode();
-        }}
-    }
+	}
 
 	function initDrag(event) {
 		event.preventDefault();
@@ -469,9 +462,9 @@ function dragMouseElement(tdDrag) {
 		}else{
 			tdDrag.style.border = "0px solid gray";
 		}
-	    offsetX = initX - event.clientX;
-	    offsetY = initY - event.clientY;
-	    initX = event.clientX;
+		offsetX = initX - event.clientX;
+		offsetY = initY - event.clientY;
+		initX = event.clientX;
 		initY = event.clientY;
 		// avoids to overlap the trackmonitor window
 		tdDrag.style.left = (collision && offsetX > 0 ? tdDrag.offsetLeft : tdDrag.offsetLeft - offsetX) + "px";// X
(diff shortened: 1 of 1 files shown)


YoRyan:
Hmm, there are a lot of indentation-related changes in this diff that make no changes to the code. We would prefer if you excluded them from the PR.

Also, there's already [a line](https://github.com/openrails/openrails/blob/b572880456aa27527b0b5e0a02224d8e0f8436e2/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js#L339) in the current code that supposedly enables the mode-changing function. I'm curious why this doesn't this work as written. A lack of an `ontouchstart` attribute, perhaps?

mbm-OR:
_Hmm, there are a lot of indentation-related changes in this diff that make no changes to the code…_
Visual Studio Code looks better with the code arranged. Are these indentations causing an unknown trouble for me? 

_there's already a line in the current code_
The line in the current code that enables the mode-changing function, works correctly when we use a mouse.
In the case of a touch screen, It doesn't work at all. This PR allows that tapping on the OR-icon we enable the mode-changing.

YoRyan:
> Visual Studio Code looks better with the code arranged. Are these indentations causing an unknown trouble for me?

Well, your indentation changes do not match the surrounding lines. Wouldn't it be preferable to stay consistent?

> The line in the current code that enables the mode-changing function, works correctly when we use a mouse.
> In the case of a touch screen, It doesn't work at all. This PR allows that tapping on the OR-icon we enable the mode-changing.

So, actually, this bug only affects Safari on iOS. MDN's documented this behavior [here](https://developer.mozilla.org/en-US/docs/Web/API/Element/click_event#safari_mobile). There are various possible fixes, but the simplest one to me seems to be to change this line:

```
	Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
```

to


```
	Str += "<tr> <td colspan='5' style='text-align: center'><img src='/or_logo.png' height='16' width='16' onclick='changeNormalTextMode()'> </td> </tr>";
```

I would prefer this fix because it's a one-liner that doesn't involve changing the global `touchstart` event.

mbm-OR:
#Wouldn't it be preferable to stay consistent?
In some cases, VS auto format to make the methods and loops indented properly, when a bracket is touched.

#I would prefer this fix because it's a one-liner that doesn't involve changing the global touchstart event.
We are working with the draggable  "TrainDrivingInfo" window selected. That is, by selecting "Track Monitor and TrainDriving" from the web interface in the option's menu.
In this case, your one-liner fix doesn't work here. (iPad Pro emulated, Android tablet).

mbm-OR:
It seems that the index.js file page encoding, was changed from UTF-8 BOM to UTF-8 by default.
Updated this pull request, to restore the original page encoding and the indentations.

YoRyan:
Thanks for reverting the indentation changes, Mauricio.

> In this case, your one-liner fix doesn't work here. (iPad Pro emulated, Android tablet).

Apologies. I have it working with:

```diff
diff --git a/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js b/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
index 8eb41cbd8..12174a14a 100644
--- a/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
+++ b/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
@@ -336,7 +336,7 @@ async function ApiTrainDriving() {//*** TrainDriving code ***
 	}
 
 	// space at bottom
-	Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
+	Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center; cursor: pointer;'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
 	Str += "</table>";
 	TrainDriving.innerHTML = Str;
 }
@@ -496,4 +496,4 @@ function isCollide(a, b) {
 		((a.x + a.width + gap) < b.x) ||
 		(a.x + gap > (b.x + b.width))
 	);
-}
\ No newline at end of file
+}
diff --git a/Source/RunActivity/Viewer3D/WebServices/Web/TrainDriving/index.js b/Source/RunActivity/Viewer3D/WebServices/Web/TrainDriving/index.js
index db7138580..7ca1185c0 100644
--- a/Source/RunActivity/Viewer3D/WebServices/Web/TrainDriving/index.js
+++ b/Source/RunActivity/Viewer3D/WebServices/Web/TrainDriving/index.js
@@ -145,7 +145,7 @@ function ApiTrainDriving() {
 				}
 				Str += "</table>";
 				// space at bottom
-				Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
+				Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center; cursor: pointer;'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
                 Str += "</table>";
 				TrainDriving.innerHTML = Str;
 			}
@@ -171,4 +171,4 @@ function changePageColor() {
 
 function changeNormalTextMode() {
 	normalTextMode = !normalTextMode;
-};
\ No newline at end of file
+};
```

YoRyan:
Okay, I owe you a double apology for posting a non-working patch.

I had neglected to test on an iDevice until now. The bug doesn't arise from the delegated click event, as I thought, but rather because `onclick` does not fire when the train driving window is dragged.

I certify that this one works. Tested on an Android phone and on an iPad Air 2:

```diff
diff --git a/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js b/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
index 8eb41cbd8..afdd2575a 100644
--- a/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
+++ b/Source/RunActivity/Viewer3D/WebServices/Web/TrackMonitorTrainDriving/index.js
@@ -336,7 +336,7 @@ async function ApiTrainDriving() {//*** TrainDriving code ***
        }

        // space at bottom
-       Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
+       Str += "<tr> <td colspan='5' onclick='changeNormalTextMode()' ontouchstart='this.onclick()' style='text-align: center'><img src='/or_logo.png' height='16' width='16'></img></td> </tr>";
        Str += "</table>";
        TrainDriving.innerHTML = Str;
 }
```

mbm-OR:
Sometimes the obvious fix is the last thing we do. It has happened to both of us.
Your patch works also here.
Pull request updated.

YoRyan (approved):
Thanks, Mauricio. This looks good.

PR #346, opened 2021-01-24
Fix mispositioned digital cab displays in letterboxing mode

YoRyan:
In letterboxing mode, digits are rendered slightly to the right of their intended positions. This is because I forgot to change the width and height computations to match the letterboxed display.

Reported by pschlik at Elvas Tower: http://www.elvastower.com/forums/index.php?/topic/34813-digital-displays-misaligned-in-letterbox-cab/


Csantucci (approved):
This seems to be OK, so merge it when you think it's appropriate.

YoRyan:
Cheers! I'll give it a week in unstable.

PR #347, opened 2021-01-26
https://blueprints.launchpad.net/or/+spec/horn-blow-sequence American horn blow sequence for AI trains

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34721-ai-train-horn-honk-american-style/page__pid__267884#entry267884

### DIFF
--- a/Source/Documentation/Manual/operation.rst
+++ b/Source/Documentation/Manual/operation.rst
@@ -1817,8 +1817,12 @@ AI Train Horn Blow
 ------------------
 
 Horn blow by AI trains is achieved by inserting into the AI train path a 
-waiting point with a waiting time value between 60011 (1 second horn blow) and 
-60020 (10 seconds horn blow).
+waiting point.
+If the waiting time value is between 60011 (1 second horn blow) and 
+60020 (10 seconds horn blow), a single horn blow is generated.
+If the waiting time value is 60021, a horn blow sequence is generated, 
+with the pattern long blow - long blow - short blow - long blow (North 
+American horn pattern at level crossings).
 
 The AI train will not stop at these waiting points, but will continue at its 
 regular speed.
@@ -1855,7 +1859,22 @@ is inserted into the activity file following the line::
 
 (note that the number in the brackets may be different), then AI trains will 
 blow their horn at level crossings for a random time between 2 and 5 
-seconds.The level crossing must be defined as such in the MSTS route editor. 
+seconds. The level crossing must be defined as such in the MSTS route editor. 
+
+If line::
+
+    ORTSAIHornAtCrossings ( 1 )
+
+is followed by line::
+
+	ORTSAICrossingHornPattern ( US )
+
+instead of a single horn blow, a horn blow sequence will be generated
+for all AI trains before crossings,  
+with the pattern long blow - long blow - short blow - long blow (North 
+American horn pattern at level crossings). 
+
+
 *Simple* road crossings, not defined as level crossings, may also be present in 
 the route. The AI train will not blow the horn at these crossings. Examining 
 the route with TrackViewer allows identification of the true level crossings. 
--- a/Source/Orts.Simulation/Simulation/LevelCrossing.cs
+++ b/Source/Orts.Simulation/Simulation/LevelCrossing.cs
@@ -18,7 +18,6 @@
 // This file is the responsibility of the 3D & Environment Team. 
 
 using Orts.Formats.Msts;
-using Orts.Formats.OR;
 using Orts.Simulation.AIs;
 using Orts.Simulation.Physics;
 using ORTS.Common;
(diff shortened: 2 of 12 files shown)


Source/Orts.Simulation/Simulation/LevelCrossing.cs
```diff
@@ -182,27 +185,40 @@ void UpdateCrossings(Train train, float elapsedTime)
...
-                // MSTS did not simulate a timeout, I introduced a simple timout using speedMpS.
+                    // MSTS did not simulate a timeout, I introduced a simple timout using speedMpS.
 
-                // Depending upon future development in this area, it would probably be best to have the current operation in its own class followed by any new region specific operations. 
+                    // Depending upon future development in this area, it would probably be best to have the current operation in its own class followed by any new region specific operations. 
 
-                // Recognizing static consists at crossings.
-                if ((train.TrainType == Train.TRAINTYPE.STATIC) && validStaticConsist)
+                    // Recognizing static consists at crossings.
+                    if ((train.TrainType == Train.TRAINTYPE.STATIC) && validStaticConsist)
```
YoRyan, line 221: Unexplained indentation changes in this chunk

YoRyan:
So, my main concern here is that there are no provisions for additional horn patterns. I cannot speak for all of the world's railways, but I suspect there are more level crossing patterns besides "one blast" and "long-long-short-long."

Rather than overloading the current property, `ORTSAIHornAtCrossings ( 2 )`, I propose introducing a new property, `ORTSAIHornCrossingPattern ( US )`, that would be easier to read for activity designers and easier to extend for additional patterns. And the code would also benefit from some dedicated horn pattern classes and syntactic sugar. That `AuxActionItem` class is looking rather bulky with all of those `const int AmHorn_SomeProperty_` variables tacked on (plus who knows how more for any future horn patterns).

If these suggestions sound good to you, I would be happy to provide a diff.

Csantucci:
OK, so I'm waiting for your diff. Important is that activity builders will continue to add only one line to get the pattern they want.

YoRyan:
Sorry for the long wait, Carlo. Attached is my [suggested patch](https://github.com/openrails/openrails/files/5957057/ryan-americanhorn.diff.txt). It's rather long and complicated, so don't hesitate to ask if you see something you'd like clarified.

I've focused on refactoring and simplifying things where possible; this is exemplified by the new AILevelCrossingHornPattern classes, which are simple to create horn patterns with. The tradeoff is that I dropped the ability to save the game while a horn is playing and have it play again on reload. That's especially difficult to write and is probably more trouble than it's worth.

For the fuller picture, try diffing against `master` after applying the patch.

Csantucci:
Ryan,
thank you for your effort.
I tested your patch and looked at it.
I have an issue: I don't think it's a clean solution to have all these hornPattern arguments in the AI Train module calls, because you're adding a quite secondary parameter in the argument list and pass it from call to call. I'd derive LevelCrossingHornPattern directly within AITrain() in module AITrain.cs.

YoRyan:
That's a fair critique. I suppose you could do so in `PostInit()`, where some activity-specific stuff currently gets handled.

Csantucci:
I've inserted the changes agreed. If the code is OK now, I proceed to the update to the OR manual.

Csantucci:
Added "not-for-unstable" tag because crashing when saving activity because LevelCrossingHornPattern may be null.

Csantucci:
Save bug  removed.

YoRyan:
No objections here! Feel free to write that manual text.

Csantucci:
OR manual updated

cjakeman:
Hi Carlo,

> OR manual updated

Thanks for finishing this.

Ryan,

If you're happy with the updates to the Manual, will you approve the work?

Thanks,

Chris

Source/Documentation/Manual/operation.rst
```diff
@@ -1855,7 +1859,22 @@ is inserted into the activity file following the line::
...
 (note that the number in the brackets may be different), then AI trains will 
 blow their horn at level crossings for a random time between 2 and 5 
-seconds.The level crossing must be defined as such in the MSTS route editor. 
+seconds. The level crossing must be defined as such in the MSTS route editor. 
+
+If line::
+
+    ORTSAIHornAtCrossings ( 1 )
+
+is followed by line::
```
YoRyan, line 1868:
```suggestion
If the following line is inserted::
```
I don't believe the order of the properties matters. Same goes for `ORTSAIHornAtCrossings ( 1 )` and `NextActivityObjectUID ( 32768 )` mentioned just above this section. If so, could we clarify that here?

YoRyan (commented):
Looks good, Chris. I just have one small question about the text.

Source/Documentation/Manual/operation.rst
Csantucci: The position of the 1 or two added lines with respect to NextActivityObjectUID ( 32768 ) must be the one defined, else the MSTS AE crashes when opening such an activity. It was not easy to find a position for the added line (it was only one up to the last change you suggested) that didn't cause crashes to MSTS. It is true that the reciprocal order of the two added lines doesn't matter, but I prefer my wording to yours, because you don't specify where the other line must be inserted. 
Moreover there  is no advantage for the activity writer to know that he can add the new line before ORTSAIHornAtCrossings ( 1 ). So I don't agree to change what I have written.

YoRyan: Oh! I had completely forgotten about the activity editor. This is really interesting information that I think deserves to be in the manual (in the form of another PR).

> ...and I also think that one should calibrate the requests for refinements to the importance of them.

I do apologize if this point seems pedantic, but it's still untrue information, and that didn't sit well with me. All I wanted was some clarification, which you gave in spades, so thanks very much.

YoRyan (approved):

PR #348, opened 2021-01-29
Adjust water scoop sounds

peternewell:
https://blueprints.launchpad.net/or/+spec/water-trough-filling - Adjust water scoop sounds so that they only come on when over trough, add extra sound events for water scoop broken, and raise/lower


cjakeman (approved):
Hi Peter,

Yes, that all looks fine.

Thanks for the comment about triggers released in MG.

Approved,

Chris

PR #352, opened 2021-02-11
Enforce EQ pressure minimal reduction at service rate

cesarBLG:
Currently the "minimal reduction" pressure drop is performed instantly. With this changes the pressure drops at MaxApplicationRate as expected, and ensures that the valve is not lapped before completing the minimal reduction.

See http://www.elvastower.com/forums/index.php?/topic/34853-equalizing-reservoir-unexpected-behavior/

Bug report: https://bugs.launchpad.net/or/+bug/1915432

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
@@ -106,27 +106,34 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Controllers
                     var type = notch.Type;
                     if (OverchargeButtonPressed()) type = ControllerState.Overcharge;
                     else if (QuickReleaseButtonPressed()) type = ControllerState.FullQuickRelease;
+
+                    if (type == ControllerState.Hold || type == ControllerState.Lap || type == ControllerState.MinimalReduction)
+                    {
+                        if (EnforceMinimalReduction)
+                            DecreasePressure(ref pressureBar, MaxPressureBar() - MinReductionBar(), ApplyRateBarpS(), elapsedClockSeconds);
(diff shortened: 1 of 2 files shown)


cesarBLG:
I've just noticed a potential issue with vacuum brakes, so I have to address it before adding it to unstable

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/MSTSBrakeController.cs
```diff
@@ -104,6 +106,17 @@ public override void UpdatePressure(ref float pressureBar, float elapsedClockSec
...
+
+                    if (type == ControllerState.Hold || type == ControllerState.Lap || type == ControllerState.MinimalReduction)
+                    {
+                        if (EnforceMinimalReduction)
+                            DecreasePressure(ref pressureBar, MaxPressureBar() - MinReductionBar(), ApplyRateBarpS(), elapsedClockSeconds);
+                    }
+                    else
+                    {
+                        EnforceMinimalReduction = false;
+                    }
```
YoRyan, line 118:
```suggestion
                    switch (type)
                    {
                        case ControllerState.Hold:
                        case ControllerState.Lap:
                        case ControllerState.MinimalReduction:
                            if (EnforceMinimalReduction)
                                DecreasePressure(ref pressureBar, MaxPressureBar() - MinReductionBar(), ApplyRateBarpS(), elapsedClockSeconds);
                            break;
                        default:
                            EnforceMinimalReduction = false;
                            break;
                    }
```
With three or more cases for the same variable, I suggest using a switch-case.

YoRyan (approved):
Thanks, this looks good to me. I'll approve, but let's wait a week until merging so that we know there haven't been any unintended changes.

cesarBLG:
Thanks, Ryan

PR #353, opened 2021-02-13
Sound section corrected

cjakeman:
See http://www.elvastower.com/forums/index.php?/topic/34876-automatic-switch-and-curve-squeal/page__st__10__gopid__268638#entry268638

### DIFF
--- a/Source/Documentation/Manual/sound.rst
+++ b/Source/Documentation/Manual/sound.rst
@@ -415,7 +415,7 @@ To enable this feature steps here below must be followed:
 
      SIMISA@@@@@@@@@@JINX0t1t______
      
-     13
+     14
      TrackType ( "Default" "EuropeSteamTrack0In.sms" "EuropeSteamTrack0Ex.sms" )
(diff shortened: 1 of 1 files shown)


Csantucci:
Chris, thanks for tackling this.
As the author of this feature, I am checking this. I see that in fact there is an error in the manual, but it is in another point: in the tsection.dat sample, there is a line that shouldn't be there, which is 
`TrackType ( "Switch" "switchtrack7in.sms" "switchtrack7ex.sms" )`
Once that removed, all original numbering present in the manual is correct.
So I suggest to remove such line from the manual and leave the original numbering in the manual unchanged. It is better to do so, because the original tsection.dat files of the 6 MSTS legacy routes have 10 entries, and not 11.
Sorry for the error I introduced.

cjakeman:
Hi Carlo,

> So I suggest to remove such line from the manual and leave the original numbering in the manual unchanged. It is better to do so, because the original tsection.dat files of the 6 MSTS legacy routes (and of many other routes) have 10 entries, and not 11.

No problem.

I've undone my changes and applied yours.

Best wishes,

Chris

Csantucci:
Thank you, Chris.

Csantucci (approved):

PR #354, opened 2021-02-17
add Learn > Activities page

cjakeman:
Initially to point visitors to the tutorials on the Siskiyou route
See http://www.elvastower.com/forums/index.php?/topic/34636-two-new-releases-at-the-siskiyou-route-website/page__p__265485__hl__activities__fromsearch__1#entry265485


YoRyan (approved):
HTML and English both look good. Will put a Markdown version on my todo list...

PR #355, opened 2021-02-18
added missing entry for Options - Simulation

cjakeman:
Simple control and physics
See http://www.elvastower.com/forums/index.php?/topic/33437-simple-vs-full-control-and-physics/page__p__251952__hl__%22simple+control%22__fromsearch__1#entry251952


Source/Documentation/Manual/options.rst
```diff
@@ -516,6 +516,21 @@ meet.
     found under :ref:`Location-Linked Passing Path Processing <operation-locationpath>`
     in the chapter  *Open Rails Train Operation*.
 
+.. _options-simple-control-and-physics:
+
```
YoRyan, line 520:
```suggestion
```
Anchors like this are not needed if there are no links to this section.

YoRyan (changes requested):
Please also address the `options-simulation.png` file, which was committed empty. Thanks!

cjakeman:
> Please also address the `options-simulation.png` file, which was committed empty. Thanks!

Thanks, but I think you might want to look again. :-)
All 144Kb are there and I've managed to get GitHub to show it as an image too.

YoRyan:
Thanks! Having checked out your branch with Git, I see that your file is indeed populated. Not sure why my GitHub interface thinks it's empty...

YoRyan (approved):

PR #357, opened 2021-02-22
Correct issues with diesel power not being applied correctly

peternewell:
Correct a number of issues identified with diesel power - https://bugs.launchpad.net/or/+bug/1916431

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
@@ -547,7 +547,7 @@ namespace Orts.Simulation.RollingStocks
         /// <summary>
         /// This function updates periodically the locomotive's motive force.
         /// </summary>
-        protected override void UpdateMotiveForce(float elapsedClockSeconds, float t, float AbsSpeedMpS, float AbsWheelSpeedMpS)
+        protected override void UpdateTractiveForce(float elapsedClockSeconds, float t, float AbsSpeedMpS, float AbsWheelSpeedMpS)
         {
             // This section calculates the motive force of the locomotive as follows:
             // Basic configuration (no TF table) - uses P = F /speed  relationship - requires power and force parameters to be set in the ENG file. 
@@ -600,23 +600,23 @@ namespace Orts.Simulation.RollingStocks
 
                     if (DieselEngines.HasGearBox)
                     {
-                        MotiveForceN = DieselEngines.MotiveForceN;
+                        TractiveForceN = DieselEngines.TractiveForceN; 
                     }
                     else
                     {
                         if (maxForceN * AbsSpeedMpS > maxPowerW)
                             maxForceN = maxPowerW / AbsSpeedMpS;
 
-                        MotiveForceN = maxForceN;
+                        TractiveForceN = maxForceN;
                         // Motive force will be produced until power reaches zero, some locomotives had a overspeed monitor set at the maximum design speed
                     }
                 }
                 else
                 {
                     // Tractive force is read from Table using the apparent throttle setting, and then reduced by the number of engines running (power ratio)
-                    MotiveForceN = TractiveForceCurves.Get(LocomotiveApparentThrottleSetting, AbsSpeedMpS) * DieselEngineFractionPower * (1 - PowerReduction);
-                    if (MotiveForceN < 0 && !TractiveForceCurves.AcceptsNegativeValues())
-                        MotiveForceN = 0;
+                    TractiveForceN = TractiveForceCurves.Get(LocomotiveApparentThrottleSetting, AbsSpeedMpS) * DieselEngineFractionPower * (1 - PowerReduction);
+                    if (TractiveForceN < 0 && !TractiveForceCurves.AcceptsNegativeValues())
+                        TractiveForceN = 0;
                 }
 
                 DieselFlowLps = DieselEngines.DieselFlowLps;
@@ -640,11 +640,11 @@ namespace Orts.Simulation.RollingStocks
 
             if (MaxForceN > 0 && MaxContinuousForceN > 0 && PowerReduction < 1)
             {
-                MotiveForceN *= 1 - (MaxForceN - MaxContinuousForceN) / (MaxForceN * MaxContinuousForceN) * AverageForceN * (1 - PowerReduction);
+                TractiveForceN *= 1 - (MaxForceN - MaxContinuousForceN) / (MaxForceN * MaxContinuousForceN) * AverageForceN * (1 - PowerReduction);
                 float w = (ContinuousForceTimeFactor - elapsedClockSeconds) / ContinuousForceTimeFactor;
                 if (w < 0)
                     w = 0;
-                AverageForceN = w * AverageForceN + (1 - w) * MotiveForceN;
+                AverageForceN = w * AverageForceN + (1 - w) * TractiveForceN;
             }
         }
 
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
@@ -581,40 +581,59 @@ namespace Orts.Simulation.RollingStocks
 
                 DieselEngineFractionPower = MathHelper.Clamp(DieselEngineFractionPower, 0.0f, 1.0f);  // Clamp decay within bounds
 
+
+                // For the advanced adhesion model, a rudimentary form of slip control is incorporated by using the wheel speed to calculate tractive effort.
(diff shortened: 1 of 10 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
```diff
@@ -581,42 +581,61 @@ protected override void UpdateMotiveForce(float elapsedClockSeconds, float t, fl
 
                 DieselEngineFractionPower = MathHelper.Clamp(DieselEngineFractionPower, 0.0f, 1.0f);  // Clamp decay within bounds
 
+
+                // For the advanced adhesion model, a rudimentary form of slip control is incorporated by using the wheel speed to calculate tractive effort.
+                // As wheel speed is increased tractive effort is decreased. Hence wheel slip is "controlled" to a certain extent.
+                // This doesn't cover all types of locomotives, for eaxmple if DC traction motors and no slip control, then the tractive effort shouldn't be reduced. This won't eliminate slip, but limits
+                // its impact. More modern locomotive have a more sophisticated system that eliminates slip in the majority (if not all circumstances).
+                // Simple adhesion control does not have any slip control feature built into it.
+                // TODO - a full review of slip/no slip control.
```
cesarBLG, line 590: What Pschlik said about using wheel speed for tractive effort calculation is not related to slip control, if I understood him. Back EM forces produced in DC motors are proportional to the speed at which the motor is running, which itself depends on wheel speed, not train speed. Therefore, using wheel speed should be the correct behavior, at least during wheelslip (in normal operation, both speeds are almost equal). I agree with the code changes, but I think the comment is misleading.

A different thing happens with slip control: it entirely depends on train speed. If train speed is much lower than wheel speed, tractive effort is reduced (by reducing the power given to the motor), but the motor itself will still have the tractive curves depending on wheel speed (just with a lower throttle setting). However, as you say, slip control needs a deeper review.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -1571,17 +1572,22 @@ public override void Update(float elapsedClockSeconds)
...
 
-            // Update dynamic brake force
             if (DynamicBrakePercent > 0 && DynamicBrakeForceCurves != null && AbsSpeedMpS > 0)
             {
-                float f = DynamicBrakeForceCurves.Get(.01f * DynamicBrakePercent, AbsSpeedMpS);
+                float f = DynamicBrakeForceCurves.Get(.01f * DynamicBrakePercent, AbsTractionSpeedMpS);
                 if (f > 0 && PowerOn)
                 {
                     DynamicBrakeForceN = f * (1 - PowerReduction);
+                    MotiveForceN -= (AbsTractionSpeedMpS > 0 ? 1 : AbsTractionSpeedMpS < 0 ? -1 : Direction == Direction.Reverse ? -1 : 1) * DynamicBrakeForceN;
```
cesarBLG, line 1590: I don't understand this. AbsTractionSpeedMpS is always positive, so why to check if it is negative?
Have you taken into account the case when the car is driven from the rear cab?

I'll approve after this is clarified.

peternewell: I have reviewed this code, and as you rightly point out it is more about getting the "sign" on the dynamic brake force correct. So I have reverted this line to as it was in earlier versions. The lines above it are the more important to change.

Thanks for pointing this out.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
peternewell: > I agree with the code changes, but I think the comment is misleading.

Back EMF may apply to a motor under load, but at the moment without further research, I am not convinced that it necessarily applies to traction motors under no load conditions, ie slipping.

One technical publication that I read suggested the following, "A  series-wound  traction  motor  cannot  be  operated  on  no-load  under  any  circumstances  because  it  increases  its  speed  to  such  a  degree  that  the  armature  is  destroyed,  i.e.  it  ‘races’ on no-load".

This maybe a a fully DC locomotive, and I appreciate that more modern locomotives have DC traction motors with AC/DC inverters which may provide some of the feature that you and Pschlik are suggesting. 

As I understand slip control, we started with none, as per the very early DC locomotives. So in this instance the increase in wheel speed with slip could destroy the motor. These locomotives relied on the driver to reduce throttle settings before the motor was destroyed.

Then automatic sanding and/or a rudimentary feedback system was used to reduce the driving force applied to the traction motor. I understand that this was based upon the motor current as a representation of the degree of wheel slip. I feel that OR is currently modelling such a rudimentary feedback control mechanism.

Finally more modern slip control systems have become more sophisticated, and now use radar readings and computer algorithms to manage slip, and almost make it impossible to get into a slip situation.

OR already models some reduction in tractive force via the axle model, using wheel speed in non slip situations results in a "double" reduction in tractive force if wheel creep is occurring.

Anyway as you suggest my main point was that I believe that OR currently doesn't model all slip scenarios correctly at the moment, and this is an area for some further work. Along the way we need to confirm the facts, either way.

I am not sure what is misleading about my comments, if you have any suggested word changes to make it clearer, then I am happy to change it.

cesarBLG: I still doubt about if it wouldn't be better to always use the wheel speed for tractive effort calculations. The ground speed is "unknown" to the motors, and shouldn't affect in any way to tractive effort calculation. Could you please give some arguments about why to use AbsSpeedMpS instead of AbsWheelSpeedMpS?

peternewell: > Could you please give some arguments about why to use AbsSpeedMpS instead of AbsWheelSpeedMpS?

The use of the AbsSpeedMpS based upon this description in a technical article:

"Maximum tractive or braking effort is obtained if each powered wheel of the vehicle is rotating at such an angular velocity that its actual peripheral speed is slightly higher (motoring)  or slightly lower (braking)  than the true vehicle speed (i.e., the linear speed at which the vehicle is traveling, usually referred to as "ground speed" or "track speed"). The difference  between  wheel  speed  and  track  (or  "ground") speed is referred to as "slip speed." There is a relatively low limit value of slip speed  at which peak tractive or braking effort is realized. This value, commonly known as maximum "creep speed," is a variable that depends on track speed and rail conditions. So long as the maximum creep speed is not exceeded, slip speed is normal and the vehicle will operate in a stable microslip or creep mode. If wheel-to-rail adhesion tends to be reduced or lost, some or all of the vehicle wheels may  slip  excessively,  i.e.,  the  actual  slip  speed  may  be greater  than the maximum  creep speed. Such  a wheel slip condition,  which is characterized  in the motoring  mode by one  or  more  spinning  axle-wheel  sets  and in  the braking mode by one or more sliding  or skidding  axle-wheel  sets, can cause accelerated wheel wear, rail damage, high mechanical stresses in the drive components  of the propul­sion  system,  and  an  undesirable  decrease  of  tractive  (or braking) effort."

So my interpretation is that creep speed (before wheel starts spinning uncontrollably) is normal. Uncontrolled wheel slip is bad.

So in the creep zone, train speed (AbsSpeedMpS) should be used as this is producing the "closest to real life" input force for the axle model. If we use wheel speed in this zone of operation then the input force to the axle module is being decreased because the  wheel speed will always be greater then the train speed. The tractive (motive) force is already being decreased in the axle module, and there is no need to further reduce the input force to the axle module. Hence if we use AbsWheelSpeedMpS we are getting a double reduction.

As AbsWheelSpeedMpS increases as the wheel goes into uncontrollable slip, then the motive force that OR reads from the Tractive force curves is going to be less then the value determined using the train speed. Hence this is acting as a "feedback and control" mechanism to try and contain the speed of the wheel. How accurately this approach mirrors real life when slip occurs, I am not still not sure without further research.

So in short, until further research, and rewriting of the slip code is done, I am comfortable that AbsSpeedMpS should be used whilst the wheel is "creeping", and that AbsWheelSpeedMpS should be used in the time being, until the wheel slip module is reviewed, for situations where the wheel experiences uncontrolled wheel slip.

cesarBLG: Thanks for the explanation, I agree.

cesarBLG (approved):
Thanks for fixing all reported regressions, Peter.

YoRyan:
Thanks for tackling this, you two. Let's give it at least a week in unstable before merging (which is our policy for complex changes that are difficult to test).

peternewell:
> Thanks for fixing all reported regressions, Peter.
Thanks for the approval, I will merge it next week sometime.

PR #358, opened 2021-02-22
Fix cab letterboxing memory leak: https://bugs.launchpad.net/or/+bug/1885342

YoRyan:
Create a black texture for pillarboxing just once instead of recreating one each frame. This avoids a gradually worsening memory leak that is known to affect the current code.

With Monogame merged, this change can now be applied.


cjakeman (approved):
Hi Ryan,

Thanks for pushing this one again.

Approved,

Chris

PR #360, opened 2021-02-22
Bug fix for https://bugs.launchpad.net/or/+bug/1916525. TimeTable mode, #settings for storage area, fails.

mbm-OR:
The code related to the #settings for storage area, does not work correctly if there is an empty field in that string.


YoRyan (approved):
A simple fix. I'm approving, but please do not merge for at least a week, so we can watch for any bugs that might crop up.

PR #361, opened 2021-02-24
Correct issue with null value

peternewell:
https://bugs.launchpad.net/or/+bug/1916688


peternewell:
Close PR to resolve conflict

PR #362, opened 2021-02-24
Correct null value bug

peternewell:
https://bugs.launchpad.net/or/+bug/1916688

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/VacuumSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/VacuumSinglePipe.cs
@@ -437,85 +437,85 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
 
             // Brake information is updated for each vehicle
 
-                if (EngineBrake && (Car.WagonType == MSTSWagon.WagonTypes.Engine || Car.WagonType == MSTSWagon.WagonTypes.Tender)) // Only apples when an engine brake is in place, otherwise processed to next loop
+            if (EngineBrake && (Car.WagonType == MSTSWagon.WagonTypes.Engine || Car.WagonType == MSTSWagon.WagonTypes.Tender)) // Only apples when an engine brake is in place, otherwise processed to next loop
+            {
+                // The engine brake can only be applied when the train brake is released or partially released. It cannot be released whilever the train brake is applied.
+                if (lead.TrainBrakeController.CurrentValue == 0 && lead.EngineBrakeController.CurrentValue > 0) // If train brake is completely released & Engine brake is applied
                 {
-                    // The engine brake can only be applied when the train brake is released or partially released. It cannot be released whilever the train brake is applied.
-                    if (lead.TrainBrakeController.CurrentValue == 0 && lead.EngineBrakeController.CurrentValue > 0) // If train brake is completely released & Engine brake is applied
+                    CylPressurePSIA = BrakeLine3PressurePSI;
+                }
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/VacuumSinglePipe.cs
```diff
@@ -547,8 +547,9 @@ public override void Update(float elapsedClockSeconds)
                     // depending upon whether steam brake fitted or not, calculate brake force to be applied
                     if (LocomotiveSteamBrakeFitted && (Car.WagonType == MSTSWagon.WagonTypes.Engine || Car.WagonType == MSTSWagon.WagonTypes.Tender))
                     {
-                        f = Car.MaxBrakeForceN * Math.Min(SteamBrakeCylinderPressurePSI / lead.MaxBoilerPressurePSI, 1);
-                    }
+                    var leadLocomotiveMaxBoilerPressurePSI = lead == null ? 200.0f : (lead.MaxBoilerPressurePSI);
+                    f = Car.MaxBrakeForceN * Math.Min(SteamBrakeCylinderPressurePSI / leadLocomotiveMaxBoilerPressurePSI, 1);
+                }
```
YoRyan, line 552:
```suggestion
                        var leadLocomotiveMaxBoilerPressurePSI = lead?.MaxBoilerPressurePSI ?? 200.0f;
                        f = Car.MaxBrakeForceN * Math.Min(SteamBrakeCylinderPressurePSI / leadLocomotiveMaxBoilerPressurePSI, 1);
                    }
```

YoRyan: Match the existing indentation and use the nullable operators.

YoRyan (changes requested):

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/VacuumSinglePipe.cs
peternewell: Indentations now corrected.

My preference is to leave the null as it is. I want to do this as I think that this is more understandable for "newbie" coders who don't have your skills. In its abbreviated form it doesn't indicate that it is linked to a null. Is there any difference in processing etc to justify the more abbreviated form?

YoRyan: No, shouldn't be a performance-related difference.

Nullables are a really handy syntax to have, particularly when you're dealing with chains of properties that may or may not be null, for example:

```C#
return Viewer?.SomeLocomotive?.SomeBoiler?.MaxBoilerPressurePSI ?? 200f;
```

versus

```C#
if (Viewer != null
        && Viewer.SomeLocomotive != null
        && Viewer.SomeLocomotive.SomeBoiler != null
        && Viewer.SomeLocomotive.SomeBoiler.MaxBoilerPressurePSI != null)
    return Viewer.SomeLocomotive.SomeBoiler.MaxBoilerPressurePSI;
else
    return 200f;
```

but the choice is up to you. I certainly won't impose my personal preferences on anyone.

peternewell: In this instance, I prefer to retain my coding for the nulls.

Are you happy to approve on this basis?

YoRyan: Certainly.

YoRyan (approved):

PR #368, opened 2021-03-01
Menu options help

cjakeman:
Help icons with links to the on-line manual added just for Options > General tab

Feature request: https://www.elvastower.com/forums/index.php?/topic/34931-links-to-the-on-line-manual/page__view__findpost__p__269252

I can add the same for other tabs, but let's see what feedback we get first.


Source/Menu/Options.cs
```diff
@@ -759,5 +760,94 @@ private void checkPerformanceTuner_Click(object sender, EventArgs e)
             labelPerformanceTunerTarget.Enabled = checkPerformanceTuner.Checked;
         }
 
+
+        #region Help for General Options
+        /// <summary>
+        /// Loads a relevant page from the manual maintained by James Ross's automatic build
+        /// Info icon taken from https://icons8.com/icon/set/info/office--blue
```
YoRyan, line 767: Icons8's terms of use [dictate](https://intercom.help/icons8-7fb7577e8170/en/articles/4725508-where-do-i-set-the-link) that we need to add a link to them in our About dialog, not just in the source code. Is there an alternative, preferably public domain set we could use?

YoRyan (changes requested):

Source/Menu/Options.cs
cjakeman: Hi Ryna,
Thanks for spotting this.
Couldn't find any suitable public domain ones, so changed Icon8 ones for homemade.
Best wishes,
Chris

YoRyan (approved):
All looks good to me, Chris. Ready to merge when you are.

PR #369, opened 2021-03-02
Copies ExtClock from OR New Year MG

cjakeman:
Adds the "external clock" which Jonas implemented for version OR New Year MG.

Feature request: https://www.elvastower.com/forums/index.php?/topic/29546-station-clocks/page__view__findpost__p__254573 which links to a route with 2 clocks.

Jonas provides that route with 2 clocks, but the second hand on one of those seems to be stuck.

Need some advice on naming, please.

Jonas uses &lt;routename&gt;\Openrails\clocks.dat to specify shape files for each clock in the route and whether digital or analog.

As this is a new file, it should be in JSON format, e.g. my_clocks.clock-or and where should it be located? I'll check these points with James.

Jonas refers to ORClock and OR-Clock but I am not aware of any clocks pre-OR, so we ought to drop the "OR" prefix.
He also uses an ExtClockFile ("Ext" for external) as opposed to the 2D cabview "dialclock" and "digitalclock" and that sounds fine. I heard of "CLOCK_DIGITAL_CLOCK" and "CLOCK_ANALOG_CLOCK" for 3D cabs but I don't find that in the code anywhere.

No entry in the documentation yet until these points are settled.


YoRyan:
All of the changes you've mentioned will have to be coordinated with @Csantucci; otherwise, we'll split the community with two incompatible formats. Most significantly, the "clock.dat" format is currently SIMIS, not JSON.

Csantucci:
I wonder if this feature was created before the JSON rule was introduced (I assume yes), and in this case, to avoid rewriting it and for compatibility, I'd suggest to accept the SIMIS format and keep the clocks.dat naming. 
If ORClock and OR-Clock are only internal code references, as it seems, I agree that the OR prefix can be dropped.

cjakeman:
Hi Carlo,

> I wonder if this feature was created before the JSON rule was introduced (I assume yes), and in this case, to avoid rewriting it and for compatibility, I'd suggest to accept the SIMIS format and keep the clocks.dat naming.
> If ORClock and OR-Clock are only internal code references, as it seems, I agree that the OR prefix can be dropped.

The policy for JSON goes back longer than you might imagine (2016 ! )

However, I know the policy isn't broadcast or well documented but buried in the forum, see 
http://www.elvastower.com/forums/index.php?/topic/28610-route-editor-activity-editor-and-json/page__view__findpost__p__202845
So we can't blame Jonas for not being aware of it and it's our fault for letting this new file type slip through.

Best wishes,

Chris

cjakeman:
> All of the changes you've mentioned will have to be coordinated with @Csantucci; otherwise, we'll split the community with two incompatible formats. Most significantly, the "clock.dat" format is currently SIMIS, not JSON.

The aim is to bring all the good stuff back from OR_NewYear_MG and so make it unnecessary. Carlo has done wonders keeping it going but I don't suppose he wants to prolong that effort any longer than he needs to. I wasn't proposing to push changes back to OR_NewYear_MG.

Fortunately, the files are small and not many exist so far, so it won't be hard to convert them or even to maintain two versions for a short while.

For example, openrails\clocks.dat:

```
  2
  ClockItem( Uhr01_OR_Jm.s analog )
  ClockItem( Uhr02_OR_Jm.s analog )
```

becomes my-name.clocks-or:

```
  { "ClockItem" : 
    [ 
      { "shape" : "Uhr01_OR_Jm.s"
      , "type" :  "analog"
      }
    ,
      { "shape" : "Uhr02_OR_Jm.s"
      , "type" :  "analog"
      }
    ]
  }
```

Csantucci:
If there is the intention to switch this to Json, I think that jonas should first give his OK. I don't know how much he or others have used this feature.

cjakeman:
> If there is the intention to switch this to Json, I think that jonas should first give his OK. I don't know how much he or others have used this feature.

Yes, you're right of course. I'll get him involved.

JSON files are not new; Open Rails includes the Newtonsoft.JSON library which Rob Roeterdink used for the Weather files (back in 2019) - see http://www.elvastower.com/forums/index.php?/topic/33262-use-of-weather-files-in-timetable-mode/

I think perhaps I will include a little convertor that looks for the JSON file. If it's missing, but clocks.dat is found, then the convertor creates the JSON file, which is used thereafter.

I am quite keen on the idea of having a convertor. That might be a painless way to migrate other MSTS data into an ORTS format (for example, if we added transition curves for track, then that might be better with a new ORTS format instead of trying to extend the MSTS format).

Thanks for your comments so far.

cjakeman:
So the plan is to create a convertor which will initially convert _openrails\clocks.dat_ into a JSON file _animated.clocks-or_ and, if that works OK, to convert more complex STF files into new JSON files where this provides new functionality. For example, extending the track to support transition curves and spirals.

The convertor would fit between Menu.exe and RunActivity.exe. Upon pressing "Start", Menu.exe checks whether _openrails\clocks.dat_ exists and is more recent than _animated.clocks-or_.

If so, the user is prompted to allow the convertor to run and create _animated.clocks-or_. (Open Rails promises not to change content without user's permission.) Then RunActivity.exe runs as normal and uses the new JSON file to animate the clocks.

If not, then RunActivity.exe runs as normal, ignores _openrails\clocks.dat_ and the clocks are not animated.

This is going to a lot of trouble to convert a very simple file, but I believe it would be a convenient proof-of-concept for more complex conversions in future.

YoRyan:
I don't think I agree, Chris.

First, what would be the point of such a converter program? At the moment we have exactly one JSON format in OR, the timetable weather definition file, which has no STF equivalent, and we have no concrete plans to add any additional ones. Second, I think this is the wrong question to present to the user. We can't expect players to understand the differences between STF and JSON formats; that's first and foremost a content creation matter. So I think a standalone tool, like a command-line program or a content manager, would be best.

As I see it, there are two reasons to introduce a new JSON format:

1. We're introducing a new data format that has no STF equivalent.
2. We're writing a new content editor, in which case it would be counterproductive to write to STF. But such an editor could support importing STF files from MSTS, in which case the editor itself would function as an STF to JSON converter. (The path editor included in the Track Viewer, which both reads and writes STF, was a missed opportunity to demonstrate this kind of workflow.)

This clock format clearly falls under category 1. My preference, which I don't know how Jonas would feel about, would be to remove the STF code and use JSON exclusively for this new format. Failing that, we could retain support for *both* STF and JSON clock formats. But I don't see any room here for a conversion process.

cjakeman:
Hi Ryan,

> I don't think I agree, Chris.

Let's see if I can convince you.

> As I see it, there are two reasons to introduce a new JSON format:
> 
>     1. We're introducing a new data format that has no STF equivalent.
> 
>     2. We're writing a new content editor, in which case it would be counterproductive to write to STF. But such an editor could support importing STF files from MSTS, in which case the editor itself would function as an STF to JSON converter. (The path editor included in the Track Viewer, which both reads and writes STF, was a missed opportunity to demonstrate this kind of workflow.)

> This clock format clearly falls under category 1. My preference, which I don't know how Jonas would feel about, would be to remove the STF code and use JSON exclusively for this new format. Failing that, we could retain support for _both_ STF and JSON clock formats. But I don't see any room here for a conversion process.

This PR removes the STF code from RunActivity.exe and replaces it with JSON as this is a file which is new to Open Rails. So that's clean.

However, it is not Category 1 because there are already animated clocks out there using OR_NewYear_MG.
As you suggest, I am using a standalone tool (which uses a command-line interface) to convert from 
openrails\clocks.data
to
animated.clocks-or

This is an existing tool Contrib.DataConverter which already offers conversion. I have just extended it to provide another conversion.

It's all coded now, so you can see how it works. It should be a simple one-time operation for anyone who has a clocks.dat file and no one else will get prompted.

Best wishes,

Chris

YoRyan:
I see, Chris. That makes sense to me. Thanks for clarifying.

@Csantucci will need to merge JSON support into NewYear so that newly converted files will continue to work in that fork.

cjakeman:
> @Csantucci will need to merge JSON support into NewYear so that newly converted files will continue to work in that fork.

The conversion does not remove the openrails\clocks.dat file. If the user leaves it in place then he can switch between official and NewYear versions without any action by Csantucci.

YoRyan:
I'm thinking of the case where the JSON file is available - say, because the author used the newer format - but the STF file is not.

Csantucci:
> 
> 
> I'm thinking of the case where the JSON file is available - say, because the author used the newer format - but the STF file is not.

Ryan and Chris, thank you for the alert and for the info. I have the intention to merge this into ORNYMG.

YoRyan:
Linking this Elvas Tower [thread](http://www.elvastower.com/forums/index.php?/topic/35098-route-extending-open-rails-files/), where there is a lot of discussion going on about adopting JSON and writing converters, for a permanent record.

cjakeman:
Hi Ryan,

After switching from "magic strings" to enums, the code no longer works.

The DataConverter is writing out the JSON object as
"ClockType": 0
but RunActivity is still looking for 
"ClockType": "analog"

I would rather the JSON contained:
"ClockType": "analog"
so that it is easier for people to follow.

What's your view?

Best wishes,
Chris

YoRyan:
Yes, of course Chris, this field should be serialized with the string value. See if 4ad652ce8aca268b3c2b9e4faff2ae8a1996a748 accomplishes that.

YoRyan (dismissed):
Chris, I still don't agree with tying the data converter to the game start button. I think we need a more explicit way to hide this from end users. But if this looks good to you, and if there are no remaining bugs, then I approve.

cjakeman:
Hi Ryan,

> Yes, of course Chris, this field should be serialized with the string value. See if [4ad652c](https://github.com/openrails/openrails/commit/4ad652ce8aca268b3c2b9e4faff2ae8a1996a748) accomplishes that.

Very nearly. Small change to make the comparison case-insensitive. The convertor delivers "analog" and the enum delivers "Analog".

Thanks,

Chris

cjakeman:
Hi Ryan,

> Chris, I still don't agree with tying the data converter to the game start button. I think we need a more explicit way to hide this from end users. But if this looks good to you, and if there are no remaining bugs, then I approve.

OK, then. I've taken it out since it causes so much disquiet. The code is simpler now as a result.

Thanks for converting the converter from static class to interface. I meant to base it on James' existing converter and forgot to re-visit that aspect.

The code passes my tests, so please renew your approval.

Best wishes,

Chris

YoRyan:
Right, then let's wrap this up. Paging @Csantucci to add JSON support to NewYear.

YoRyan (approved):

PR #371, opened 2021-03-05
Bug fix for https://bugs.launchpad.net/or/+bug/1917917 Turntable not turns in timetable mode.

mbm-OR:
This solves a timetable mode issue.
More info [here](http://www.elvastower.com/forums/index.php?/topic/34926-turntable-problems/page__view__findpost__p__269128).

### DIFF
--- a/Source/RunActivity/Viewer3D/Shapes.cs
+++ b/Source/RunActivity/Viewer3D/Shapes.cs
@@ -1071,23 +1071,25 @@ namespace Orts.Viewer3D
 
         public override void PrepareFrame(RenderFrame frame, ElapsedTime elapsedTime)
         {
+            float nextKey;
+            var animation = SharedShape.Animations[0];
             if (Turntable.GoToTarget || Turntable.GoToAutoTarget)
             {
-                AnimationKey = (Turntable.TargetY / (float)Math.PI * 1800.0f + 3600) % 3600.0f;
+                nextKey = Turntable.TargetY / (2 * (float)Math.PI) * animation.FrameCount;
             }
-
-            else if (Turntable.Counterclockwise)
+            else
             {
-                AnimationKey += SharedShape.Animations[0].FrameRate * elapsedTime.ClockSeconds;
+                float moveFrames;
+                if (Turntable.Counterclockwise)
+                    moveFrames = animation.FrameRate * elapsedTime.ClockSeconds;
+                else if (Turntable.Clockwise)
+                    moveFrames = -animation.FrameRate * elapsedTime.ClockSeconds;
+                else
+                    moveFrames = 0;
+                nextKey = AnimationKey + moveFrames;
             }
-            else if (Turntable.Clockwise)
-            {
-                AnimationKey -= SharedShape.Animations[0].FrameRate * elapsedTime.ClockSeconds;
-            }
-            while (AnimationKey > SharedShape.Animations[0].FrameCount) AnimationKey -= SharedShape.Animations[0].FrameCount;
-            while (AnimationKey < 0) AnimationKey += SharedShape.Animations[0].FrameCount;
-
-            Turntable.YAngle = MathHelper.WrapAngle(AnimationKey / 1800.0f * (float)Math.PI);
+            AnimationKey = nextKey % animation.FrameCount;
+            Turntable.YAngle = MathHelper.WrapAngle(nextKey / animation.FrameCount * 2 * (float)Math.PI);
 
             if ((Turntable.Clockwise || Turntable.Counterclockwise || Turntable.AutoClockwise || Turntable.AutoCounterclockwise) && !Rotating)
             {
--- a/Source/RunActivity/Viewer3D/Shapes.cs
+++ b/Source/RunActivity/Viewer3D/Shapes.cs
@@ -1089,6 +1089,8 @@ namespace Orts.Viewer3D
                 nextKey = AnimationKey + moveFrames;
             }
             AnimationKey = nextKey % animation.FrameCount;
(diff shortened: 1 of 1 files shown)


YoRyan:
Am I missing something, or is
```C#
AnimationKey %= SharedShape.Animations[0].FrameCount
```
what we're looking for here? Looping to keep a value within a range is an anti-pattern. @Roeterdink

cesarBLG:
I haven't tested this code, but I think it might work if the problem is that the frame count is not exactly 3600. The idea is to do a linear map between [0,FrameCount) and [0,2*pi), thus not requiring a 3600 fixed frame count.
```csharp
if (Turntable.GoToTarget || Turntable.GoToAutoTarget)
{
    AnimationKey = Turntable.TargetY / (2 * (float)Math.PI) * SharedShape.Animations[0].FrameCount;
}
else if (Turntable.Counterclockwise)
{
    AnimationKey += SharedShape.Animations[0].FrameRate * elapsedTime.ClockSeconds;
}
else if (Turntable.Clockwise)
{
    AnimationKey -= SharedShape.Animations[0].FrameRate * elapsedTime.ClockSeconds;
}
while (AnimationKey > SharedShape.Animations[0].FrameCount) AnimationKey -= SharedShape.Animations[0].FrameCount;
while (AnimationKey < 0) AnimationKey += SharedShape.Animations[0].FrameCount;

Turntable.YAngle = MathHelper.WrapAngle(AnimationKey / SharedShape.Animations[0].FrameCount * 2 * (float)Math.PI);
```
Ryan, unfortunately the modulus operator won't work for negative values.

Roeterdink:
> 
> 
> ```cs
> AnimationKey %= SharedShape.Animations[0].FrameCount
> ```
> 
> is probably what we're looking for here. Looping to keep a value within a range is an anti-pattern. Am I missing something? @Roeterdink

Yes, you're missing something. Please check out the related post on Elvas Tower which has all the details.
Your suggestion is only another way to write the same code and would therefor have the same problem.

YoRyan:
I got it with c28f3331564c6da8a768910a0f4c4aafb60b1fbd. The solution is to convert from radians to degrees assuming the actual frame count, not just 3600.

To verify the change, I wrote some Python code to simulate the original logic:

```python
import math

targety = -0.0005*math.pi
framecount = 3599


def wrapangle(a):
    if a > -math.pi and a <= math.pi:
        return a
    a = a % (2*math.pi)
    if a <= -math.pi:
        return a + 2*math.pi
    elif a > math.pi:
        return a - 2*math.pi
    else:
        return a

animationkey = (targety / math.pi * 1800 + 3600) % 3600
animationkey = animationkey % framecount
yangle = wrapangle(animationkey / 1800 * math.pi)

print("animationkey =", animationkey)
print("yangle =", yangle / math.pi, "* pi")
```

| TargetY | FrameCount | -> | AnimationKey | YAngle | Ok |
| --- | --- | --- | --- | --- | --- |
| -0.001*pi | 3599 || 3598.2 | -0.001*pi | ✔️ |
| -0.0005*pi | 3599 || 0.1 | 0.00006*pi | ❌ |
| 0.0005*pi | 3599 || 0.9 | 0.0005*pi | ✔️ |
| 0.001*pi | 3599 || 1.8 | 0.001*pi | ✔️ |

---

And with the new logic:

```python
import math

targety = -0.0005*math.pi
framecount = 3599


def wrapangle(a):
    if a > -math.pi and a <= math.pi:
        return a
    a = a % (2*math.pi)
    if a <= -math.pi:
        return a + 2*math.pi
    elif a > math.pi:
        return a - 2*math.pi
    else:
        return a

nextkey = targety/(2*math.pi)*framecount
animationkey = nextkey % framecount
yangle = wrapangle(nextkey/framecount*2*math.pi)

print("animationkey =", animationkey)
print("yangle =", yangle / math.pi, "* pi")
```

| TargetY | FrameCount | -> | AnimationKey | YAngle | Ok |
| --- | --- | --- | --- | --- | --- |
| -0.001*pi | 3599 || 3597.2 | -0.001*pi | ✔️ |
| -0.0005*pi | 3599 || 3598.1 | -0.0005*pi | ✔️ |
| 0.0005*pi | 3599 || 0.8998 | 0.0005*pi | ✔️ |
| 0.001*pi | 3599 || 1.7995 | 0.001*pi | ✔️ |

YoRyan:
> Ryan, unfortunately the modulus operator won't work for negative values.

Fortunately, it does. The result of a modulus operation is always positive.

> Yes, you're missing something. Please check out the related post on Elvas Tower which has all the details.

Yes, it turned out to be a bit more complex than a simple modulus operation! My apologies.

cesarBLG:
I've just tested and python and C# have different implementations for the modulus:
Python -5%3 returns +1
C# -5%3 returns -2

YoRyan:
> I've just tested and python and C# have different implementations for the modulus:

Ah! Mathematicians and Python use the floored version of modulus, which does not return a negative value, while C# uses the truncated form of modulus. [There is a difference.](https://torstencurdt.com/tech/posts/modulo-of-negative-numbers/)

Fixed by 51da8155eb6e8e2ad2ac73e3c6a6ae9313854b80.

cesarBLG:
> Ah! Mathematicians and Python use the floored version of modulus, which does not return a negative value, while C# uses the truncated form of modulus.

I would have expected all languages to behave like Python, which seems a better generalization. However, CPU division instruction was implemented with the truncated version.

It should work fine with your last commit.

YoRyan:
It's been long enough to merge this now. Thanks for your patience.

YoRyan (approved):

PR #373, opened 2021-03-07
https://trello.com/c/Ek3wMQRs/484-enable-rotation-of-gauges-in-2d-cabs Enable inclined gauges in 2D cabs

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34517-enable-inclined-gauges-in-2d-cabs/


Csantucci:
OR Manual will be updated when code approved.

Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -446,6 +446,13 @@ protected virtual float ParseSwitchVal(STFReader stf)
             stf.SkipRestOfBlock();
             return switchVal;
         }
+        protected virtual float ParseRotation(STFReader stf)
+        {
+            stf.MustMatch("(");
+            var rotation = -MathHelper.ToRadians((float)stf.ReadDouble(0));
```
YoRyan, line 452: The angle is inverted here, which means the rotation is clockwise instead of counterclockwise. Was this intentional?

Csantucci: Hi Ryan,
this is coherent with what is already included in OR for the rotation of digital controls. And the rotation is counterclockwise for positive angles, at least for vertical bars, as you can see on the white gauge at the left of the attached picture http://www.interazioni-educative.it/Varie/GaugeRotation.jpg , where ORTSAngle has been set to 12. So, yes, the minus sign is intentional

YoRyan: Ok, that makes sense. Thanks for clarifying!

YoRyan:
No objections to the code here, Carlo.

Csantucci:
Ryan, thank you. OR manual updated.

YoRyan (approved):
Merge pending final approval by @cjakeman, since this is a new feature.

cjakeman:
Hi Carlo and Ryan,

Yes, I'm happy too, so I've merged the PR.

Thanks,

Chris

PR #374, opened 2021-03-13
Bug fix for ttps://bugs.launchpad.net/or/+bug/1918921 Crash after restore in timetable mode

Csantucci:
An additional parameter related to AI trains horn blow has been added in the AI train save method. In this case, the same parameter must be added for in the Save method for TTTrains, because AI trains and TTTrains share the restore method, but don't share the save method. Apart correcting the bug, now a warning message has been added in the AITrain save method.


YoRyan (approved):
Looks good. Thanks for looking into this, Carlo.

PR #375, opened 2021-03-14
Fix ReadTheDocs by adding an optional configuration option back

twpol:
This was easier than getting ReadTheDocs to use a newer version of Sphinx, which seems to involve adding a RTD cofig file and a complete Python `requirements.txt`


cjakeman (approved):
Hi James,

I found the Unstable version on ReadTheDocs.org and that worked just fine.

Approved with thanks,

Chris

PR #376, opened 2021-03-16
Localize HUD brake information

cesarBLG:
Localize the "BP 4 bar  BC 3 bar EOT BP 3.5 bar" message shown in HUD.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -124,29 +124,29 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
         public override string GetStatus(Dictionary<BrakeSystemComponent, PressureUnit> units)
         {
             string s = string.Format(
-                " BC {0}",
+                Simulator.Catalog.GetString(" BC {0}"),
                 FormatStrings.FormatPressure(CylPressurePSI, PressureUnit.PSI, units[BrakeSystemComponent.BrakeCylinder], true));
-                s += string.Format(" BP {0}", FormatStrings.FormatPressure(BrakeLine1PressurePSI, PressureUnit.PSI, units[BrakeSystemComponent.BrakePipe], true));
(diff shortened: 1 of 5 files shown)


YoRyan:
Good idea, but let's remove the format codes from the localized strings. Translators who aren't programmers are likely going to be confused by codes like `{0:F0}`.

Also, let's prefer the newer `$"{someVariable}"` syntax over `string.Format()`.

cesarBLG:
> Good idea, but let's remove the format codes from the localized strings. Translators who aren't programmers are likely going to be confused by codes like `{0:F0}`.
> 
> Also, let's prefer the newer `$"{someVariable}"` syntax over `string.Format()`.

Much cleaner, thanks. I'm waiting for @Weter-ORTS feedback to see if all braking messages are localized now.

Weter-ORTS:
Hello, Cesar and Ryan. As well as James.
I've downloaded Source of 26.03.21, updated templates, but there were no new strings...
Maybe I missed something...

cesarBLG:
> Hello, César and Ryan. As well as James.
> I've downloaded Source of 26.03.21, updated templates, but there were no new strings...
> Maybe I missed something...
> PR#376 is not mentioned at OR.org as merged, while here, it said, that was done to X1.3.1-331 TV

Hi Weter,
The new strings are only available in the unstable version of the source code. It has not been merged yet just in case there are more strings to include. If you have trouble downloading the unstable source, we can merge this and start another PR for further changes.

YoRyan (approved):
Approve and merged.

Weter-ORTS:
Thank you, Cesar!
The brake pressures names are accessible, but see ET forum about misplaced spaces.
There are some more things with HUD to improve.
WRT BC pressure, the first digit (head of the train) is not affected by pressure, referenced by the Engine brake lever (shows 0)
Looks like it's tied with the first car, but not with leading locomotive.

Another things are signaling information and wind characteristics-authors forgot to make them accessible for gettext as well.

cesarBLG:
> The brake pressures names are accessible, but see ET forum about misplaced spaces.

Could you please provide a link for the related thread? I am unable to find it.

> WRT BC pressure, the first digit (head of the train) is not affected by pressure, referenced by the Engine brake lever (shows 0)

Will check it, it's probably more useful that the pressure in the lead locomotive is displayed.

PR #377, opened 2021-03-19
API CabControls

cjakeman:
I've added another API, for Cab Controls.
This is primarily to support players with real instruments, dials, gauges etc to make the simulator values available.
There is also a webpage so you can see the values in a table.

Feature request: https://www.elvastower.com/forums/index.php?/topic/33642-full-sized-cab-controls-at-ferroviaire-show/page__view__findpost__p__269777


cjakeman:
I've moved this into Draft until I've had a chance to update the Manual.

YoRyan:
I would prefer if we kept any web server API glue in the WebServices/ directory, so I've pushed some commits to that effect. Otherwise, this looks like a good first step towards driving over HTTP.

cjakeman:
> I've moved this into Draft until I've had a chance to update the Manual.

Manual now updated.

cjakeman:
Hi Ryan,

> I would prefer if we kept any web server API glue in the WebServices/ directory, so I've pushed some commits to that effect. Otherwise, this looks like a good first step towards driving over HTTP.

Thanks for tidying this up.

"Driving" in the sense that players can more easily hook up real hardware gauges and dials to Open Rails. This is useful for museum exhibits etc..

Chris

YoRyan (approved):
Looks good to me, Chris. Merge when ready.

PR #378, opened 2021-03-20
Bug fix for https://bugs.launchpad.net/or/+bug/1919354 In isolated locos rear brakeman view is not available

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34936-no-6-rear-camera-for-player-locomotive/


YoRyan (approved):
Looks good, Carlo.

PR #379, opened 2021-03-20
Bug fix for https://bugs.launchpad.net/or/+bug/1919337 In activities after midnight boarding times are not respected

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/27958-midnight-traffic/page__p__269676#entry269676


YoRyan (approved):
Thanks for the fix, Carlo.

PR #380, opened 2021-03-22
Fix operation.rst build warning

YoRyan:
Resolves a build warning caused by #340.


cjakeman (approved):
Hi Ryan,

Something that  can be approved immediately.

Thanks.

Chris

PR #381, opened 2021-03-22
Add Visual Studio Code files to gitignore

YoRyan:
Adds .vscode and friends to gitignore so we won't commit them by accident.


cjakeman (approved):
Hi Ryan,

Great idea.

Approved with thanks,

Chris

PR #382, opened 2021-03-22
Replace for loops with explicit array copies and comparisons

YoRyan:
Replaces `for (i = 0; i < length; i++)` with `Array.Copy()`, `Array.Clone(),` and `Enumerable.SequenceEqual()`.

Array.Copy() and and Array.Clone(), in addition to being easier to read, are also an order of magnitude [faster](https://stackoverflow.com/a/18639042) than looping by index.

### DIFF
--- a/Source/Orts.Formats.Msts/ShapeFile.cs
+++ b/Source/Orts.Formats.Msts/ShapeFile.cs
@@ -901,8 +901,7 @@ namespace Orts.Formats.Msts
         {
             flags = copy.flags;
             ishader = copy.ishader;
-            tex_idxs = new int[copy.tex_idxs.Length];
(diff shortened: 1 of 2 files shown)


cjakeman (commented):
Hi Ryan,

Looks clean and effective.

Shall we leave it a few days more?

Are there any other bottlenecks that might benefit from this approach?

Chris

YoRyan:
Sure, we can leave it in unstable for a week.

I ran a grep for lines with both the `++` and `[` tokens, so this should cover all of the array copy operations that were done with for loops. For further performance improvements, I think I'd need to spend some time with a profiler.

cjakeman (approved):
Hi Ryan,

Approved as agreed.

Thanks,

Chris

PR #383, opened 2021-03-24
Add a video options setting for MultiSampleCount anti-aliasing

YoRyan:
Now that Monogame has landed, we can introduce proper anti-aliasing support. This PR adds a 6-step slider that ranges from no anti-aliasing to 32x MSAA. It is based on the equivalent feature already present in NewYear MG, except this patch is agnostic to different methods of anti-aliasing - which would make it easier to introduce FXAA, SSAA, and other alternative methods in the future. I have also located the slider at the bottom of the video tab, next to the ambient brightness slider.

image

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -367,6 +367,17 @@ Ambient daylight brightness
 
 With this slider you can set the daylight brightness.
 
+Anti-aliasing
+-------------
+
+Controls the anti-aliasing method used by Open Rails. Anti-aliasing is a
+computer graphics technique that smooths any harsh edges, otherwise known as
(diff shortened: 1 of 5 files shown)


cjakeman (approved):
Hi Ryan,

Looks tidy and good to see useful enhancements coming across from OR_NewYear_MG.

I try to explain enhancements in advance either with a Blueprint or a forum post, so that others get a chance to comment before the work goes ahead. Sometimes I forget, but it's good practice.

Approved with thanks,

Chris

YoRyan:
Thanks, Chris!

Normally I'd make an ET thread to open the floor for comments and discussion, but anti-aliasing seemed small enough (and essential enough) that I thought it was okay to implement right away. It'll be a surprise for this week's testing release.

PR #388, opened 2021-04-01
Save settings immediately upon changing in RunActivity.exe

YoRyan:
Changes the code to save whitelisted settings immediately upon change. Also changes the code to respect the skip-user-settings command-line flag.

Bug report: https://github.com/openrails/openrails/pull/306#issuecomment-803592877

### DIFF
--- a/Source/ORTS.Common/SettingsBase.cs
+++ b/Source/ORTS.Common/SettingsBase.cs
@@ -46,6 +46,11 @@ namespace ORTS.Common
         protected readonly Dictionary<string, Source> Sources = new Dictionary<string, Source>();
 
         /// <summary>
+        /// True when the user settings store is in use and will be read from and written to.
+        /// </summary>
+        protected bool AllowUserSettings { get; private set; }
+
+        /// <summary>
         /// Constructor
         /// </summary>
         /// <param name="settings">The store for the settings</param>
@@ -76,9 +81,8 @@ namespace ORTS.Common
         /// <summary>
         /// Load all settings, possibly partly from the given options
         /// </summary>
-        /// <param name="allowUserSettings">Are user settings allowed?</param>
         /// <param name="optionsDictionary">???</param>
-		protected abstract void Load(bool allowUserSettings, Dictionary<string, string> optionsDictionary);
+		protected abstract void Load(Dictionary<string, string> optionsDictionary);
 
         /// <summary>
         /// Save all settings to the store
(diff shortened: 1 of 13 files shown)


Source/ORTS.Common/SettingsBase.cs
```diff
@@ -43,7 +43,15 @@ protected enum Source
         /// <summary>The store of the settings</summary>
         protected SettingsStore SettingStore { get; private set; }
         /// <summary>Translates name of a setting to its source</summary>
-        protected readonly Dictionary<string, Source> Sources = new Dictionary<string, Source>();
+        protected readonly IDictionary<string, Source> Sources = new Dictionary<string, Source>();
+        /// <summary>
+        /// Translates name of a setting to its value when first loaded.
+        /// </summary>
+        protected readonly IDictionary<string, object> InitialValues = new Dictionary<string, object>();
```
twpol, line 50: The save system (`SettingsBase.Save` IIRC) already handles the prospect of saving a setting with a value equal to its default (it removes it), so this `InitialValues` might be redundant?

Source/RunActivity/Program.cs
```diff
@@ -65,7 +65,7 @@ static void Main(string[] args)
             game.PushState(new GameStateRunActivity(args));
             game.Run();
 
-            settings.Save();
+            settings.SaveInGameChanges();
```
twpol, line 68: An objection to this whole approach is that saving on exit requires the program to exit cleanly, including the computer not crashing.

And targeting these specific settings is contrary to what the existing code does, which is explicitly saving when things are changed: https://github.com/openrails/openrails/blob/69600d7daa88e3340bebe894aaa0f226b81d3495/Source/RunActivity/Viewer3D/Popups/Window.cs#L105 (granted that also doesn't respect `/skip-user-settings` but that would be an easy fix too)

While this PR improves the behaviour of the newly added saving of settings, it is still disjoint from the pre-existing method and I would like to see that addressed somewhere.

YoRyan:
Thanks for your feedback, James. I've adopted the save-immediately-upon-change approach you suggested. To accomplish this, I created a new class; I do wish I had a niftier way, but the C# typing system makes it difficult to get fancy.

twpol (approved):
Thanks very much, this is great!

YoRyan:
Thanks! I'll give it a week in unstable.

PR #390, opened 2021-04-04
Adds the signal ID and the aspect (including text aspect) of each signal head to the signalling debug window

Sharpe49:
To make signal debugging easier.


YoRyan (approved):
Thanks. This looks like a simple addition, so I don't think it needs management approval.

PR #392, opened 2021-04-13
Blueprint https://blueprints.launchpad.net/or/+spec/trainset-elevator Trainset elevator

Csantucci:
See also http://www.elvastower.com/forums/index.php?/topic/35082-locomotive-and-wagon-elevator/

### DIFF
--- a/Source/Documentation/Manual/features-route.rst
+++ b/Source/Documentation/Manual/features-route.rst
@@ -313,6 +313,98 @@ The animation block for the above transfertable is as follows::
 3600 is not a mandatory value, however to have a reasonable transfer speed a number of 
 animation keys equal to 60 - 90 every meter should be selected. 
 
+
+Locomotive and wagon elevators
+------------------------------
+.. index::
+   single: Elevator
+
+The elevator is managed by ORTS as a  vertically moving transfertable. So files needed 
+are the same as used for a transfertable, with content modified where needed.
+
+Info to identify an elevator in a route is stored in file turntables.dat, as it is 
+for turntables and transfertables. The same file can store info for moving tables of 
+different types. Here a turntables.dat file that contains info for an elevator::
+
+  1
+  Transfertable(
+  WFile ( "w-005578+014976.w" )
+  UiD ( 75 )
+  XOffset ( 0 )
+  YOffset ( -0.18 )
+  ZOffset ( 13.405)
+  VerticalTransfer ( 1 )
+  TrackShapeIndex ( 37301 )
+  Animation ( "TRACKPIECE" )
+  Length ( 26.81 )
+  )
+
+What identifies this as an elevator is the presence of the VerticalTransfer parameter 
+with value 1. The other difference to a transfertable is the presence of the YOffset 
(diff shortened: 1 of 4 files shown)


Csantucci:
Manual will be updated after code approved.

YoRyan:
Code looks good, Carlo! You may proceed with updating the manual.

Csantucci:
Having CenterOffsetComponent becoming a property doesn't convince me. CenterOffsetComponent is fixed for a transfertable, which is either horizontal or vertical. Having it becoming a property means that every time that you call it you have to retest for VerticalTransfer. I created that variable with the aim to avoid to have to retest every time.

YoRyan:
> Having CenterOffsetComponent becoming a property doesn't convince me. CenterOffsetComponent is fixed for a transfertable, which is either horizontal or vertical. Having it becoming a property means that every time that you call it you have to retest for VerticalTransfer, which is exactly what I wanted to avoid.

What would you think about encoding this information into the type system with a separate class, a la 09af7ab? This solution is a little more verbose, but it consolidates the various horizontal-or-vertical checks.

Csantucci:
With your last force push transfertables don't work any more,, neither horizontal nor vertical. I don't see a significant advantage either in your solution: my original code has three tests for VerticalTransfer, and two of them are active only in the initialization phase
Ryan, while I appreciate your big review work, which has dramatically accelerated the way features become available in the testing releases, I'd prefer if you'd provide suggested changes rather than force pushes, except in simple cases. I think it is more coherent with a peer review approach, where one involved part creates something  and the other one accepts it, or requests changes to it. Else the reviewer becomes also the approver of his own changes.

YoRyan:
My apologies for not replying to you sooner, Carlo. I absolutely should have tested such a dramatic change. I have reverted the last set of commits and set up your testing route on my own machine (it took me a few days to realize it required XTracks).

> I'd prefer if you'd provide suggested changes rather than force pushes, except in simple cases.

Agreed. From now on, I will only force push small changes. For more substantial refactors, I will attach a patch. (I'm not going to use GitHub suggestions anymore - they are way too limiting and cumbersome to work with.)

> Having CenterOffsetComponent becoming a property doesn't convince me. CenterOffsetComponent is fixed for a transfertable, which is either horizontal or vertical. Having it becoming a property means that every time that you call it you have to retest for VerticalTransfer, which is exactly what I wanted to avoid.

The test is a simple boolean test, so the performance impact would be negligible. I suggest dropping the variable because its existence is just one more thing to keep track of when reading or modifying the code—two values to set in the constructor, two values to set in the restorer. As Linus Torvalds says, programmers can generally only keep track of up to 7 variables at a time. (That might seem like a moot point in a codebase with thousands of variables per class, but it's something to consider, IMO.)

Csantucci:
Ryan, thank you and sorry for the XTracks inconvenience.
I'm still not quite convinced about the CenterOffsetComponent issue, but I accept your position. So I don't see obstacles towards a positive review now.

YoRyan (approved):
Yes, I agree. Now waiting for @cjakeman's approval.

cjakeman:
Sorry I kept you waiting,

Chris

PR #397, opened 2021-04-18
Webserver API/TIME

cjakeman:
Feature request: https://www.elvastower.com/forums/index.php?/topic/34945-web-server/

This is the simplest possible API since it returns a single floating point value for the simulation time.
It's illustrated by some sample webpages showing digital and analogue clocks.


YoRyan (approved):
Looks good, Chris!

PR #398, opened 2021-04-19
Ignore VS Code preferences (again)

YoRyan:
Turns out GitHub's gitignore template is for projects that would like to unify VS Code preferences among all of their developers. We don't do that here, so just ignore the .vscode directory entirely.


cjakeman (approved):
Hi Ryan,

Thanks for catching this and refining it.

Approved

PR #391, opened 2021-04-05
Animated wipers for 2D cabs

Csantucci:
Feature request: https://www.elvastower.com/forums/index.php?/topic/34604-2d-cabs-wipers-animation/

PR #389, opened 2021-04-04
Russian Localization Stage III

Weter-ORTS:
The goals are:
- finding sence of unclear strings and translate them;
- ask some programmist(s) to modify GUI spaces for fit russian-translated text on some buttons and fields,
- as well as to make all in-game text strings processable by gettext;
- keep the localization up-to date with the new ORTS features implemented, as soon, as it possible for me;
- trying to translate, if not entry Manual, but at least its Time-table section.

PR #384, opened 2021-03-24
ETCS Driver Machine Interface

cesarBLG:
The display used by the European Train Control System has some areas that cannot be rendered with existing cab controls. This implements the planning area (similar to the Track Monitor) and the text message area.

Now that the master branch includes Monogame, it can be ported back from OR NewYear version.

Discussion: http://www.elvastower.com/forums/index.php?/topic/34577-planning-area-of-the-etcs-dmi/
Blueprint: https://blueprints.launchpad.net/or/+spec/etcs-dmi

PR #399, opened 2021-04-19
Power supply package

Sharpe49:
Including :
- Master key
- Battery switch
- Service retention
- Cabview controls shown only when battery power supply or cab power supply is switched on
- Electric train supply
- Passenger car power supply and HVAC
- Diesel power supply (with power consumption from electric train supply)
- Dual mode power supply (in order to implement dual mode locomotives later)
- Traction cut-off relay
- TCS additions linked to power supplies


YoRyan:
Woah, this is a massive PR. Does this represent all of the power supply additions in the NewYear fork? I think it's too large for a single person to review.

By comparison, the Monogame upgrade, #40, was abut one-fifth the size, and it was rejected by @twpol for being overly complex and containing too many unexplained changes.

Csantucci:
To have some test I locally tried to merge the PR into a branch derived from the actualized master branch, but I got many rejected hunks, all related to .pot files. Not important to have tests, but I wanted to share this to you.

Sharpe49:
> To have some test I locally tried to merge the PR into a branch derived from the actualized master branch, but I got many rejected hunks, all related to .po or .pot files. Not important to have tests, but I wanted to share this to you. Probably you forked from the master branch before the latest changes to the locals files by Weter.

The parent commit is https://github.com/openrails/openrails/commit/f497b6b6da69357020c92a9e214b379c5d98f792
which is the latest commit on the master branch (I always rebase my branch on the master branch before pushing).
I regenerated the POT files in my working directory, no modifications were created. So your branch is probably the one having some problems with the translation files (probably not up to date).

Sharpe49:
> Woah, this is a massive PR. Does this represent all of the power supply additions in the NewYear fork? I think it's too large for a single person to review.
> 
> By comparison, the Monogame upgrade, #40, was abut one-fifth the size, and it was rejected by @twpol for being overly complex and containing too many unexplained changes.

I know, but many of the added functionalities are interlinked. I makes no sense separating them. I could probably create 2, 3 or 4 separate commits, but it would take a long time to separate them logically and the commits would still be quite huge.

cesarBLG:
I've had a quick look at the code and, although introducing a considerable amount of new code, it's not complex. Also, not taking into account locales, documentation or moved code, the size of the commit is reduced more than half. There's a lot of initialization and save/restore code, which doesn't need a deep review.

Separating the different traction subsystems from the Electric/Diesel locomotive classes is necessary to correctly reproduce some locomotives (for example, dual locomotives, electric trains without pantographs or with gearboxes), and I believe is in line with [James' architecture draft](https://github.com/twpol/openrails/blob/feature/organise-github-files/Docs/Architecture.md).

YoRyan:
This PR *does* merge cleanly, but the code bot is excluding it from the unstable build, possibly because of a conflict with #389 (which itself is currently failing CI).

cesarBLG:
> This PR _does_ merge cleanly, but the code bot is excluding it from the unstable build, possibly because of a conflict with #389 (which itself is currently failing CI).

It might be something else, because the Code Bot doesn't have Weter in the list of members, so that PR has to be explicitly included in the unstable branch.

Csantucci:
1) In Train.cs, starting from line 167, there are some lines defining some variables cancelled, like e.g. 
immagine
Is this wanted, and where have been moved such variables?
2) Some user commands added use special keys, like Insert and Enter: Can't this have drawbacks?
3) What must added/modified to implement dual mode locomotives?
4) Is dual mode general enough to allow also for other dual modes (e.g. pantograph/battery powered)?

Sharpe49:
> > This PR _does_ merge cleanly, but the code bot is excluding it from the unstable build, possibly because of a conflict with #389 (which itself is currently failing CI).
> 
> It might be something else, because the Code Bot doesn't have Weter in the list of members, so that PR has to be explicitly included in the unstable branch.
> 
> The console output from James website shows that there are conflicts with CabViewFile.cs and TrainControlSystem.cs, so probably it's conflicting with #384 (and maybe #391).

I just sent a rebase from master. This should be fixed.
I also added a commit including some small modifications I made this week (two new HandleEvent functions used when other cars are receiving orders from the lead locomotive).
I will also push to my script repository some classes that shows the changes of behaviour of the power supply according to the French rules.

> 1. In Train.cs, starting from line 167, some lines defining some variables have been deleted, like e.g.
>    public float TrainSteamHeatLossWpT;             // Total Steam Heat loss of train
>    Is this wanted, and where have been such variables moved to?
> 2. Some user commands added use special keys, like Insert and Enter: Can't this have drawbacks?
> 3. What must still be added/modified to implement dual mode locomotives?
> 4. Is dual mode general enough to allow also for other dual modes (e.g. pantograph/battery powered)?

1. They are removed because they are unused. Most variables used in the heating calculations are located in TrainCar.
2. I don't think so.
3. Basically, creating a new locomotive type able to store parameters for both electric and diesel traction. I think, we should wait for the JSON format for cars before implementing this.
4. Currently, no. But this is indeed something to think about when we will create dual mode locomotives.

Csantucci:
Thanks for replies. This crash http://www.elvastower.com/forums/index.php?/topic/35128-error-with-or-u20210425-0921/page__pid__271465#entry271465 is caused by following code line modified by this PR
`                        if (!LocomotivePowerSupply.MainPowerSupplyOn)`
and related to an AI train and to a steam locomotive.

Sharpe49:
> Thanks for replies. This crash http://www.elvastower.com/forums/index.php?/topic/35128-error-with-or-u20210425-0921/page__pid__271465#entry271465 is caused by following code line modified by this PR
> ` if (!LocomotivePowerSupply.MainPowerSupplyOn)`
> and related to a steam locomotive of an AI train. LocomotivePowerSupply is null.

It should be fixed with latest commit.

Sharpe49:
Ok, since there are a lot of variables coming from the power supply in MSTSLocomotive, instead of null checking everything in the case of the steam locomotive, I will create a stub (a SteamPowerSupply class that returns always the same values => PowerOn).

peternewell:
I have been working on a bug with the resume of the advanced adhesion -

http://www.elvastower.com/forums/index.php?/topic/35096-issue-with-advanced-adhesion-after-saveload/page__view__findpost__p__272082

There is an outstanding issue with electric locomotives that appears to be related to the resumption of the Power circuit. When initially resumed Power is indicated as being **Off**, and hence no MotiveForce is produced. (See screen shot of electric locomotive in the post above mine)

This issue seems to be more related to the work that you are doing on power circuits, and I don't want to interfere with the code related to your changes, hence could you have a look at it and correct it?

If you need any more details then please ask.

Thanks

Source/ORTS.Settings/UserSettings.cs
```diff
@@ -251,6 +251,8 @@ public enum DirectXFeature
         public bool OverrideNonElectrifiedRoutes { get; set; }
         [Default(true)]
         public bool HotStart { get; set; }
+        [Default(false)]
+        public bool NoDieselEngineStart { get; set; }
```
YoRyan, line 255: Tying the power mode to a global setting like this seems too inflexible; surely this should be a property of the selected consist and/or the selected activity. For example, imagine an activity that starts your locomotive in diesel mode in a non-electrified portion of track, with the expectation that you will change over to electric mode upon reaching the electrified section. The activity designer would want to ensure that the player's locomotive starts up in the correct mode.

You also have to consider other kinds of locomotives that can't be captured with a simple diesel on/diesel off toggle, such as dual overhead/third-rail units.

In the interest of getting this PR through, I would support keeping this setting for now, but I would like to move it to the Experimental tab. In the long-term, the goal should be to make it easy for the player to select which mode to start up in in explorer mode, and for activity and timetable designers to be able to select which modes they want their trains to start up in, thus allowing us to remove it.

Source/Orts.Simulation/Common/Scripting/Common.cs
```diff
@@ -90,6 +91,11 @@ public Timer(AbstractScriptClass asc)
         {
             CurrentValue = asc.GameTime;
         }
+
+        public Timer(TrainCar car)
+        {
+            CurrentValue = () => (float)car.Simulator.GameTime;
+        }
```
YoRyan, line 98: When does any content script have access to the TrainCar class? It should be for internal use only.

Source/RunActivity/Viewer3D/Viewer.cs
```diff
@@ -547,14 +547,18 @@ public void SetCommandReceivers()
...
             if (PlayerLocomotive is MSTSElectricLocomotive)
             {
-                CircuitBreakerClosingOrderCommand.Receiver = (MSTSElectricLocomotive)PlayerLocomotive;
-                CircuitBreakerClosingOrderButtonCommand.Receiver = (MSTSElectricLocomotive)PlayerLocomotive;
-                CircuitBreakerOpeningOrderButtonCommand.Receiver = (MSTSElectricLocomotive)PlayerLocomotive;
-                CircuitBreakerClosingAuthorizationCommand.Receiver = (MSTSElectricLocomotive)PlayerLocomotive;
+                CircuitBreakerClosingOrderCommand.Receiver = (PlayerLocomotive as MSTSLocomotive).LocomotivePowerSupply;
+                CircuitBreakerClosingOrderButtonCommand.Receiver = (PlayerLocomotive as MSTSLocomotive).LocomotivePowerSupply;
+                CircuitBreakerOpeningOrderButtonCommand.Receiver = (PlayerLocomotive as MSTSLocomotive).LocomotivePowerSupply;
+                CircuitBreakerClosingAuthorizationCommand.Receiver = (PlayerLocomotive as MSTSLocomotive).LocomotivePowerSupply;
```
YoRyan, line 553: There's no benefit to the `as MSTSLocomotive` construction unless you also test for `null` when indexing into it. If the cast cannot be performed, both
```C#
(PlayerLocomotive as MSTSLocomotive).Xxx
```
and
```C#
(MSTSLocomotive)PlayerLocomotive.Xxx
```
will result in a runtime error.

Not going to change it, as this entire section needs to be refactored, really - just for future reference.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/ILocomotivePowerSupply.cs
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
+
+using ORTS.Scripting.Api;
+
+namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
+{
+    public interface ILocomotivePowerSupply : IPowerSupply
```
YoRyan, line 22: It would be helpful if you included a quick summary of what this interface is useful for.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/IPowerSupply.cs
```diff
@@ -0,0 +1,42 @@
...
+// GNU General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
+
+using ORTS.Scripting.Api;
+
+namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
+{
+    public interface IPowerSupply : ISubSystem<IPowerSupply>, IParsable
```
YoRyan, line 22: As above.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+
+    public class DefaultDieselPowerSupply : DieselPowerSupply
+    {
+        private Timer PowerOnTimer;
+        private Timer AuxPowerOnTimer;
+
+        private DieselEngineState PreviousFirstEngineState;
+        private DieselEngineState PreviousSecondEngineState;
```
YoRyan, line 167: Why should we be limited to two diesel engines? Perhaps a dictionary would be the way to go here.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/Pantograph.cs
```diff
@@ -342,20 +347,30 @@ public void HandleEvent(PowerSupplyEvent evt)
...
 
                             case 4:
                                 soundEvent = Event.Pantograph4Up;
+                                Confirm(CabControl.Pantograph4, CabSetting.On);
                                 break;
                         }
+
+
+                        if (!Simulator.TRK.Tr_RouteFile.Electrified)
+                            Simulator.Confirmer.Warning(Simulator.Catalog.GetString("No power line!"));
```
YoRyan, line 371: I'm probably being ambitious here, but: If we ever get our own route editor, this logic may need to be changed. We can expect OR to support electrified and non-electrified sections on the same route.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/Pantograph.cs
```diff
@@ -342,20 +347,30 @@ public void HandleEvent(PowerSupplyEvent evt)
...
+                        if (Simulator.Settings.OverrideNonElectrifiedRoutes)
+                            Simulator.Confirmer.Information(Simulator.Catalog.GetString("Power line condition overridden."));
```
YoRyan, line 373: I think a warning like this is redundant. Players should already know that they're forcing electrification on.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/PassengerCarPowerSupply.cs
```diff
@@ -0,0 +1,452 @@
...
+            firstUpdate = false;
+        }
+
+        public virtual void Update(float elapsedClockSeconds)
+        {
+            CarId = Train?.Cars.IndexOf(Wagon) ?? 0;
+
+            if (firstUpdate)
+            {
+                firstUpdate = false;
```
YoRyan, line 198: Testing for the first update like this is brittle and unlikely to survive a simulator save/restore. Could we possibly move this logic to Initialize()?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/ElectricPowerSupply.cs
```diff
@@ -15,65 +15,40 @@
...
-        }
-
-        public float LineVoltageV {
-            get
-            {
-                return (float)Simulator.TRK.Tr_RouteFile.MaxLineVoltage;
-            }
-        }
+        public bool RouteElectrified => Simulator.TRK.Tr_RouteFile.Electrified || Simulator.Settings.OverrideNonElectrifiedRoutes;
+        public float LineVoltageV => (float)Simulator.TRK.Tr_RouteFile.MaxLineVoltage;
```
YoRyan, line 38: Again, assuming route-wide electrification is not particularly future-proof.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/SteamPowerSupply.cs
```diff
@@ -0,0 +1,129 @@
...
+using System;
+using System.IO;
+
+namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
+{
+    /// <summary>
+    /// Basic power supply class for steam locomotives
+    /// For electrical systems powered by battery
+    /// </summary>
+    public class SteamPowerSupply : ILocomotivePowerSupply, ISubSystem<SteamPowerSupply>
```
YoRyan, line 12: So, no scriptable steam locomotive power supply? @peternewell would be disappointed. :)

YoRyan (commented):
Right, this PR has sat in the queue long enough. I've performed a review and left some comments. Thanks!

Source/ORTS.Settings/UserSettings.cs
Sharpe49: Ok to put it in the Experimental tab for the moment.
A parameter should be added to the timetables later.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/SteamPowerSupply.cs
Sharpe49: Do you really have a complex electric power supply system on steam locomotives?
I don't think so! :D

Source/Orts.Simulation/Common/Scripting/Common.cs
Sharpe49: It is used for BatterySwitch and MasterKey classes which are not scriptable.
In the script, you don't have access to TrainCar objects anyway.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/PassengerCarPowerSupply.cs
Sharpe49: We can't:
The Train member is not initialized when the Initialize is called.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/Pantograph.cs
Sharpe49: This code was moved from a MSTSElectricLocomotive function:
public override void SignalEvent(PowerSupplyEvent evt, int id) (function has been removed)

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/Pantograph.cs
Sharpe49: This code was moved from a MSTSElectricLocomotive function:
public override void SignalEvent(PowerSupplyEvent evt, int id) (function has been removed)

Source/RunActivity/Viewer3D/Viewer.cs
Sharpe49: I think that, in this case, this should be cleaner:
`if (PlayerLocomotive is MSTSElectricLocomotive electricLocomotive)` and use electricLocomotive in the code after.

About the runtime error, I don't think it can happen: we test that the locomotive is a MSTSElectricLocomotive and LocomotivePowerSupply is always initialized in the constructor of this class.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
Sharpe49: We are only using these variables for the EnginePowerOn/Off and SecondEnginePowerOn/Off events.

YoRyan (approved):
Approved after a few more changes. LTGM. @cjakeman

PR #407, opened 2021-05-06
added Web\CabControls\ files to RunActivity projects

cjakeman:
Follow up to https://github.com/openrails/openrails/pull/377/files
Doesn't stop anything working, but meant that files were not listed in Visual Studio's Solution Explorer.

The need for default values for CabControl.MinValue and MaxValue were exposed when ORTS_POWERKEY and ALERTER_DISPLAY were stuck at 0 in the API/CABCONTROLS


YoRyan (approved):
Good catch, Chris.

PR #408, opened 2021-05-07
Updating RailDriver SDK library to prepare for 64bit mode

perpetualKid:


YoRyan:
I am not having much luck with your new Raildriver package. ReadData() seems to fail to return any data, so OR never enters the body of [this loop](https://github.com/perpetualKid/openrails/blob/RailDriverUpdate/Source/RunActivity/Viewer3D/UserInputRailDriver.cs#L107).

That is, until I set a breakpoint inside of this function. Doing so allows OR to read _one_ single update after I trip the breakpoint and then resume execution. Seems like some kind of timing glitch. Any ideas? I'm stumped.

YoRyan (approved):
Thanks, that got it working! This has my seal of approval.

PR #409, opened 2021-05-08
Bug fix for https://bugs.launchpad.net/or/+bug/1927811 Dynamic weather not working

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/35110-no-weather-change/


YoRyan (approved):
I wouldn't classify this as a "bug fix" so much as an accommodation for TSRE.

I'll approve as long as this change doesn't affect handcrafted .act files. Can we leave this in unstable for another week? Hopefully that'll be enough time to catch any compatibility issue—if there is any for such a small change.

Csantucci:
OK, let's wait a week.

PR #412, opened 2021-05-11
New TCS function to get speed post information (including warning speed posts)

Sharpe49:
[Trello Card](https://trello.com/c/S38d2RoV/482-further-tcs-handles-related-to-signals-and-sound-events)
Initially part of PR 293 but now split into its own PR.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -110,6 +110,12 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<string, int, float, SignalFeatures> NextGenericSignalFeatures;
         /// <summary>
+        /// Features of next speed post
+        /// int: position of speed post in the speed post sequence along the train route, starting from train front; 0 for first speed post;
+        /// float: max testing distance
+        /// </summary>
+        public Func<int, float, SpeedPostFeatures> NextSpeedPostFeatures;
+        /// <summary>
         /// Next normal signal has a repeater head
         /// </summary>
         public Func<bool> DoesNextNormalSignalHaveRepeaterHead;
@@ -664,6 +670,24 @@ namespace ORTS.Scripting.Api
             SpeedLimitMpS = speedLimitMpS;
             AltitudeM = altitudeM;
             TextAspect = textAspect;
+        }
+    }
+
+    public struct SpeedPostFeatures
+    {
+        public readonly string SpeedPostTypeName;
+        public readonly bool IsWarning;
+        public readonly float DistanceM;
+        public readonly float SpeedLimitMpS;
+        public readonly float AltitudeM;
+
+        public SpeedPostFeatures(string speedPostTypeName, bool isWarning, float distanceM, float speedLimitMpS, float altitudeM)
+        {
+            SpeedPostTypeName = speedPostTypeName;
+            IsWarning = isWarning;
+            DistanceM = distanceM;
+            SpeedLimitMpS = speedLimitMpS;
+            AltitudeM = altitudeM;
         }
     }
 
--- a/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
@@ -68,6 +68,8 @@ namespace Orts.Simulation.Signalling
         private List<SignalWorldObject> SignalWorldList = new List<SignalWorldObject>();
         private Dictionary<uint, SignalRefObject> SignalRefList;
         private Dictionary<uint, SignalObject> SignalHeadList;
+        private List<SpeedPostWorldObject> SpeedPostWorldList = new List<SpeedPostWorldObject>();
+        private Dictionary<int, int> SpeedPostRefList = new Dictionary<int, int>();
         public static SIGSCRfile scrfile;
         public static CsSignalScripts CsSignalScripts;
         public int ORTSSignalTypeCount { get; private set; }
@@ -467,6 +469,7 @@ namespace Orts.Simulation.Signalling
 
             var Tokens = new List<TokenID>();
             Tokens.Add(TokenID.Signal);
+            Tokens.Add(TokenID.Speedpost);
             Tokens.Add(TokenID.Platform);
 
             // loop through files, use only extention .w, skip w+1000000+1000000.w file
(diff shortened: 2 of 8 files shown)


Csantucci:
This problem http://www.elvastower.com/forums/index.php?/topic/35165-problems-with-speed-restriction-since-u20210511-0251/ has arisen when this PR has become part of the unstable release.

Sharpe49:
This commit should improve the situation.

Sharpe49:
The problem has been fixed (http://www.elvastower.com/forums/index.php?/topic/35165-problems-with-speed-restriction-since-u20210511-0251/page__view__findpost__p__272364)

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
```diff
@@ -667,6 +673,24 @@ public SignalFeatures(string mainHeadSignalTypeName, Aspect aspect, float distan
         }
     }
 
+    public struct SpeedPostFeatures
+    {
+        public readonly string SpeedPostTypeName;
```
YoRyan, line 678: I suggest using `SpeedPostShapeName` as the property name here. It would be more descriptive.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
```diff
@@ -1231,22 +1245,39 @@ private void AddWorldInfo()
...
+                        int speedPostIndex;
+
+                        if (SpeedPostRefList.TryGetValue(head.TDBIndex, out speedPostIndex))
                         {
-                            uint signalIndex = thisRef.SignalWorldIndex;
-                            if (signal.WorldObject == null)
+                            if (signal.SpeedPostWorldObject == null)
                             {
-                                signal.WorldObject = SignalWorldList[(int)signalIndex];
+                                signal.SpeedPostWorldObject = SpeedPostWorldList[speedPostIndex];
```
YoRyan, line 1278: Indexing a list like this is a linear O(n) operation. A dictionary could accomplish the same thing in constant-time.

I realize you were copying the existing signal ref code, so I'm not going to press the issue. It might be something to revisit later.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
Sharpe49: If, one day, we add the possibility to add new signals or speed posts from packages (which seems to be something that the virtual filesystem concept can manage to do), I think we would have config files (in JSON for example) for each new signal or each new speed post.
In that context, I think the term "speed post type name" is more generic and can be reused later for speed posts with a JSON config file.

Source/Orts.Simulation/Simulation/Signalling/Signals.cs
Sharpe49: I am not really good at optimizing code. If you have any proposals, I will gladly take them into account.

Sharpe49:
Ok, I agree with your modifications.

Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
YoRyan: What gets me is that you have an equivalent SignalFeatures property, MainHeadSignalTypeName, that communicates the MSTS signal type: "NORMAL", "DISTANCE", "SPEED", etc. To a user who doesn't understand your intentions for SpeedPostTypeName, it looks like a completely different piece of data that, for some reason, shares the same name.

To me, the speed post **type name** would be something more like the **TrainObjectType.SpeedItemType** enum, which can be one of "Standard", "TempRestrictedStart", and "TempRestrictedResume". If you want access to the shape name, then I still think it would be better to use an unambiguous "SpeedPostShapeName" property. (Eventually, when we get new file formats and editor support (ugh), we can introduce custom identifiers.)

Sharpe49: > What gets me is that you have an equivalent SignalFeatures property, MainHeadSignalTypeName, that communicates the MSTS signal type: "NORMAL", "DISTANCE", "SPEED", etc. To a user who doesn't understand your intentions for SpeedPostTypeName, it looks like a completely different piece of data that, for some reason, shares the same name.

No, it contains the signal type name of the SIGCFG file, not the signal function type:
image

> To me, the speed post **type name** would be something more like the **TrainObjectType.SpeedItemType** enum, which can be one of "Standard", "TempRestrictedStart", and "TempRestrictedResume". If you want access to the shape name, then I still think it would be better to use an unambiguous "SpeedPostShapeName" property. (Eventually, when we get new file formats and editor support (ugh), we can introduce custom identifiers.)

That would be the function type of the speed post.

YoRyan: Huh! I totally misread that code.

Okay, then no objections from me. Thanks for clarifying this.

YoRyan (approved):
Approving! Do new TCS functions need ORMT approval before merging? @cjakeman

Sharpe49:
> Approving! Do new TCS functions need ORMT approval before merging? @cjakeman

Any news about merging this?

cjakeman:
Hi Ryan,

> Approving! Do new TCS functions need ORMT approval before merging? @cjakeman

That's not quite how it works (or should work). New features need a Trello card or a Blueprint and ORMT approves the new feature as set out in [the Contributing document](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md). 
Contributors should be linking their PRs back to a Trello card or a blueprint so it is easy for a reviewer, such as yourself, to check that the feature has ORMT approval.

Hi Cedric,

I don't see a link here, but I find a rather dated [Trello card](https://trello.com/c/FlBnKn2h/446-expand-potential-of-tcs-scripts) and a similar [Blueprint](https://blueprints.launchpad.net/or/+spec/expand-potential-of-tcs-scripts). 

I think it's better if we follow the procedures as far as possible, so please would you write a more focussed card or Blueprint and link the PR to it. Then I will be happy to approve it and merge the PR.

I did notice that no update to the Manual was included. Is that not needed for this function?

Best wishes,

Sharpe49:
Hi @cjakeman,

I agree, but one of the problem is that we don't see a lot of updates of the Trello cards and blueprints. So, we tend to think that those are a bit left abandonned. Many functionnalities have been integrated to the master branch and most of them are in a pending approval state on Launchpad.
If we respect as much as possible the procedures, the development will be slowed down a lot. For example, one of my blueprints of 2020 is still not approved while the code is already in the master branch. The ORMT must be more reactive to what's happening on the different platforms for the procedures to be used properly.
If we (developers and ORMT) can't respect those procedures, maybe they have to change (for example, delegating blueprints validation and cards update for some subsystems to people that work a lot on these subsystems).

About this pull request, this is more linked to the pull request #293 and in this message (https://github.com/openrails/openrails/pull/293#issuecomment-736365911), we moved the speed post information subject to a later commit.

About the manual, this function works similarly as the NextGenericSignalFeatures function. And this function was not described in the manual.

cjakeman:
Hi Cedric,

Yes, I can see that a backlog of Trello cards and Blueprints is not encouraging and slows down the work. We'll try to be more responsive and, now that Open Rails has gathered so much functionality, we might consider delegating sub-systems. Another option is to enlarge ORMT.

I've approved [your Blueprint from 2020](https://blueprints.launchpad.net/or/+spec/power-supply-enhancements).

Perhaps documenting the TCS functions deserves its own project.

I've added the link for this PR to #293 and to the Trello card and merged this PR.

Thanks and best wishes,

PR #413, opened 2021-05-11
Correct issue with wheels not rotating when simple control/physics is set

peternewell:
https://bugs.launchpad.net/or/+bug/1926988 - correct application of switch for simple physics/control.

http://www.elvastower.com/forums/index.php?/topic/35153-wheels-do-not-rotate/

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -4468,7 +4468,7 @@ namespace Orts.Simulation.Physics
                 if (car.CouplerSlackM < 0 || car.CouplerForceB >= 1)
                     continue;
 
-                if (car.IsPlayerTrain && Simulator.UseAdvancedAdhesion && car.IsAdvancedCoupler) // "Advanced coupler" - operates in three extension zones
+                if (car.IsPlayerTrain && !Simulator.Settings.SimpleControlPhysics && car.IsAdvancedCoupler) // "Advanced coupler" - operates in three extension zones
                 {
                     float MaxZ1TensionM = car.GetMaximumCouplerTensionSlack1M() * AdvancedCouplerDuplicationFactor;
                     // If coupler in Zone 1 tension, ie ( -ve CouplerForceU ) then set coupler forces to zero, as coupler faces not touching yet
@@ -4500,7 +4500,7 @@ namespace Orts.Simulation.Physics
                 if (car.CouplerSlackM > 0 || car.CouplerForceB >= 1) 
                     continue;
 
-                if (car.IsPlayerTrain && Simulator.UseAdvancedAdhesion && car.IsAdvancedCoupler) // "Advanced coupler" - operates in three extension zones
+                if (car.IsPlayerTrain && !Simulator.Settings.SimpleControlPhysics && car.IsAdvancedCoupler) // "Advanced coupler" - operates in three extension zones
                 {
                     float MaxZ1CompressionM = -car.GetMaximumCouplerCompressionSlack1M() * AdvancedCouplerDuplicationFactor;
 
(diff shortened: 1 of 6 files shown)


YoRyan:
Pardon me, but doesn't this solution seem a little bit backward? Instead of adding another check for `Simulator.Settings.SimpleControlPhysics`, we should just make sure that `Simulator.UseAdvancedAdhesion` is false when simple physics are enabled. That would allow the existing code to work without the additional checks.

peternewell:
> 
> Pardon me, but doesn't this solution seem a little bit backward? Instead of adding another check for `Simulator.Settings.SimpleControlPhysics`, we should just make sure that `Simulator.UseAdvancedAdhesion` is false when simple physics are enabled. That would allow the existing code to work without the additional checks.
>
`UseAdvancedAdesion`  parameter is specifically only to enable the Advanced Adhesion feature.

`SimpleControlPhysics` is a generic switch to put everything back into a "simple physics or control" mode.

Reviewing the code, using the AdvancedAdhesion switch for the checks in the Advanced coupler code was probably not the best approach.  It is better suited to have the "generic" switch as the control flag. I therefore think that this makes the code more "readable".

YoRyan (approved):
I see, Peter. I'm concerned that we now have two checks to remember to put in our code—one for the advanced adhesion flag, and another one for the simple physics flag. But as long as it's clear that both are independent settings that have their own distinct purposes, I suppose we can keep both.

peternewell:
> 
> I see, Peter. I'm concerned that we now have two checks to remember to put in our code—one for the advanced adhesion flag, and another one for the simple physics flag. But as long as it's clear that both are independent settings that have their own distinct purposes, I suppose we can keep both.
>

It should be a very rare occasion where the Advanced Adhesion flag is needing to be set/tested. In fact perhaps it could be removed all together so that the number of options are reduced.

My preference would be for the Simple Control/Physics flag to be the principal flag.

Thanks for the approval.

YoRyan:
> It should be a very rare occasion where the Advanced Adhesion flag is needing to be set/tested. In fact perhaps it could be removed all together so that the number of options are reduced.
> 
> My preference would be for the Simple Control/Physics flag to be the principal flag.

Yes, I would support folding advanced adhesion into the "simple physics" setting. That would simplify things a lot for new players, and I doubt many people would be interested in running with simple physics, but without advanced adhesion. Can we float this idea on Elvas Tower?

peternewell:
> 
> 
> Yes, I would support folding advanced adhesion into the "simple physics" setting. That would simplify things a lot for new players, and I doubt many people would be interested in running with simple physics, but without advanced adhesion. Can we float this idea on Elvas Tower?
>

I disagree that advanced adhesion should be folded into the "simple physics" setting. My original concept is described at the start of this post  -  http://www.elvastower.com/forums/index.php?/topic/33437-simple-vs-full-control-and-physics/

To me the simple physics control should dumb OR operation down across all physics so that the player does not have to worry about, wheel slip, train brakes, etc. It is more about getting the train to run without having to wonder why it can't be started or is not running properly. For example, in vacuum braked trains, if air braked cars are attached to the train, then the train brakes will not release, and hence the train cannot be started. If simple mode is selected then all air braked cars will be converted to vacuum braked cars and the train should start. This allows users who just want to run trains to run them without getting into all the issues that might impact a trains operation.  

Currently the simple physics setting will disable advanced adhesion (as it should since users have selected to operate with simple physics, in this case Simple Adhesion).

I agree that this switch would be a bonus for new users. At least they could see trains moving very quickly without have to fault find to work out why a train wouldn't start. Once they have see OR in operation they could delve into the detail later.

Personally as a developer, I don't agree with having thousands of different options for a user to select as it makes OR more and more complex to debug ( and operate) as each particular operational scenario needs to be clarified (ie what options have or have not been set). Or then needs to be operated to reproduce the bug. Thus it would make it a lot simpler to have only a simple or advanced physics option (switch). However I suspect that we have created a situation where users want the capability to turn individual features on and off.

I am happy for you to float the concept on Elvis Tower, but perhaps we should be talking about the broader context of simple/advanced physics/control operation rather then just about simple/advanced adhesion.

PR #416, opened 2021-05-14
De locales 1.4

jonas-or:


Csantucci:
Hi jonas,
thank you for this. I have only one point: you translate  "Blow horn for" with "Horn betätigen wegen". 
"Blow horn for" is followed by the number of seconds and then by the "seconds" word. So "wegen" does not seem correct to me. Maybe "für"?

jonas-or:
Hi Carlo,
yes, the "für" is correct then. Thank you for noticing that!

I'm working this way via Git Commits for the first time. How do I proceed now? Do I now make a further commit here with the change?

Csantucci:
> I'm working this way via Git Commits for the first time. How do I proceed now? Do I now make a further commit here with the change?
If you modify the .po file in your remote branch de-locales-1.4, the change should automatically be applied also to your PR.

jonas-or:
> If you modify the .po file in your remote branch de-locales-1.4, the change should automatically be applied also to your PR.

Yes, it worked, thanks! I've commited and push my remoted branch.

Made the change from "wegen" to "für" and "Drehscheibe verriegelt" (TT locked) for both wording "backward" and "forward".

Csantucci:
That's good, but I don't understand why you have translated "connected" with "verriegelt" (locked) instead of "verbunden". Is this used for turntables?

jonas-or:
A good question, I think so. I went from videos like this https://www.youtube.com/watch?v=-TagL5WwBG8 - right at minute 3:12 or 5:17.
There you can see the latch ("Riegel") that has to be extended before the loco is allowed to move.
Here is another video https://www.youtube.com/watch?v=IbYzoTuUJA4 .

It could also be the german "verschlossen" (locked), but I found that too misleading, because one might think that means the loco is not allowed to run on or off the turntable at all, so locked in the sense of blocked or not allowed to drive on.

The origin possibility to use the german "verbunden" (connected) is comparatively weaker than "verriegelt", because it somehow only means the opposite of the track ends, but not that the connection ("Verbindung") in the railroad sense is secured.
That's why I finally choose "verriegelt" - locked by a latch.

Csantucci (approved):
OK, I've approved the changes. You can merge them.

jonas-or:
Is my OR Git status right to merge it? If so, what do I have to do next here to merge it? What should a Button be named to merge and where to find it here?
Or do I have to do it via my local Git context menu in the win explorer by clicking on "Merge..."? There I could select the master branch and then klick "ok" to merge it?

Csantucci:
Probably you don't have the rights to merge. I'll do it for you.

jonas-or:
Thank you reviewing and merged it!

PR #417, opened 2021-05-14
Animated Clocks recognized by ClockHands without extern Files

jonas-or:

### DIFF
--- a/Source/RunActivity/Viewer3D/Scenery.cs
+++ b/Source/RunActivity/Viewer3D/Scenery.cs
@@ -51,6 +51,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text.RegularExpressions;
 
 namespace Orts.Viewer3D
 {
@@ -327,7 +328,6 @@ namespace Orts.Viewer3D
 
                 var shadowCaster = (worldObject.StaticFlags & (uint)StaticFlag.AnyShadow) != 0 || viewer.Settings.ShadowAllShapes;
                 var animated = (worldObject.StaticFlags & (uint)StaticFlag.Animate) != 0;
-                var isAnimatedClock = false; //Declare and preset shape is not an animated clock 
                 var global = (worldObject is TrackObj) || (worldObject is HazardObj) || (worldObject.StaticFlags & (uint)StaticFlag.Global) != 0;
 
                 // TransferObj have a FileName but it is not a shape, so we need to avoid sanity-checking it as if it was.
@@ -480,27 +480,10 @@ namespace Orts.Viewer3D
                     }
                     else if (worldObject.GetType() == typeof(StaticObj))
                     {
-                        isAnimatedClock = false;                                                              //Preset
                         //          preTestShape for lookup if it is an animated clock shape with subobjects named as clock hands 
                         StaticShape preTestShape = (new StaticShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
-                        if (preTestShape.SharedShape.Animations != null)                                      // shape has an Animation at all
-                        {
-                            if (preTestShape.SharedShape.Animations[0].anim_nodes.Count > 1)                  // shape has more than 1 anim node
-                            {
-                                //lookup in all anim nodes of the shape for subobjects named as clock hands
-                                foreach (var animNodes in preTestShape.SharedShape.Animations[0].anim_nodes)
-                                {
-                                    if (animNodes.Name.ToLowerInvariant().IndexOf("hand_clock") == 6)        //Shape anim node name contains "hand_clock"
-                                    {
-                                        if (animNodes.Name.ToLowerInvariant().IndexOf("orts_") == 0)         //Shape anim node name begins with "ORTS_"
-                                        {
-                                            isAnimatedClock = true;                                          //Shape is animated clock
-                                            break;                                                           //Exit because of find a clock hand subobject
-                                        }
-                                    }
-                                }
-                            }
-                        }
+                        var animNodes = preTestShape.SharedShape.Animations?[0]?.anim_nodes ?? new List<anim_node>();
+                        var isAnimatedClock = animNodes.Exists(node => Regex.IsMatch(node.Name, @"^orts_[hmsc]hand_clock", RegexOptions.IgnoreCase));
                         if (isAnimatedClock)
                         {
                             sceneryObjects.Add(new AnalogClockShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
--- a/Source/RunActivity/Viewer3D/Scenery.cs
+++ b/Source/RunActivity/Viewer3D/Scenery.cs
@@ -327,7 +327,7 @@ namespace Orts.Viewer3D
 
                 var shadowCaster = (worldObject.StaticFlags & (uint)StaticFlag.AnyShadow) != 0 || viewer.Settings.ShadowAllShapes;
                 var animated = (worldObject.StaticFlags & (uint)StaticFlag.Animate) != 0;
-                var isAnalogClock = GetClockType(worldObject.FileName) == ClockType.Analog;
+                var isAnimatedClock = false; //Declare and preset shape is not an animated clock 
                 var global = (worldObject is TrackObj) || (worldObject is HazardObj) || (worldObject.StaticFlags & (uint)StaticFlag.Global) != 0;
 
                 // TransferObj have a FileName but it is not a shape, so we need to avoid sanity-checking it as if it was.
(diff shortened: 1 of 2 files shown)


jonas-or:
Animated Clocks recognized by ClockHands without extern Files

Source/RunActivity/Viewer3D/Scenery.cs
```diff
@@ -480,7 +480,28 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
...
+                        //          preTestShape for lookup if it is an animated clock shape with subobjects named as clock hands 
+                        StaticShape preTestShape = (new StaticShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
+                        if (preTestShape.SharedShape.Animations != null)                                      // shape has an Animation at all
+                        {
+                            if (preTestShape.SharedShape.Animations[0].anim_nodes.Count > 1)                  // shape has more than 1 anim node
+                            {
+                                //lookup in all anim nodes of the shape for subobjects named as clock hands
+                                foreach (var animNodes in preTestShape.SharedShape.Animations[0].anim_nodes)
+                                {
+                                    if (animNodes.Name.ToLowerInvariant().IndexOf("hand_clock") == 6)        //Shape anim node name contains "hand_clock"
```
YoRyan, line 493: This checks that the string "hand_clock" is in the sixth position, not just that it exists. I think you meant to use .Contains("hand_clock") here.

YoRyan:
Hi Jonas,

Do you have some test content anywhere? I'd like to suggest the following patch. It uses Linq for brevity and also uses the .Contains() method I've suggested.

Also, adding this feature will require a manual update.

```diff
diff --git a/Source/RunActivity/Viewer3D/Scenery.cs b/Source/RunActivity/Viewer3D/Scenery.cs
index 144728444..1c5a74b5b 100644
--- a/Source/RunActivity/Viewer3D/Scenery.cs
+++ b/Source/RunActivity/Viewer3D/Scenery.cs
@@ -327,7 +327,6 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
 
                 var shadowCaster = (worldObject.StaticFlags & (uint)StaticFlag.AnyShadow) != 0 || viewer.Settings.ShadowAllShapes;
                 var animated = (worldObject.StaticFlags & (uint)StaticFlag.Animate) != 0;
-                var isAnimatedClock = false; //Declare and preset shape is not an animated clock 
                 var global = (worldObject is TrackObj) || (worldObject is HazardObj) || (worldObject.StaticFlags & (uint)StaticFlag.Global) != 0;
 
                 // TransferObj have a FileName but it is not a shape, so we need to avoid sanity-checking it as if it was.
@@ -480,27 +479,14 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
                     }
                     else if (worldObject.GetType() == typeof(StaticObj))
                     {
-                        isAnimatedClock = false;                                                              //Preset
                         //          preTestShape for lookup if it is an animated clock shape with subobjects named as clock hands 
                         StaticShape preTestShape = (new StaticShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
-                        if (preTestShape.SharedShape.Animations != null)                                      // shape has an Animation at all
+                        var animNodes = preTestShape.SharedShape.Animations?[0]?.anim_nodes ?? new List<anim_node>();
+                        var isAnimatedClock = animNodes.Exists(node =>
                         {
-                            if (preTestShape.SharedShape.Animations[0].anim_nodes.Count > 1)                  // shape has more than 1 anim node
-                            {
-                                //lookup in all anim nodes of the shape for subobjects named as clock hands
-                                foreach (var animNodes in preTestShape.SharedShape.Animations[0].anim_nodes)
-                                {
-                                    if (animNodes.Name.ToLowerInvariant().IndexOf("hand_clock") == 6)        //Shape anim node name contains "hand_clock"
-                                    {
-                                        if (animNodes.Name.ToLowerInvariant().IndexOf("orts_") == 0)         //Shape anim node name begins with "ORTS_"
-                                        {
-                                            isAnimatedClock = true;                                          //Shape is animated clock
-                                            break;                                                           //Exit because of find a clock hand subobject
-                                        }
-                                    }
-                                }
-                            }
-                        }
+                            var lowerName = node.Name.ToLowerInvariant();
+                            return lowerName.IndexOf("orts_") == 0 && lowerName.Contains("hand_clock");
+                        });
                         if (isAnimatedClock)
                         {
                             sceneryObjects.Add(new AnalogClockShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
```

jonas-or:
Hallo Ryan,

Thanks for your reviewing and working on that. 
Actually, I intentionally made the strict query for the 6th position here. To me, it's conceivable that a shape modeller might be using the subobject name "hand_clock" in future or even already previous shapes, without having an animated OR clock in mind - so he just happens to use "hand_clock" in one of his subobjects.
But I admit that it is hard to imagine that someone would randomly start a subobject name with "orts_" and then somehow use "hand_clock" in a non-6th position. But better safe than sorry ... I thought :-). You can decide it.
Otherwise I find your code suggestion very good and elegantly shortening!
I could not test your code patch yet.

Do you mean by "some test content" an OR-Clock as shape? I can send you one.

Greetings
Jonas

YoRyan:
> But I admit that it is hard to imagine that someone would randomly start a subobject name with "orts_" and then somehow use "hand_clock" in a non-6th position. But better safe than sorry ... I thought :-). You can decide it.

I see. I would lean toward keeping the check simple and predictable. I've updated the patch to use `.StartsWith("orts_shand_clock")`.

> Otherwise I find your code suggestion very good and elegantly shortening!
> I could not test your code patch yet.
> 
> Do you mean by "some test content" an OR-Clock as shape? I can send you one.

If can spare the time to test the patch, that would be great! Otherwise, if you can put together a test route, I can test it for you. Go with whichever way you feel would be easier.

jonas-or:
>... I've updated the patch to use `.StartsWith("orts_shand_clock")`.
> 

Oh, only one test with ".StartsWith("orts_shand_clock")" will not work. There are 4 clock hand types, not each of which must occur in a clock. "orts_shand_clock" is only the hand showing the seconds. There are:

"orts_**h**hand_clock" - **h**our hand.
"orts_**m**hand_clock" - **m**inute hand
"orts_**s**hand_clock" - **s**econd hand (ticking)
"orts_**c**hand_clock" - second hand (**c**ontinuously moving)

The hands differ only at the 5th index digit in the name of the subobject.
However, a clock can also contain only one hand, for example only the hour hand, as in sundials. In this case ".StartsWith("orts_shand_clock")" would not recognize the shape as a clock at all.

jonas-or:
Hi Ryan,

I have your last line:
var isAnimatedClock = animNodes.Exists(node => node.Name.ToLowerInvariant().StartsWith("orts_shand_clock"));
expanded to this line:
var isAnimatedClock = animNodes.Exists(node => { return node.Name.ToLowerInvariant().StartsWith("orts_") && node.Name.ToLowerInvariant().Contains("hand_clock"); });

(very similar to your original patch suggestion, but without the "lowerName" variable, which of course can go back in too for all I care to avoid the duplication of "node.Name.ToLowerInvariant()".)

I tested it and it works fine I would say. What do you think?

YoRyan:
Ok, I wasn't aware of the multiple hand types. Then let's go with
```c#
Regex.IsMatch(node.Name, @"^orts_[hmsc]hand_clock", RegexOptions.IgnoreCase)
```
...and I've pushed e8f7541e96c9da3e7ddeb3f340ae212bda6360e2 to your branch accordingly.

All this needs is a manual update (and approval from the management team), and this will be ready to go!

jonas-or:
> ... Then let's go with
> 
> ```cs
> Regex.IsMatch(node.Name, @"^orts_[hmsc]hand_clock", RegexOptions.IgnoreCase)
> ```
I have test it, it works fine. I had not heard of a Regex object before, seems perfect to me now, thanks!

> ...and I've pushed [e8f7541](https://github.com/openrails/openrails/commit/e8f7541e96c9da3e7ddeb3f340ae212bda6360e2) to your branch accordingly.
> 
> All this needs is a manual update (and approval from the management team), and this will be ready to go!

Am I asked to do something active concerning the manual update? (Sorry Ryan for my ignorance about the next Git step here)

YoRyan:
Yes, our policy is that any new features require:

1. A corresponding update to the Open Rails manual, which is written in Sphinx. One of us could do this for you, but it would be convenient if you could do so since, as the developer of the feature, you know how exactly it should work.
2. Approval from the OR management team.

When Chris Jakeman made the initial addition of animated clocks, he updated the manual in 755d56c73be59e1a92a44d7eda898d183d35b2bd.

jonas-or:
A simple misunderstanding on my part :-))). I understood "manual update" as something to do here on the Git website - so clicking a button or something - by hand so to say. I didn't come up with manual in the sense of user guide.
I will contact Chris about the text for the manual.
Thank you Ryan so far!

cjakeman:
Hi Jonas and Ryan,

> I will contact Chris about the text for the manual.

I will sort out the manual. Just a bit pre-occupied at the moment with preparations for v1.4. (Which will include your animated clocks).

Chris

cjakeman (approved):
Hi Jonas,

Thanks for this. I've updated the manual.

Approved,

Chris

PR #418, opened 2021-05-22
Fix index out of range bug in ETCS display

cesarBLG:
Using Enumerable.Range() throws a exception if the second argument is less than 0, which is the case if the list is empty and List.Count-1 is calculated.

Bug report: http://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__272321


YoRyan:
Yeah, that may have been some premature optimization on my part. My apologies, and thanks for investigating!

YoRyan (approved):

PR #415, opened 2021-05-12
It locales 1.4

strawberryfield:
IT locales update

PR #419, opened 2021-05-27
Revert changes from #415 that only bumped the POT timestamps

YoRyan:
See [discussion here](https://github.com/openrails/openrails/pull/415#issuecomment-840463085).


Sharpe49:
That's probably too late. I already merged the POT files modifications in my branches.
For the Russian language branch, it probably needs a rebase and some rework of the commits because there are unwanted file deletions in this branch.

YoRyan:
The 9 files I reverted have not been modified since merging #415. It might not be necessary anymore, but if we want to preserve the timestamps, we still have the opportunity to do so.

With regard to the Russian locale, #389 has been broken since April and likely needs a complete rework.

YoRyan:
Closing. I'm sorry for failing to block such changes and I'll make sure to do so in the future.

PR #420, opened 2021-05-27
Delete Russian timetable.rst

YoRyan:
It's a nice gesture on the part of @Weter-ORTS, but our infrastructure is not currently set up to actually build translated manual files.


Weter-ORTS:
Hello, Ryan.
Do I ungerstand right, that this part of work can't be supported/included nowadays, so you have to remove all the things, that causing conflicts?

YoRyan:
It's dead code, so it should be removed from the repository.

This deletion would not be permanent; you can always retrieve old files from Git history. We can restore it if/when we support translated manual texts.

Weter-ORTS:
Okay.
Thanks for explanation; and good, that it is just shifted aside, instead of deleting.
Then, i'll continue translation of manual offline - with text editor.

cjakeman (approved):
Agreed.

We nned to add support for non_English version of the manual and I have created a B;ueprint for this at https://blueprints.launchpad.net/or/+spec/non-english-manuals

PR #422, opened 2021-05-31
Change alignment of couplers as they travel through curve.

peternewell:
This is part of the work being done on the Advanced Coupler - https://blueprints.launchpad.net/or/+spec/advanced-coupler

This segment rotates and aligns the coupler as the car moves along the track.


cjakeman (approved):
Hi Peter,

The changes look fine.

Approved with thanks,

Chris

PR #424, opened 2021-06-02
Bug fix for https://bugs.launchpad.net/or/+bug/1930630. The web TrainDriving window is not available.

mbm-OR:
The web Train Driving window is not available, since the locomotiveStatus contains consecutive escape sequences, e.g. “\n\r”.


mbm-OR:
This issue occurs only in the Unstable version, since the **Power supply package** was added. The fix is simple.

YoRyan (approved):
I'd prefer our logic be clearer and more general than simply testing for string length. Otherwise, this looks good. Thanks for tracking this down.

PR #425, opened 2021-06-08
updates Paths and References section

cjakeman:
Update to document based on posts in this thread: http://www.elvastower.com/forums/index.php?/topic/16633-overall-file-system-and-data-structure/page__view__findpost__p__262013


YoRyan (approved):
This has my blessing, Chris!

PR #426, opened 2021-06-08
Update to German translation of Menu and Commons

markus-GE:
Should not interfere with Jonas' changes as we agreed on non-overlapping work batches


Sharpe49 (changes requested):
Please remove the POT files that have only a change of the dates.
These can provoke conflicts with other pull requests and can remove some pull requests from the unstable version.

markus-GE:
> Please remove the POT files that have only a change of the dates.

Sorry, didn't know that. Done now.

YoRyan:
You deleted the .pot files from the repository, which is also something we don't want. 😉 What Cedric was asking you to do was to copy the unchanged versions from the master branch, so that the timestamp bumps aren't included in the final diff. I have done this for you and pushed the commits to your branch.

@Sharpe49, does this look good to your eyes now?

markus-GE:
> What Cedric was asking you to do was to copy the unchanged versions from the master branch, so that the timestamp bumps aren't included in the final diff. 

I'm sorry, that's not what he initially wrote (or at least, how I understood it). Makes sense, so thanks for the clarification and for doing this for me.

I'll also have a chat with Chris Jakeman to include this information in his Git-guide for translators.

Sharpe49:
> You deleted the .pot files from the repository, which is also something we don't want. 😉 What Cedric was asking you to do was to copy the unchanged versions from the master branch, so that the timestamp bumps aren't included in the final diff. I have done this for you (`git checkout master -- <path/to/pot>`) and pushed the commits to your branch.
> 
> @Sharpe49, does this look good to your eyes now?

If you use TortoiseGit, use the Git diff command in the context menu to look at differences generated by the Update batch files.
If only 2 lines are changed, it means that only the dates have been changed. You can then use the Git revert command to restore the previous version for those files.

@YoRyan Seems good to me too.

Sharpe49 (approved):

markus-GE:
Thanks for the kind explanation, but I'll have yet to see how to translate that to GitHub desktop. I'll be sure to make it work one or the other way, however.

PR #427, opened 2021-06-10
Correct an Issue with Motive Force being incorrectly reduced by wagon braking force.

peternewell:
https://bugs.launchpad.net/or/+bug/1931520


peternewell:
A better way to handle this issue has been developed, and will be added through a new PR.

PR #428, opened 2021-06-13
Update to German translation

jonas-or:
final update to German translation for version 1.4


YoRyan:
I've reverted one of the .pot changes that only bumped the timestamp. Otherwise, this looks good. I assume we hold this until #426 is merged?

jonas-or:
I probably just changed the status ("Requires editing") in individual lines in the poedit in the de.po in "RunActivity". One is then prompted to save when closing, even though no word has changed.
Yes, of course we can wait until [#426](https://github.com/openrails/openrails/pull/426).
Thank you!

YoRyan (approved):

PR #429, opened 2021-06-15
Czech translations updated by Howky

cjakeman:


YoRyan (approved):
Looks good, Chris. Special thanks to Howky for the translation work.

PR #430, opened 2021-06-21
Train Driving UI window

mbm-OR:
Blueprint https://blueprints.launchpad.net/or/+spec/train-driving-ui-window


Source/RunActivity/Viewer3D/Viewer.cs
```diff
@@ -85,6 +85,8 @@ public class Viewer
         public HelpWindow HelpWindow { get; private set; } // F1 window
         public TrackMonitorWindow TrackMonitorWindow { get; private set; } // F4 window
         public HUDWindow HUDWindow { get; private set; } // F5 hud
+        public TrainDrivingWindow TrainDrivingWindow { get; private set; } // F5 train driving window
+        public MultiPlayerWindow MultiPlayerWindow { get; private set; } // MultiPlayer data windowed
```
YoRyan, line 89: Is there a reason the multiplayer window has to be separate, but tied to the new train driving window? Why not bake its information into the existing train list window?

Source/ORTS.Settings/InputSettings.cs
```diff
@@ -463,6 +463,7 @@ static void InitializeCommands(UserCommandInput[] Commands)
             Commands[(int)UserCommand.DisplayCompassWindow] = new UserCommandKeyInput(0x0B);
             Commands[(int)UserCommand.DisplayHelpWindow] = new UserCommandModifiableKeyInput(0x3B, Commands[(int)UserCommand.DisplayNextWindowTab]);
             Commands[(int)UserCommand.DisplayHUD] = new UserCommandModifiableKeyInput(0x3F, Commands[(int)UserCommand.DisplayNextWindowTab]);
+            Commands[(int)UserCommand.DisplayTrainDrivingWindow] = new UserCommandModifiableKeyInput(0x3F, KeyModifiers.Control);
```
YoRyan, line 466: Not keen on Ctrl+F5 as the shortcut. For our existing bindings, the general rule of thumb is that Ctrl represents some kind of mode toggle (Ctrl+F4 track monitor mode, Ctrl+F7 label display mode) while Alt opens extra windows (Alt+F9 train list window). How about Alt+F8 for the default key?

YoRyan:
I'm also having trouble rendering the bottom text. It's black on black and impossible to read:

image

Source/ORTS.Settings/InputSettings.cs
Csantucci: F5 is used for the main HUD, so it seems mnemonically useful to use Ctrl-F5 for the Train Driving Info, that displays more or less the same information.

mbm-OR:
> 
> It's black on black and impossible to read:
> 
It's a trick to prevent the window height from changing every time that an item is displayed. 
We could replace the black text by empty items, while keeping the required spaces, to let unmodified the window height.

Source/ORTS.Settings/InputSettings.cs
mbm-OR: The Train Driving Info is a small HUD, with just the information required for the driver.

Source/RunActivity/Viewer3D/Viewer.cs
mbm-OR: The new train driving info window inherits the same principle as the HUD window. Like the original, the multiplayer information was displayed within the HUD window. 
Now the proposal is to show the multiplayer information in a dedicated window.
Maybe we could create a web-multiplayer window, if possible.

Csantucci:
> 
> 
> > It's black on black and impossible to read:
> 
> It's a trick to prevent the window height from changing every time that an item is displayed.
> We could replace the black text by empty items, while keeping the required spaces, to let unmodified the window height.
This is used only for steam locomotives. So for electrics and diesels these lines will always remain black; IMHO for such locomotives it would be better if these lines weren't displayed. As long as the player doesn't switch player locomotvie, the window height would remain the same and would be smaller, with no unused lines.

mbm-OR:
> > This is used only for steam locomotives. So for electrics and diesels these lines will always remain black; IMHO for such locomotives it would be better if these lines weren't displayed.
Now a new code display items based on the engine type.

Source/RunActivity/Viewer3D/Viewer.cs
YoRyan: Well, the HUD is a debugging tool not meant for use by ordinary players, so it makes sense to communicate the state of the multiplayer subsystem there.

But dedicating an entire window for the multiplayer state? One that can't be closed, shows just two lines, and contains information that seemingly would fit very well within the existing Train List (Alt+F9) window?

image

I'm just having trouble seeing the justification here.

Source/ORTS.Settings/InputSettings.cs
YoRyan: Okay, so I get the F5 mnemonic now. But it's unfortunate that this would be the only window that toggles with the Ctrl mod key. So far, for window popups, we've always used Alt:

image

On a second look, I would actually suggest assigning the new window to Alt+F5, while moving "Basic HUD Toggle" to Shift+F5.

I should say that, in my thought process, I'm thinking about brand-new players who are downloading OR for the very first time. "Because it's the way we've always done things" isn't a convincing argument to me because, if you already have a NewYear or mainline config with Ctrl+F5 already assigned, OR will preserve, not overwrite, that assignment. Meanwhile, any new players will be left wondering about this seeming inconsistency.

Definitely a controversial decision; maybe we need the management team's opinion here? @cjakeman

Source/RunActivity/Viewer3D/Popups/TrainDrivingWindow.cs
```diff
@@ -0,0 +1,1214 @@
...
+        int WindowHeightMax = 0;
+        int WindowWidthMin = 0;
+        int WindowWidthMax = 0;
+
+        Label indicator;
+        LabelMono indicatorMono;
+        Label ExpandWindow;
+        Label LabelFontToBold;
+
+        public bool normalTextMode = true;// Standard text
```
YoRyan, line 194: We now have the ability to save options like this to the settings store so they will be preserved across games, by the way. [See the Track Monitor code.](https://github.com/openrails/openrails/blob/5b55e3fb5fd00412323cc61b73f014084615f277/Source/RunActivity/Viewer3D/Popups/TrackMonitorWindow.cs#L193)

YoRyan:
We should also keep in mind that @twpol has previously raised [some objections](http://www.elvastower.com/forums/index.php?/topic/33401-f5-hud-scrolling/page__view__findpost__p__252005) about the abbreviated text mode included in this window. He felt that non-standard abbreviations could be confusing and that the use of fixed-width font is inappropriate.

I have to say, this feature give me some pause, too. The four-letter abbreviations could be difficult to translate into other languages, and they are not even standard railway terminology.

Source/RunActivity/Viewer3D/Viewer.cs
Csantucci: > 
> 
> Well, the HUD is a debugging tool not meant for use by ordinary players, so it makes sense to communicate the state of the multiplayer subsystem there.
> 
> But dedicating an entire window for the multiplayer state? One that can't be closed, shows just two lines, and contains information that seemingly would fit very well within the existing Train List (Alt+F9) window?
> 
> image
> 
> I'm just having trouble seeing the justification here.

The Multiplayer info is used also by ordinary players, both by the dispatcher and by the clients (I have experienced that in a good amount of multiplayer sessions), so IMHO it makes sense to have that info together the Train Driving Info window. The info might however be included in the Train Driving Info window, instead of being in a separate, small window, although also the actual layout, with a separate add-on window, provides the same functionality from the point of view of the user.

mbm-OR:
> I have to say, the existence of this feature gives me some pause, too. The four-letter abbreviations could be difficult to translate into other languages, and they are not even standard railway terminology.

The abbreviated text mode seems to be the preferred mode of expert users. 
By default, the extended text mode is displayed, and is not mandatory to toggle the abbreviated text mode, if not desired by the user.
Here is an example of the German translation, suggested by Jonas. 
Diesel-Locomotive-Extended-20210710 Diesel-Locomotive-Small-20210710

For a better user experience, it would be convenient to update some Locales files (.pot/.po), but this is actually not convenient until the PR "Power supply package" is approved. 
Would it be advisable to add the “draft” label to this PR, until the "Power supply package" PR is approved?

Source/ORTS.Settings/InputSettings.cs
mbm-OR: After many years using a familiar key sequence for the HUD and extended windows. I have the feeling that it won't be welcome by current users.

Source/RunActivity/Viewer3D/Viewer.cs
mbm-OR: This is a multiplayer session, Dispatcher-player side:
Multiplayerinfo-Dispatcher
Similar info is replicated Client-player side.

YoRyan: This would be the first window that is divided into multiple, interlinked windows, so the impetus is on you to explain why this would be desirable for players.

YoRyan:
I would be very cautious in assuming that these 4-letter fixed-width abbreviations can be easily translated across all languages. @Weter-ORTS has had one heckuva time working on the Russian translation of OR; he is constantly requesting more space for the UI controls. He does this even *after* attempting to come up with Cryllic abbreviations. And then you have the logographic languages, like Chinese, Japanese, Korean, etc.

> For a better user experience, it would be convenient to update some Locales files (.pot/.po), but this is actually not convenient until the PR "Power supply package" is approved.
> Would it be advisable to add the “draft” label to this PR, until the "Power supply package" PR is approved?

This is no longer a problem as that PR has just been merged. However, I do think we need the management team's opinion on the abbreviations feature here. @twpol @cjakeman

Source/RunActivity/Viewer3D/Viewer.cs
mbm-OR: Initially, the MP information was located after the Train Driving Info (TDI), elements, within the same window.
But the information of each new player who joined the MP session increased the width and height of the TDI window.
Taking into account, that the purpose of the abbreviated mode is to display a window as narrow as possible, keeping the information about the MP session in the same window, makes infeasible to obtain a narrow TDI window.
This was the reason to implement a dedicated Multiplayer window.

mbm-OR:
> This is no longer a problem as that PR has just been merged. However, I do think we need the management team's opinion on the abbreviations feature here.

Meanwhile, for a better user experience, the German, French and Italian translations related with the Train Driving Info window, they are going to be published.
Thanks to Jonas for the German translation.

Source/ORTS.Settings/InputSettings.cs
cjakeman: > Definitely a controversial decision; maybe we need the management team's opinion here? @cjakeman

Sorry to hold you up. This mention had passed me by. Will confer and get back to you all shortly.

cjakeman:
> This is no longer a problem as that PR has just been merged. However, I do think we need the management team's opinion on the abbreviations feature here. @twpol @cjakeman

Sorry to hold you up. This mention had passed me by. Will confer and get back to you all shortly.

cjakeman (commented):
Hi Mauricio,

I don't see any changes to the Manual in this PR. Do you need help with making those changes?

Chris

cjakeman (changes requested):
Thanks for asking for ORMT's opinion and for looking at this so carefully.

James and I have been studying the PR and all the points raised.
We are keen that this new window should have the same stature as the other windows that the player needs to operate the train (F4, F8, F9 and F10). 

The Blueprint for this feature is not yet approved and we will be very happy to approve it subject to some conditions to make it fit better with the current version. These are:

1. The F5 key is re-assigned to open and close the Train Driving Info window.
2. The Multi-player Info window is made independent of the Train Driving Info window and should open and close on a different key combination.
3. In addition to the yellow dot action, a key combination is provided to switch between the full width and abbreviated versions of the window.
4. The full width version should be offered as the default.
5. The Manual should document with screenshots all the variations of the window (same as our existing windows such as F4), including a table for [field names, 4-character abbreviations, description]
6. There should not be any additions to Menu > Options (don't think any are intended). 

In addition to conditions 1-6, we can make the following recommendations:
For key combinations:
•	F5 toggles the Train Driving Info (keeping wide/narrow state)
•	Shift-F5 toggles the Train Driving Info window between full and abbreviated mode
•	Alt-F5 toggles the existing HUD (keeping page state)
•	Shift-Alt-F5 cycles through HUD pages
We can drop the Alt-F5 function that hides and reveals the extra page of HUD data.

Since Ctrl-9 is used for the multi-player Dispatcher Window, it might be appropriate to use:
•	Shift-9 toggles the Multiplayer Info
•	Shift-Alt-9 toggles the Multiplayer Info window between full and abbreviated mode

How about changing the yellow dot to a yellow triangle pointing left (in the full version) or right (in the abbreviated version)?

Hope that helps to give a clear direction. If anything is not clear, please ask.

mbm-OR:
> 
> 
> 
> In addition to conditions 1-6, we can make the following recommendations:
> For key combinations:
> 1• F5 toggles the Train Driving Info (keeping wide/narrow state)
> 2• Shift-F5 toggles the Train Driving Info window between full and abbreviated mode
> 3• Alt-F5 toggles the existing HUD (keeping page state)
> 4• Shift-Alt-F5 cycles through HUD pages
> 5-We can drop the Alt-F5 function that hides and reveals the extra page of HUD data.
> 
> Since Ctrl-9 is used for the multi-player Dispatcher Window, it might be appropriate to use:
> 6• Shift-9 toggles the Multiplayer Info
> 7• Shift-Alt-9 toggles the Multiplayer Info window between full and abbreviated mode
> 
> 8-How about changing the yellow dot to a yellow triangle pointing left (in the full version) or right (in the abbreviated version)?
> 

Steps 1 to 5 and step 8 were implemented.
Steps 6 and 7, still a work in progress. Proposals are welcome. Meanwhile, the Multiplayer Info window it visible when the "Train Driving Info" window is also visible.

About, the  “_Window Glass can be disabled to help user with ageing eyesight_” feature, any suggestions?

mbm-OR:
> 
> 
> Hi Mauricio,
> 
> I don't see any changes to the Manual in this PR. Do you need help with making those changes?
> 
> Chris

A draft with the changes to the Manual was generated. The screenshots for each different locomotive type was added. But it looks better in the HTML format than the PDF document. Any help is welcome.
The Multiplayer Info window is not included in this draft, because right now, it was not implemented as suggested.

mbm-OR:
> 
>  But it looks better in the HTML format than the PDF document. Any help is welcome.
> 

The PDF document has been improved.

mbm-OR:
It has been detected that the feature to toggle the WindowGlass (transparency) at runtime, works well only, when the ToggleFullScreen() function (Alt + Return) is not called, otherwise, after switching from windowed to full-screen mode, sometimes a transparent window, like a layered window, is visible. That is why this feature is removed.

cjakeman (commented):
Mauricio, 

In the OR_NewYear_MG version, clicking on the yellow dot switches the window between wide and narrow format.

I was hoping that clicking on the yellow triangles would do the same (and duplicate the Shift+F5 keystroke), but I find that nothing at all happens which is a shame.

Chris

Source/Documentation/Manual/driving.rst
```diff
@@ -418,6 +415,105 @@ the locations of signals, mileposts, permanent speed limits, sidings, and
...
+   :align: center
+
+The Diesel locomotive:
+
+.. image:: images/tdi-diesel-loco.png
+   :align: center
+   :scale: 80%
+
+To help the user with ageing eyesight, the Time value is clickable, to toggle between the Regular
+and Bold font-style. Once toggled, this window and external Messages-window changes all them
```
cjakeman, line 460:
```suggestion
Once toggled, the Bold font style is used in this window and also the Multiplayer Info window if it’s open.
```

Source/Documentation/Manual/driving.rst
```diff
@@ -418,6 +415,105 @@ the locations of signals, mileposts, permanent speed limits, sidings, and
...
+to Bold font-style.
```
cjakeman, line 461:
```suggestion

```

cjakeman (commented):
Mauricio,

I've suggested changes to the Manual text in 3 places so that it reads as better English. Hope that's OK.

Chris

mbm-OR:
> 
> 
> I was hoping that clicking on the yellow triangles would do the same (and duplicate the Shift+F5 keystroke), but I find that nothing at all happens which is a shame.
> 
> Chris

Fixed. It was a misunderstanding by my part.

mbm-OR:
> 
> 
> I've suggested changes to the Manual text in 3 places so that it reads as better English. Hope that's OK.
> 
> Chris

Thank you for your help.
I understand, that I must apply all yours suggested changes to the current driving.rst file. Is it right?

Source/RunActivity/Viewer3D/Popups/TrainDrivingWindow.cs
mbm-OR: Please, could you extend this suggestion?
Does this replace the `protected internal override void Save(BinaryWriter outf)` and the `protected internal override void Restore(BinaryReader inf)` procedures?.

cjakeman:
Hi Mauricio,

> I understand, that I must apply all yours suggested changes to the current driving.rst file. Is it right?

Not "must" but "may" - these changes are not a condition for approval. I think the Manual would be better if you made these changes to driving.rst. If you wish, you can do this from right here within GitHub, by pressing the Commit Suggestion buttons.

P.S. I am very thankful that you have worked on this PR and quickly enough so that it will become part of the v1.4 Release (expect a code freeze very soon).

Chris

cjakeman (approved):
Hi Mauricio,

Thanks for those changes.

Approved.

Source/RunActivity/Viewer3D/Popups/MultiPlayerWindow.cs
```diff
@@ -0,0 +1,420 @@
...
+                    {
+                        FirstCol = $"{t}",
+                        LastCol = ""
+                    });
+                }
+                AddLabel(new ListLabel());
+            }
+            else if(Orts.MultiPlayer.MPManager.Simulator.Confirmer != null)
+            {
+                var status = $"{Viewer.Catalog.GetString("Status")}: {MPManager.Catalog.GetString("Connection to the server is lost, will play as single mode")}";
```
Sharpe49, line 370: During initialization in timetable mode, this line is executed and triggers an exception because MPManager.Catalog is null.

mbm-OR: Thank you very much for detecting it so quickly.
A possible solution would be the following:

protected internal override void Initialize()
{
base.Initialize();
// Reset window size
**if( !Owner.Viewer.Simulator.TimetableMode )**
UpdateWindowSize();
}

I will generate a bug report and a PR.

PR #432, opened 2021-06-26
ci: Automatically create merge-to-master PR on release branch push

twpol:
**For the upcoming stable release**

This is the best I could manage for automatically merging `release/*` pushes into `master`, as GitHub branch restrictions cannot be bypassed - we cannot allow the action to push directly to `master` without also allowing everyone else

For each push, this will create a new PR which someone with write access needs to approve and merge manually:

image

### DIFF
--- a/.github/workflows/release-pr.yml
+++ b/.github/workflows/release-pr.yml
@@ -1,0 +1,12 @@
+on:
+  push:
+    branches:
+      - "release/*"
+jobs:
+  release-pr:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Create PR
+        uses: funivan/github-autopr@0.2.0
+        env:
+          GITHUB_TOKEN: ${{ github.token }}


YoRyan:
Not every PR filed against a release branch should also be merged into the master branch. Some hotfixes, for example, may only make sense in the context of the release branch's code. How about introducing a label or similar mechanism to opt-in to auto-generating a master PR?

twpol:
> Not every PR filed against a release branch should also be merged into the master branch. Some hotfixes, for example, may only make sense in the context of the release branch's code. How about introducing a label or similar mechanism to opt-in to auto-generating a master PR?

You cannot pick and choose which branch PRs to be merged into the mainline without also using rebase, cherry-pick, or branching before the divergence. I do not wish to introduce such complexity into the process, as some people are struggling as it is.

If the code has changed and a fix applied to the release branch no longer applies to the mainline, the automatically generated PR is where we resolve that conflict, which could include a merge with no changes being made.

YoRyan:
Okay, that makes sense to me. Hopefully it won't be a big issue when we start filing release PR's.

cjakeman (approved):

PR #438, opened 2021-07-11
Russian localization files correction (2-nd attempt)

Weter-ORTS:
Correction of mistakes in menu-localization, making menu.exe erroneous while launched.


Weter-ORTS:
Correction of mistakes in menu-localization, making menu.exe erroneous while launched.

YoRyan:
I'm sorry Weter, but this PR is still only changing the .pot timestamps. There's no reason to think this has any bearing on your issues with Menu.exe.

Try [downloading](https://github.com/openrails/openrails/archive/refs/heads/master.zip) a clean copy of the testing branch and building it with Visual Studio. If this version builds and runs successfully, then the problem is not with our code and you need to investigate your personal copy of OR.

Weter-ORTS:
Hello, Ryan.
Clear unpack of T1.3.1-1890 behaves the same way, if russian language choosen at Options.
Replacing of ru folder of it, with reworked *.dll-s causes normal working.

It's not so, that only difference is *.pot timestamps.
At least there was whitespaces at menu/po removed (which, as i suspect, was the problem for Menu.exe to run)

Anyway, Ihave advanced my translation, adding new strings, needed for 1.3.1-1890 release, so it demands new PR

PR #439, opened 2021-07-13
Bug fix for https://bugs.launchpad.net/or/+bug/1935997 Missing null check

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/35300-constant-crashes-on-most-routes/


YoRyan (approved):
LGTM!

PR #441, opened 2021-07-14
Fix obstructed block for non-normal signal on a track circuit at the end of the track

Sharpe49:
https://bugs.launchpad.net/or/+bug/1936263


YoRyan (approved):
Looks good to me!

PR #442, opened 2021-07-20
Correct input discrepancy for tractive force tables

peternewell:
https://bugs.launchpad.net/or/+bug/1936890

Input UoM changed.


cjakeman (approved):
Hi Peter,

A necessary correction, I think. Thanks.

I see you have also taken the chance to correct the Temperature conversions. As a general rule, I think it wiser to have 2 PRs, so that each one gets an appropriate title. But I won't push for that in this case.

Aprroved with thanks.

PR #443, opened 2021-07-27
CVCDiscrete extended to report correct MaxValue

cjakeman:
Optional parameter added to CVCDiscrete constructor so it can distinguish Lever|TwoState|TriState|MultiState and can set MaxValue correctly for use in CabControls API.

Specifically, the FRONT_HLIGHT is a tristate control (Off|Dim|Full) and was reported incorrectly.
See PR https://github.com/openrails/openrails/pull/377

### DIFF
--- a/Source/Orts.Formats.Msts/CabViewFile.cs
+++ b/Source/Orts.Formats.Msts/CabViewFile.cs
@@ -855,7 +855,7 @@ namespace Orts.Formats.Msts
         private int numPositions;
         private bool canFill = true;
 
-        public CVCDiscrete(STFReader stf, string basepath, DiscreteStates discreteStates)
+        public CVCDiscrete(STFReader stf, string basepath, DiscreteStates discreteState)
         {
 //            try
             {
@@ -1160,7 +1160,7 @@ namespace Orts.Formats.Msts
                 if (ControlType == CABViewControlTypes.DIRECTION && Orientation == 0)
                     Direction = 1 - Direction;
 
-                switch (discreteStates)
+                switch (discreteState)
                 {
                     case DiscreteStates.TRI_STATE:
                         MaxValue = 2.0f; // So that LocomotiveViewerExtensions.GetWebControlValueList() returns right value to web server
@@ -1312,8 +1312,8 @@ namespace Orts.Formats.Msts
     #region other controls
     public class CVCSignal : CVCDiscrete
     {
-        public CVCSignal(STFReader inf, string basepath, DiscreteStates discreteStates)
-            : base(inf, basepath, discreteStates)
+        public CVCSignal(STFReader inf, string basepath, DiscreteStates discreteState)
+            : base(inf, basepath, discreteState)
         {
             FramesCount = 8;
             FramesX = 4;
--- a/Source/Orts.Formats.Msts/CabViewFile.cs
+++ b/Source/Orts.Formats.Msts/CabViewFile.cs
@@ -311,6 +311,14 @@ namespace Orts.Formats.Msts
         LBS
     }
 
+    public static class DiscreteStates
+    {
+        public const string Lever = "lever";
+        public const string TwoState = "twostate";
+        public const string TriState = "tristate";
+        public const string MultiState = "multistate";
+    }
+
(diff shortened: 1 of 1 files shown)


Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -1149,14 +1157,23 @@ public CVCDiscrete(STFReader stf, string basepath)
...
-//                else                       // Unexpected error, so provide a hint
-//                    throw new STFException(stf, "Problem with NumPositions/NumValues/NumFrames/ScaleRange");
-//            } // End of Need check the Values collection for validity
+            //            catch (Exception error)
+            //            {
+            //                if (error is STFException) // Parsing error, so pass it on
+            //                    throw;
+            //                else                       // Unexpected error, so provide a hint
+            //                    throw new STFException(stf, "Problem with NumPositions/NumValues/NumFrames/ScaleRange");
+            //            } // End of Need check the Values collection for validity
```
Sharpe49, line 1176: Don't keep commented code inside the repository.
Either remove it or put it inside preprocessor directives (#if DEBUG) if it is needed for debug purposes.

Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -311,6 +311,14 @@ public enum CABViewControlUnits
         LBS
     }
 
+    public static class DiscreteStates
+    {
+        public const string Lever = "lever";
+        public const string TwoState = "twostate";
+        public const string TriState = "tristate";
+        public const string MultiState = "multistate";
+    }
```
Sharpe49, line 320: I think an enumerated value would have been better here.
And we don't need to use this const string inside the first parameter TokenProcessor constructor below. It is less readable than before.

Sharpe49 (changes requested):

Source/Orts.Formats.Msts/CabViewFile.cs
cjakeman: Thanks, Cedric. I'll try it like that.

Source/Orts.Formats.Msts/CabViewFile.cs
cjakeman: Agreed, but my intention is to fix a bug, not to review this class. That really requires a new PR.
I've now undone the format change that highlighted this section.

Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -845,7 +855,7 @@ public class CVCDiscrete : CVCWithFrames
         private int numPositions;
         private bool canFill = true;
 
-        public CVCDiscrete(STFReader stf, string basepath)
+        public CVCDiscrete(STFReader stf, string basepath, DiscreteStates discreteStates)
```
Sharpe49, line 858: The name of the parameter should be singular.

Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -1293,8 +1312,8 @@ protected void ParseCustomParameters(STFReader stf)
     #region other controls
     public class CVCSignal : CVCDiscrete
     {
-        public CVCSignal(STFReader inf, string basepath)
-            : base(inf, basepath)
+        public CVCSignal(STFReader inf, string basepath, DiscreteStates discreteStates)
+            : base(inf, basepath, discreteStates)
```
Sharpe49, line 1316: The name of the parameter should be singular.

Sharpe49 (changes requested):
Seems good besides some parameter names that should be singular.

Source/Orts.Formats.Msts/CabViewFile.cs
cjakeman: Yes, of course it should.

Good catch.

Source/Orts.Formats.Msts/CabViewFile.cs
cjakeman: Yes, of course it should.

Good catch.

Sharpe49 (approved):

PR #445, opened 2021-08-01
Add a PreUpdate variable to scripts

Sharpe49:
Timers don't work really well during the pre-update phase because the time step is huge during this phase.
For example, in C# signal scripts, they are causing huge delays to AI trains because the signals are opening too late.
Having access to the PreUpdate variable allows for the scripts to disable timers during pre-update.

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -54,7 +54,6 @@ namespace Orts.Simulation.AIs
         public List<AITrain> AutoGenTrains = new List<AITrain>(); // auto-generated trains
         public double clockTime; // clock time : local time before activity start, common time from simulator after start
         private bool localTime;  // if true : clockTime is local time
-        public bool PreUpdate; // if true : running in pre-update phase
         public List<AITrain> TrainsToRemove = new List<AITrain>();
         public List<AITrain> TrainsToAdd = new List<AITrain>();
         public List<AITrain> TrainsToRemoveFromAI = new List<AITrain>();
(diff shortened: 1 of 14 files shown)


peternewell:
Is there a blueprint or Trello card for this PR?

Is there a discussion thread on ET to announce these changes?

Sharpe49:
This is a small extension to https://blueprints.launchpad.net/or/+spec/csharp-signal-scripts

Source/Orts.Simulation/Simulation/AIs/AI.cs
```diff
@@ -54,7 +54,18 @@ public class AI
...
         private bool localTime;  // if true : clockTime is local time
-        public bool PreUpdate; // if true : running in pre-update phase
+        public bool PreUpdate // if true : running in pre-update phase
+        {
+            get
+            {
+                return Simulator.PreUpdate;
+            }
+
+            set
```
cjakeman, line 64: Giving read access to Simulator.PreUpdate is easy to understand.
But why is write access also needed?

cjakeman:
Hi Cedric,

> This is a small extension mainly to https://blueprints.launchpad.net/or/+spec/csharp-signal-scripts and but also other scripting interfaces such as TCS.

Please consider airing this feature on the forum - as recommended in the Contributing.md. It would help to raise awareness of your work and flush out any issues.

Thanks,

Chris

Source/Orts.Simulation/Simulation/AIs/AI.cs
Sharpe49: The PreUpdate variable is set from the AI class. Perhaps this should be protected, but the variable in Simulator is completely public anyway.

Sharpe49:
Done!
http://www.elvastower.com/forums/index.php?/topic/34792-c-signal-scripts/page__view__findpost__p__275090

Source/Orts.Simulation/Simulation/AIs/AI.cs
twpol: Apparently, this is only set in one place, so how about we make that location set via `Simulator` and remove the setter here?

Sharpe49: It is set in 4 places: 3 in AI.cs and 1 in Simulator.cs.
Ok, I will change this since this is a low number of instances.

Sharpe49: Well... in the end, I removed it from AI and changed all of the calls to this variable.

cjakeman (approved):
Hi Cedric,

Thanks for making the changes and posting to the forum.

Approved

PR #446, opened 2021-08-03
Fix for Copy function not being called in ScriptedDieselPowerSupply and ScriptedElectricPowerSupply

Sharpe49:
I should have kept it simple...
https://bugs.launchpad.net/or/+bug/1938744


cjakeman (approved):
Hi Cedric,

Looks fine to me.

Approved with thanks.

PR #450, opened 2021-08-15
Bug fix for https://bugs.launchpad.net/or/+bug/1939995 DynamicBrakeAvailable not copied to further loco instances

Csantucci:


cjakeman (approved):
Hi Carlo,

Good catch.

Approved with thanks,

Chris

PR #451, opened 2021-08-16
Fixed the wrong states of battery switch, master key and ETS switch after restoring from a save

Sharpe49:
https://bugs.launchpad.net/or/+bug/1940094
http://www.elvastower.com/forums/index.php?/topic/35377-restoring-the-master-key-variables-does-not-work-correctly/


cjakeman (commented):
Hi Cedric,

Looks good.

I'm going to leave it a week before merging in case any issues arise.

Thanks,

Chris

cjakeman (approved):
Hi Cedric,

No more issues have surfaced, so approved with thanks.

Chris

PR #452, opened 2021-08-16
Fix : Passenger car power supply should not be available for locomotives

Sharpe49:
https://bugs.launchpad.net/or/+bug/1938611


cjakeman (approved):
Hi Cedric,

Thanks for fixing this.

Approved,

Chris

PR #453, opened 2021-08-17
Fix tractive force not equal to zero during emergency braking

Sharpe49:
(using the same code as MSTSLocomotive)
https://bugs.launchpad.net/or/+bug/1940163


cjakeman (approved):
Hi Cedric,

Thanks for fixing this. Approved.

Chris

PR #455, opened 2021-08-17
Fix mismatch between saved and restored values in PassengerCarPowerSupply and added missing values

Sharpe49:
https://bugs.launchpad.net/or/+bug/1938607


cjakeman (approved):
Hi Cedric,

Thanks for fixing this. Approved,

Chris

PR #457, opened 2021-08-21
Increase upper limit for World Object Density

cjakeman:
https://blueprints.launchpad.net/or/+spec/object-density-upper-limit

Increases the limit from 10 to 199 and the default to 99 to suit modern content builders.

image

### DIFF
--- a/Source/Menu/Options.Designer.cs
+++ b/Source/Menu/Options.Designer.cs
@@ -1195,7 +1195,7 @@
             // 
             this.numericWorldObjectDensity.Location = new System.Drawing.Point(6, 299);
             this.numericWorldObjectDensity.Maximum = new decimal(new int[] {
-            199,
+            99,
             0,
             0,
             0});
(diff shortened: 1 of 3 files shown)


Source/Menu/Options.Designer.cs
```diff
@@ -1194,11 +1194,6 @@ private void InitializeComponent()
             // numericWorldObjectDensity
             // 
             this.numericWorldObjectDensity.Location = new System.Drawing.Point(6, 299);
-            this.numericWorldObjectDensity.Maximum = new decimal(new int[] {
-            10,
-            0,
-            0,
-            0});
```
Sharpe49, line 1201: Shouldn't there be a maximum set to 100?

cjakeman: Good catch, Cedric.

Not sure what went wrong there, but I've used the time to re-consider.
I've now changed the default, the maximum and the Manual text.

Sharpe49 (approved):

Sharpe49 (approved):

PR #460, opened 2021-08-23
fix: Fix squashed logo on menu

twpol:
This is a very tiny fix for an issue which has bugged me for ages... the logo on the menu is squashed:

image

After this change it looks much better:

image


cjakeman (approved):
Looks so much better, James.

Approve with thanks,

Chris

PR #462, opened 2021-08-29
Revised manual text for some options

cjakeman:
As discussed at http://www.elvastower.com/forums/index.php?/topic/35301-menu-options/
See https://blueprints.launchpad.net/or/+spec/manual-revisions

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -227,16 +227,22 @@ Video Options
 Dynamic shadows
 ---------------
 
-With this option it is possible to enable or disable the display of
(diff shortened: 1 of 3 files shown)


Sharpe49:
OK for the manual update.
But why is there UI modifications in this commit?

cjakeman:
> OK for the manual update.
> But why is there UI modifications in this commit?

Excellent question, Cedric.

There aren't supposed to be any changes to the UI.
Let's see if I can undo that part.

cjakeman (commented):
Couldn't find a way to remove just the unwanted changes, so I've added a commit that restores them back to their previous state.

PR #463, opened 2021-08-30
Add SME brake function

peternewell:

Forum thread - http://www.elvastower.com/forums/index.php?/topic/35435-smee-brake-system/


cjakeman (approved):
Hi Peter,

All looks good and the Blueprint has the detail.

Approved with thanks.

Chris

PR #464, opened 2021-08-30
Bug fix for https://bugs.launchpad.net/or/+bug/1942061. WindResistance locales.

mbm-OR:
At the Force Information extended page (HUD). The TrainWindResistanceDependent string can't be extracted to *.pot files for translation.
https://bugs.launchpad.net/or/+bug/1942061

### DIFF
--- a/Source/RunActivity/Viewer3D/Popups/HUDWindow.cs
+++ b/Source/RunActivity/Viewer3D/Popups/HUDWindow.cs
@@ -1009,11 +1009,13 @@ namespace Orts.Viewer3D.Popups
 
                 if (train.TrainWindResistanceDependent) // Only show this information if wind resistance is selected
                 {
-                    TableAddLine(table, $"{Viewer.Catalog.GetString("Wind Speed:")} {Me.ToMi(pS.TopH(train.PhysicsWindSpeedMpS)):N2} mph   {Viewer.Catalog.GetString("Wind Direction:")} {train.PhysicsWindDirectionDeg:N2} Deg   {Viewer.Catalog.GetString("Train Direction:")} {train.PhysicsTrainLocoDirectionDeg:N2} Deg    {Viewer.Catalog.GetString("ResWind:")} {train.ResultantWindComponentDeg:N2} Deg   {Viewer.Catalog.GetString("ResSpeed:")} {Me.ToMi(pS.TopH(train.WindResultantSpeedMpS)):N2} mph");
+                    TableAddLine(table, $"{Viewer.Catalog.GetString("Wind Speed:")} {Viewer.Catalog.GetStringFmt("{0:N2} mph", Me.ToMi(pS.TopH(train.PhysicsWindSpeedMpS)))}   " +
+                                        $"{Viewer.Catalog.GetString("Wind Direction:")} {Viewer.Catalog.GetStringFmt("{0:N2} Deg", train.PhysicsWindDirectionDeg)}   " +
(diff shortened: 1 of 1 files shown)


Source/RunActivity/Viewer3D/Popups/HUDWindow.cs
```diff
@@ -1009,8 +1009,7 @@ void TextPageForceInfo(TableData table)
 
                 if (train.TrainWindResistanceDependent) // Only show this information if wind resistance is selected
                 {
-                    TableAddLine(table, "Wind Speed: {0:N2} mph   Wind Direction: {1:N2} Deg   Train Direction: {2:N2} Deg    ResWind: {3:N2} Deg   ResSpeed {4:N2} mph", Me.ToMi(pS.TopH(train.PhysicsWindSpeedMpS)), train.PhysicsWindDirectionDeg, train.PhysicsTrainLocoDirectionDeg, train.ResultantWindComponentDeg, Me.ToMi(pS.TopH(train.WindResultantSpeedMpS)));
-
+                    TableAddLine(table, $"{Viewer.Catalog.GetString("Wind Speed:")} {Me.ToMi(pS.TopH(train.PhysicsWindSpeedMpS)):N2} mph   {Viewer.Catalog.GetString("Wind Direction:")} {train.PhysicsWindDirectionDeg:N2} Deg   {Viewer.Catalog.GetString("Train Direction:")} {train.PhysicsTrainLocoDirectionDeg:N2} Deg    {Viewer.Catalog.GetString("ResWind:")} {train.ResultantWindComponentDeg:N2} Deg   {Viewer.Catalog.GetString("ResSpeed:")} {Me.ToMi(pS.TopH(train.WindResultantSpeedMpS)):N2} mph");
```
Sharpe49, line 1012: You can also put the old string with the parameters inside within a GetStringFmt function.
The new string seems really complicated and doesn't translate the units.

mbm-OR: Formatted the new string. Now it looks easier to read and added the GetStringFmt () function.

YoRyan: Maybe [FormatStrings.FormatSpeed()](https://github.com/openrails/openrails/blob/4a5757f158356a6cd9444b355d9e51d281e41fa3/Source/ORTS.Common/Conversions.cs#L580) is what we want here?

mbm-OR: This error is simpler and is related to the strings: "Wind speed:", "Wind direction:", "Train direction:", "ResWind:", "ResSpeed:" that they do not appear in the *. pot for translation, because the Viewer.Catalog.GetString () function was missing.
Here is the proposed code: https://github.com/openrails/openrails/pull/464/commits/1d09b79d180b02903267fddf3462e7d1cf80628c

Is it mandatory to translate the units "Deg" and "mph"?  I have not found information about it.

cjakeman: Hi Mauricio,

I can find no previous translations from "Deg" or "mph" either, so I guess we can leave that alone.

Chris

cjakeman (approved):
Hi Mauricio,

Fixes the bug and is much tidier than the original.

Approved with thanks,

Chris

Sharpe49:
> Hi Mauricio,
> 
> Fixes the bug and is much tidier than the original.
> 
> Approved with thanks,
> 
> Chris

I would have preferred either using a SI unit or being able to select the desired unit. @cjakeman

PR #465, opened 2021-08-30
feat: Add latest version information to registry

twpol:
Storing the latest version number in the registry has been requested by content creators, so that their installers can validate the user's version of Open Rails is compatible with the content

This will be stored per-user, next to the existing content folders storage, which installers already modify to add content

The version will be stored in a machine-readable format (so that it may be updated automatically) and human-readable text for display purposes

**Discussion:** http://www.elvastower.com/forums/index.php?/topic/35458-store-latest-version-number-in-registry/
**Trello card:** https://trello.com/c/6ybEXrey/508-store-latest-version-number-in-registry


cjakeman (approved):
Hi James,

All looks good.

Approved with thanks,

Chris

PR #471, opened 2021-09-08
Correct issue with control car wheels not rotating

peternewell:
https://blueprints.launchpad.net/or/+spec/unpowered-control-car


cjakeman (approved):
HI Peter,

Looks good.

Approved with thanks,

Chris

PR #473, opened 2021-09-10
Further enhancements to Control Car

peternewell:
Add some gauges that were not visible

https://blueprints.launchpad.net/or/+spec/unpowered-control-car


cjakeman (approved):
Hi Peter,

Looks fine. Approved with thanks,

Chris

PR #474, opened 2021-09-10
Drops XNA references from .appveyor.yml

cjakeman:
The Appveyor build is failing when building the Unstable version.
The Chocolatey installer is having trouble downloading and installing XNA.
As this is no longer needed, I have removed it from the script so it more closely matches the release/1.4 script.


twpol:
This change was already in #472 and the Unstable Version build failure was a different, but temporary, problem

PR #480, opened 2021-09-13
Blueprint https://blueprints.launchpad.net/or/+spec/digital-alignment-in-3dcabs

Csantucci:
Alignment of digitals in 3D cabs.
See also http://www.elvastower.com/forums/index.php?/topic/35498-justification-alignment-for-digitals-in-3dcabs/#entry275959

### DIFF
--- a/Source/RunActivity/Viewer3D/Popups/WindowControls.cs
+++ b/Source/RunActivity/Viewer3D/Popups/WindowControls.cs
@@ -115,6 +115,9 @@ namespace Orts.Viewer3D.Popups
         Left,
         Center,
         Right,
+        Cab3DLeft,
+        Cab3DCenter,
+        Cab3DRight,
     }
 
     public class Label : Control
--- a/Source/RunActivity/Viewer3D/Popups/WindowControls.cs
+++ b/Source/RunActivity/Viewer3D/Popups/WindowControls.cs
@@ -115,9 +115,6 @@ namespace Orts.Viewer3D.Popups
         Left,
         Center,
         Right,
-        Cab3DLeft,
-        Cab3DCenter,
-        Cab3DRight,
     }
 
     public class Label : Control
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2431,7 +2431,18 @@ namespace Orts.Viewer3D.RollingStock
     /// </summary>
     public class CabViewDigitalRenderer : CabViewControlRenderer
     {
-        readonly LabelAlignment Alignment;
+        public enum CVDigitalAlignment
+        {
+            Left,
+            Center,
+            Right,
+            // Next ones are used for 3D cabs; digitals of old 3D cab will continue to be displayed left aligned for compatibility
+            Cab3DLeft,
+            Cab3DCenter,
+            Cab3DRight
+        }
+
+        public readonly CVDigitalAlignment Alignment;
         string Format = "{0}";
         readonly string Format1 = "{0}";
         readonly string Format2 = "{0}";
@@ -2452,10 +2463,10 @@ namespace Orts.Viewer3D.RollingStock
 
             // Clock defaults to centered.
             if (Control.ControlType == CABViewControlTypes.CLOCK)
-                Alignment = LabelAlignment.Center;
-            Alignment = digital.Justification == 1 ? LabelAlignment.Center : digital.Justification == 2 ? LabelAlignment.Left : digital.Justification == 3 ? LabelAlignment.Right : Alignment;
+                Alignment = CVDigitalAlignment.Center;
+            Alignment = digital.Justification == 1 ? CVDigitalAlignment.Center : digital.Justification == 2 ? CVDigitalAlignment.Left : digital.Justification == 3 ? CVDigitalAlignment.Right : Alignment;
             // Used for 3D cabs
-            Alignment = digital.Justification == 4 ? LabelAlignment.Cab3DCenter : digital.Justification == 5 ? LabelAlignment.Cab3DLeft : digital.Justification == 6 ? LabelAlignment.Cab3DRight : Alignment;
+            Alignment = digital.Justification == 4 ? CVDigitalAlignment.Cab3DCenter : digital.Justification == 5 ? CVDigitalAlignment.Cab3DLeft : digital.Justification == 6 ? CVDigitalAlignment.Cab3DRight : Alignment;
             Format1 = "{0:0" + new String('0', digital.LeadingZeros) + (digital.Accuracy > 0 ? "." + new String('0', (int)digital.Accuracy) : "") + "}";
             Format2 = "{0:0" + new String('0', digital.LeadingZeros) + (digital.AccuracySwitch > 0 ? "." + new String('0', (int)(digital.Accuracy + 1)) : "") + "}";
         }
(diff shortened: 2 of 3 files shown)


Sharpe49:
This bug report will also be resolved by this change:
https://bugs.launchpad.net/or/+bug/1891210

Sharpe49 (changes requested):
I tested this PR. It works properly.
But I think the new getter method is superfluous.

Sharpe49:
Also, these new values for the Justification parameter can be documented in the manual, I think.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2679,6 +2680,10 @@ public string GetDigits(out Color DrawColor)
             return "";
         }
 
+        public LabelAlignment GetAlignment() //used in 3D cab, to get alignment
+        {
+            return Alignment;
+        }
```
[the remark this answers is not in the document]
Csantucci: I did so for analogy with the other .cvf variables used. But I agree that it is simpler as you propose and will change this once the blueprint is approved.

Csantucci:
> 
> 
> Also, these new values for the Justification parameter can be documented in the manual, I think.

Yes, that's correct. I'll do that once the blueprint gets approved.

Source/RunActivity/Viewer3D/Popups/WindowControls.cs
```diff
@@ -115,6 +115,9 @@ public enum LabelAlignment
         Left,
         Center,
         Right,
+        Cab3DLeft,
+        Cab3DCenter,
+        Cab3DRight,
```
twpol, line 120: This change is not okay; this `enum` is for the popup windows and adding values that are **not** used by the popup windows is confusing and wrong – `CabViewDigitalRenderer` should be using its own `enum` if you want additional/different `enum` values

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2454,7 +2454,8 @@ public CabViewDigitalRenderer(Viewer viewer, MSTSLocomotive car, CVCDigital digi
             if (Control.ControlType == CABViewControlTypes.CLOCK)
                 Alignment = LabelAlignment.Center;
             Alignment = digital.Justification == 1 ? LabelAlignment.Center : digital.Justification == 2 ? LabelAlignment.Left : digital.Justification == 3 ? LabelAlignment.Right : Alignment;
-
+            // Used for 3D cabs
+            Alignment = digital.Justification == 4 ? LabelAlignment.Cab3DCenter : digital.Justification == 5 ? LabelAlignment.Cab3DLeft : digital.Justification == 6 ? LabelAlignment.Cab3DRight : Alignment;
```
twpol, line 2458: I'm sure you have a reason for using 4, 5, 6 here instead of 1, 2, 3 so please can you put that reason into a comment here (I would rather use 1, 2, 3)

But, whatever that reason, I definitely do not see why you need those three additional values for `Alignment`. Why not read 4, 5, 6 and store as `LabelAlignment.Center`, `LabelAlignment.Left`, `LabelAlignment.Right`?

twpol (changes requested):
The feature request itself – alignment of digital displays in 3D cabs – is good

Source/RunActivity/Viewer3D/Popups/WindowControls.cs
Csantucci: I will put in a comment the explanation that I have written here http://www.elvastower.com/forums/index.php?/topic/35498-justification-alignment-for-digitals-in-3dcabs/page__view__findpost__p__275959 : as of now existing 3D cabs have left-aligned digitals independently from the value of Justification. If I now respect the value of Justification,, existing 3D cabs will change the alignment of their digitals, which is not desirable.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
Csantucci: I can put in a comment the explanation that I have written here http://www.elvastower.com/forums/index.php?/topic/35498-justification-alignment-for-digitals-in-3dcabs/page__view__findpost__p__275959 : as of now existing 3D cabs have left-aligned digitals independently from the value of Justification. If I now respect the value of Justification,, existing 3D cabs will change the alignment of their digitals, which is not desirable. 
Your proposal does not solve the above problem. If you prefer, I can use a new field in the .cvf control, e.g. ORTS3DAlignment, which can have values 1, 2 or 3, and check that one in the 3D cab code.

twpol: Are there many 3D cabs that will seriously break if you used the same 1, 2, 3 values?

We are adding a lot of confusion complexity with 4, 5, 6 here so we need to justify that it is absolutely necessary.

Source/RunActivity/Viewer3D/Popups/WindowControls.cs
twpol: This response does not match the issue here. You **cannot** add these three values to this particular `enum`. It is completely and utterly the wrong thing to do here.

If you absolutely must keep track of 6 different alignment values, please create a _new_ `enum` in the cab view code.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
Csantucci: James, I see that especially commercial content developers are offering 3D cabs. I suppose they aren't happy if we break compatibility.
As written in my alternate proposal at the end of my message above yours, inserting ORTS3DAligment in the .cvf file there is no more need for values 4, 5 and 6. ORTS3DAlignment would have values 1, 2 and 3, and within the code it would be used to assign a LabelAlignment to a new variable Alignment3D, which would be the one checked in the 3D cab code.

Csantucci:
> 
> 
> Hi Carlo,
> 
> I see that your code is still adding new enums to the file Source/RunActivity/Viewer3D/Popups/WindowControls.cs
> 
> This is surely the wrong file for them as it's a file for Pop-up Windows, not Cab Controls. Will you be moving them elsewhere?
> 
> Thanks.

Hi Chris,
there are two issues.
That enum is already being used for cabview controls, independently from my PR, and therefore I decided to add the two new entries there. However, if it is preferred that I generate a new enum, I haven't problems to that. 
There is the other issue: I understood from James' comments, that in general he doesn't like having these two new entries, independently from the location of the enum. So, in my comments above, I presented another solution. Before proceeding I need to know whether the original solution (with duplicated enum of course) or the new solution is preferred.

twpol:
> That enum is already being used for cabview controls, independently from my PR, and therefore I decided to add the two new entries there. However, if it is preferred that I generate a new enum, I haven't problems to that.

Please create a new enum for the cab view controls – reusing an enum is only okay when they share exactly the same semantics and values, otherwise there will be confusion (what does a popup label with alignment `Cab3DRight` mean?)

> There is the other issue: I understood from James' comments, that in general he doesn't like having these two new entries, independently from the location of the enum. So, in my comments above, I presented another solution. Before proceeding I need to know whether the original solution (with duplicated enum of course) or the new solution is preferred.

I dislike the values 4, 5, 6 about the same as I dislike `ORTS3DAligment` so you'll have to pick – either one is okay :)

Sharpe49:
@twpol Can you check if the changes are OK for you?

twpol (approved):

twpol:
> @twpol Can you check if the changes are OK for you?

The changes look much better now, thanks all

Csantucci:
Nice that we arrived at a shared conclusion. Thanks.

PR #482, opened 2021-09-15
Fix Toggle Player Engine command and refactor

Sharpe49:
https://bugs.launchpad.net/or/+bug/1943648

### DIFF
--- a/Source/Orts.Simulation/Common/Commands.cs
+++ b/Source/Orts.Simulation/Common/Commands.cs
@@ -1740,8 +1740,7 @@ namespace Orts.Common
 
         public override void Redo()
         {
-            if (Receiver == null) return;
-            Receiver.LocomotivePowerSupply.HandleEvent(Receiver.DieselEngines.PowerOn ? PowerSupplyEvent.StopEngine : PowerSupplyEvent.StartEngine);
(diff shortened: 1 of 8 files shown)


twpol:
@Sharpe49 Please can you post on [the forum thread](http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/page__view__findpost__p__275780) explaining the impact and scope of these changes and why the deserve to be a release candidate fix

I am already hesitant because of the refactoring - that is not appropriate for a release candidate fix IMHO

Sharpe49:
@twpol This is one of the problems mentioned by Carlo in this message of the private forum: http://www.elvastower.com/forums/index.php?/topic/35296-release-plan/page__pid__275757#entry275757

Basically, in the power supply package, there was an error in the way that the Control Player Engine key was handled.
Basically, it started all of the engines of the train whereas it should only start the first engine of the player locomotive.
The other engines must start only when the Control Helper Engine key is used.
This fix would have been tricky to fix if the refactoring was not done because a part of the old code was not going through the new power supply classes. So the refactoring consist in moving the old code in the power supply classes.

twpol (approved):
Okay, I'll approve this, thanks!

PR #484, opened 2021-09-15
fix: Do not act recursively on documentation or things go wrong

twpol:
In particular, different language versions of the manual overwrite each other


cjakeman (approved):
Hi James,

Thanks for this. Approved,

Chris

PR #488, opened 2021-09-17
Correct error introduced that stopped some brake tokens from operating

peternewell:
Corrects an error introduced as part of this blueprint -  https://blueprints.launchpad.net/or/+spec/brake-enhancements

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
@@ -37,40 +37,44 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
             float demandedAutoCylPressurePSI = 0;
 
             // Only allow EP brake tokens to operate if car is connected to an EP system
-            if (lead != null && lead.BrakeSystem is EPBrakeSystem && Car.BrakeSystem is EPBrakeSystem && (lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPFullServ || lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPOnly || lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPApply))
+            if (lead == null || !(lead.BrakeSystem is EPBrakeSystem))
             {
-
-                if (BrakeLine3PressurePSI >= 1000f || Car.Train.BrakeLine4 < 0)
-                {
-                    HoldingValve = ValveState.Release;
-                }
-                else if (Car.Train.BrakeLine4 == 0)
-                {
-                    HoldingValve = ValveState.Lap;
-                }
-                else
-                {
-                    demandedAutoCylPressurePSI = Math.Min(Math.Max(Car.Train.BrakeLine4, 0), 1) * MaxCylPressurePSI;
-                    HoldingValve = AutoCylPressurePSI <= demandedAutoCylPressurePSI ? ValveState.Lap : ValveState.Release;
-                }
+                HoldingValve = ValveState.Release;
+                base.Update(elapsedClockSeconds);
+                return;
             }
+
+            // process valid EP brake tokens
+
+            if (BrakeLine3PressurePSI >= 1000f || Car.Train.BrakeLine4 < 0)
+            {
+                HoldingValve = ValveState.Release;
+            }
+            else if (Car.Train.BrakeLine4 == 0)
+            {
+                HoldingValve = ValveState.Lap;
+            }
+            else
+            {
+                demandedAutoCylPressurePSI = Math.Min(Math.Max(Car.Train.BrakeLine4, 0), 1) * MaxCylPressurePSI;
+                HoldingValve = AutoCylPressurePSI <= demandedAutoCylPressurePSI ? ValveState.Lap : ValveState.Release;
+            }
+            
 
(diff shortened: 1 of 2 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
```diff
@@ -37,40 +37,44 @@ public override void Update(float elapsedClockSeconds)
             float demandedAutoCylPressurePSI = 0;
 
             // Only allow EP brake tokens to operate if car is connected to an EP system
-            if (lead != null && lead.BrakeSystem is EPBrakeSystem && Car.BrakeSystem is EPBrakeSystem && (lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPFullServ || lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPOnly || lead.TrainBrakeController.TrainBrakeControllerState == ControllerState.EPApply))
+            if (lead == null || !(lead.BrakeSystem is EPBrakeSystem))
```
cesarBLG, line 40:
```suggestion
            if (!(lead?.BrakeSystem is EPBrakeSystem))
```
The null conditional operator could be used here.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/SMEBrakeSystem.cs
```diff
@@ -41,38 +41,43 @@ public override void Update(float elapsedClockSeconds)
...
-                else if (Car.Train.BrakeLine4 == 0)
-                {
-                    HoldingValve = ValveState.Lap;
-                }
-                else
-                {
-                    demandedAutoCylPressurePSI = Math.Min(Math.Max(Car.Train.BrakeLine4, 0), 1) * MaxCylPressurePSI;
-                    HoldingValve = AutoCylPressurePSI <= demandedAutoCylPressurePSI ? ValveState.Lap : ValveState.Release;
-                }
+            if (lead == null || !(lead.BrakeSystem is SMEBrakeSystem))
```
cesarBLG, line 44:
```suggestion
            if (!(lead?.BrakeSystem is SMEBrakeSystem))
```
Same here.

cesarBLG (commented):
Thanks, Peter. Brakes work as expected now.

peternewell:
> 
> 
> Thanks, Peter. Brakes work as expected now.

Thanks for the feedback and confirmation

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
peternewell: If I saw some code like what you have suggested, I would need to spend time researching (googling) to try and understand what it is doing.

Is there any significant benefit in OR performance from using the shortened version?

So whilst I agree that it could be used, I personally prefer to use the "long hand" way rather then the "short hand" way as I find that it is more "readable" for inexperienced coders like myself.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/SMEBrakeSystem.cs
peternewell: See above comment.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
cesarBLG: I agree that using this kind of 'shortcuts' need more attention until we all get used to them. However, from my experience in other PRs the shorter version is usually preferred.

The performance gain, if any, would be negligible, so I have no objections to keeping things as they are.

peternewell: > 
> 
> However, from my experience in other PRs the shorter version is usually preferred.
> 
I agree that this seems to be the case. To me it is about readability for all levels of coders working on OR.

Perhaps there is a requirement for a guideline and some direction from the ORMT on this. If the desire is to use the "short hand" approach then the guideline would need to offer some directions and interpretations so that coders understand what they mean, and how to use them so that there is a level of consistency.

Anyway thanks for the feedback.

Csantucci: I also prefer the "longer" version;  the shorter version is more cryptic to me.

twpol: For a single null-check as we have here, I would not express a strong preference for or against the null conditional operator.

However, once you start getting two, three, or more null-checks (as we have in a few places), I would certainly recommend using it. It is not just _a lot_ shorter, but it is also faster, because it does not fetch each level (e.g. A.B.C) multiple times.

In the long run, I would rather we have cleaner, simpler-looking code with the null conditional operator, and some people new to C# will likely have to learn about it. This also applies to a lot of other features we have in C# 7.3 over, say, the more basic features in C# 2/3

cjakeman:
Thanks, Peter

PR #498, opened 2021-09-24
Fixed wrong integer values for cabview controls related to power supplies

Sharpe49:
https://bugs.launchpad.net/or/+bug/1944783

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/PowerSupply/CircuitBreaker.cs
+++ b/Source/Orts.Simulation/Common/Scripting/PowerSupply/CircuitBreaker.cs
@@ -46,7 +46,7 @@ namespace ORTS.Scripting.Api
 
     public enum CircuitBreakerState
     {
-        [GetString("Unavailable")] Unavailable,
+        [GetString("Unavailable")] Unavailable = -1,
         [GetParticularString("CircuitBreaker", "Open")] Open,
         [GetParticularString("CircuitBreaker", "Closing")] Closing,
         [GetParticularString("CircuitBreaker", "Closed")] Closed
--- a/Source/Orts.Simulation/Common/Scripting/PowerSupply/PowerSupply.cs
+++ b/Source/Orts.Simulation/Common/Scripting/PowerSupply/PowerSupply.cs
@@ -160,7 +160,7 @@ namespace ORTS.Scripting.Api
 
     public enum PowerSupplyState
     {
-        [GetString("Unavailable")] Unavailable,
+        [GetString("Unavailable")] Unavailable = -1,
         [GetParticularString("PowerSupply", "Off")] PowerOff,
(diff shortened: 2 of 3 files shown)


twpol:
Please could you explain in http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/ why this needs to be fixed for 1.4? What's the impact and risk?

Sharpe49:
I added a message on the forum, but this is already explained in the bug report.
Cabview controls are not showing the right aspect that corresponds to the state of the variables in the power supply classes.
So what is displayed on the cabview is wrong.

twpol:
> @twpol I added a message on the forum, but this is already explained in the bug report.

Please can you assume that people have read such linked items in future.

Sharpe49:
@twpol If nobody reads the bug report... then the bug reports are useless.

twpol:
> @twpol If nobody reads the bug report... then the bug reports are useless.

I read the bug report **before** I commented on here. It did not answer my questions, which is why I asked for more information

I am asking that you **assume** people have done that instead of simply telling them to read the bug report

Sharpe49:
Sorry, I misunderstood the sentence in your second message.
Remember, I am French. English may be my second language but I may sometimes interpret things wrongly. And this is what happened here.
What I understood in French was: "Do you assume that people have read the bug report?"
(It may also be due to the fact that I am still a bit exhausted from what I did during my week-end)

By the way, shouldn't comments about a bug report be sent directly on Launchpad?
Why not use the comments in Launchpad to talk about this instead of the forums? The information is scattered on 3 different platforms currently.

twpol:
> Sorry, I misunderstood the sentence in your second message. Remember, I am French. English may be my second language but I may sometimes interpret things wrongly. And this is what happened here. What I understood in French was: "Do you assume that people have read the bug report?" (It may also be due to the fact that I am still a bit exhausted from what I did during my week-end)

I am sorry about the confusion :(

> By the way, shouldn't comments about a bug report be sent directly on Launchpad? Why not use the comments in Launchpad to talk about this instead of the forums? The information is scattered on 3 different platforms currently.

I was specifically asking about the bug's **inclusion in 1.4**. For this particular question, we have chosen to keep everything in [this one forum thread](http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/) so that everyone can see what is going on and keep track of things

All other comments about a bug should go in Launchpad (and comments about code should go in GitHub)

I do understand the problems caused by the separation of these three systems (using GitHub for code instead of Launchpad was mostly accidental). GitHub can perform all of these operations in theory (pull requests/code, issues, discussions), but it would be painful to switch. You are welcome to start a discussion on the forums about it, though.

twpol (approved):

PR #507, opened 2021-09-30
Correct an issue with independent brake not working

peternewell:

Corrects an issue with the independent brake not working, and identified in this post -

Bug report: https://www.elvastower.com/forums/index.php?/topic/35363-preparing-for-release-or-v14/page__view__findpost__p__276504

(Does this automatically merge, how do I get the user to confirm that the fix has worked?)

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -827,9 +827,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
             int last = -1;
             train.FindLeadLocomotives(ref first, ref last);
             float sumpv = 0;
-            float summainrespv = 0;
             float sumv = 0;
-            float summainresv = 0;
             int continuousFromInclusive = 0;
             int continuousToExclusive = train.Cars.Count;
             for (int i = 0; i < train.Cars.Count; i++)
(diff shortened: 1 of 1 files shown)


twpol:
> (Does this automatically merge, how do I get the user to confirm that the fix has worked?)

Yes, as you might've spotted, this PR gets included in the Unstable Version just the same as ones targeting `master` :)

peternewell:
>
> Yes, as you might've spotted, this PR gets included in the Unstable Version just the same as ones targeting `master` :)
>
Ok, thanks for that.

Should this patch be in rc5 now?

I asked somebody to confirm that it was working, and whilst it appears to be working in the unstable version, it isn't in rc5.

http://www.elvastower.com/forums/index.php?/topic/35363-preparing-for-release-or-v14/page__view__findpost__p__276609

twpol:
> Should this patch be in rc5 now?

I assume you mean 1.4-rc6, because we never replace an existing version :)

The release candidates are similar to the Testing Version - they are only updated after the PR is approved and merged - but involve some extra, manual steps. I don't create a new release candidate for every merge, otherwise we'd be up to rc24 or so, and we haven't got it automated this time around

There are two things you need:

- Approval via the pull request
- Acceptance for 1.4 via http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/ (please post or link to an existing post explaining why it needs to go into 1.4 instead of waiting, how badly it'll affect users if not fixed, etc.)

peternewell:
> 
> I assume you mean 1.4-rc6, because we never replace an existing version :)
> 
> The release candidates are similar to the Testing Version - they are only updated after the PR is approved and merged - but involve some extra, manual steps. I don't create a new release candidate for every merge, otherwise we'd be up to rc24 or so, and we haven't got it automated this time around
>
Thanks for the explanation, this now makes sense.

> 
>     * Acceptance for 1.4 via http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/ (please post or link to an existing post explaining why it needs to go into 1.4 instead of waiting, how badly it'll affect users if not fixed, etc.)
>
I have added information to the thread that you have asked me to -

http://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/page__view__findpost__p__276647

twpol (approved):

PR #509, opened 2021-09-30
fix: Swap colour bytes from MSTS/XNA order to Monogame/DX order

twpol:
Bug report: https://www.elvastower.com/forums/index.php?/topic/35481-open-rails-14-is-coming-soon/page__view__findpost__p__276585


cjakeman (approved):
Hi James,

All good.

Approved with thanks,

Chris

PR #512, opened 2021-10-03
Fixed dynamic brake being applied during standstill when dynamic brake blending is enabled

Sharpe49:
(which blocks throttle during hill start)
https://bugs.launchpad.net/or/+bug/1945886


Sharpe49:
@cjakeman I made a small modification: SpeedMpS can be negative. We must use absolute value here.

PR #514, opened 2021-10-07
Some strings should not be extracted for translation

mbm-OR:
Some strings are used as internal keywords and should not be extracted to the *.pot file for translation.
Fix for https://bugs.launchpad.net/or/+bug/1946373.

### DIFF
--- a/Source/Locales/RunActivity/RunActivity.pot
+++ b/Source/Locales/RunActivity/RunActivity.pot
@@ -1,8 +1,8 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2021-09-15 03:48:21+0200\n"
-"PO-Revision-Date: 2021-09-15 03:48:22+0200\n"
+"POT-Creation-Date: 2021-10-11 16:55:19+0200\n"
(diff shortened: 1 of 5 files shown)


Sharpe49 (changes requested):
You should probably update the POT files to check if some keywords are still translated and then add the updated POT files to this pull request.

mbm-OR:
> 
> 
> You should probably update the POT files to check if some keywords are still translated and then add the updated POT files to this pull request.

Currently, no keywords pointed by this PR, have been translated in any "* .PO" files.
Once updated with update.bat, these keywords appear in the POT file, so if they don't need to be translated, they should not appear in the POT file either.
That is the reason for this PR, suggested by Weter, [here.](http://www.elvastower.com/forums/index.php?/topic/26556-translations-unclear-strings-other-problems/page__view__findpost__p__276547)

Sharpe49:
What I mean is that the previously translated keywords are still in the POT file. For example : https://github.com/openrails/openrails/blob/fab59f576535f027a2ccac044b9de4cbd4d74900/Source/Locales/RunActivity/RunActivity.pot#L1912
You need to remove those keywords from the POT file.

Sharpe49 (approved):

PR #515, opened 2021-10-08
Correct issue with compressor not turning off

peternewell:

Corrects an issue identified -

http://www.elvastower.com/forums/index.php?/topic/35558-compressor-doesnt-shut-off/


peternewell:
> 
> 
> Seems good to me. You need a bug report to merge this fix. This bug also affects 1.4.

Thanks for that.

Based upon posting, I assume that v1.4 is closed unless it is a major issue.

This is probably more a cosmetic issue I suspect, rather then a major performance issue.

PR #521, opened 2021-10-24
Further strings translation

Weter-ORTS:


Weter-ORTS (commented):
have just removed that two ocassionally saved files:
i only closed PoEdit with "save" button instead of cross
i.e. no actual modifications, but new date of modification

Sharpe49:
You deleted the files instead of remove them from the commit.
You can go back to the first commit using the git reset function.
Then amend the commit in order to remove the files from the commit.
Then force push the new commit to Github.

Weter-ORTS:
> removing them from the commit.
how to do that correct?

Sharpe49:
Which software are you using for Git?

Weter-ORTS:
firefox browser only

Weter-ORTS:
> You can go back to the first commit using the git reset function.
> Then amend the commit in order to remove the files from the commit.
new words for me, sorry

Sharpe49:
So the Github website directly? OK, you will need a Windows application to do this change. You can use TortoiseGit (but it is a bit complicated) or SourceTree (which is simpler).

Weter-ORTS:
Ryan did show to me fetch button:
as i understood, this updates all material with newer dates from master branch,
after removing that two ones, i fetched my branch
I expect, that missing files will be restored

Sharpe49:
No, this won't work here.

Weter-ORTS:
so what must i do?

Sharpe49:
As I said, you need to download a Windows Git application as the ones mentioned before.

Weter-ORTS:
but what if closing that PR and starting new one with only three "right" files?

Weter-ORTS:
TBH i, being no programmist would prefer a simplest way, that worked all last year.

Sharpe49:
> but what if closing that PR and starting new one with only three "right" files?

This could work.

Weter-ORTS:
so, new PR...

Weter-ORTS:
restarting...

PR #524, opened 2021-10-30
Further strings translation

Weter-ORTS:


Sharpe49:
In this branch the files are still deleted.

Weter-ORTS:
Have just restored files, been missing...

pzgulyas:
Hey Weter, I see you have trouble submitting the PR. :-)
I would also recommend using a Windows client for managing these. I personally use [TortoiseGit](https://tortoisegit.org/download/), I can assist in installing that, and making the first steps, if you feel ready for that. If you okay with that, then download the installation package, and install it in a next-next-finish style. At a point it will ask for the install path of Git itself. There it will show you a download location of the Git-for-Windows, download and install that as well in a next-next-finish style. When you are ready, just tell me, we can proceed with cloning your github repository to your local disk.

pzgulyas:
@Weter-ORTS : Or, alternatively, if you don't want to bother with these, I can submit the ru-locales for you in a PR, so you don't need to learn Git. Just let me know.

YoRyan:
We could certainly use these translations, so I've gone ahead and reverted the .pot diffs that only modified the timestamp. I trust that makes this PR suitable for merging?

For future reference, the command to retrieve a version of a file from a different branch (in this case, master) is:

```
git checkout master -- path/to/file.pot
```

Weter-ORTS:
I would check for new strings, which appeared during last three month.
Should I delete my branch completely, or fetch button does enough?

pzgulyas (approved):
Lets merge this.

PR #526, opened 2021-10-31
Bug fix for https://bugs.launchpad.net/or/+bug/1949292 AI train disappears after coupling and reversing

Csantucci:
The fix follows the same lines used for the player train


Sharpe49 (commented):
Code changes seems OK.
But I don't have a test case to check if it works properly.

Csantucci:
> 
> 
> Code changes seems OK. But I don't have a test case to check if it works properly.

If you want to test, here http://www.elvastower.com/forums/index.php?/topic/35624-ai-train-disappearing-after-coupling-and-reversing/ you find a test activity. With the actual master release the AI train in the activity will disappear after coupling. With the patch this doesn't occur.

Sharpe49 (approved):

Csantucci:
Thanks for approval.

PR #527, opened 2021-10-31
Brake cuts power refactor and new parameters

Sharpe49:
https://blueprints.launchpad.net/or/+spec/brake-cuts-power-parameters

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -4485,12 +4485,56 @@ the tables below.
 .. index::
    single: DoesBrakeCutPower
    single: BrakeCutsPowerAtBrakeCylinderPressure
+   single: ORTSDoesVacuumBrakeCutPower
+   single: ORTSBrakeCutsPowerAtBrakePipePressure
+   single: ORTSBrakeRestoresPowerAtBrakePipePressure
+   single: ORTSBrakeCutsPowerForMinimumSpeed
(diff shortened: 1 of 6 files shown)


Sharpe49:
Ok, removed one file that shouldn't have been in the commit. It will probably fix the automerge problem on my other pull request.

peternewell:
Has this change created the issue raised in this post?

http://www.elvastower.com/forums/index.php?/topic/35678-ortsdoesvacuumbrakecutpower-is-no-longer-working/

Sharpe49:
Yes, this change has provoked this issue. I will take a look at it.

pzgulyas (dismissed):
Seems to be a careful consideration behind the code (although I have not much knowledge about the real world systems behind)

Sharpe49:
In the LOC&PAS TSI, this functionnality is partly defined here (when cut-off is mentionned): https://eur-lex.europa.eu/legal-content/EN/TXT/HTML/?uri=CELEX:02014R1302-20200311&from=EN#tocId102 (§4.2.4.4)
More details were present in a French document.

This situation still needs to be clarified: http://www.elvastower.com/forums/index.php?/topic/35678-ortsdoesvacuumbrakecutpower-is-no-longer-working/

Sharpe49:
I added a variable to the abstract class in order for the script migration to be easier: script developers only have to call UpdateTractionCutOff() in the Update function and use the TractionCutOffRequested variable inside the SetTractionAuthorization() function (with other variables that may trigger a traction cut-off).

PR #530, opened 2021-11-10
Steam adhesion not applied to all locomotives in the player train.

peternewell:
See this thread for more details -

Bug report - https://bugs.launchpad.net/or/+bug/1950408

http://www.elvastower.com/forums/index.php?/topic/35337-curtius-kniffler-adhesion-parameter/page__view__findpost__p__277792


Csantucci (commented):
In the comment below your correction in MSTSSteamLocomotive.cs you write:
"and is the player locomotive"
Accordingly to your fix I'd expect that this comment should be modified.

peternewell:
> 
> 
> In the comment below your correction in MSTSSteamLocomotive.cs you write: "and is the player locomotive" Accordingly to your fix I'd expect that this comment should be modified.

Thanks for picking that up. Corrected as suggested.

Sharpe49 (approved):

Sharpe49:
You need to add a link to a bug report.

peternewell:
Thanks for approving this.

PR #533, opened 2021-11-20
fix for shapes hidden by animations( 0 )

cjakeman:
See https://bugs.launchpad.net/or/+bug/1951677

First reported by Tim Muir and Chris Gerlach after work on the City Point and Army Line http://www.elvastower.com/forums/index.php?/topic/33552-work-started-along-city-point-army-line/ found that Stable Release 1.4 failed to show certain shapes.

Investigation by Jonas found that the test for an animated clock shape fails when the shape includes
animations( 0 )
and this is automatically inserted by MSTSexporter for Blender 2.8+ Release V4.0 or older.

### DIFF
--- a/Source/RunActivity/Viewer3D/Scenery.cs
+++ b/Source/RunActivity/Viewer3D/Scenery.cs
@@ -463,10 +463,10 @@ namespace Orts.Viewer3D
                         if (Program.Simulator.CarSpawnerLists != null && ((CarSpawnerObj)worldObject).ListName != null)
                         {
                             ((CarSpawnerObj)worldObject).CarSpawnerListIdx = Program.Simulator.CarSpawnerLists.FindIndex(x => x.ListName == ((CarSpawnerObj)worldObject).ListName);
-                            if (((CarSpawnerObj)worldObject).CarSpawnerListIdx < 0 || ((CarSpawnerObj)worldObject).CarSpawnerListIdx > Program.Simulator.CarSpawnerLists.Count-1) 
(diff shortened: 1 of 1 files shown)


Source/RunActivity/Viewer3D/Scenery.cs
```diff
@@ -480,10 +480,16 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
                     }
                     else if (worldObject.GetType() == typeof(StaticObj))
                     {
-                        //          preTestShape for lookup if it is an animated clock shape with subobjects named as clock hands 
+                        // preTestShape for lookup if it is an animated clock shape with subobjects named as clock hands 
                         StaticShape preTestShape = (new StaticShape(viewer, shapeFilePath, worldMatrix, shadowCaster ? ShapeFlags.ShadowCaster : ShapeFlags.None));
-                        var animNodes = preTestShape.SharedShape.Animations?[0]?.anim_nodes ?? new List<anim_node>();
```
pzgulyas, line 485: Wouldn't it be shorter to do simply
`var animNodes = preTestShape.SharedShape.Animations?.FirstOrDefault()?.anim_nodes ?? new List<anim_node>();`

Sharpe49: I agree, this is much simpler and fits what needs to be fixed.

cjakeman: Thanks, Peter.
Your suggestion now implemented.

Sharpe49 (approved):

PR #534, opened 2021-11-20
deletes RunActivityLAA.* post-build

cjakeman:
Small extension to the RunActivity project file to prevent old version of RunActivityLAA.exe being used accidentally.

See https://blueprints.launchpad.net/or/+spec/delete-runactivitylaa


Csantucci (approved):
Good idea, approved!

PR #535, opened 2021-11-21
Correction of OR manual

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/35676-game-time-acceleration-in-dispatcher-window/page__view__findpost__p__278145

### DIFF
--- a/Source/Documentation/Manual/driving.rst
+++ b/Source/Documentation/Manual/driving.rst
@@ -1606,8 +1606,8 @@ key assignment list shown by pressing ``<F1>``.
 In addition, Open Rails offers functionality similar to the time 
 acceleration switch for MSTS. 
 
-Use ``<Ctrl+Alt+PgUp>`` or ``<Ctrl+Alt+PgDn>`` keys to increase or decrease the 
-speed of the game clock. ``<Ctrl+Alt+Home>`` resets the speed.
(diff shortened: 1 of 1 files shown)


Source/Documentation/Manual/driving.rst
```diff
@@ -1606,8 +1606,8 @@ key assignment list shown by pressing ``<F1>``.
 In addition, Open Rails offers functionality similar to the time 
 acceleration switch for MSTS. 
 
-Use ``<Alt+PgUp>`` or ``<Alt+PgDn>`` keys to increase or decrease the 
-speed of the game clock.
+Use ``<Ctrl+Alt+PgUp>`` or ``<Ctrl+Alt+PgDn>`` keys to increase or decrease the 
+speed of the game clock. ``<Ctrl+Alt+Home>`` resets the speed.
```
Sharpe49, line 1610: You should say it is the 9, 3 and 7 keys from the numeric keypad in order to not confuse them with the other keys called PgUp, PgDn and Home.

Sharpe49 (changes requested):

Source/Documentation/Manual/driving.rst
Csantucci: Modified.

Sharpe49 (approved):

Source/Documentation/Manual/driving.rst
twpol: @Sharpe49 Actually it was designed for the main keys; it _also_ works on the numeric keypad, but only with Num Lock off

PR #537, opened 2021-11-23
C# signal script extensions

Sharpe49:
https://blueprints.launchpad.net/or/+spec/csharp-signal-script-extensions
https://blueprints.launchpad.net/or/+spec/override-speed-limit-from-csharp-signal-scripts

### DIFF
--- a/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
@@ -2,6 +2,7 @@
 using ORTS.Scripting.Api;
 using System;
 using System.Collections.Generic;
+using System.IO;
 
 namespace Orts.Simulation.Signalling
 {
@@ -108,6 +109,10 @@ namespace Orts.Simulation.Signalling
         /// Name of this signal type, as defined in sigcfg
         /// </summary>
         public string SignalTypeName => SignalHead.SignalTypeName;
+        /// <summary>
+        /// Name of the signal shape, as defined in sigcfg
+        /// </summary>
+        public string SignalShapeName => Path.GetFileNameWithoutExtension(SignalObject.WorldObject.SFileName);
         /// <summary>
         /// Local storage of this signal, which can be accessed from other signals
         /// </summary>
--- a/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
@@ -6,6 +6,13 @@ using System.IO;
 
 namespace Orts.Simulation.Signalling
 {
+    public enum SignalEvent
+    {
+        RequestMostRestrictiveAspect,
+        RequestApproachAspect,
+        RequestLeastRestrictiveAspect,
+    }
+
     // The exchange of information is done through the TextSignalAspect property.
     // The MSTS signal aspect is only used for TCS scripts that do not support TextSignalAspect.
     public abstract class CsSignalScript : AbstractScriptClass
@@ -81,6 +88,10 @@ namespace Orts.Simulation.Signalling
         /// True if the signal link is activated
         /// </summary>
         public bool RouteSet => SignalHead.route_set() > 0;
+        /// <summary>
+        /// Hold state of the signal
+        /// </summary>
+        public HoldState HoldState => SignalObject.holdState;
         /// <summary>
         /// Set this variable to true to allow clear to partial route
         /// </summary>
@@ -359,5 +370,10 @@ namespace Orts.Simulation.Signalling
         /// <param name="message">Message sent to signal</param>
         /// <returns></returns>
         public virtual void HandleSignalMessage(int signalId, string message) {}
+        /// <summary>
+        /// Called when the simulator
+        /// </summary>
+        /// <param name="evt"></param>
+        public virtual void HandleEvent(SignalEvent evt, string message = "") { }
     }
 }
--- a/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
@@ -3,6 +3,7 @@ using ORTS.Scripting.Api;
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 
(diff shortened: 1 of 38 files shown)


Sharpe49:
The bot doesn't want to merge this in the unstable version while Github says there is no conflict... weird.

Sharpe49:
I finally integrated the speed limit override with a refactoring consisting in:
- Moving the signal functions to a class and referencing the signal function by reference instead of an integer index
- Replacing the isSignal and isSpeedSignal variables by a enumerated property

This will probably conflict with #525 so, after one of the pull request is merged, I will have to rebase.
I think #525 is the easiest to review and so it should be merged first.

Roeterdink:
I find it difficult to review these changes. Code has been changed, but the split over multiple files makes the .diff files completely useless. It is therefor just impossible to find out what code changes have been made.
What I did notice is that all //===// lines separating methods have been removed. I'm very disapointed by that. I have always used these separations in all my coding to emphasize the beginning of a method, it helps me to orientate when scrolling through the code. Obviously these lines have nothing to do with the actual code so I cannot understand why these have been removed.
Regards, Rob Roeterdink

cjakeman:
Hi Cedric,

The Code Bot (https://james-ross.co.uk/projects/or) reports a Merge conflict in Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs

Are you able to sort this out?

Thanks,

Chris

Source/Orts.Simulation/Simulation/Signalling/SignalObject.cs
```diff
@@ -4223,18 +4224,38 @@ public bool[] RequestHoldSignalDispatcher(bool requestResetSignal)
...
+            }
+        }
+
+        public void RequestLeastRestrictiveAspect()
+        {
+            holdState = HoldState.ManualPass;
+            foreach (var sigHead in SignalHeads)
+            {
+                sigHead.RequestLeastRestrictiveAspect();
+                sigHead.draw_state = sigHead.def_draw_state(sigHead.state);
```
cesarBLG, line 4235: Why are you forcing the draw state here? This will override the draw aspect set by the C# script after the RequestLeastRestrictiveAspect event is received.

Sharpe49: I probably messed up something during rebase.
I have sent a fix.

Source/Orts.Simulation/Simulation/Signalling/CsSignalScript.cs
```diff
@@ -359,21 +376,87 @@ internal void AttachToHead(SignalHead signalHead)
...
+        /// <summary>
+        /// Indicates if the signal script has taken control over the speed limit (overrides the speed limit set in the SignalType).
+        /// </summary>
+        public bool SpeedLimitSetByScript { get => SignalHead.SpeedInfoSetBySignalScript; set => SignalHead.SpeedInfoSetBySignalScript = value; }
+
+        /// <summary>
+        /// Returns the parameters of the speed limit set by the script.
+        /// </summary>
+        /// <returns>A tuple containing the parameters of the speed limit, or null if the speed limit is not set</returns>
+        public (float PassengerSpeedLimitMpS, float FreightSpeedLimitMpS, bool Asap, bool Reset, bool NoSpeedReduction, bool IsWarning)? SpeedLimit()
```
cesarBLG, line 407: Is there a reason for using a tuple instead of a struct? I think a struct might be more future-proof in case new flags are added.

Sharpe49: They are equivalent in terms of future proofing. The only difference is the length of the function definition.

Tuples in C# are structs that don't have a predefined structure.
Here, I used it because it doesn't make really sense creating a new structure just to pass information to another function.

If I change something, I would not use structs but record structs (which are much shorter to declare than structs). But this would require C# 10 which is not yet enabled in the project files.

cesarBLG: Thanks for your clarification. I had the wrong idea that extending the tuple would break existing scripts.

cesarBLG (approved):
Thanks for this useful additions to signal scripts!
I reviewed the PR commit by commit, but I didn't had a deep look into refactoring commits, as they are huge. I assume they are only moving things around and reformatting the code.
I didn't spot more issues, so I'm approving. This has been around for quite a long time.

Sharpe49:
@cesarBLG Well, I have to rebase once more... ;)

cesarBLG (approved):
OK, I'm merging before more conflicts arise

cesarBLG:
I tested a different route and I'm getting some null reference exceptions:
SignalObject.cs:97
```csharp 
if (SignalHeads.Exists(x => x.Function.MstsFunction != MstsSignalFunction.SPEED))
```
It crashes during startup. Changing to Function?.MstsFunction solves the issue, but I believe Function should always be non-null.

Also, NextSpeedPostFeatures returns a null SpeedPostTypeName.

Probably this is caused by errors in the route, but it used to work.

Sharpe49:
I will take a look at it.
=

Sharpe49:
@cesarBLG Can you send a log when it fails?

Sharpe49:
@cesarBLG It was for this PR that I needed some log. ;)

cesarBLG:
Oops, I thought it was for the other PR. Here you are:
[OpenRailsLog.txt](https://github.com/openrails/openrails/files/9392706/OpenRailsLog.txt)

Sharpe49:
You have a missing signal type (etcs_estacion_pere_comas) and a lot of weird things in your SIGSCR. Can you check it?
Maybe the new code is less tolerant to SIGCFG errors.

cesarBLG:
There was a phantom signal or something like that. Removing it solves the null pointer error at SignalFunction.

I am still investigating why SpeedPostFeatures.SpeedPostTypeName is null, as everything seems correct with that speedpost.

cesarBLG:
It seems that AddWorldInfo() is calling the signal.Type property, which internally requires signal.SpeedPostWorldObject to be initialized. However, the initialization of signal.SpeedPostWorldObject only happens in AddWorldInfo() after the signal.Type property is checked. This results in an incorrectly initialized speed post.

PR #540, opened 2021-12-03
Correct issue created in timetable mode

peternewell:
See this bug - 

https://bugs.launchpad.net/or/+bug/1953068


Sharpe49 (approved):
Ok, comparing code with TortoiseGitMerge ignoring whitespace changes made things clearer : besides the added null check, LocomotiveIdentification was replaced by LocomotiveParameters which is exactly the same so that's one duplicate variable less.
That's approved for me.

peternewell:
Thanks for the approval

twpol:
> Ok, comparing code with TortoiseGitMerge ignoring whitespace changes made things clearer

FYI: you can also ignore whitespace with GitHub, but it's a little hidden:

image

PR #538, opened 2021-11-25
Delete binary paths if the ascii version or the tdb was modified. Blueprint: https://blueprints.launchpad.net/or/+spec/binary-timetable-paths

pzgulyas:
Delete binary paths if the ascii version or the tdb was modified. Blueprint: https://blueprints.launchpad.net/or/+spec/binary-timetable-paths

PR #541, opened 2021-12-04
Initial Virtual File System development

pzgulyas:
Added support for virtual file system, manageable by config files. Se the [discussion thread](http://www.elvastower.com/forums/index.php?/topic/16633-overall-file-system-and-data-structure/)

### DIFF
--- a/Source/Menu/Options.Designer.cs
+++ b/Source/Menu/Options.Designer.cs
@@ -207,6 +207,8 @@
             this.ElevationText = new System.Windows.Forms.Label();
             this.checkPreferDDSTexture = new System.Windows.Forms.CheckBox();
             this.toolTip1 = new System.Windows.Forms.ToolTip(this.components);
+            this.checkVfsLog = new System.Windows.Forms.CheckBox();
+            this.buttonContentFile = new System.Windows.Forms.Button();
             ((System.ComponentModel.ISupportInitialize)(this.numericBrakePipeChargingRate)).BeginInit();
             this.tabOptions.SuspendLayout();
             this.tabPageGeneral.SuspendLayout();
@@ -1782,6 +1784,7 @@
             // 
             this.groupBoxContent.Anchor = ((System.Windows.Forms.AnchorStyles)(((System.Windows.Forms.AnchorStyles.Bottom | System.Windows.Forms.AnchorStyles.Left) 
             | System.Windows.Forms.AnchorStyles.Right)));
+            this.groupBoxContent.Controls.Add(this.buttonContentFile);
             this.groupBoxContent.Controls.Add(this.buttonContentBrowse);
             this.groupBoxContent.Controls.Add(this.textBoxContentPath);
             this.groupBoxContent.Controls.Add(this.label20);
@@ -1799,9 +1802,9 @@
             this.buttonContentBrowse.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Right)));
             this.buttonContentBrowse.Location = new System.Drawing.Point(428, 19);
             this.buttonContentBrowse.Name = "buttonContentBrowse";
-            this.buttonContentBrowse.Size = new System.Drawing.Size(75, 23);
+            this.buttonContentBrowse.Size = new System.Drawing.Size(35, 23);
             this.buttonContentBrowse.TabIndex = 2;
-            this.buttonContentBrowse.Text = "Change...";
+            this.buttonContentBrowse.Text = "Dir";
             this.buttonContentBrowse.UseVisualStyleBackColor = true;
             this.buttonContentBrowse.Click += new System.EventHandler(this.buttonContentBrowse_Click);
             // 
@@ -1980,6 +1983,7 @@
             this.tabPageExperimental.Controls.Add(this.trackAdhesionFactorChange);
             this.tabPageExperimental.Controls.Add(this.trackAdhesionFactor);
             this.tabPageExperimental.Controls.Add(this.checkAdhesionPropToWeather);
+            this.tabPageExperimental.Controls.Add(this.checkVfsLog);
             this.tabPageExperimental.Controls.Add(this.checkCircularSpeedGauge);
             this.tabPageExperimental.Controls.Add(this.checkSignalLightGlow);
             this.tabPageExperimental.Controls.Add(this.checkUseMSTSEnv);
@@ -2567,6 +2571,27 @@
             this.checkPreferDDSTexture.Text = "Load DDS textures in preference to ACE";
             this.checkPreferDDSTexture.UseVisualStyleBackColor = true;
             // 
+            // checkVfsLog
+            // 
+            this.checkVfsLog.AutoSize = true;
+            this.checkVfsLog.Location = new System.Drawing.Point(6, 235);
+            this.checkVfsLog.Name = "checkVfsLog";
+            this.checkVfsLog.Size = new System.Drawing.Size(120, 17);
+            this.checkVfsLog.TabIndex = 21;
+            this.checkVfsLog.Text = "Log VFS file access";
+            this.checkVfsLog.UseVisualStyleBackColor = true;
+            // 
+            // buttonContentFile
+            // 
+            this.buttonContentFile.Anchor = ((System.Windows.Forms.AnchorStyles)((System.Windows.Forms.AnchorStyles.Top | System.Windows.Forms.AnchorStyles.Right)));
+            this.buttonContentFile.Location = new System.Drawing.Point(468, 19);
+            this.buttonContentFile.Name = "buttonContentFile";
+            this.buttonContentFile.Size = new System.Drawing.Size(35, 23);
+            this.buttonContentFile.TabIndex = 2;
+            this.buttonContentFile.Text = "File";
+            this.buttonContentFile.UseVisualStyleBackColor = true;
+            this.buttonContentFile.Click += new System.EventHandler(this.buttonContentFile_Click);
+            // 
             // OptionsForm
             // 
             this.AcceptButton = this.buttonOK;
@@ -2831,5 +2856,7 @@
         private System.Windows.Forms.TrackBar trackAntiAliasing;
         private System.Windows.Forms.Label labelAntiAliasingValue;
         private System.Windows.Forms.Label labelAntiAliasing;
+        private System.Windows.Forms.CheckBox checkVfsLog;
+        private System.Windows.Forms.Button buttonContentFile;
     }
 }
--- a/Source/ORTS.Common/VirtualFileSystem.cs
+++ b/Source/ORTS.Common/VirtualFileSystem.cs
@@ -1,0 +1,500 @@
+﻿// COPYRIGHT 2021 by the Open Rails project.
+// 
+// This file is part of Open Rails.
+// 
+// Open Rails is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+// 
+// Open Rails is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+// 
+// You should have received a copy of the GNU General Public License
+// along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
+
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading;
+using System.Diagnostics;
+using System.Text.RegularExpressions;
+using System.IO;
+using SharpCompress.Archives;
+using SharpCompress.Readers;
+
+namespace ORTS.Common
+{
+    /// <summary>
+    /// Virtual File System. Can be <see cref="Initialize(string)"/>'d from a config file with lines e.g. C:\dir\archive.zip\subdir\ /MSTS/mountpoint/,
+    /// or from a directory. Found subsequent archives are auto-mounted to their respective locations.
+    /// <br><see cref="Path.Combine"/> like functions continue to work on virtual paths, as well as both the "\" and "/" separators.
+    /// But <see cref="Path.GetFullPath"/> must never be used on them, because that adds an unwanted drive letter.</br>
+    /// <br><see cref="File.Exists"/> becomes <see cref="Vfs.FileExists"/>, <see cref="Directory.Exists"/> becomes <see cref="Vfs.DirectoryExists"/>,
+    /// <see cref="Directory.GetFiles(string, string)"/> becomes <see cref="Vfs.GetFiles"/>.</br>
+    /// </summary>
+    public static class Vfs
+    {
+        static VfsNode VfsRoot;
+        static readonly Stack<(string, VfsNode)> Stack = new Stack<(string, VfsNode)>();
+        static readonly Dictionary<string, IArchive> OpenArchives = new Dictionary<string, IArchive>();
+        static readonly List<string> SupportedArchiveExtensions = new List<string> { ".zip", ".rar", ".7z" };
+
+        ///////////////////////////////////////////////////////////////////////////////////////
+        ///
+        /// The configuration file consits of lines of format:
+        /// Drive:\SystemPath\ /MountPoint/
+        ///
+        /// SystemPath - may use "/" or "\", but MountPoint may use "/" only as a directory separator.
+        ///            - with spaces in it must be quoted, e.g. "C:\My Routes\USA85\" /MSTS/ROUTES/USA85/
+        ///            - referring to a directory must end with "/" or "\" to assure that only _below_ dirs and files will be visible _below_ the mount point.
+        ///            - referring to an archive must not end with "/" or "\", e.g. C:\TEMP\MSTS1.2.zip /MSTS/
+        ///            - may refer to an archive subdirectory, e.g.: C:\routes.zip\USA3\ /MSTS/ROUTES/USA3/
+        ///            - may _not_ refer to a single non-archive file, neither a non-archive file within an archive.
+        ///            - is case-insenditive.
+        /// MountPoint - must end with "/" to avoid confusion and to assure that mounting will be done _below_ the given path.
+        ///            - may not contain spaces, may not be quoted.
+        ///            - is case-insensitive, will be converted to all-uppercase internally.
+        ///
+        /// Found subsequent archives are auto-mounted to their respective VFS locations, but nested archives are unsupported.
+        ///
+        /// System.IO.Path.*() functions can be used on virtual paths, but make sure that
+        /// NormalizeVirtualPath() is used on them _afterwards_ in the processing pipeline somewhere.
+        ///
+        ///////////////////////////////////////////////////////////////////////////////////////
+        ///
+        /// Public interfaces
+        /// 
+
+        public const string MstsBasePath = "/MSTS/";
+        public const string ExecutablePath = "/EXECUTABLE/";
+
+        public static bool AccessLoggingEnabled { get; set; }
+
+        public static void Initialize(string initPath, string executablePath)
+        {
+            if (File.Exists(initPath))
+                Initialize(File.ReadAllLines(initPath), executablePath);
+            else if (Directory.Exists(initPath))
+                Initialize(new[] { $"\"{initPath}/\" {MstsBasePath}" }, executablePath);
+            else
+                Trace.TraceError($"VFS: Could not initialize from {initPath}, aborting.");
+        }
+
+        public static StreamReader StreamReader(string vfsPath, bool detectEncodingFromByteOrderMarks)
+        {
+            switch (PrepareForRead(vfsPath))
+            {
+                case Stream stream: return new StreamReader(stream, detectEncodingFromByteOrderMarks);
+                case string path: return new StreamReader(path, detectEncodingFromByteOrderMarks);
+                default: throw new FileNotFoundException($"VFS reading failed: {vfsPath}");
+            }
+            // Looks like the rest of the OR code expects to throw an exception at these places instead of just returning null.
+        }
+
+        public static StreamReader StreamReader(string vfsPath, Encoding encoding)
+        {
+            switch (PrepareForRead(vfsPath))
+            {
+                case Stream stream: return new StreamReader(stream, encoding);
+                case string path: return new StreamReader(path, encoding);
+                default: throw new FileNotFoundException($"VFS reading failed: {vfsPath}");
+            }
+        }
+
+        public static StreamReader OpenText(string vfsPath)
+        {
+            switch (PrepareForRead(vfsPath))
+            {
+                case Stream stream: return new StreamReader(stream, Encoding.UTF8);
+                case string path: return File.OpenText(path);
+                default: throw new FileNotFoundException($"VFS reading failed: {vfsPath}");
+            }
+        }
+
+        public static Stream OpenRead(string vfsPath)
+        {
+            switch (PrepareForRead(vfsPath))
+            {
+                case Stream stream: return stream;
+                case string path: return File.OpenRead(path);
+                default: throw new FileNotFoundException($"VFS reading failed: {vfsPath}");
+            }
+        }
+
+        public static Stream OpenCreate(string vfsPath)
+        {
+            switch (PrepareForWrite(vfsPath))
+            {
+                case Stream stream: return stream;
+                case string path: return new FileStream(path, FileMode.Create);
+                default: throw new FileNotFoundException($"VFS writing failed: {vfsPath}");
+            }
+        }
+
+        public static string[] GetDirectories(string vfsPath) => GetDirectoriesBase(vfsPath) ?? Array.Empty<string>();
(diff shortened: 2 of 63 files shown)


pzgulyas:
I have no more open issues with this PR, so removing the not-for-unstable label. Now waiting for the conceptual review by @twpol and @cjakeman .

cjakeman (commented):
Hi Péter,

Good to see this is now ready for Unstable so non-devs can explore it.

There is a merge conflict at present (see https://james-ross.co.uk/projects/or) with Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs which is preventing your work appearing in the Unstable Version.

Are you able to look at this?

Thanks,

Chris

pzgulyas:
Hi Chris @cjakeman,

Yes two of my PRs (this and #538) conflict with each other, and whichever gets merged first into the master, the other will have a conflict that needs to be resolved, I am aware of that. However I'm not sure how to resolve it at this stage. I cannot commit the fix into either of the PRs, because then that could not be merged without the other.

The best I can do now is that I mark #538 as not-for-unstable, however that is also ready for the review and the merge.

cjakeman:
Hi Peter, @pzgulyas 

> However I'm not sure how to resolve it at this stage. I cannot commit the fix into either of the PRs, because then that could not be merged without the other.

I am happy to approve PR 538 once I've removed from the Manual the instruction to delete binary files by hand.
Hope that will unblock this mutual conflict. 
Looking at that now.

cjakeman:
Hi Peter, @pzgulyas

> I am happy to approve PR 538

PR 538 is now approved and merged into the master branch.
You should be able to update your local master branch and then resolve the conflict.

Best wishes

pzgulyas:
Conflict resolved, back on track.

cjakeman:
> Conflict resolved, back on track.

That's great. I see the PR has made it into the Unstable build.

In trying things out, I came across a case which vfs.FileExists() rejects:

"\\MSTS\\trains\\trainset\\MT_DD_CLASS_27_27102\\CABVIEW\\..\\..\\MT_Common.cabs\\HP\\Class26\\Class26.cvf"

although the file exists in that location. I hope you can come up with a simple fix.

Best wishes

pzgulyas:
Could you please double-check if you really mount the zip containing the MT_Common.cabs to the right mount point? Vfs.FileExists() flattens the above path to "/MSTS/TRAINS/TRAINSET/MT_COMMON.CABS/HP/CLASS26/CLASS26.CVF" virtual file path, so it must reside there after the mount. You might check the log file, at the beginning all the exact mount paths are logged, so you can see if everything is all right there.

Anyway, can I download this rolling stock and the cabview from somewhere to test?

pzgulyas:
Are you using the version later than the commit efffead3c1179acbb58d921675fff69c189334f0 ? I've added that specifically to flatten such paths to avoid browsing the imaginary directories, like probably the "CABVIEW" in this case.

cjakeman:
> Are you using the version later than the commit [efffead](https://github.com/openrails/openrails/commit/efffead3c1179acbb58d921675fff69c189334f0) ? I've added that specifically to flatten such paths to avoid browsing the imaginary directories, like probably the "CABVIEW" in this case.

Yes, I am. Also I tried adding a real CABVIEW directory and it made no difference.

The loco I am using is the one from the activity in Demo Model 1 from http://openrails.org/download/content/ 

Thanks for looking at this.

pzgulyas:
Hi Chris, @cjakeman, My bad, I should have thought of testing it with our own test asset. :-) However on my part everything works well. Now I think your vfs configuration might miss an important part. Actually this route and its assets is an excellent test bed for this case. I needed to add this route by the following line:
`"C:\packages\DemoModel1.zip\Demo Model 1\" /MSTS/`
Please note the additional directory part after the "DemoMode1.zip". This is inevitable, because this package did not zippped the contents of the "C:\whatever\Microsoft Train Simulator\" folder, but added one more directory level, the "\Demo Model 1\", which needs to be ruled out from our virtual directory structure.

Could you double-check if your vfs config set up like this?

cjakeman:
> Hi Chris, @cjakeman, My bad, I should have thought of testing it with our own test asset. :-) 

Not your mistake, Péter, but mine. I didn;t start from your branch but taook a diff and thought that was from after efffead was committed. Now I have looked at the changes that commit brings, I can see I was hasty. I've taken a fresh diff and that works just fine.

Well done with that commit and sorry to have caused you unnecessary work.

pzgulyas:
No problem Chris, glad you sorted this out!

pzgulyas:
Quite a big change: I modified the code to use the dotnet builtin ZipArchive class to handle the zip format archives, while leaving SharpCompress to handle the others (rar). (BTW we will need to decide if we want to support other formats than zip at all.)

cjakeman (commented):
Hi Peter,

I've set your PR status back to "draft" and add the label "not for unstable" as it appears to be breaking the build of the Unstable Version. I've written about it on the forum at 

Do you agree with my diagnosis?

How do you want to proceed?

Best wishes,

Chris

pzgulyas:
Hi Chris,

The AppVeyor builds is correctly, and the problem is only on the other server, so I need to go with trial-and-error method, which is impossible if the not-for-unstable label is set. So I added some commits, and deleted the label you added to see, if it now builds.

If not, then tomorrow please add the label back, because I will be travelling, and I will be available only in the evening.

cjakeman:
Hi Peter,

> The AppVeyor builds it correctly, and the problem is only on the other server, so I need to go with trial-and-error method, which is impossible if the not-for-unstable label is set. So I added some commits, and deleted the label you added to see, if it now builds.

All seems good now. Many thanks..

cjakeman (commented):
Hi Peter,

As requested, James will be reviewing this valuable contribution from an architectural viewpoint.

In the meantime, may I make some other comments?

Deployment
It's great that this PR can be published to users without them needing to change anything at all. Their current installation just works and they can make use of VFS or not as they wish.

Archive Formats
I don't think it's a good idea to support formats other than the Zip format.
1. Having a single format encourages content publishers to use it. That makes life easier for everyone, publishers, users and code maintainers.
2. The Zip format is on every Windows PC and, I think, does all that we need.

User Errors
It would be good if all possible *user* errors result in a helpful message in the log. I've not checked this yet.

User Interface
I guess you are anticipating a user interface which fits on top of this initial work, so that users cannot create invalid VFS files.
You will remember that our standard for new file types is JSON, so I am hoping the user interface (whoever builds it) will give us a single JSON file to replace the collection of VFS files that this version supports.

Best wishes,

Chris

pzgulyas:
Hi Chris,

Thank you guys for doing the review of my PR. I think this is too fundamental – with taking over almost all file reads as a layer of indirection – for being possibly architectually wrong, so it would be reassuring to know this is all right.

Deployment
I hoped that it will work for most of the people almost without noticing the change, and indeed there was only a problem for guys doing some development (software or asset) with the zip automounting, that appeared to mount their intermediate-temporary archives left in the directories. I think end users will not have such a problem, so I would like not to disable the automounting function, since that makes deploying the asset packages much easier.

Archive Formats
I have no objections about leaving only the zip format as the only one supported. I have added rar only because sometime in the past 10-20 years of MSTS history rar was so popular in distributing content – at least in my country – that I have quite a few of them on my disk, and I thought it would be convenient not to repackage them. But for the future I agree, that contetnt publishers should not be encouraged to use that. Zip can be supported without any 3rd party libraries getting involved, MS dotnet ones have everything needed for that.

User Errors
I tried to write as much information into log as possible. I added a three-level log system, even the individual file accesses can be logged in case of someone needed that. If you guys can think of anything left out, I am happy to add that.

User Interface
Json is a viable alternative to the current format, I have no objections either. Just keep in mind, the content of the config file is just string-pairs in a definite order, no other structure is needed (at least yet), so introducing json here would make our config file more complex than the unix /etc/fstab. :-)

Cheers,

Peter

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+        ///
+        /// SystemPath - may use "/" or "\", but MountPoint may use "/" only as a directory separator.
+        ///            - with spaces in it must be quoted, e.g. "C:\My Routes\USA85\" /MSTS/ROUTES/USA85/
+        ///            - referring to a directory must end with "/" or "\" to assure that only _below_ dirs and files will be visible _below_ the mount point.
+        ///            - referring to an archive must not end with "/" or "\", e.g. C:\TEMP\MSTS1.2.zip /MSTS/
+        ///            - may refer to an archive subdirectory, e.g.: C:\routes.zip\USA3\ /MSTS/ROUTES/USA3/
+        ///            - may _not_ refer to a single non-archive file, neither a non-archive file within an archive.
+        ///            - is case-insensitive.
+        /// MountPoint - must end with "/" to avoid confusion and to assure that mounting will be done _below_ the given path.
+        ///            - may not contain spaces, may not be quoted.
```
twpol, line 63: There are many problems with using a plain text file for this (and a file at all), but this line is particularly egregious to me. Why can't we have spaces in the mounted paths?

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+        /// The exception is the Path.GetFullPath(), because that adds an unwanted drive letter,
+        /// so must never be used on virtual paths.
+        ///
+        ///////////////////////////////////////////////////////////////////////////////////////
+        ///
+        /// Public interfaces
+        /// 
+
+        public const string MstsBasePath = "/MSTS/";
+        public const string ExecutablePath = "/EXECUTABLE/";
```
twpol, line 80: What is this for? I'm feeling like we've skipped the _design_ phase with the architecture of the VFS here.

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+        {
+            switch (PrepareForWrite(vfsPath))
+            {
+                case Stream stream: return stream;
+                case string path: return new FileStream(path, FileMode.Create);
+                default: throw new DirectoryNotFoundException($"VFS writing failed: {vfsPath}");
+            }
+        }
+
+        public static void FileDelete(string vfsPath)
```
twpol, line 212: Do not implement this. We never delete content files.

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+
+        static bool Attach(string line)
+        {
+            Match match;
+            string message;
+            if (string.IsNullOrWhiteSpace(line))
+                return true; // Blank lines are accepted silently.
+            if (line.Trim().StartsWith("#")) // commented out line
+                return true; // Comment lines are accepted silently.
+            if (!(match = Regex.Match(line, @"^("".+""|\S+) +(/MSTS|/OR)([\S/]+)+(\s+#.*)?", RegexOptions.IgnoreCase)).Success)
```
twpol, line 301: And this is why a text file is problematic, yuk :)

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+                if (LogLevel > 2)
+                    Trace.TraceInformation($"VFS reading system file: {foundNode.AbsolutePath} => {vfsPath}");
+                return foundNode.AbsolutePath;
+            }
+            if (LogLevel > 2)
+                Trace.TraceInformation($"VFS reading failed: null => {vfsPath}");
+            return null;
+        }
+
+        static object PrepareForWrite(string vfsPath)
```
twpol, line 535: What are we doing with this? We do not write to the content areas, **ever**.

Source/RunActivity/Viewer3D/Processes/GameStateRunActivity.cs
```diff
@@ -104,6 +104,11 @@ internal override void Update(RenderFrame frame, double totalRealSeconds)
 
         internal override void Load()
         {
+            // The virtual file system must be initialized before loading anything.
+            Vfs.LogLevel = Game.Settings.VfsLogLevel;
+            Vfs.NoAutoMount = Game.Settings.VfsNoAutoMount;
+            Vfs.Initialize(Game.Settings.Menu_Selection[0], Path.GetDirectoryName(System.Windows.Forms.Application.ExecutablePath));
```
twpol, line 110: This will fail if you run `RunActivity` directly, as developers and tester do. You **must** pass all information needed to load the correct content via the command line.

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+                    if (IsArchiveSupported(file))
+                    {
+                        if (NoAutoMount)
+                        {
+                            message = $"VFS skipped auto-mounting archive by user settings: {file}";
+                            Trace.TraceInformation(message);
+                            InitLog.Enqueue(message);
+                        }
+                        else
+                            MountArchive(file, null, vfsNode.GetVfsPath());
```
twpol, line 423: Hang on, is this going to mount every zip file in a content folder **recursively**? That's not a thing we want to happen by default, and possibly at all

Source/ORTS.Settings/UserSettings.cs
```diff
@@ -309,6 +309,10 @@ public enum DirectXFeature
         public bool PerformanceTuner { get; set; }
         [Default(true)]
         public bool SuppressShapeWarnings { get; set; }
+        [Default(1)]
+        public int VfsLogLevel { get; set; }
+        [Default(false)]
+        public bool VfsNoAutoMount { get; set; }
```
twpol, line 315: In addition to removing the negative in the name, please can we have this default to **not** auto-mounting?

We want the first version of this code to behave as close to the old code as possible by default, but allow for experimentation with the additional capabilities the VFS offers.

Source/ORTS.Common/VirtualFileSystem.cs
```diff
@@ -0,0 +1,717 @@
...
+    }
+
+    internal class VfsNode
+    {
+        readonly Dictionary<string, VfsNode> Children = new Dictionary<string, VfsNode>();
+        readonly VfsNode Parent;
+        readonly string Name;
+
+        public string AbsolutePath { get; private set; }
+        public string SubPath { get; }
```
twpol, line 614: We could really do with some documentation on what the difference here is

twpol (changes requested):
Overall, this is a fantastic piece of work @pzgulyas!

However, I have some concerns with how it currently behaves:

- Auto-mount should default to off
- All modification actions (write/delete) must be removed, they are completely banned on content files/directories; any code that needs these must be fixed or removed separately
- The plain text file is a poor choice IMHO and a burden I do not want to go forward with; I would be happy to go forward with no VFS configuration initially (so only mounting directories with or without archives)
- Switching to my main MSTS installation profile in the menu has slowed from four seconds to sixty seconds! We should figure out if we can selectively scan for VFS files or otherwise improve this

Source/Menu/Options.Designer.cs
```diff
@@ -2567,6 +2620,16 @@ private void InitializeComponent()
...
             this.checkPreferDDSTexture.UseVisualStyleBackColor = true;
             // 
+            // checkVfsNoAutoMount
+            // 
+            this.checkVfsNoAutoMount.AutoSize = true;
+            this.checkVfsNoAutoMount.Location = new System.Drawing.Point(6, 212);
+            this.checkVfsNoAutoMount.Name = "checkVfsNoAutoMount";
+            this.checkVfsNoAutoMount.Size = new System.Drawing.Size(256, 17);
+            this.checkVfsNoAutoMount.TabIndex = 21;
+            this.checkVfsNoAutoMount.Text = "Disable virtual file system auto-mounting archives";
```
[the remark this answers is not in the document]
pzgulyas: Done.

Source/ORTS.Common/packages.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="SharpCompress" version="0.30.1" targetFramework="net472" />
```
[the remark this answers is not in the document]
pzgulyas: Done, removed the rar support.

Source/ORTS.Common/VirtualFileSystem.cs
pzgulyas: Nothing special, just didn't want to further complicate the parsing regex, and it seemed not be inevitable to have spaces there.

Source/ORTS.Common/VirtualFileSystem.cs
pzgulyas: Need to mount the executable "Content" path for being able to access the stored images there. All texture loading got transferred to be done via the VFS, and I didn't want to add a SharedTextureManager.GetNonVfsImage()-like method.

Source/ORTS.Common/VirtualFileSystem.cs
pzgulyas: We do. The generated *.or-binpat files need to be deleted, and they currently reside in the ROUTES\route\PATHS\ folder.

pzgulyas:
> * All modification actions (write/delete) must be removed, they are completely banned on content files/directories; any code that needs these must be fixed or removed separately

This might really be a valid argument, but currently the *.or-binpat files are written to the content directory, and had to choose to either break that functionality, or implement writing. I've chosen the latter. I think changing that is not in the scope of the VFS.

> * The plain text file is a poor choice IMHO and a burden I do not want to go forward with; I would be happy to go forward with no VFS configuration initially (so only mounting directories with or without archives)

Please note that the configuration file is optional only, one can switch on automount and go with no configuration file. But in cases like the Openrails's downloadable DemoModel1.zip it is necessary, because it is not zipped to have the MSTS root-level folders into the zip's own root-level. Some means are necessary in such cases for binding the right archive directory to the right VFS directory.

I don't adhere to the plain text file, however besides its contras it has pros too. For example, because the configuration need to contain only pairs of strings, a json structure is actually more than what is needed here. I had the unix /etc/fstab in my mind, but of course, json didn't yet exist when that got invented. An example json structure could look like this:

`
{
  "vfs": [
    {
      "source": "C:/something/MSTS_Base.zip/Train Simulator/",
      "mountPoint": "/MSTS/"
    },
    {
      "source": "C:/something/DemoModel1.zip/Demo Model 1/",
      "mountPoint": "/MSTS/"
    }
  ]
}`

> * Switching to my main MSTS installation profile in the menu has slowed from four seconds to sixty seconds! We should figure out if we can selectively scan for VFS files or otherwise improve this

Interesting point, I didn't experience any comparable slowdown, but your concern might be valid. But I don't have an idea currently how to solve this exactly.

Source/ORTS.Common/VirtualFileSystem.cs
pzgulyas: My idea was not only to support future well-defined and well-built zip packages, but also the 20 years log inheritance we have. To me it was convenient to just drom a downloaded zip into subdirectory, and get that auto-mounted. The only case when this cannot be done, is when the zip has some inside fantasy-named subfolder that I didn't want to use. In that case registering that into the vfs-config file is the only choice. In short, the purpose of doing it recursively was to be able to auto-mount a zip that is "too short" in paths inside.

Source/ORTS.Common/VirtualFileSystem.cs
twpol: I would argue that it is always inevitable to have spaces in a path/filename :)

Source/ORTS.Common/VirtualFileSystem.cs
twpol: Okay, that code is going to have to be fixed before we can proceed with VFS :(

Source/ORTS.Common/VirtualFileSystem.cs
twpol: Okay, so I'm not against recursively looking for ZIPs to mount, but it must be opt-in IMHO

twpol:
> This might really be a valid argument, but currently the *.or-binpat files are written to the content directory, and had to choose to either break that functionality, or implement writing. I've chosen the latter. I think changing that is not in the scope of the VFS.

Yeah, I don't expect you to fix that here. I will get the 7-year-old (!) code which uses these files fixed.

It has been a foundational principal of Open Rails that it never modifies your content and, while I think these files skirt around that a bit by being cache files, I am quite sure most people would be surprised that we're doing it. :)

> I don't adhere to the plain text file, however besides its contras it has pros too. For example, because the configuration need to contain only pairs of strings, a json structure is actually more than what is needed here. I had the unix /etc/fstab in my mind, but of course, json didn't yet exist when that got invented. An example json structure could look like this:

It's a combination of:

1. We have agreed as a project to always use JSON for new (non-binary) file formats
2. Plain text requires you to write a parser which is where all the bugs slip in :)

The update code with a JSON file looks better, thanks for that

> Interesting point, I didn't experience any comparable slowdown, but your concern might be valid. But I don't have an idea currently how to solve this exactly.

I've had a look, and my main MSTS install (which is where A LOT of content has been installed) has the following properties compared to a clean MSTS install (with all base routes):

- 75.7GB vs 1.6GB
- 880k vs 20k files
- 17k vs 1k directories
- 75 ZIP files not in the root (0 in the root)

If I understood the code right, and you did scan recursively through for ZIP files to mount, the VFS would have had to scan through all those directories and then try to mount all those ZIPs - it takes `dir /s > nul` a full 25 seconds on a warmed-up SSD :)

And at this point, I'm not even sure mounting ZIPs anywhere without opt-in is a good plan, but let's at least cut out the recursive part to start with :)

twpol (changes requested):
Thanks very much for the updates, Peter, we're getting closer!

With PR #647 merged, we can now remove all the write/delete-related VFS code

My biggest concern is still that this is doing _a lot_; and consequently, this PR is too big for me to comfortably review (I know I've missed a load of things)

The aim of the first version was always to implement the VFS with no new features and no visible impact to users; hopefully, the auto-mount off will make the impact minimal (but it still scans 100% of files to initialise the VFS which people **will** notice), but the UI changes to allow for the configuration file are above and beyond what I'm comfortable with right now

I really hope that this doesn't deter you too much; even if we add back all of the things I'd like removed from the PR, splitting them to out - particularly to allow for discussion - will benefit everyone

pzgulyas:
> Thanks very much for the updates, Peter, we're getting closer!
> 
> With PR #647 merged, we can now remove all the write/delete-related VFS code
> 
> My biggest concern is still that this is doing _a lot_; and consequently, this PR is too big for me to comfortably review (I know I've missed a load of things)
> 
> The aim of the first version was always to implement the VFS with no new features and no visible impact to users; hopefully, the auto-mount off will make the impact minimal (but it still scans 100% of files to initialise the VFS which people **will** notice), but the UI changes to allow for the configuration file are above and beyond what I'm comfortable with right now
> 
Thanks James for the fix you've committed! I couldn't yet take the chance to check that out, and I will be unavailable for the weekend, so I am just answering shortly now, but I will answer all the other comments in detail later this month.

To me it looks like you biggest two concerns are the configuration file and the automounter. Shortly speaking, I've added these into my initial development to serve the same purpose: to support existing non-precisely packaged content. 

Take for example the following two downloadable files:

1) The "Demo Model 1" from our own website: http://openrails.org/download/content/
2) I picked the first ET locomotive download by searching files with keyword "Open Rails": http://www.elvastower.com/forums/index.php?/files/file/3117-bo-4-4-0-no270-mason/

As you can see, neither of them can be mounted into the VFS without either using a config file or the automounter.

(1) is – lets say – "too long" in its packaged path. There must be a way to say to ORTS to "omit the first directory level, and mount it as if was packaged without the \Demo Model 1\ directory". This can be achieved by a configuration file.

(2) is "too short". It must be mounted under /MSTS/TRAINS/TRAINSET/. This can be achieved either by a config file or by the automounter (without config!).

So the automounter is one way to get rid of the config file, but it can help only in cases the packaged path is "too short". For the "too long" cases only the config file can be the solution. You are repeatedly argueing against the config file and the automounter, but never explained a possible replacement solution for these problems. What would be your suggestion? (I can tell in advance, that a solution like "do not use mispackaged or before-this-date content packages" is not the thing that was in my mind when I developed this feature, but rather is exactly the opposite.)

twpol:
> You are repeatedly arguing against the config file and the automounter, but never explained a possible replacement solution for these problems. What would be your suggestion?

For the automounter, it is an acceptable feature to include, but its performance is not (https://github.com/openrails/openrails/pull/541#pullrequestreview-847281027), so we must either reduce its functionality or turn it off by default - which you've done, and that's good with me :)

For the configuration file, I have only requested that we do not include it **in this version**; I wanted/expected the VFS to be implemented in smaller steps than you've implemented here, and since we have not actually discussed anything about the configuration file, it is premature to include it IMHO

I have spent far too much of my involvement with Open Rails untangling the messes from prematurely implemented features, so I am extremely wary of things like the configuration file as a result - it needs to be discussed on the forums and agreed upon before code is written, or we might be stuck with something bad for years and years

pzgulyas:
Honestly, I don't like the idea of stripping the configurability at all, because without that most of the existing downloadable packages will be unusable. Also the VFS cannot be demonstrated with either of the examples (1) or (2) I've mentioned above. Actually your answer didn't address my concerns stated above about removing the configuration file. But I'm not against delaying the inclusion of the VFS into the official version until we all feel comfortable with how it works, it is a bad feeling to be stuck with something for years with premature functions.

twpol:
> Honestly, I don't like the idea of stripping the configurability at all, because without that most of the existing downloadable packages will be unusable.

If you don't want to remove the configuration part of this PR, we had better start discussing the design and format in the forum :)

PR #545, opened 2021-12-10
Fix VRAM usage

Sharpe49:
https://bugs.launchpad.net/or/+bug/1954541
http://www.elvastower.com/forums/index.php?/topic/35644-memory-using-monogame-or/

Please tell me in this pull request if you see textures not being rendered because that means that they are not yet marked in their renderer.


twpol (changes requested):
I'd like to request not merging this until I've had a chance to review the forum thread and confirmed why we need this change (please keep all discussion there). In particular, whether we need to include it in a new 1.4 release.

twpol:
> I'd like to request not merging this until I've had a chance to review the forum thread and confirmed why we need this change (please keep all discussion there). In particular, whether we need to include it in a new 1.4 release.

Oh, I did not see this is still a draft - that's useful, so thanks for that!

Sharpe49:
As proposed on the forum, this pull request is no longer a draft.

twpol (approved):

PR #550, opened 2021-12-17
Adjust diesel fuel consumption

peternewell:
Adjust fuel consumption so that it will operate whilever the diesel engine is running.

https://bugs.launchpad.net/or/+bug/1953251

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
@@ -636,18 +636,6 @@ namespace Orts.Simulation.RollingStocks
                         TractiveForceN = 0;
                 }
 
-                DieselFlowLps = DieselEngines.DieselFlowLps;
-                partialFuelConsumption += DieselEngines.DieselFlowLps * elapsedClockSeconds;
-                if (partialFuelConsumption >= 0.1)
-                {
-                    DieselLevelL -= partialFuelConsumption;
-                    partialFuelConsumption = 0;
-                }
-                if (DieselLevelL <= 0.0f)
-                {
-                    SignalEvent(Event.EnginePowerOff);
-                    DieselEngines.HandleEvent(PowerSupplyEvent.StopEngine);
-                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -1085,7 +1085,8 @@ public void Update(float elapsedClockSeconds)
             if ((State != DieselEngineState.Starting) && (RealRPM == 0f))
                 State = DieselEngineState.Stopped;
 
-            if ((State == DieselEngineState.Stopped) || (State == DieselEngineState.Stopping) || ((State == DieselEngineState.Starting) && (RealRPM < StartingRPM)))
+            // fuel consumption will occur when engine is running above the starting rpm
+            if ((State == DieselEngineState.Stopped) || ((State == DieselEngineState.Stopping) && (RealRPM < StartingRPM)) || ((State == DieselEngineState.Starting) && (RealRPM < StartingRPM)))
```
pzgulyas, line 1089: && takes precedence over ||, and == takes precedence over both of these, so brackets are totally unnecessary in this line except the if()'s brackets. See: [Operator precedence](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/#operator-precedence)

pzgulyas (approved):

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
peternewell: Whilst I agree that it is possible to remove the brackets, sometimes I prefer a long hand approach as I believe that it makes the code more readable for all.

So I have removed some of the brackets and left a couple on this basis.

Thanks for highlighting this.

Could you please approve this PR now.

pzgulyas (approved):
No question I give my thubms-up to you!

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
```diff
@@ -662,6 +650,24 @@ protected override void UpdateTractiveForce(float elapsedClockSeconds, float t,
                     w = 0;
                 AverageForceN = w * AverageForceN + (1 - w) * TractiveForceN;
             }
+
+            // Calculate fuel consumption will occur unless diesel engine is stopped
+            if (LocomotivePowerSupply.MainPowerSupplyOn)
```
Sharpe49, line 655: Main power supply on means that power is available for traction, not that the engine is running.
You should use diesel engine state.

Sharpe49 (changes requested):

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
peternewell: Ok, I have removed this condition as the diesel engine state should manage diesel fuel flow.

Sharpe49: The functionnality is correct, but the indentation needs to be fixed.

peternewell: Indentation fixed

Sharpe49 (approved):

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
Sharpe49: Thank you. It's approved.

peternewell:
Thanks for that.

PR #552, opened 2021-12-18
Fix out of range index when changing cabs

cesarBLG:
https://bugs.launchpad.net/or/+bug/1955294

See http://www.elvastower.com/forums/index.php?/topic/35636-3d-cab-change-error/


pzgulyas (approved):
This looks good to me.

cesarBLG:
Thanks, Peter.

PR #553, opened 2021-12-21
ScriptManager migration to the CodeAnalysis class. This is in preparation for the net 6 migration.

pzgulyas:
@Sharpe49 Preliminary ScriptManager migration. LoadFolder() is commented out yet. Could you please try it on .net5? I don't have a working tree of that currently.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
+++ b/Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
@@ -41,6 +41,7 @@ namespace Orts.Common.Scripting
             typeof(System.Diagnostics.Debug).GetTypeInfo().Assembly.Location,
             typeof(ORTS.Common.ElapsedTime).GetTypeInfo().Assembly.Location,
             typeof(ORTS.Scripting.Api.Timer).GetTypeInfo().Assembly.Location,
+            typeof(System.Linq.Enumerable).GetTypeInfo().Assembly.Location,
         };
         static MetadataReference[] References = ReferenceAssemblies.Select(r => MetadataReference.CreateFromFile(r)).ToArray();
         static CSharpCompilationOptions CompilationOptions = new CSharpCompilationOptions(
@@ -61,7 +62,7 @@ namespace Orts.Common.Scripting
             if (pathArray == null || pathArray.Length == 0 || name == null || name == "")
                 return null;
 
-            if (Path.GetExtension(name) != ".cs")
+            if (Path.GetExtension(name).ToLower() != ".cs")
                 name += ".cs";
 
             var path = ORTSPaths.GetFileFromFolders(pathArray, name);
@@ -74,20 +75,19 @@ namespace Orts.Common.Scripting
             var type = String.Format("{0}.{1}", nameSpace, Path.GetFileNameWithoutExtension(path).Replace('-', '_'));
 
             if (!Scripts.ContainsKey(path))
-                Scripts[path] = CompileScript(path);
+                Scripts[path] = CompileScript(new string[] { path });
             return Scripts[path]?.CreateInstance(type, true);
         }
 
-        private static Assembly CompileScript(string path)
+        private static Assembly CompileScript(string[] path)
         {
             try
             {
-                var scriptName = Path.GetFileName(path);
-                var scriptCode = File.ReadAllText(path);
-                var syntaxTree = CSharpSyntaxTree.ParseText(scriptCode);
+                var scriptName = Path.GetFileName(path[0]);
+                var syntaxTrees = path.Select(file => CSharpSyntaxTree.ParseText(File.ReadAllText(file)));
                 var compilation = CSharpCompilation.Create(
                     scriptName,
-                    new[] { syntaxTree },
+                    syntaxTrees,
                     References,
                     CompilationOptions);
                 var ms = new MemoryStream();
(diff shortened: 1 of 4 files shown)


cesarBLG:
I created the LoadFolder() function for Signal Scripts so that all of them were compiled at once, thus allowing the share of common enums and such. I tested your code, and it is straightforward to get it working for signals too, removing the code duplication by the way. It's enought to create a syntax tree for each file:
```diff
             if (!Scripts.ContainsKey(path))
-                Scripts[path] = CompileScript(path);
+                Scripts[path] = CompileScript(new string[] { path });
             return Scripts[path]?.CreateInstance(type, true);
         }
 
-        private static Assembly CompileScript(string path)
+        public static Assembly CompileScript(string[] path)
         {
             try
             {
-                var scriptName = Path.GetFileName(path);
-                var scriptCode = File.ReadAllText(path);
-                var syntaxTree = CSharpSyntaxTree.ParseText(scriptCode);
+                var scriptName = Path.GetFileName(path[0]);
+                var syntaxTrees = path.Select(file => CSharpSyntaxTree.ParseText(File.ReadAllText(file)));
                 var compilation = CSharpCompilation.Create(
                     scriptName,
-                    new[] { syntaxTree },
+                    syntaxTrees,
                     References,
```
I'm experiencing a problem with the referenced assemblies though. I had to add the following line
```csharp
            typeof(System.Linq.Enumerable).Assembly.Location,
```
to the list of referenced assemblies because I was getting compile errors due to some methods for dictionaries and lists not being found.

pzgulyas:
Thank you, @cesarBLG, I've edded a commit with the modified code.

Sharpe49:
I don't have a .Net 5 build currently.

Source/Orts.Simulation/Orts.Simulation.csproj
```diff
@@ -54,6 +61,31 @@
...
+    </Reference>
+    <Reference Include="System.Runtime.CompilerServices.Unsafe, Version=5.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.Runtime.CompilerServices.Unsafe.5.0.0\lib\net45\System.Runtime.CompilerServices.Unsafe.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Text.Encoding.CodePages, Version=4.1.1.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.Text.Encoding.CodePages.4.5.1\lib\net461\System.Text.Encoding.CodePages.dll</HintPath>
+    </Reference>
+    <Reference Include="System.Threading.Tasks.Extensions, Version=4.2.0.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51, processorArchitecture=MSIL">
+      <HintPath>..\packages\System.Threading.Tasks.Extensions.4.5.4\lib\net461\System.Threading.Tasks.Extensions.dll</HintPath>
+    </Reference>
```
Sharpe49, line 88: Are those assembly references really needed?

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -34,22 +35,18 @@ public class ScriptManager
...
-            cp.ReferencedAssemblies.Add("System.Core.dll");
-            cp.ReferencedAssemblies.Add("ORTS.Common.dll");
-            cp.ReferencedAssemblies.Add("Orts.Simulation.dll");
-            return cp;
-        }
+            typeof(System.Object).GetTypeInfo().Assembly.Location,
+            typeof(System.Diagnostics.Debug).GetTypeInfo().Assembly.Location,
+            typeof(ORTS.Common.ElapsedTime).GetTypeInfo().Assembly.Location,
+            typeof(ORTS.Scripting.Api.Timer).GetTypeInfo().Assembly.Location,
+            typeof(System.Linq.Enumerable).GetTypeInfo().Assembly.Location,
```
Sharpe49, line 44: GetTypeInfo() doesn't seem to be needed: the Type object returned by typeof() already has an Assembly property.

Sharpe49 (commented):
It works properly with the scripts I use.
I don't really like when "var" is used... it makes it difficult to know rapidly which type are objects made of.

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
```diff
@@ -34,22 +35,18 @@ public class ScriptManager
     {
         readonly Simulator Simulator;
```
Sharpe49, line 36: The Simulator variable doesn't seem to be used.

Source/Orts.Simulation/Orts.Simulation.csproj
pzgulyas: They got added by the VS automatically together with the CodeAnalysis assembly.

pzgulyas:
> I don't really like when "var" is used... it makes it difficult to know rapidly which type are objects made of.

I find the code with using "var" far more readable when it comes to understanding the algorithm. Unnecessary verbosity just distracts the attention. And I can always hover the mouse cursor over the "var" word when I need to know the exact type (but it is rarely needed).

Source/Orts.Simulation/Common/Scripting/ScriptManager.cs
pzgulyas: Agreed, keep it simple.

Source/Orts.Simulation/Orts.Simulation.csproj
Sharpe49: That's weird because, when I merged it in my work branch (with .Net SDK project files), I only had to add Microsoft.CodeAnalysis.CSharp to the references.

pzgulyas: I think that's fine, if works. Probably the rest of the references are added silently by some background process in that case. For some reason VS felt it must mention them explicitly here, don't know why.

cesarBLG:
The Roslyn package is not needed anymore, is it?

Sharpe49:
You mean the Microsoft.CodeDom.Providers.DotNetCompilerPlatform package from the ASP.Net team?
No, it's no longer needed.

Sharpe49:
I added the pull request to my work branch. I have an issue with the script debug. It doesn't seem to work with this pull request. We probably don't have the debug symbols of the compiled scripts.

pzgulyas:
> I added the pull request to my work branch. I had an issue with the script debug. It doesn't seem to work with this pull request. The debug symbols of the compiled scripts are not loaded
> 
> Here's my version of ScriptManager with debug symbols loading: https://gist.github.com/Sharpe49/93a7904ac2d0771e193e0e28ed3f0846

pzgulyas:
Opps, accident

PR #554, opened 2021-12-22
Bug fix for Crash when switching cabs and changing screen page  https://bugs.launchpad.net/or/+bug/1955575

Csantucci:
After a cab change cabrenderers are not yet alive at the moment that the command receivers are set.
Changed command receiver and having also a better selection between 2D and 3D cabrenderer as command receiver.


Sharpe49 (approved):
Seems good to me.

Csantucci:
Thanks for approval!

PR #555, opened 2021-12-22
Bug fix for crash with diesel locos with more than an engine, https://bugs.launchpad.net/or/+bug/1955582

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/35811-mg-1121-crashes-on-loading-advanced-deltic-from-ctn/

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
@@ -968,15 +968,27 @@ namespace Orts.Simulation.RollingStocks
             // Throttle
             status.AppendFormat("{0}\t", throttle);
             // Load
+            var avgLoadPercent = 0f;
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
```diff
@@ -968,15 +968,27 @@ public string GetDPDebugStatus()
             // Throttle
             status.AppendFormat("{0}\t", throttle);
             // Load
+            var avgLoadPercent = 0f;
+            var totalLoadW = DieselEngines.NumOfActiveEngines > 0 ? LocomotivePowerSupply.ElectricTrainSupplyPowerW : 0f;
+            var totalAvailablePowerW = 0f;
             foreach (var eng in DieselEngines.DEList)
-                status.AppendFormat("{0:F1}%\t", eng.LoadPercent);
+            {
+                totalLoadW += eng.CurrentDieselOutputPowerW <= 0f ? 0f : eng.OutputPowerW;
```
Sharpe49, line 976: In the same line, you use CurrentDieselOutputPowerW and OutputPowerW.
Is that normal?

PR #556, opened 2021-12-24
Improve memory statistics for CPU and GPU

twpol:


cjakeman (commented):
Hi James,

Nice to see this.

Will you be updating the section in the Manual on HUD Display to match? (I wonder if that should go into a separate chapter on Debug Tools?)

Best wishes,

Chris

twpol:
> Will you be updating the section in the Manual on HUD Display to match? (I wonder if that should go into a separate chapter on Debug Tools?)
The section was quite bare, so I've expanded it to explain all the values and included an example of how you might check performance

cjakeman:
> The section was quite bare, so I've expanded it to explain all the values and included an example of how you might check performance

I can see how very bare it was from the new material you've added. Good job.

cjakeman (approved):
HI James,

Approved with thanks,

Chris

PR #560, opened 2021-12-28
https://blueprints.launchpad.net/or/+spec/label-for-cab-controls , by Jindrich

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/34855-label-feature-for-cabin-controls/page__view__findpost__p__268341

### DIFF
--- a/Source/Orts.Formats.Msts/CabViewFile.cs
+++ b/Source/Orts.Formats.Msts/CabViewFile.cs
@@ -397,6 +397,7 @@ namespace Orts.Formats.Msts
         
         public double OldValue;
         public string ACEFile = "";
+        public string Label = "";
 
         public int Display;
         public List<string> Screens;
@@ -598,6 +599,11 @@ namespace Orts.Formats.Msts
                     ToDegree = stf.ReadFloat(STFReader.UNITS.None, null);
                     stf.SkipRestOfBlock();
                 }),
+                new STFReader.TokenProcessor("label", ()=>{
+                    stf.MustMatch("(");
+                    Label = stf.ReadString();
+                    stf.SkipRestOfBlock();
+                }),
                 new STFReader.TokenProcessor("ortsdisplay", ()=>{ParseDisplay(stf); }),
                 new STFReader.TokenProcessor("ortsscreenpage", () => {ParseScreen(stf); }),
                 new STFReader.TokenProcessor("ortscabviewpoint", ()=>{ParseCabViewpoint(stf); }),
@@ -687,6 +693,11 @@ namespace Orts.Formats.Msts
                     }
                 }),
                 new STFReader.TokenProcessor("ortsangle", () =>{ Rotation = ParseRotation(stf); }),
+                new STFReader.TokenProcessor("label", ()=>{
+                    stf.MustMatch("(");
+                    Label = stf.ReadString();
+                    stf.SkipRestOfBlock();
+                }),
                 new STFReader.TokenProcessor("ortsdisplay", ()=>{ParseDisplay(stf); }),
                 new STFReader.TokenProcessor("ortsscreenpage", () => {ParseScreen(stf); }),
                 new STFReader.TokenProcessor("ortscabviewpoint", ()=>{ParseCabViewpoint(stf); }),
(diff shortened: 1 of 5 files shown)


Source/Orts.Formats.Msts/CabViewFile.cs
```diff
@@ -598,6 +599,11 @@ public CVCDial(STFReader stf, string basepath)
                     ToDegree = stf.ReadFloat(STFReader.UNITS.None, null);
                     stf.SkipRestOfBlock();
                 }),
+                new STFReader.TokenProcessor("label", ()=>{
```
Sharpe49, line 602: The parameter should be called ortslabel since it does not exist in MSTS.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -1560,6 +1560,7 @@ public interface ICabViewMouseControlRenderer
         bool IsMouseWithin();
         void HandleUserInput();
         string GetControlName();
+        string GetControlLabel();
```
Sharpe49, line 1563: You should use a property with a getter in the interface instead of a getter method.

Sharpe49 (changes requested):

Source/Orts.Formats.Msts/CabViewFile.cs
Csantucci: There are recently added fields without the leading ORTS, like disablediflowvoltagepowersupplyoff and disabledifcabpowersupplyoff.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
Csantucci: What is the advantage? There is a GetControlName() since years above it.

Source/Orts.Formats.Msts/CabViewFile.cs
Sharpe49: Yeah... that might not have been a good idea.
In your case, you have other parameters around that are called ORTS<something> so it would make sense to use ORTS here.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
Sharpe49: Using the getter syntax makes code simpler.
And repeating what is done in the old code doesn't make it better. It will probably be refactored later.

Source/Orts.Formats.Msts/CabViewFile.cs
Csantucci: The reason why I didn't add "ORTS" is that this parameter was created as "Label" by Jindrich for ORNYMG, and there surely is content using it. If I switch to ORTSLabel I don't have a problem for ORNYMG, because I can make ORNYMG accepting both namings, but some existing content wouldn't show the label under the official OR versions.

cjakeman: > The reason why I didn't add "ORTS" is that this parameter was created as "Label" by Jindrich for ORNYMG, and there surely is content using it. If I switch to ORTSLabel some existing content wouldn't show the label under the official OR versions.

So this is a conflict between 
- tidy code and documentation for official Open Rails
and
- content written and shared by early adopters using ORNYMG not showing correctly. 

I'm with Sharpe49 on this - Official Open Rails should aim for consistency, so content written for ORNYMG will have to change to work on Open Rails.

It will help, Carlo, if you go ahead and change ORNYMG to accommodate both forms as early adopters can run Open Rails-compatible content if they want to.

Csantucci: OK, I've renamed "Label" to "ORTSLabel".

Sharpe49: Ok, thank you.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
Csantucci: Getter replaed with property now.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2189,6 +2189,8 @@ public string GetControlName()
             return (Locomotive as MSTSLocomotive).TrainControlSystem.GetDisplayString(GetControlType().ToString());
         }
 
+        public string ControlLabel { get { return Control.Label; } }
```
Sharpe49, line 2192:
```suggestion
        public string ControlLabel => Control.Label;
```
You can also write it like that when you only have a getter.

Source/RunActivity/Viewer3D/RollingStock/SubSystems/ETCS/DriverMachineInterface.cs
```diff
@@ -856,7 +856,7 @@ public string GetControlName()
             }
             return "";
         }
-
+        public string ControlLabel { get { return GetControlName(); } }
```
Sharpe49, line 859:
```suggestion
        public string ControlLabel => GetControlName();
```

Csantucci: OK, done.

Sharpe49 (approved):

Csantucci:
Thank you. The approval of the blueprint is still missing...

PR #561, opened 2021-12-28
Manual call-on is reset when passing an advanced signal - https://bugs.launchpad.net/or/+bug/1955907

cesarBLG:
When the train is allowed to "call on" from the dispatcher window, but the signal is not the first one for the train, the CallOn is reset. Therefore, it won't work with TrainHasCallOn_Advanced and TrainHasCallOn_Restricted_Advanced sigscr functions.


cjakeman (approved):
Hi Cesar,

This looks good. The new version seems simpler than the original, which is always a positive.

Approved with thanks,

Chris

cesarBLG:
Thanks, Chris

PR #562, opened 2022-01-03
Remove the locomotive handle from the script, because no interface function can allow access to any internal structure.

pzgulyas:
It was a huge mistake adding this function to the scripting interface. It must be (must have been, actually) removed ASAP.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -300,10 +300,6 @@ namespace ORTS.Scripting.Api
         /// Get distance of next station if any, else max float value
         /// </summary>
         public Func<float> NextStationDistanceM;
-        /// <summary>
-        /// Get locomotive handle
-        /// </summary>
-        public Func<MSTSLocomotive> Locomotive;
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


cesarBLG:
What do you think about doing something like this?
```csharp
class DummyLocomotive
{
    float AccelerationMpSpS;
    float CurrentElevationPercent;
    string WagFilePath;

    class DummyTrainBrakeController
    {
         void SetValue(float value) {}
    };
    DummyTrainBrakeController TrainBrakeController;
    class DummyThrottleController
    {
        readonly float CurrentValue = 0;
    }
    DummyThrottleController ThrottleController;
}
public Func<DummyLocomotive> Locomotive;
```
With this we could prevent further uses of Locomotive() without completely breaking existing scripts, until users are notified and upgrade to more recent script versions. It is not a clean solution, but I think it's better than just removing it and break everything.

pzgulyas:
When I saw a script last time using this function, that script immediately went towards MSTSLocomotive -> MSTSWagon -> Train, and used some signalling references there directly. But a script theoretically get any path in the internal structures, and you cannot keep everything working without duplicating EVERYTHING. This is why it must be removed immediately.

Sharpe49:
On my side, the Locomotive function is no longer used.

By the way, for the needed information, I can integrate the needed functions in #525 if you want.

cjakeman (approved):
Approved.

Thanks, Peter

PR #571, opened 2022-01-18
01: Manual entries for Menu > Options corrected

cjakeman:
and a missing one added.

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -156,8 +156,8 @@ The option *Player's Location* sets the units according to the Windows
 The option *Route* sets the units based on the data in the route files.
 The other options are self-explanatory.
 
-These windows use the abbreviations *stn* for short tons (2000 lb),
-*t* for long tons (2,240 lb) or *tn* for metric tons (tonnes).
+These windows use the abbreviations *t-us* for short tons (2,000 lb),
+*t-uk* for long tons (2,240 lb) and *t* for metric tons (1,000 kg).
 
 Note: The units displayed by the :ref:`F4 Track Monitor <driving-track-monitor>` (e.g. velocity and
 distance) are always based on data read from the route files.
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -156,8 +156,8 @@ The option *Player's Location* sets the units according to the Windows
 The option *Route* sets the units based on the data in the route files.
 The other options are self-explanatory.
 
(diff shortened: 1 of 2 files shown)


Source/Documentation/Manual/options.rst
```diff
@@ -148,18 +148,18 @@ Other units
...
 The option *Player's Location* sets the units according to the Windows
 *Language and Region* settings on the player's computer.
 
 The option *Route* sets the units based on the data in the route files.
 The other options are self-explanatory.
 
-The F5 HUD uses the abbreviations *stn* for short tons (2000 lb) and
-*t* or *tn* for metric tons (tonnes).
+These windows use the abbreviations *stn* for short tons (2000 lb),
+*t* for long tons (2,240 lb) or *tn* for metric tons (tonnes).
```
pzgulyas, line 160: Was the ton sign also changed in the code? The SI standard explicitly mentions that the metric ton sign is "t". So I would prefer having "t" for the metric, and e.g. "lt" for long ton.

cjakeman: Hi Peter,

Thanks for catching this. The original text was wrong, but so was my correction. Guess I meant to check it and never did.

Now matches the code.

Best wishes,
Chris

Source/Documentation/Manual/options.rst
```diff
@@ -156,8 +156,8 @@ The option *Player's Location* sets the units according to the Windows
 The option *Route* sets the units based on the data in the route files.
 The other options are self-explanatory.
 
-These windows use the abbreviations *stn* for short tons (2000 lb),
-*t* for long tons (2,240 lb) or *tn* for metric tons (tonnes).
+These windows use the abbreviations *t-us* for short tons (2,000 lb),
+*t-uk* for long tons (2,240 lb) and *t* for metric tons (1,000 kg).
```
pzgulyas, line 160: Chris, I'm really sorry for being so picky, but I would have one more comment, please forgive me.

In almost all other non-English languages the comma (,) is used as a decimal separator. For a non native English reader it is very confusing to read numbers with commas in that, because they never know if that is a decimal separator or a thousand separator. Also in mathematics and physics the thousand separators are never used exactly because of the international nature of the science. Could we possibly also follow a rule not using these in the documentation? I would prefer writing 2,000 as 2000, 2,240 as 2240 and 1,000 as 1000, if possible.

cjakeman: OK Peter, I'll do my bit for internationalisation.

I've also refined the Hot Start text as posted on the forum.

Thanks,
Chris

pzgulyas (approved):
This looks fine to me now.

PR #573, opened 2022-01-23
Fix wrong variable used in battery button

cesarBLG:
The CommandButtonOn variable was incorrectly used for both battery open and battery close buttons


Sharpe49 (approved):
Oops... that's right.

PR #576, opened 2022-01-24
DaylightOffsetHrs control: Fix for offset not effective soon after midnight

cjakeman:
https://bugs.launchpad.net/or/+bug/1958893
Also code was difficult to read.

### DIFF
--- a/Source/RunActivity/Viewer3D/MSTSSky.cs
+++ b/Source/RunActivity/Viewer3D/MSTSSky.cs
@@ -69,11 +69,7 @@ namespace Orts.Viewer3D
         public Orts.Viewer3D.SkyViewer.Date date;
 
         private SkySteps skySteps = new SkySteps();
-        // Size of the sun- and moon-position lookup table arrays.
-        // Must be an integral divisor of 1440 (which is the number of minutes in a day).
-        //private int maxSteps = 72;
-        //private double mstsskyoldClockTime;
-        //private int step1, step2;
+
         // Phase of the moon
         public int mstsskymoonPhase;
         // Wind speed and direction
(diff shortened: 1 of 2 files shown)


Source/RunActivity/Viewer3D/Sky.cs
```diff
@@ -137,12 +137,10 @@ public void PrepareFrame(RenderFrame frame, ElapsedTime elapsedTime)
             // Current solar and lunar position are calculated by interpolation in the lookup arrays.
             // The arrays have intervals of 1200 secs or 20 mins.
             // Using the Lerp() function, so need to calculate the in-between differential
-            float diff = GetCelestialDiff();
             // The rest of this increments/decrements the array indices and checks for overshoot/undershoot.
-            while (Viewer.Simulator.ClockTime >= (oldClockTime + 1200)) // Plus key, or normal forward in time; <CSComment> better so in case of fast forward
+            while (Viewer.Simulator.ClockTime >= (oldClockTime - DaylightOffsetS + 1200)) // Plus key, or normal forward in time; <CSComment> better so in case of fast forward
```
pzgulyas, line 141: The Sky.cs has a brother-file: MSTSSky.cs. Everywhere the Sky.cs is changed, the MSTSSky.cs should follow. For example the pair of these lines are: https://github.com/openrails/openrails/blob/71c4e89ab1cc0bb5110274617cd1fdd16e3dec9c/Source/RunActivity/Viewer3D/MSTSSky.cs#L216-L218

Actually the two files should be united or inreited one from the other, to have the common code pieces in a common class. I think code copying is a poor programming technique, I don't know why these were implemented this way.

pzgulyas (changes requested):

Source/RunActivity/Viewer3D/Sky.cs
cjakeman: Thanks, Peter.

Neither do I, but I'll try to re-unite the two files.

Source/RunActivity/Viewer3D/MSTSSky.cs
```diff
@@ -209,58 +210,59 @@ public void PrepareFrame(RenderFrame frame, ElapsedTime elapsedTime)
...
-            mstsskylunarDirection.Y = MathHelper.Lerp(mstsskylunarPosArray[step1].Y, mstsskylunarPosArray[step2].Y, diff);
-            mstsskylunarDirection.Z = MathHelper.Lerp(mstsskylunarPosArray[step1].Z, mstsskylunarPosArray[step2].Z, diff);
+            //////////////////////////////////////////////////////////////////////
+
+            //// Current solar and lunar position are calculated by interpolation in the lookup arrays.
+            //// Using the Lerp() function, so need to calculate the in-between differential
+            //float diff = (float)(MSTSSkyViewer.Simulator.ClockTime - mstsskyoldClockTime) / 1200;
+            //// The rest of this increments/decrements the array indices and checks for overshoot/undershoot.
+            //if (MSTSSkyViewer.Simulator.ClockTime >= (mstsskyoldClockTime + 1200)) // Plus key, or normal forward in time
+            //{
```
pzgulyas, line 220: Thank you Chris for your efforts, I think this is absolutely fine. For now it is just important that the sky and mstssky don't diverge more, because later noone would know which is the corrected and which is the old part. However I think instead of commenting out code sections it is better just to delete these lines. The old will always remain available in git anyway, so there is no need to keep theese lines in the live code.

Source/RunActivity/Viewer3D/MSTSSky.cs
```diff
@@ -68,11 +68,12 @@ public class MSTSSkyDrawer
...
 
+        private SkySteps skySteps = new SkySteps();
         // Size of the sun- and moon-position lookup table arrays.
         // Must be an integral divisor of 1440 (which is the number of minutes in a day).
-        private int maxSteps = 72;
-        private double mstsskyoldClockTime;
-        private int step1, step2;
+        //private int maxSteps = 72;
+        //private double mstsskyoldClockTime;
+        //private int step1, step2;
```
pzgulyas, line 76: There are commented out lines here too that could be deleted.

cjakeman: Thanks, Peter.

I didn't mean to leave in this old code. Thanks for spotting it.

Guess I'm trying to do too much at once !

Chris

pzgulyas (approved):
To mee this looks good now.

PR #580, opened 2022-01-28
SamplerStates can only be set after the ShaderPasses.Current.Apply().

pzgulyas:
When dealing with the graphics subsystem I bumped into this bug, that probably has been in the code for a long time. This bug falls into the category "less than 10 additions and deletions each" (not counting the comments).

### DIFF
--- a/Source/RunActivity/Viewer3D/Forest.cs
+++ b/Source/RunActivity/Viewer3D/Forest.cs
@@ -426,7 +426,6 @@ namespace Orts.Viewer3D
 
             // Enable alpha blending for everything: this allows distance scenery to appear smoothly.
             graphicsDevice.BlendState = BlendState.NonPremultiplied;
-            graphicsDevice.SamplerStates[0] = SamplerState.LinearClamp;
         }
 
         public override void Render(GraphicsDevice graphicsDevice, IEnumerable<RenderItem> renderItems, ref Matrix XNAViewMatrix, ref Matrix XNAProjectionMatrix)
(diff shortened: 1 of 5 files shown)


cjakeman (approved):
Hi Peter,

Looks good and it's been in Unstable long enough.

Approved with thanks,

Chris

twpol:
> When dealing with the graphics subsystem I bumped into this bug, that probably has been in the code for a long time

I'm curious, what told you that this was a bug?

pzgulyas:
@twpol, while implementing glTF and going through the test models, I came accross this test model:
https://github.com/KhronosGroup/glTF-Sample-Models/tree/master/2.0/TextureSettingsTest

This is a specific test for checking if sampler states are set correctly. And it didn't work for the god's sake, until I figured out that the ShaderPasses.Current.Apply() overwrites any preset sampler state with the one written in the .fx file.

twpol:
> This is a specific test for checking if sampler states are set correctly. And it didn't work for the god's sake, until I figured out that the ShaderPasses.Current.Apply() overwrites any preset sampler states with those written in the .fx file.

Ah wow, kinda surprised we didn't notice anything breaking until now. Thanks for fixing it!

pzgulyas:
@twpol, This fix of yours from 5+ years ago might also be an attempt to overcome this bug:
https://github.com/openrails/openrails/blob/aa9c76bac8992144f73ea021c8a4960732d3e440/Source/RunActivity/Content/SceneryShader.fx#L462-L467
(Although I'm not exactly sure, because that was done for XNA 3,1 yet, and I'm not sure whether that overwrote the sampler states too at ShaderPasses.Current.Begin(), and I'm not in a mood to go back and try. :-) )

PR #582, opened 2022-01-31
Bug fix for https://bugs.launchpad.net/or/+bug/1959617 Positive bars can't have two colours

Csantucci:
The second colour is never shown.


pzgulyas (approved):
I see no problems with this change

PR #587, opened 2022-02-06
02d: Removed menu option Curve-Dependent Resistance

cjakeman:


pzgulyas (approved):
I can't spot any errors in this

cjakeman:
> I can't spot any errors in this

Thanks, Peter

PR #588, opened 2022-02-06
02ef: Removed menu option Tunnel&Wind-Dependent Resistance

cjakeman:


Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2108,46 +2101,38 @@ public void UpdateWindComponents()
...
+            else if (PhysicsTrainLocoDirectionDeg > PhysicsWindDirectionDeg)
+                ResultantWindComponentDeg = PhysicsTrainLocoDirectionDeg - PhysicsWindDirectionDeg;
+            else
+                ResultantWindComponentDeg = 0.0f;
 
-                // Wind angle should be kept between 0 and 180 the formulas do not cope with angles > 180. If angle > 180, denotes wind of "other" side of train
-                if (ResultantWindComponentDeg > 180)
-                    ResultantWindComponentDeg = 360 - ResultantWindComponentDeg;
+            // Correct wind direction if it is greater then 360 deg, then correct to a value less then 360
+            if (Math.Abs(ResultantWindComponentDeg) > 360)
```
pzgulyas, line 2126: Adding Math.Abs() seems to be unnecessary here, after all if it is negative, it will surely not be > 360

pzgulyas (approved):

PR #592, opened 2022-02-06
02j: Removed menu option ViewDispatcher . Now always available.

cjakeman:

Includes some re-organising of DebugViewerForm.cs so that initialisation is only done when Ctrl+9 is pressed.

not-for-unstable currently as conflicts with previous PRs
6-Mar Removed not-for-unstable as no longer conflicts with previous PRs


pzgulyas (approved):
Seems all right

cjakeman:
@pzgulyas Hi Peter,

This one also now no longer conflicts with other PRs.

Please re-approve.

Thanks,

Chris

pzgulyas (approved):
Approved!

PR #595, opened 2022-02-10
Release locomotive brake when player train enters game with speed > 0

Csantucci:
At the moment the locomotive brake is set at the value defined in the .eng file, which may have a % higher than 0. This is acceptable only if the train has speed zero at game start.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -1653,6 +1653,7 @@ public List<CabView> CabViewList = new List<CabView>();
                 TrainControlSystem.InitializeMoving();
                 TrainBrakeController.InitializeMoving();
                 BrakeSystem.LocoInitializeMoving();
+                EngineBrakeController?.SetPercent(0);
             }
         }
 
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -1653,6 +1653,7 @@ public override void InitializeMoving()
                 TrainControlSystem.InitializeMoving();
                 TrainBrakeController.InitializeMoving();
                 BrakeSystem.LocoInitializeMoving();
+                EngineBrakeController?.SetPercent(0);
```
Sharpe49, line 1656: The MSTSBrakeController class has an InitializeMoving function with the following code:
`        public override void InitializeMoving()
        {
            NotchController.SetValue(0);
            NotchController.CurrentNotch = 0;
        }`
Shouldn't it be enough to call that function?

Sharpe49:
```suggestion
                EngineBrakeController?.InitializeMoving();
```

Sharpe49 (approved):
Seems good to me

Csantucci:
> Seems good to me

Thanks for approval.

PR #596, opened 2022-02-10
Fix for https://bugs.launchpad.net/or/+bug/1960519. Crash, diesel locomotive with vacuum brakes.

mbm-OR:
When the Player locomotive is a diesel with vacuum brakes, a crash occurs, when <Shift+F9> is pressed.
It is related to the GetDpuStatus() function.
Fix for https://bugs.launchpad.net/or/+bug/1960519


pzgulyas (approved):
I see no problems with this.

mbm-OR:
> I see no problems with this.

Thank you.

PR #598, opened 2022-02-12
02k: Removed menu option Override non-electrified route line-voltage

cjakeman:

Can now run electric locos on non-electricified routes.
An information message is shown when the pantograph is raised.


Source/ORTS.Settings/UserSettings.cs
```diff
@@ -247,8 +247,8 @@ public enum DirectXFeature
         public bool TunnelResistanceDependent { get; set; }
         [Default(false)]
         public bool WindResistanceDependent { get; set; }
-        [Default(false)]
-        public bool OverrideNonElectrifiedRoutes { get; set; }
+        //[Default(false)]
+        //public bool OverrideNonElectrifiedRoutes { get; set; }
```
Sharpe49, line 251:
```suggestion
```

Sharpe49 (changes requested):

Source/ORTS.Settings/UserSettings.cs
cjakeman: Thanks, Cedric. Missed that.

pzgulyas (dismissed):
Seems good

Sharpe49 (approved):

Sharpe49 (changes requested):
The merge was incomplete. You probably should rebase your branches instead of merging.

cjakeman:
@Sharpe49 Hi Cedric,

Thanks for that.

I've overcome the conflicts, now, so it would be great if you could re-approve this PR.

Thanks,

Chris

Sharpe49 (approved):

PR #599, opened 2022-02-12
02l: Removes web-server option. Now always available.

cjakeman:

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -168,14 +168,12 @@ Disable TCS scripts
 This option disables the train control system scripts for locomotives where
 these have been implemented.
 
-Enable web server
+Web server port
 -----------------
 
-This option enables an internal web server that can be used to display game and
-train status information in a web browser, intended for use on secondary screens.
-
-When activated, the server can be accessed from a browser on the local machine at
+The web server can be accessed from a browser on the local machine at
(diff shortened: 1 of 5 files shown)


cesarBLG:
Hi Chris,
Some days ago I had to disable the web server in this situation:
I was running two instances of Open Rails for a Multiplayer test, and the second one crashed at startup because there was a server already running.

I don't know if that crash would affect someone else, but if that is the case we should try to gracefully handle that situation.

cjakeman:
> Hi Chris, Some days ago I had to disable the web server in this situation: I was running two instances of Open Rails for a Multiplayer test, and the second one crashed at startup because there was a server already running.
> 
> I don't know if that crash would affect someone else, but if that is the case we should try to gracefully handle that situation.

Hi Cesar,

Thanks for reporting this.

The code will have to cope with this situation once the option to enable the WebServer is removed.

I'll add a try . . catch to the WebServer with a message in the log if the port is already in use.

Best wishes,

Chris

cjakeman:
@cesarBLG Hi Cesar,

I've added teh change you needed, so please review this PR when you have time.

Thanks,

Chris

cesarBLG (approved):
Thanks, Chris. I've tested it and it works as expected.

pzgulyas (approved):
Seems to be alright.

PR #600, opened 2022-02-12
fix: Store DirectX feature level as the enum internally for performance

twpol:
This has impact on performance because the Enum.Parse function allocates a lot of memory each time it is called, so instead of calling that repeatedly it is now only called when the feature level is set.

This results in a drop of 21% in TOTAL allocations in a simple profiling run (load activity and exit after 1000 frames are rendered), and a drop of 57% (!) in allocations post-loading.


pzgulyas (approved):
Looks like a good catch! :-)

PR #601, opened 2022-02-13
Video-RAM message to logfile fails under VMWare

cjakeman:
https://bugs.launchpad.net/or/+bug/1960746


cjakeman:
> I see no problem here.

Thanks, Peter

PR #444, opened 2021-07-28
First phase of wheel slide protection

peternewell:
Adding a number of brake enhancements - https://blueprints.launchpad.net/or/+spec/brake-enhancements

PR #602, opened 2022-02-15
New model for MR pipe air propagation

cesarBLG:
This allows having different pressures in main reservoirs than in the main reservoir train pipe.
This feature was initially introduced in #444, but was reverted in #507. The current model used for air propagation in the MR pipe isn't realistic, and cannot handle this situation. I've modified the code to use the same logic as for brake pipe.

See http://www.elvastower.com/forums/index.php?/topic/35988-air-brakes-bug/

Solves https://bugs.launchpad.net/or/+bug/1961089


pzgulyas (approved):
I see no problems with the code

cesarBLG:
Thanks, Peter

PR #604, opened 2022-02-18
Fix headlight flickering when switching down from bright to dim

pzgulyas:


cesarBLG (approved):
Not too much to say about this tiny change, it's been in unstable for enough time.

PR #605, opened 2022-02-19
Fix problem when curve speed dependent not selected

peternewell:
https://bugs.launchpad.net/or/+bug/1961497


peternewell:
> Hi Peter, i view of a positive review of your PR, can you explain me how your patch solves the issue?

In basic terms there were two "if (CurveSpeedDependent)" statements, and the first one which was removed in this PR, was stopping the calculation of some parameters required for other calculations associated with curve resistance (which should occur all the time).

The remaining "if (CurveSpeedDependent)" parameter should be sufficient to disable notifications with "roll over speeds", which is what the option selection was designed to do.

cjakeman (approved):
Hi Peter,

Thanks for finding and fixing this.

Approved,

Chris

PR #606, opened 2022-02-20
Bug fix for http://www.elvastower.com/forums/index.php?/topic/36021-crash-on-switching-to-autopilot/#entry282465

Csantucci:


pzgulyas (approved):
Seems okay.

PR #608, opened 2022-02-23
Copy DynamicBrakeMaxCurrentA to replicas of same locomotive

Csantucci:


Sharpe49 (approved):
Yep, that was missing indeed

Csantucci:
> Yep, that was missing indeed

Wow, that was fast, thanks! The problem came out using the DP table display.

PR #614, opened 2022-03-05
Initialize EP and twin pipe brakes to correct pressure

cesarBLG:
See http://www.elvastower.com/forums/index.php?/topic/36055-auxiliary-reservoir-not-fully-charged-on-brake-initialisation/


Csantucci (approved):
OK for me.

cesarBLG:
Thanks, Carlo

PR #616, opened 2022-03-10
Improvements in graduated/full release braking

cesarBLG:
Discussion: http://www.elvastower.com/forums/index.php?/topic/36073-triple-valves-on-air-twin-pipe-systems/
Roadmap: https://trello.com/c/B8ml6Hyu

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -2191,24 +2191,8 @@ The operation of air single-pipe brakes is described in general below.
 
 The auxiliary reservoir needs to be charged by the brake pipe and,
 depending on the WAG file parameters setting, this can delay the brake
-release. When the *Graduated Release Air Brakes* box is not checked, the
-auxiliary reservoir is also charged by the emergency reservoir (until
-both are equal and then both are charged from the pipe). When the
-*Graduated Release Air Brakes* box is checked, the auxiliary reservoir is
-only charged from the brake pipe. The Open Rails software implements it
-this way because the emergency reservoir is used as the source of the
-reference pressure for regulating the brake cylinder pressure.
-
-The end result is that you will get a slower release when the *Graduated
-Release Air Brakes* box is checked. This should not be an issue with two
-pipe air brake systems because the second pipe can be the source of air
-for charging the auxiliary reservoirs.
-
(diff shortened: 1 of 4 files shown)


cjakeman (commented):
Hi Cesar,

With 5 pages on the forum already, I was expecting to see substantial changes to the Manual. Is the change you've made here really adequate for a complex topic?

Just my thoughts,

Chris

cesarBLG:
Hi Chris,
Most of the code changes are things that are moved from one place to another. The only changes are:
1) Moved distributor configuration from AirTwinPipe.cs to SinglePipe.cs, so full/partial release works independently of the brake system. A further step would be creating separate classes for each brake valve (triple valve or distributor), as in http://www.elvastower.com/forums/index.php?/topic/29263-refactoring-the-air-brake-code/page__view__findpost__p__209707
In the past BrakeEquipmentType was sometimes ignored in OR, now the behavior is consistent with MSTS TechDoc's. :
```
BrakeEquipmentType:
“triple_valve" – The standard triple valve with partial application and full release. Plus auxilary res. charging.
"graduated_release_triple_valve" – The vehicle has a triple valve that permits partial releasing of the brakes.
```
2) Added a STF token for an already existing variable (NoMRPAuxResCharging), that I renamed to MRPAuxResCharging as recommended by coding guidelines.
3) Corrected a wrong behavior of triple valves as reported here some time ago: http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__264905
This is an experimental feature yet, I'll need some feedback to see if the results are OK. By default it is disabled, when it works OK we can add safe estimates using UIC timings.

The thread at ET includes a lot of comments about how to correctly setup different brake systems, but I believe the brake controller token table made by Darwin and already present in the manual is enough.

If you think an specific topic that has to be extended, I'll try to write about it.

cjakeman:
Hi Cesar,

> If you think an specific topic that has to be extended, I'll try to write about it.

Thanks for explaining how the changes work out and for your offer to write some more. I can now see why the manual entry is shorter than it was before, so nothing more needs to be written.

Well done with this work, 

Chris

cjakeman (approved):
Hi Cesar,

Looks very good, especially the comments.

Approved with thanks,

Chris

PR #620, opened 2022-03-16
Adjust kinietic friction calculations

peternewell:
https://bugs.launchpad.net/or/+bug/1965071

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -3034,10 +3034,10 @@ public List<CabView> CabViewList = new List<CabView>();
                 // Exponential curve is used to transition between static friction and dynamic friction when wheel slips
                 // Exponential constant calculated between two points, using this tool - https://mathcracker.com/exponential-function-calculator#results
                 // Google search suggests that Steel on steel has a static coeff = 0.74, and a dynamic coeff = 0.57. Hence reduction = 0.77.
-                // Constant points facilitate a decrease from 1 to 0.7 in 3 seconds - P1 = (0, 1), P2 = (3, 0.77). Hence exp constant = −0.0871
-                var expAdhesion = -0.0871;
+                // Constant points facilitate a decrease from 1 to 0.7 in 3 seconds - P1 = (0, 1), P2 = (5, 0.77). Hence exp constant = −0.0523
+                var expAdhesion = -0.0523;
                 WheelSlipTimeS += elapsedClockSeconds;
-                WheelSlipTimeS = MathHelper.Clamp(WheelSlipTimeS, 0.0f, 3.0f); // Ensure that time to transition between the two friction cases is maintained - currently set to 3 secs
+                WheelSlipTimeS = MathHelper.Clamp(WheelSlipTimeS, 0.0f, 5.0f); // Ensure that time to transition between the two friction cases is maintained - currently set to 3 secs
 
                 float adhesionMultiplier = (float) Math.Exp(expAdhesion * WheelSlipTimeS);
                 CurrentWheelSlipAdhesionMultiplier = adhesionMultiplier;
@@ -3053,12 +3053,12 @@ public List<CabView> CabViewList = new List<CabView>();
 
                 if (SlipFrictionCoefficientFactor < BaseFrictionCoefficientFactor && SlipFrictionCoefficientFactor != 0) // Once these two are equal then assume that wheels have stopped slipping.
                 {
-//                    Trace.TraceInformation("SlipFriction {0} Base {1}", SlipFrictionCoefficientFactor, BaseFrictionCoefficientFactor);
+                    //                    Trace.TraceInformation("SlipFriction {0} Base {1}", SlipFrictionCoefficientFactor, BaseFrictionCoefficientFactor);
                     // Exponential curve is used to transition between dynamic friction and static friction when wheel stops slipping
-                    // Constant points facilitate an increase from 0.7 to 1 in 3 seconds - P1 = (3, 0.77), P2 = (0, 1). Hence exp constant = 0.0871
-                    var expAdhesion = 0.0871;
+                    // Constant points facilitate an increase from 0.7 to 1 in 3 seconds - P1 = (5, 0.77), P2 = (0, 1). Hence exp constant = 0.0523
+                    var expAdhesion = 0.0523;
                     WheelStopSlipTimeS += elapsedClockSeconds;
-                    WheelStopSlipTimeS = MathHelper.Clamp(WheelStopSlipTimeS, 0.0f, 3.0f); // Ensure that time to transition between the two friction cases is maintained - currently set to 3 secs
+                    WheelStopSlipTimeS = MathHelper.Clamp(WheelStopSlipTimeS, 0.0f, 5.0f); // Ensure that time to transition between the two friction cases is maintained - currently set to 3 secs
 
                     float adhesionMultiplier = CurrentWheelSlipAdhesionMultiplier * (float)Math.Exp(expAdhesion * WheelStopSlipTimeS);
 
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -2920,18 +2920,15 @@ public List<CabView> CabViewList = new List<CabView>();
                 }
                 if (Simulator.WeatherType == WeatherType.Rain) // Wet weather
                 {
-                    if (Simulator.Settings.AdhesionProportionalToWeather && AdvancedAdhesionModel && !Simulator.Paused)  // Adjust clear weather for precipitation presence - base friction value will be approximately between 0.15 and 0.2
-                    // ie base value between 0.8 and 1.0 (TODO) 
-                    // note lowest friction will be for drizzle rain; friction will increase for precipitation both higher and lower than drizzle rail
+                    if (Simulator.Settings.AdhesionProportionalToWeather && AdvancedAdhesionModel && !Simulator.Paused)  // Adjust clear weather for precipitation presence
+                    // ie base value between 60% and 80% (TODO) 
+                    // note lowest friction will be for drizzle (light) rain; friction will increase for precipitation higher than drizzle rail
                     {
                         float pric = Simulator.Weather.PricipitationIntensityPPSPM2 * 1000;
-                        // precipitation will calculate a value between 0.15 (light rain) and 0.2 (heavy rain) - this will be a factor that is used to adjust the base value - assume linear value between upper and lower precipitation values
-                        if (pric >= 0.5)
-                            BaseFrictionCoefficientFactor = Math.Min((pric * 0.0078f + 0.45f), 0.8f); // should give a minimum value between 0.8 and 1.0
-                        else
-                            BaseFrictionCoefficientFactor = Math.Min((0.4539f + 1.0922f * (0.5f - pric)), 0.8f); // should give a minimum value between 0.8 and 1.0
+                        // precipitation will calculate a base coefficient value between 60% (light rain) and 80% (heavy rain) - this will be a factor that is used to adjust the base value - assume linear value between upper and lower precipitation values
+                        BaseFrictionCoefficientFactor = Math.Min((pric * 0.0078f + 0.6f), 0.8f); // should give a minimum value between 60% and 80%
                     }
-                    else // if not proportional to precipitation use fixed friction value of 0.8 x friction coefficient of 0.33
+                    else // if not proportional to precipitation use fixed friction value of 0.8 x friction coefficient value
                     {
                         BaseFrictionCoefficientFactor = 0.8f;
                     }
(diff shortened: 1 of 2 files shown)


cesarBLG:
Hi Peter,
I wonder if it could be possible to use the wheel creep model in Axle.cs also for steam locomotives. The idea would be to use the inertia moments and tangential forces calculated in UpdateTractiveForce() in MSTSSteamLocomotive.cs as an input for the axle model, instead of the simple wheel slip model used in steam locomotives. Do you think it will improve the results?

peternewell:
>I wonder if it could be possible to use the wheel creep model in Axle.cs also for steam locomotives. The idea would be to use the inertia moments and tangential forces calculated in UpdateTractiveForce() in MSTSSteamLocomotive.cs as an input for the axle model, instead of the simple wheel slip model used in steam locomotives. Do you think it will improve the results?

Diesel/Electric locomotive have different wheel slip characteristics compared to steam locomotives. In a diesel/electric locomotive the force applied to the wheels is constant throughout the full rotation of the wheels. Which is what the axle model calculates.

Whereas for the steam locomotive, the force varies throughout the full revolution of the wheel - see here for an explanation -
https://www.coalstonewcastle.com.au/physics/steam-adhesion/#slip

Hence they need to be treated differently.

I think that the adhesion model needs some work as well (which will impact both the diesel/electric and steam locomotives). Once I have got my thoughts organised I will post into the ET thread some more detail.

cesarBLG:
I agree that the forces applied are radically different, but I wonder if that affects the wheel-rail dynamics at differential intervals. The axle model uses as an input an instant force, then integrates the revolutions as if the wheel was flying, and finally calculates the effective motive force that is exerted to the rails. This adhesion model (wheel creep) is more sophisticated than the one used in steamers (i.e. just comparing friction force and applied force, and slip if the latter is greater), but I don't know if the model is better or not.
What I was saying is if connecting the angle-dependent force produced by the cylinders and transmitted to the wheels to the axle model would produce more accurate results.

peternewell:
>The axle model uses as an input an instant force, then integrates the revolutions as if the wheel was flying, and finally calculates the effective motive force that is exerted to the rails. This adhesion model (wheel creep) is more sophisticated than the one used in steamers (i.e. just comparing friction force and applied force, and slip if the latter is greater), but I don't know if the model is better or not. 

I would need to study the axle model more to be confident that it could do it. For example if it averages the force over a full wheel rotation, then it might not pick up when the wheel tread force exceeds the adhesive force momentarily during part of the wheel cycle.

If it could do this, then perhaps you might be correct.

The other issue, that I am trying to model more accurately is the change in adhesion from the static coefficient to dynamic coefficient. This appears to be an external input to the axle model, and I am leaning towards a model that has some level of transition between the two values (though it is hard to find a definitive physics explanation on the web).

Have a look at the second graph in this section of CTN - https://www.coalstonewcastle.com.au/physics/steam-adhesion/#slip

cesarBLG:
I'm not really convinced that it is correct to modify the adhesion values due to the change from static to dynamic friction coefficient. The axle model considers the base friction coefficient as an input from which it calculates a dynamic friction coefficient that takes into account the possible slipping, so modifying the base coefficient will introduce the friction coefficient reduction twice. I've sent to you at ET a document where I try to explain how the axle model works, as well as some graphs.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -3016,13 +3020,54 @@ public virtual void UpdateFrictionCoefficient(float elapsedClockSeconds)
...
+                    WheelStopSlipTimeS = MathHelper.Clamp(WheelStopSlipTimeS, 0.0f, 5.0f); // Ensure that time to transition between the two friction cases is maintained - currently set to 3 secs
+
+                    float adhesionMultiplier = CurrentWheelSlipAdhesionMultiplier * (float)Math.Exp(expAdhesion * WheelStopSlipTimeS);
+
+//                    Trace.TraceInformation("adhesion {0} StopTime {1} Base {2} Current {3}", adhesionMultiplier, WheelStopSlipTimeS, BaseFrictionCoefficientFactor, CurrentWheelSlipAdhesionMultiplier);
+
+                    BaseFrictionCoefficientFactor *= adhesionMultiplier;  // Descrease friction to take into account dynamic (kinetic) friction, typically kinetic friction is approximately 50% of static friction.
+                    SlipFrictionCoefficientFactor = BaseFrictionCoefficientFactor;
+                    BaseFrictionCoefficientFactor = MathHelper.Clamp(BaseFrictionCoefficientFactor, 0.05f, 1.0f); // Ensure friction coefficient never exceeds a "reasonable" value
+                }
```
cesarBLG, line 3070: Peter, the base friction coefficient is also used for the axle model, which already takes into account this kind of variations, so in fact we are introducing twice the same effects. Could you move this calculations to SteamLocomotive.cs, or add a check to avoid it for diesel&electric locos?

cesarBLG (changes requested):

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
peternewell: > the base friction coefficient is also used for the axle model, which already takes into account this kind of variations, so in fact we are introducing twice the same effects. Could you move this calculations to SteamLocomotive.cs, or add a check to avoid it for diesel&electric locos?

I have added a temporary patch to apply this only to steam locomotives (subject to further discussion).

I strongly believe that adhesion coefficients should only be calculated in one location, ie MSTSLocomotive with the outcome being passed to the axle model, or steam model as required. I think that it is confusing to be calculating adhesion in two different places, and it also runs the risk that any future code changes to the adhesion coefficient may not be altered in both locations if they remain separate.

The other aspect to be considered is that the adhesion coefficient is displayed in the HuD, is the value calculated in the Axle module being displayed in the HuD?

I would also like to see conversion factors used where possible, rather then using discrete numbers. For example, 

`speed = Math.Abs(3.6f*speed);`

should be 

`speedkmph = (MpS.ToKpH(AbsSpeedMpS);`

Given the fact that this PR and your recently opened PR (#627) have some degree of overlap, and thus are interacting with each other, and also given your stated desire to use the axle model for steam locomotives, I would suggest that we co-ordinate these activities, close one of these PRs and combine the relevant code changes.

Perhaps we can discuss this work consolidation in the PM that we currently have going.

cesarBLG: > I have added a temporary patch to apply this only to steam locomotives (subject to further discussion).
Thanks for that.
> I strongly believe that adhesion coefficients should only be calculated in one location, ie MSTSLocomotive with the outcome being passed to the axle model, or steam model as required. I think that it is confusing to be calculating adhesion in two different places, and it also runs the risk that any future code changes to the adhesion coefficient may not be altered in both locations if they remain separate.

I more or less agree. In my opinion the _maximum_ adhesion coefficient should be calculated in a global class (probably better in MSTSWagon or TrainCar, so wagon brakes are also affected). With maximum adhesion coefficient I mean the static coefficient, given by Curtius-Kniffler formula and corrected due to weather and sanding. The change of the adhesion factor with time or slip percentage will depend on the model being used. Ideally we should calculate also the dynamically-varying adhesion factor outside the model too, but unfortunately the formula used in Axle.cs does not allow doing so (Polach 2005 formula does need a varying adhesion factor, but as I've said it has too many free parameters).

> The other aspect to be considered is that the adhesion coefficient is displayed in the HuD, is the value calculated in the Axle module being displayed in the HuD?

I have no idea. Strictly speaking, the axle model does not calculate an effective adhesion coefficient, but it calculates the  tangential force per unit of weight, so I don't know.

>  would also like to see conversion factors used where possible, rather then using discrete numbers. For example,
> speed = Math.Abs(3.6f*speed);
> should be
> speedkmph = (MpS.ToKpH(AbsSpeedMpS);

Yes, I believe this is required by OR coding conventions. Another thing that I like when reading code (although I don't know if it is mandatory) is to have uppercase names for variables when they produce side effects outside the current method (e.g. public global variables), and lowercase for local variables created in intermediate steps.

> Given the fact that this PR and your recently opened PR (https://github.com/openrails/openrails/pull/627) have some degree of overlap, and thus are interacting with each other, and also given your stated desire to use the axle model for steam locomotives, I would suggest that we co-ordinate these activities, close one of these PRs and combine the relevant code changes.

I don't mind resolving git conflicts, so I suggest that you keep going with this PR, and I'll correct mine if there are conflicts. I have to study if the integration time in the axle model is good enough to allow the particularities of steam traction (variable force during rotation), so meanwhile we can keep the two different models.

What else do you have in mind for including in this PR?

peternewell: > I more or less agree. In my opinion the _maximum_ adhesion coefficient should be calculated in a global class (probably better in MSTSWagon or TrainCar, so wagon brakes are also affected). With maximum adhesion coefficient I mean the static coefficient, given by Curtius-Kniffler formula and corrected due to weather and sanding. 
>

I believe that wheel adhesion during braking and braking skid may be different to the adhesion for a locomotive, especially where wheel slip correction is used to increase the adhesion of the locomotive.

As long as it caters correctly for the different scenarios, and is all together, I would have no problems if it is in TrainCar, or a suitable other class.

>The change of the adhesion factor with time or slip percentage will depend on the model being used. Ideally we should calculate also the dynamically-varying adhesion factor outside the model too, but unfortunately the formula used in Axle.cs does not allow doing so (Polach 2005 formula does need a varying adhesion factor, but as I've said it has too many free parameters).
> 
What values from the SlipCharacteristics calculations cannot be bought out of the axle model, and why can't it be calculated elsewhere.

> I have no idea. Strictly speaking, the axle model does not calculate an effective adhesion coefficient, but it calculates the tangential force per unit of weight, so I don't know.
> 
Regardless, it is still a "dynamic adhesion coefficient" and as such should be visible to the user for debug purposes as the other adhesion coefficients already are.

> I don't mind resolving git conflicts, so I suggest that you keep going with this PR, and I'll correct mine if there are conflicts. I have to study if the integration time in the axle model is good enough to allow the particularities of steam traction (variable force during rotation), so meanwhile we can keep the two different models.
> 
> What else do you have in mind for including in this PR?

At the moment I am still not 100% happy that the steam model is performing in a similar fashion as video evidence suggests. Until I can explain the difference or get it to align, I want to reflect on a few aspects of the model, and do additional testing.

cesarBLG: > I believe that wheel adhesion during braking and braking skid may be different to the adhesion for a locomotive, especially where wheel slip correction is used to increase the adhesion of the locomotive.

I don't think so, the wheel-rail contact is the same. From a physical point of view, a tractive force applied to the wheel is the same as a brake force slowing the wheel, the only differences being on how does that force changes with time. Slip/skid control does not increase the maximum possible adhesion, but modifies applied forces so the adhesion limit is not reached.

> What values from the SlipCharacteristics calculations cannot be bought out of the axle model, and why can't it be calculated elsewhere.

The formula in axle.cs is:
image
Where the mu_0 parameter is the maximum (static) adhesion factor. So this formula requires a constant friction coefficient, i.e. not dependent on slip percentage.

On the other hand, for example in Polach model:
image
The "mu" coefficient is dependent on the wheel speed, so this formula accepts a friction factor varying during the wheelslip process.

Both approaches are correct from my point of view, so that's why I prefer to calculate the maximum factor in TrainCar.cs, and include the dependence with wheelslip in the specific adhesion model being used.

Do you have any references where I can find some information about Curtius&Kniffler equation?

peternewell: Rather then spreading our conversation over two places, I will respond to some of this in the PM that we already have going.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2916,18 +2920,15 @@ public virtual void UpdateFrictionCoefficient(float elapsedClockSeconds)
...
+                    if (Simulator.Settings.AdhesionProportionalToWeather && AdvancedAdhesionModel && !Simulator.Paused)  // Adjust clear weather for precipitation presence
+                    // ie base value between 60% and 80% (TODO) 
+                    // note lowest friction will be for drizzle (light) rain; friction will increase for precipitation higher than drizzle rail
                     {
                         float pric = Simulator.Weather.PricipitationIntensityPPSPM2 * 1000;
-                        // precipitation will calculate a value between 0.15 (light rain) and 0.2 (heavy rain) - this will be a factor that is used to adjust the base value - assume linear value between upper and lower precipitation values
-                        if (pric >= 0.5)
-                            BaseFrictionCoefficientFactor = Math.Min((pric * 0.0078f + 0.45f), 0.8f); // should give a minimum value between 0.8 and 1.0
-                        else
-                            BaseFrictionCoefficientFactor = Math.Min((0.4539f + 1.0922f * (0.5f - pric)), 0.8f); // should give a minimum value between 0.8 and 1.0
```
cesarBLG, line 2928: I'd prefer to keep the if clause, so there's a small transition curve from dry to wet.

```csharp
if (pric >= 0.5)
    BaseFrictionCoefficientFactor = Math.Min((pric-0.5f) * 0.0078f + 0.6f, 0.8f);
else
    BaseFrictionCoefficientFactor = 0.6f + 0.8f * (0.5f - pric);
```
weather

What do you think?

peternewell: Ok, I agree.

I have also looked at the code, and it only seemed to be applying this if the weather was raining or snow. I don't believe that this should be the case as it should vary from Clear - "Snow" - Rain and back again.

So I have modified the code so that it will default to the weather selected in the launch screen, but the precipitation and fog can be adjusted in game (or not as desired by the user).

I believe that the option of linking adhesion to precipitation can be removed, and will post in the forums accordingly.

cesarBLG (approved):
Thanks, Peter. I agree that the "adhesion proportional to weather" checkbox is not needed.

peternewell:
Thanks for that.

PR #623, opened 2022-03-20
Russian localization for T 1.4-413 version

Weter-ORTS:
Have actualized translations, corrected Options menu's entries and added gearbox&electrification warnings.
Also, have started with Timetable editor localization.


cesarBLG (approved):
Thanks, Weter. It seems to be OK. I don't speak Russian so I didn't look for typos, though.

Weter-ORTS:
> Thanks, Weter. It seems to be OK. I don't speak Russian so I didn't look for typos, though.
Thank You in return.
Don't worry about typos: as I launch ORTS from time to time, I notice them sooner or later, correct, and even find more suitable expressions.

PR #624, opened 2022-03-20
Menu option04 - invert controls

cjakeman:
and http://www.elvastower.com/forums/index.php?/topic/35301-menu-options

"Disable TCS Scripts" => "Use TCS Scripts"
"Diesel engines stopped after startup" => "At game start, Diesel - run engines"

This PR changes the UI, but not the defaults and the user's settings are preserved too.


Weter-ORTS:
> At game start, Diesel - run engines
Internal combustion engines running at game start

cesarBLG:
Hi Chris,
Is there a reason for changing the UI, but not the internal variable naming?
For example, could Settings.DisableTCSScripts be changed to Settings.EnableTCSScripts, so there's a correspondence between UI and code?

cesarBLG:
Hi Chris,
Is there a reason for changing the UI, but not the internal variable naming?
For example, could Settings.DisableTCSScripts be changed to Settings.EnableTCSScripts, so there's a correspondence between UI and code?

cjakeman:
Hi Cesar,

Thanks for looking at this.

I agree that it's not ideal, but I didn't do as you suggest in order to keep the user's saved settings.
The current PR does not require the user to do anything. If his "Disable TCS" was checked then his new "Enable TCS" will be unchecked. 

Chris

cesarBLG (approved):
I see, in that case it's OK to me.

PR #625, opened 2022-03-20
Menu > Options - Change defaults for best graphics

cjakeman:
and forum from http://www.elvastower.com/forums/index.php?/topic/35301-menu-options/page__view__findpost__p__273843

Changes default to checked for:
Dynamic Shadows
Vertical Sync
Model Instancing

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -200,11 +200,11 @@ Video Options
 Dynamic shadows
 ---------------
 
-Check this option to cast shadows from movable objects such as trains.
+This option causes movable objects such as trains to cast shadows.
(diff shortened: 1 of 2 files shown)


Weter-ORTS:
Model Instancing can cause lags on weak systems, as engmod told

cjakeman:
Hi Weter,

> Model Instancing can cause lags on weak systems, as engmod told

You are quite right. The aim is to change the defaults so that new users get the best quality graphics if their PC is fast enough. If not, then they will have to dial back the settings. 

I plan to discuss this on the public forum shortly.

Best wishes,
Chris

Weter-ORTS:
best wishes in turn

cesarBLG (approved):
No objections from my side

PR #627, opened 2022-03-25
Improve advanced adhesion model

cesarBLG:

- [x] Include wheel radius in calculations to convert from torque to tangential forces, and from angular speed to linear speed.
- [x] Allow lower wheel inertia values for electric and diesel locomotives.
- [x] Ensure that dissipative forces (e.g. braking) never accelerate the wheels (i.e. only deceleration).
- [x] Correct the slip characteristics so that adhesion tends to a fixed value (usually 40% of static friction). With the current model adhesion tends to zero.
- [x] Slip control.

http://www.elvastower.com/forums/index.php?/topic/36144-improving-adhesion/


cjakeman:
Hi Cesar,

Thanks for your continued work on this topic. This PR is still in draft and I can see from your initial list that there is more that you want to do.

Is the PR at a suitable point to take it out of draft, get it into the Testing Version and start a new PR for the rest of the work?

Will the Manual need any changes?

Best wishes,

Chris

cesarBLG:
I've now modified the manual, and set the PR 'ready for review'.

I expect a number of complaints, as the adhesion calculations are quite non-linear and dependent on frame-rates. Feedback from other users will be needed.

I don't know the timings for v1.5, but depending on the number of complaints when this gets into Testing, we might have to wait before including it into a stable version.

cjakeman:
Hi César,

Thanks for picking this up and getting it ready for review. I’ve not seen many complaints while it’s been in Unstable but we’ll see how it goes once it’s in Testing.

Best wishes,

Chris

cjakeman (approved):
Hi Cesar,

This all looks good and thanks for updating the Manual too.

I'm sorry I didn't follow up earlier on my comment of 17th June.

Approved with thanks,

Chris

PR #631, opened 2022-03-31
Get correct signal id if signal ahead is in same Track Circuit

cesarBLG:
https://bugs.launchpad.net/or/+bug/1967313
When looking for not-normal signals from a not-normal signal, the search currently does not take into account signals ahead that are in the same Track Circuit.


Csantucci (approved):
Seems to do what it should.

cesarBLG:
Thanks, Carlo.

PR #633, opened 2022-03-31
Removed ViewDispatcher setting

cjakeman:
To fix bug reported by Hannes.
See http://www.elvastower.com/forums/index.php?/topic/35301-menu-options/page__view__findpost__p__283958


cesarBLG (approved):
The change seems reasonable to me.

PR #634, opened 2022-03-31
Out of range exception with >= 256 cab controls

cesarBLG:
This bug does not yet affect the official OR code, as the CABViewControlTypes enum has less than 256 elements, but may affect in the future.

However, @Csantucci New Year version already exceeds that limit, causing crashes.


Csantucci (approved):
Thanks for that, César.

PR #615, opened 2022-03-05
 Management of EOT devices  https://blueprints.launchpad.net/or/+spec/eot

Csantucci:
See also here http://www.elvastower.com/forums/index.php?/topic/35912-eot/

PR #636, opened 2022-04-03
Remove venting of MRP because conflicts with EOT

cesarBLG:
This is required to get #615 working as expected. EOT devices were developed in parallel to some brake changes that are incompatible. The conflicting brake functionality isn't really useful (venting the main reservoir pipe), so I remove it.


Csantucci (approved):
César,
thank you for this.

PR #639, opened 2022-04-09
fix: Use correct randomness source for viewer weather

twpol:
This fixes a few places in the `Viewer3D` code which use `Simulator` randomness - which is a problem when you want to compare a `Viewer3D` run (i.e. the normal application) with a `Simulator`-only run (i.e. to test only the simulation, physics, etc.)

It fits into the +10/-10 rule but is also partially part of https://trello.com/c/rA9gu11Q/526-refactor-traincar-to-simplify-inheritance


cesarBLG (approved):
Seems logical to me. Thanks, James.

PR #640, opened 2022-04-09
feat: Initial implementation of Simulation Tester

twpol:
Roadmap: https://trello.com/c/rA9gu11Q/526-refactor-traincar-to-simplify-inheritance

This new contributed program loads a save file and runs the `Simulator` on it just like the "Replay from start" button in the menu. It prints out how close the final player train position is to the saved position.

This is intended for regression testing to ensure any refactored code has not changed the overall behaviour of the simulation.

### DIFF
--- a/Source/Contrib/SimulatorTester/App.config
+++ b/Source/Contrib/SimulatorTester/App.config
@@ -1,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.2" />
+    </startup>
+</configuration>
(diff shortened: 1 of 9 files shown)


cjakeman (commented):
Hi James,

I got it working based on what you told me, but I found several ways to make it crash :-(

- If no parameters are provided, then it would be nice to get a Help message. Instead the program crashes.
- If the Save file is not an activity then the program crashes.
- If the Save file contains camera movements, then the program crashes.

Didn't travel very far, but the accuracy wasn't great either.

Best wishes,
Chris

cjakeman (approved):
Hi James,

My comments earlier don't stop it being used for the intended purpose, so I'm approving this.

Chris

PR #641, opened 2022-04-10
Removing unnecessary spaces for the language template generator

FrannDzs:
The "menu" language template had some, I think unnecessary, spaces.
I was checking this file because in the translation I made into Spanish based on the .pot template there were some strings that didn't work, as shown in the screenshot below

Captura de pantalla 2022-04-09 190319

This would fix the problem. 
*Language files need to be updated


cesarBLG:
There's currently an ongoing process of reorganizing the menu options. Until the reorganization is finished, the new strings can't be translated.

PR #643, opened 2022-04-10
Brake controller buttons are not animated

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/36157-dedicated-buttons-for-brake-controllers


Csantucci (approved):
Thanks for that! Florian will be happy.

cesarBLG:
Thanks, Carlo

PR #646, opened 2022-04-26
Fix issue with brake controller pressure

cesarBLG:
Use the pressure in the local "Main Reservoir" for the brake controllers, instead of the pressure in Main Res Pipe, which is not always available.

https://bugs.launchpad.net/or/+bug/1970622


Csantucci (approved):
Seems OK to me, thanks.

cesarBLG:
Thanks, Carlo

PR #647, opened 2022-05-06
fix: Put all cache files into the designated area away from user files

twpol:
Blueprint: https://trello.com/c/qdZ0yzjy/285-virtual-filesystem

This change is necessary for the VFS, since that should not have to deal with writing changes, and this specialised feature (binary paths, opt-in timetable feature) contradicts that

Binary paths should behave as they did before, but stored in a new location

Loading information is also moved to be consistent


cjakeman (dismissed):
Looks tidy and no bad reports seen.

Approved with thanks,

Chris

PR #648, opened 2022-05-07
adds AsBoolean, AsVector3 to JsonReader

cjakeman:
also propagates warnings so they can be reported easily
Blueprint at https://blueprints.launchpad.net/or/+spec/extend-json-reader
Forum thread at http://www.elvastower.com/forums/index.php?/topic/36162-container-management/

### DIFF
--- a/Source/Orts.Parsers.OR/JsonReader.cs
+++ b/Source/Orts.Parsers.OR/JsonReader.cs
@@ -18,13 +18,14 @@
 // Use this define to diagnose issues in the JSON reader below.
 //#define DEBUG_JSON_READER
 
+using Microsoft.Xna.Framework;
+using Newtonsoft.Json;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text;
-using Newtonsoft.Json;
 
 namespace Orts.Parsers.OR
 {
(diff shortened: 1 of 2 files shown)


Source/Orts.Parsers.OR/JsonReader.cs
```diff
@@ -194,9 +222,36 @@ public float AsTime(float defaultValue)
...
+                        vector3.Z = AsFloat(0f);
+                    if (!_reader.Read() || _reader.TokenType != JsonToken.EndArray)
+                        goto default; // We did not have exactly 3 items in the array
+                    _path.Length = _pathPositions.Pop();
+                    return vector3;
+                default:
+                    TraceWarning($"Expected Vector3 (3 item array) in {Path}; got {_reader.TokenType}");
+
+                    // If the end of the array is not found in the right position, then parsing of subsequence objects also fails.
+                    TraceWarning($"Subsequent objects may be skipped");
```
twpol, line 245: This warning should not be here

I can only think of two reasons subsequent objects may be skipped:

1.  A bug in `AsVector3` or another part of `JsonReason` leaves it in an incorrect state compared to the JSON
2. Mismatched object/array brackets in the JSON

For one, we should fix the bug, and for two there should already be another warning or error emitted due to the problem (it might be inscrutable but when you mismatch brackets that's what you get - both in SIMIS and JSON)

twpol (changes requested):

Source/Orts.Parsers.OR/JsonReader.cs
cjakeman: Agreed

twpol (approved):

PR #651, opened 2022-05-29
fix: Update existing JsonReader loading code for new array handling

twpol:


Csantucci (approved):
That's OK. Before seeing this PR, I came to the same conclusion for LoadStationsOccupancyFile.cs

PR #653, opened 2022-06-03
Blueprint for container management https://blueprints.launchpad.net/or/+spec/containers

Csantucci:
Specs from here http://www.elvastower.com/forums/index.php?/topic/36162-container-management/page__view__findpost__p__285313

### DIFF
--- a/Source/ORTS.Common/Coordinates.cs
+++ b/Source/ORTS.Common/Coordinates.cs
@@ -85,6 +85,20 @@ namespace ORTS.Common
             Location = copy.Location;
         }
 
+
+        /// <summary>
+        /// MSTS WFiles represent some location with a position, quaternion and tile coordinates
+        /// This converts it to the ORTS WorldPosition representation
+        /// </summary>
+        public WorldPosition (int tileX, int tileZ, Vector3 xnaPosition, Quaternion xnaQuaternion)
+        {
+            XNAMatrix = Matrix.CreateFromQuaternion(xnaQuaternion);
+            XNAMatrix *= Matrix.CreateTranslation(xnaPosition);
+
+            TileX = tileX;
+            TileZ = tileZ;
+        }
+
         /// <summary>
         /// The world-location in MSTS coordinates of the current position
         /// </summary>
@@ -168,6 +182,7 @@ namespace ORTS.Common
             return WorldLocation.ToString();
         }
     }
+
 
     /// <summary>
     /// Represents the position of an object within a tile in MSTS coordinates.
@@ -314,6 +329,15 @@ namespace ORTS.Common
         {
             return new Vector2((float)(locationTo.Location.X - locationFrom.Location.X + (locationTo.TileX - locationFrom.TileX) * TileSize),
                 (float)(locationTo.Location.Z - locationFrom.Location.Z + (locationTo.TileZ - locationFrom.TileZ) * TileSize));
+        }
+
+        public static float ApproximateDistance(WorldLocation a, WorldLocation b)
+        {
+            var dx = a.Location.X - b.Location.X;
+            var dz = a.Location.Z - b.Location.Z;
+            dx += (a.TileX - b.TileX) * 2048;
+            dz += (a.TileZ - b.TileZ) * 2048;
+            return Math.Abs(dx) + Math.Abs(dz);
         }
 
         /// <summary>
--- a/Source/ORTS.Common/Input/UserCommand.cs
+++ b/Source/ORTS.Common/Input/UserCommand.cs
@@ -203,6 +203,7 @@
         [GetString("Control Cylinder Compound")] ControlCylinderCompound,
         [GetString("Control Firing")] ControlFiring,
         [GetString("Control Refill")] ControlRefill,
+        [GetString("Control Discrete Unload")] ControlDiscreteUnload,
         [GetString("Control Water Scoop")] ControlWaterScoop,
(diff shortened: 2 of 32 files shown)


cjakeman (commented):
Hi Carlo,	
	
Just to let you know that I'm currently looking at your PR.	
	
I've installed your ContainerDemoPack and, as I've said before, what you've achieved looks terrific.	
	
Thanks for providing all the detail in the Manual. When I build a PDF, there are some problems with the images though with no obvious cause. Two of them show up too small to read:
Source/Documentation/Manual/images/features-loading-positions.png	
Source/Documentation/Manual/images/features-stack-locations-code.png	

and one fails to render and the PDF shows the text:
.. image:: images/features-sample-load-entry.png

I'll be looking at the code next and get back to you.

Best wishes,
Chris

cjakeman (commented):
Hi Carlo,

I was wondering what are the consequences of performing a Save (key ``<F2>``) when a loading or unloading operation is ongoing?

Thanks,
Chris

Csantucci:
> Hi Carlo,
> 
> Just to let you know that I'm currently looking at your PR.
> 
> I've installed your ContainerDemoPack and, as I've said before, what you've achieved looks terrific.
> 
> Thanks for providing all the detail in the Manual. When I build a PDF, there are some problems with the images though with no obvious cause. Two of them show up too small to read: Source/Documentation/Manual/images/features-loading-positions.png Source/Documentation/Manual/images/features-stack-locations-code.png
> 
> and one fails to render and the PDF shows the text: .. image:: images/features-sample-load-entry.png
> 
> I'll be looking at the code next and get back to you.
> 
> Best wishes, Chris

Hi Chris,
thanks for taking your time to review this.
The reference to the third picture had an error in the .rst file (CRLF missing). I've uploaded the fix and It should appear now. About the other two, unfortunately I'm able to build only the .html version of the manual, and in that one the two pictures appear OK. I'll check if I find the reason why the pictures appear unreadable.

Csantucci:
Chris,
could you provide me the screenshots of the two pages with the two unreadable pictures,  so that I can see how I could improve them?

cjakeman:
> Chris, could you provide me the screenshots of the two pages with the two unreadable pictures, so that I can see how I could improve them?

I see no reason why these images should be undersized. I've tinkered a little but to no effect.

Not tried images before in a GitHub post. Here is a drag and drop:
2022-06-19 14_02_33-MS Excel with extensions - Container Management xlsx   Compatibility Mode

and here is a link to OneDrive:
https://1drv.ms/u/s!AqyUzuEHyizFotpz5QoJ5s-uPgpvBg?e=C5EhKm

Hope one of these works.

I found an error in the Manual where you have changed the key combination for ImmediateRefill from Shift+T to Ctrl+T. This is a reasonable change, but the Manual needs to change to match in section 7.3.5.

I find nothing in the Manual about multiplayer texting, so no need to worry about the use of  Alt+T.

Best wishes,

Chris

cjakeman (commented):
HI Carlo,

I see nothing in the code that troubles me :)

I saw that you had to add a new constructor for WorldPosition and I was surprised that we had not needed it before now. I looked, but could find nowhere else that should take advantage of this new method.

Chris

cjakeman (changes requested):
Hi Carlo,

Something I I don't think we can accept is the extension you've provided for the JSON files.

When Rob added the timetable JSON, we adopted a convention that these files would have an extension which was not truncated and ends in "-or".

By this convention,
"TRAINSET\Common.ContainerData\20cmacgm.loa"
should become
"TRAINSET\Common.ContainerData\20cmacgm.load-or"

and
"Activities\Openrails\BigContainerStationPopulation.lsp"
should become
"Activities\Openrails\BigContainerStationPopulation.load-stations-population-or"

"Population" is a bit long-winded, so an alternative might be:
"Activities\Openrails\BigContainerStationLoads.load-stations-loads-or"

Please would you make the change and update the Manual to match?

Thanks,
Chris

cjakeman (changes requested):
Hi Carlo,

In the Manual you advise not performing a Save (key <F2>) when a loading or unloading operation is ongoing.

I tried this out by Saving and Resuming and all was well until I tried to load another container. Sometimes it works and sometimes I get a NullReferenceException.

We could perhaps:
- trap that and continue
- save more data so that the animation is resumed from exactly where it was saved
- block or delay the Save until the (un)load operation is complete

Which approach would you favour?
I don't think we should ship with a known bug.

Best wishes,

Chris

Csantucci:
Hi Chris,
thanks for all your time. I hadn't time to devote to this, and will try to find from now on.
In short words:
- I will change .loa to .load-or
- I will add an -or suffix to the load population file, but I confess that a .load-stations-loads-or extension seems to me extremely long; maybe i'll come with another proposal
- when saving and resuming during load/unload operations  I found cases where a container disappeared or was no more recognized, although visible; I checked the code again, and increasing the save/resume parameters seems quite complicated to me; as you are so kind to leave to me the choice, I'll prepare a patch to disable saving during load/unload and to display on the monitor a line if the player tries to save during that; I have studied a bit and it should be easy
- I'll modify the manual about the Immediate Refill command
- About the pictures, I'm without ideas; I have an alternate picture for the first one, and I can replace the second one with text, so I'll proceed that way.

Csantucci:
Chris, I believe I have now satisfied all your requests.

PR #655, opened 2022-06-16
Menu block for bad content location

cjakeman:
Bug report: https://www.elvastower.com/forums/index.php?/topic/36313-or-update-deleting-its-own-folder-structure

Not really a bug - just plugging a hole that users could fall into.


peternewell (approved):
Hi Chris,

This seems fine.

Thanks

PR #657, opened 2022-06-18
fix: Correctly merge exception and current stack traces in logger

twpol:
I noticed in http://www.elvastower.com/forums/index.php?/topic/36331-or-crashclose-after-loading-a-route/ that the hang details were not looking correct, so I have fixed them with this change


cjakeman (approved):
James,

Approved and merged with thanks.

Chris

PR #658, opened 2022-06-20
style: Add .NET editorconfig from Visual Studio templates

twpol:
Adding this file will:

- **Override** most user settings regarding code style *when* editing C# files in Visual Studio, Visual Studio Code, and other supported editors
- **Ensure** new code is automatically written in a more consistent and expected style
- **Not** produce any build messages *unless* you run the analysers manually
- **Not** reformat any existing code *unless* you run the code clean-up manually

The settings in this file are the default Visual Studio 2022 "editorconfig File (.NET)" template and have **not** been modified by me in any way

We can always come back and change any of these options - with agreement - if they do not seem to match our preferred code style

### DIFF
--- a/Source/.editorconfig
+++ b/Source/.editorconfig
@@ -1,0 +1,132 @@
+# To learn more about .editorconfig see https://aka.ms/editorconfigdocs
+###############################
+# Core EditorConfig Options   #
+###############################
+# All files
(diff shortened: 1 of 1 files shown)


cjakeman (approved):
Good idea. Approved

PR #660, opened 2022-06-24
perf: Improved performance counter initialisation

twpol:
This improves the performance of the performance counter initialisation by loading all the data at once (`ReadCategory()`) instead of repeatedly reading a single item (`new PerformanceCounter()` + `NextValue()`)

It does not change the behaviour

In testing, this improves the time taken for the three try blocks for performance counter initialisation:

* 0.81 seconds before
* 0.58 seconds after (28% improvement)


cjakeman (approved):
Hi James,

No problems reported and code chnge is clean.

Approved witht thanks,

Chris

PR #661, opened 2022-06-25
Corrected text that had become out of date

cjakeman:
Reflects changes in managing the map Window (Ctrl+9). 
Also references in chapter on Physics.


peternewell (approved):
Hi Chris,

All seems ok.

PR #662, opened 2022-06-26
Brake pipe pressure is reduced instantly to EQ pressure

cesarBLG:
This falls into the category of "Bug fix with less than +-10 additions"


peternewell (approved):
Thanks for fixing this issue.

cesarBLG:
Thanks, Peter

PR #665, opened 2022-06-27
02n: Removes some redundant experimental options

cjakeman:

Load DDS textures in preference to ACE is now always true.
Adhesion is now always reduced by rain/snow/fog.
Precipitation Box values are hidden awaiting any user feedback.


peternewell (approved):
Hi Chris,

As far as I can see this looks fine.

PR #666, opened 2022-06-27
Website - correction to link for unstable versions

cjakeman:
Link to unstable versions led to change summary but not to downloads.
Now leads to both.


peternewell (approved):
Hi Chris,

Thanks for that.

Cheers

peternewell (approved):
Hi Chris,

Thanks for the changes.

PR #632, opened 2022-03-31
Menu options07a - Re-do System Tab

cjakeman:
See https://blueprints.launchpad.net/or/+spec/update-menu-options
and http://www.elvastower.com/forums/index.php?/topic/35301-menu-options

Pulls system controls, including Updater, into a System tab. Drops Updater tab.

PR #668, opened 2022-07-16
fix: Swap defaults for instancing and glass back

twpol:
It looks like the default values of these two settings got accidentally swapped in #632 so this swaps them back


cjakeman (approved):
Hi James,

Thanks for catching this.

Approved,

Chris

PR #669, opened 2022-07-22
refactor: Separate host performance data collection and processing to another thread

twpol:
Not sure we've got a bug on file for this, but it is certainly part of bug-fixing and similar problems reported recently

Bug report: https://www.elvastower.com/forums/index.php?/topic/36331-or-crashclose-after-loading-a-route/


cjakeman (approved):
Hi James,

Looks much cleaner.

Approved with thanks,

Chris

PR #672, opened 2022-07-25
Calculate distance to signal in explorer mode

cesarBLG:
https://bugs.launchpad.net/or/+bug/1982755
While in explorer mode, distance to next signal is not calculated. That prevents approach control from working.


Csantucci (approved):
Approved, thank you.

cesarBLG:
Thanks, Carlo

PR #673, opened 2022-07-26
hu locales update

pzgulyas:


Csantucci (approved):
Peter, thanks for updating this.

PR #674, opened 2022-08-04
Mark particle drawer texture for wagons

Sharpe49:
Should fix : http://www.elvastower.com/forums/index.php?/topic/29799-steam-effects-for-diesel-locomotives-and-passenger-cars/page__view__findpost__p__286963

### DIFF
--- a/Source/RunActivity/Viewer3D/ParticleEmitter.cs
+++ b/Source/RunActivity/Viewer3D/ParticleEmitter.cs
@@ -148,7 +148,7 @@ namespace Orts.Viewer3D
         internal void Mark()
         {
             if (Material != null) // stops error messages if a special effect entry is not a defined OR parameter
-            Material.Mark();
+                Material.Mark();
         }
     }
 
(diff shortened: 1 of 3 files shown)


cjakeman:
Hi Cedric,

This looks like a very simple and effective change.

I'll leave it for Peter to check that it works on his test cases.

Many thanks,

Chris

peternewell:
> I'll leave it for Peter to check that it works on his test cases.

It appears that the Unstable version has stopped building as the last build was on 26/07, so this correction hasn't made it through to the Unstable version. When the build starts again, I can check it.

Peter

peternewell (approved):
This patch seems to have fixed the bug, so I have approved it.

Is this a universal fix for all textures? Does there need to be some developer notes for any future issues detected or new features added?

Thanks

cjakeman:
> Is this a universal fix for all textures? Does there need to be some developer notes for any future issues detected or new features added?

Not for all textures, but for all ParticleEmitters. These are now handled in the MSTSWagonViewer instead of the MSTSLocomotiveViewer, which makes lots of sense. Why didn't I think of that?

Yes, we could do with some documentation or helpful comments, especially since textures and shapes are managed differently.

Thanks to Cedric for sorting this.

PR #675, opened 2022-08-07
Bug fix for https://bugs.launchpad.net/or/+bug/1983511 When an AI train disappears at end of run, containers remain visible

Csantucci:


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
```diff
@@ -863,6 +863,22 @@ public bool CheckForMerge(int i)
             }
             return false;
         }
+
+        /// <summary>
+        /// Removes discrete freight animations and containers when AI train is deleted
+        /// </summary>
+
+        public void RemoveDiscreteFreightAnimations(MSTSWagon wagon)
```
Sharpe49, line 871: Shouldn't the function by called RemoveContainers?
Can't you use the Wagon member which is already present in the class?

Sharpe49 (changes requested):

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
Csantucci: OK, I have updated the code.

Sharpe49 (approved):

Csantucci:
Cédric, thank you.

Sharpe49:
You should probably try to merge this branch in release/1.5

PR #676, opened 2022-08-07
adds content hosted on GitHub.com

cjakeman:
Changes the Download > Content page to include 3 routes hosted on GitHub.com and updates the home page news


peternewell (approved):
Hi Chris,

This seems ok.

Do these three routes include the stock as well, ie can you run a train once the packs have been installed?

If they can't then I would suggest that an appropriate note is added. Both the demonstration packs include stock, and can be used straight away. The user should be aware that some additional work is required before stock can be run on them if this is the case.

Cheers

cjakeman:
> Do these three routes include the stock as well, ie can you run a train once the packs have been installed?
> 
> If they can't then I would suggest that an appropriate note is added. Both the demonstration packs include stock, and can be used straight away. The user should be aware that some additional work is required before stock can be run on them if this is the case.

Yes, they do include stock and can all be used straight away. I believe Chiltern is intended to get more activities soon and MECoast to get more stock soon.

Best wishes,

PR #569, opened 2022-01-09
Upgraded project files to .Net SDK project file format

Sharpe49:
Removed AssemblyInfo files (integrated in the project file)
Removed unused files (not referenced in the old project files)
Removed resource files in ActivityEditor which were causing execution problems

1st step of https://blueprints.launchpad.net/or/+spec/dependencies-update
http://www.elvastower.com/forums/index.php?/topic/32819-upgrading-net-c/page__view__findpost__p__278324

PR #680, opened 2022-08-11
build: Fix the file metadata after the .NET SDK project switch

twpol:
Looks like the project metadata fields in #569 did not match what was previously in the `AssemblyInfo.cs` files, so this PR puts it all back exactly as it was, except for:

- Copyright is updated to 2022 from 2019
- Contributed Timetable Editor copyright format changed to match the other files

Most of this is purely aesthetic, however, certain fields are used by the updater and menu:

- [Updater matches the file description and product name fields to find the main executable](https://github.com/openrails/openrails/blob/14559529ca8750365e021206c31ed04f06b4fce7/Source/ORTS.Updater/UpdateManager.cs#L76)
- [Menu checks for matching product names to show the correct tools](https://github.com/openrails/openrails/blob/14559529ca8750365e021206c31ed04f06b4fce7/Source/Menu/MainForm.cs#L195)

### DIFF
--- a/Source/ContentChecker/ContentChecker.csproj
+++ b/Source/ContentChecker/ContentChecker.csproj
@@ -5,9 +5,11 @@
     <UseWindowsForms>true</UseWindowsForms>
     <ImportWindowsDesktopTargets>true</ImportWindowsDesktopTargets>
     <IsPublishable>False</IsPublishable>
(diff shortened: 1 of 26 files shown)


Sharpe49 (approved):
Seems good to me

PR #681, opened 2022-08-12
test: Fix tests when StfReader tests run before JsonReader tests

twpol:
Because the `StfReader` tests modify global state (`Trace.Listeners`), we need to counter with more global state modifications in `JsonReader` tests

Ultimately, though, the `AssertWarnings` class needs to go


Sharpe49 (approved):
Yep, this fix seems to work.

PR #683, opened 2022-08-13
fix: Clean up Simulator Tester help output to match dotnet-style

twpol:


Sharpe49 (approved):

Sharpe49:
BTW, most of the dotnet tools are using options starting with a dash instead of a slash.

twpol:
> BTW, most of the dotnet tools are using options starting with a dash instead of a slash.

Yeah, we allow `/` (and sometimes `-`) but `dotnet` is exclusively `--` for long options I believe - changing that will be a bigger follow-up PR :)

Sharpe49:
OK ! :)

PR #693, opened 2022-08-17
French locales

Sharpe49:


twpol:
@Sharpe49 Thanks for the updated locale, I have rebased them onto the 1.5 release branch so that we can include them successfully

PR #694, opened 2022-08-17
Upgrade libraries to netstandard 2.0

Sharpe49:
(only for libraries without Windows Forms)

### DIFF
--- a/Source/Contrib/ActivityEditor/LibAE/Formats/PathEvent.cs
+++ b/Source/Contrib/ActivityEditor/LibAE/Formats/PathEvent.cs
@@ -22,24 +22,12 @@
 /// 
 
 using System;
-using System.Windows;
-using System.IO;
-using System.Collections.Generic;
-using System.Linq;
-using System.Text;
 using System.Drawing;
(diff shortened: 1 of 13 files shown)


cesarBLG:
There are reports that TCS scripts are not compiling: http://www.elvastower.com/forums/index.php?/topic/36471-tcs-script-problems/page__pid__287979#entry287979

CS0012: The type 'Enum' is defined in an assembly that is not referenced. You must add a reference to assembly 'netstandard, Version=2.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'., line: 15, column: 27

Sharpe49:
Weird, for me it includes .Net Framework 4.7.2 during script compilation so it's still working fine...
I'll add a reference to the assembly that contains System.Enum...

cesarBLG:
I posted my log in the forum. Your last commit didn't solve the issue (I suppose System.Object and System.Enum are in the same assembly, which was already included)

Sharpe49:
@cesarBLG Ok, removing this commit. Seems like we will have to switch to .Net 5 directly.

PR #700, opened 2022-08-24
Fix for crash during signal loading

Sharpe49:
See https://github.com/openrails/openrails/pull/537#issuecomment-1219961043

https://bugs.launchpad.net/or/+bug/1987564


cesarBLG (approved):
Hi Cédric,
Thanks for looking at it. The code seems OK, and fixes the problem.

PR #421, opened 2021-05-29
Added a reset command in order to reset a TCS emergency braking triggered by the simulator

Sharpe49:
https://trello.com/c/QvCgMnVW/256-command-to-release-train-control-system-emergency-braking

PR #701, opened 2022-08-25
Prevent train from going to manual mode from an explorer activity

cesarBLG:
https://bugs.launchpad.net/or/+bug/1987634

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -487,6 +487,10 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Action RequestToggleManualMode;
         /// <summary>
+        /// Requests reset of Out of Control Mode.
+        /// </summary>
+        public Action ResetOutOfControlMode;
+        /// <summary>
         /// Get bool parameter in the INI file.
         /// </summary>
         public Func<string, string, bool, bool> GetBoolParameter;
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -10193,15 +10193,13 @@ namespace Orts.Simulation.Physics
             }
             else if (ControlMode == TRAIN_CONTROL.EXPLORER)
             {
-                if (LeadLocomotive is MSTSLocomotive locomotive &&
-                    (locomotive.TrainBrakeController.TCSEmergencyBraking || locomotive.TrainBrakeController.TCSFullServiceBraking))
-                {
-                    locomotive.TrainControlSystem.HandleEvent(TCSEvent.EmergencyBrakingReleasedBySimulator);
-                    ResetExplorerMode();
-                    return;
-                }
-                else if (Simulator.Confirmer != null) // As Confirmer may not be created until after a restore.
+                if (Simulator.Confirmer != null) // As Confirmer may not be created until after a restore.
                     Simulator.Confirmer.Message(ConfirmLevel.Warning, Simulator.Catalog.GetString("Cannot change to Manual Mode while in Explorer Mode"));
+            }
+            else if (ControlMode == TRAIN_CONTROL.OUT_OF_CONTROL && ControlModeBeforeOutOfControl == TRAIN_CONTROL.EXPLORER)
+            {
+                if (Simulator.Confirmer != null) // As Confirmer may not be created until after a restore.
+                    Simulator.Confirmer.Message(ConfirmLevel.Warning, Simulator.Catalog.GetString("Cannot change to Manual Mode. Use the Reset Out Of Control Mode command to release brakes"));
             }
             else
             {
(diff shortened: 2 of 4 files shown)


cesarBLG:
@Csantucci IIRC you are using ToggleManualMode in your TCS scripts to reset out of control mode. However, this incorrectly changes the mode from Explorer to Manual, which is what I'm fixing here. In order to allow switching back to explorer mode (or Manual Mode if in an activity) I added a hook called ResetOutOfControlMode(). You can use this new hook for future releases of your scripts.

However, this causes that your already released scripts will do nothing when calling to RequestToggleManualMode(). I am open to modify the behaviour of that function to internally call ResetOutOfControlMode if coming from Explorer Mode.

Csantucci:
César,
Thanks for your message.  It's not clear to me if your proposal in the last lines of your message to solve the problem with my scripts is to modify the code of RequestToggleManualMode(), or to modify what occurs when Script.RequestToggleManualMode is called. At the moment both solutions seem OK to me, at least in an intermediate phase.

cesarBLG:
I see in your scripts that you have:
```csharp
if (SPAD) RequestToggleManualMode();
```
and
```csharp
if (SPAD && ControlMode != TRAIN_CONTROL.EXPLORER) RequestToggleManualMode();
```
What did you want to do here?
a. Reset Out of Control Mode and switch back to manual or explorer mode.
b. Toggle Manual Mode, but only in activity mode (not explorer mode)

If you wanted to do a), then I will modify Script.RequestToggleManualMode like this:
```csharp
Script.RequestToggleManualMode = () => 
{
    if (Locomotive.Train.ControlMode == TRAIN_CONTROL.OUT_OF_CONTROL && Locomotive.Train.ControlModeBeforeOutOfControl == TRAIN_CONTROL.EXPLORER)
    {
        Trace.TraceWarning("RequestToggleManualMode() is deprecated for explorer mode. Please use ResetOutOfControlMode() instead");
        Locomotive.Train.ManualResetOutOfControlMode();
    }
    else Locomotive.Train.RequestToggleManualMode();
}
```
Otherwise, if you only want b), this PR does not affect your script and I don't need to do further modifications.

Csantucci:
Considering TCS_SCMT_IT.cs, there are 4 calls to RequestToggleManualMode():1) in line 1864, which have different aims:
1) in line 1864 of CheckSpad() the aim is to exit the OutOfControl mode after an unwanted Spad, while anyhow generating an emergency braking
2) In line 825 of Update() the aim is to re-enter Auto mode, after driver pressure of the re-arm button, from the manual mode entered by execution of line 184
3) in line 1828 of UpdateAttivazioneSR() (the method that handles the driver controlled passing over a signal at Stop) the aim is to automatically exit the Out of Control mode due to the Spad.
4) in line 1842 of UpdateAttivazioneSR() the aim is to automatically return to the original Auto mode after some seconds from the Manual mode triggered by line 1838.

cesarBLG:
OK, so you you can change situations 1 and 4 to use the new function ResetOutOfControlMode(). For situations 2 and 3 you are fine with RequestToggleManualMode().

PR #421 allowed trains in Explorer mode to go Out of Control. This means that you could do an undesired Explorer->OutOfControl->Manual sequence by calling RequestToggleManualMode().

As a transitory step I will add the lines I mentioned above, so your script works as expected in Explorer mode.

Csantucci:
OK, it seems that this should work this way, that is my scripts will continue running as now both when starting from an Auto mode and from an Explorer mode.
I know it won't occur often during an OR run, but maybe the deprecation trace line should be generated only once, and not every time the script generates the call to RequestToggleManualMode().
I have seen that you have added also a patch in SignalObject.cs. What is the aim of such patch?

Something out of scope: this is not the first time where there is the problem of compatibility of a script with evolving OR releases. I wonder if it were possible to insert in a script conditional compilation variables that enable/disable the runtime compilation of certain parts of script code depending from the OR release that starts the compilation.

cesarBLG:
> OK, it seems that this should work this way, that is my scripts will continue running as now both when starting from an Auto mode and from an Explorer mode.

Just note that your scripts weren't working in explorer mode since since #421 was merged. Hopefully they will start working again with this PR.

> I know it won't occur often during an OR run, but maybe the deprecation trace line should be generated only once, and not every time the script generates the call to RequestToggleManualMode().

I don't know if it is worthy to add a new variable and make the code bigger just for that warning, as it will happen few times.

> I have seen that you have added also a patch in SignalObject.cs. What is the aim of such patch?

It is fixing another undesired transition from explorer mode to auto signal. It is a very rare situation and it usually doesn't happen. The propagateRequest() method calls requestClearSignal, which isn't valid for explorer mode. For explorer mode, the clear request is propagated from CheckExplorerPath() in Train.cs.

> Something out of scope: this is not the first time where there is the problem of compatibility of a script with evolving OR releases. I wonder if it were possible to insert in a script conditional compilation variables that enable/disable the runtime compilation of certain parts of script code depending from the OR release that starts the compilation.

It is a pity that C# does not allow integer #defines like C/C++, as it would allow using things like #if VERSION > 5. Something similar can be obtained with the boolean #define in C#, but I don't know what would be the best way to include that into scripts.

Csantucci:
I have performed some first test with my script in activity mode, and all seems OK. In Explore mode I seem to have a problem with the script, but it does not depend from your changes. So your changes are OK to me.

Csantucci:
It works also in Explore mode

Csantucci (approved):
OK for me. Thank you for the improvement.

PR #702, opened 2022-08-26
Bug fix for https://bugs.launchpad.net/or/+bug/1987504 Server can't throw switches for clients in multiplayer

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/36481-behavior-of-the-switches-in-multiplayer/


cesarBLG (approved):
Thanks for looking at it. It seems OK to me.

PR #703, opened 2022-08-27
Add default signal function in case of SIGCFG errors

cesarBLG:
Prevents a crash if signal type is not defined in SIGCFG.


Sharpe49 (changes requested):
You should set this value directly during property initialization (at line 53).
`public SignalFunction Function { get; protected set; } = SignalFunction.UNKNOWN;`

Sharpe49 (approved):

PR #706, opened 2022-09-04
TCS Extensions

cesarBLG:
Add TCS function to get door state. It can be useful to cut-off traction if doors are open.

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Common/Scripting/TrainControlSystem.cs
@@ -188,6 +188,14 @@ namespace ORTS.Scripting.Api
         /// </summary>
         public Func<bool> IsRearCab;
         /// <summary>
+        /// True if left doors are open
+        /// </summary>
+        public Func<bool> AreLeftDoorsOpen;
+        /// <summary>
+        /// True if right doors are open
+        /// </summary>
+        public Func<bool> AreRightDoorsOpen;
+        /// <summary>
         /// True if train brake controller is in emergency position, otherwise false.
         /// </summary>
         public Func<bool> IsBrakeEmergency;
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
@@ -343,6 +343,8 @@ namespace Orts.Simulation.RollingStocks.SubSystems
                 Script.CurrentTrainMUDirection = () => Locomotive.Train.MUDirection; // Direction of train
                 Script.IsFlipped = () => Locomotive.Flipped;
                 Script.IsRearCab = () => Locomotive.UsingRearCab;
+                Script.AreLeftDoorsOpen = () => Locomotive.DoorLeftOpen;
+                Script.AreRightDoorsOpen = () => Locomotive.DoorRightOpen;
                 Script.IsBrakeEmergency = () => Locomotive.TrainBrakeController.EmergencyBraking;
                 Script.IsBrakeFullService = () => Locomotive.TrainBrakeController.TCSFullServiceBraking;
                 Script.PowerAuthorization = () => PowerAuthorization;


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Door.cs
```diff
@@ -0,0 +1,218 @@
...
+    {
+        
+        // Parameters
+        public float OpeningDelayS { get; set; } = 0f;
+        public float ClosingDelayS { get; set; } = 0f;
+
+        // Variables
+        readonly MSTSWagon Wagon;
+        protected Simulator Simulator => Wagon.Simulator;
+        public readonly bool RightSide;
```
Sharpe49, line 113: Maybe a DoorSide enum would be better.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Door.cs
```diff
@@ -0,0 +1,218 @@
...
+            }
+        }
+        public void SetDoorLock(bool lck)
+        {
+            Locked = lck;
+            if (lck) SetDoor(false);
+        }
+        public void SetDoor(bool open)
+        {
+            if (!Locked && open && (State == DoorState.Closed || State == DoorState.Closing))
```
Sharpe49, line 195: The State condition can be put inside a switch

Sharpe49 (changes requested):

cesarBLG:
Hi Cédric,
Thanks for your feedback. I've corrected everything you pointed out. I chose the lazy option of keeping the old code that used a boolean to determine door side, but this way is much more transparent.

Sharpe49 (changes requested):

Sharpe49:
Ok, there's still some remarks remaining and I added some suggestions.

cesarBLG:
Unfortunately the change from GetDoorState() to DoorState() for TCS scripts conflicts with the name of the DoorState enum. Any suggestion for a new name?

Sharpe49:
> Unfortunately the change from GetDoorState() to DoorState() for TCS scripts conflicts with the name of the DoorState enum. Any suggestion for a new name?

Maybe CurrentDoorState ?
I already used this kind of formulation in the PowerSupply classes.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Door.cs
```diff
@@ -0,0 +1,244 @@
...
+
+namespace Orts.Simulation.RollingStocks.SubSystems
+{
+    // Door state: do not change the order of this enum
+    public enum DoorState
+    {
+        Closed,
+        Closing,
+        Open,
+        Opening,
```
Sharpe49, line 30: It would have been more logic to keep Opening between Closing and Open.
Opening is an intermediate state between Closed and Open.

cesarBLG: Yeah, that was my original intention. However, for the Train Driving Window @mbm-OR wanted to show "Doors opening" until all doors were open. The reasoning was that if a locomotive has no doors it would immediately report "Doors open" even if the rest of the doors in the train weren't open yet. Another solution could be to define the number of doors available in each vehicle, and exclude vehicles without doors.

Csantucci:
César,
inserting this PR in ORNYMG and building I'm getting two warnings about fields 'MultiPlayerViewer.DoorLeftCount' and the analogous one for the right doors no more used. Checking the modifications I've seen that you have eliminated from MultiplayerViewer.cs the line
            if (UserInput.IsPressed(UserCommand.ControlDoorLeft)) MPManager.Notify((new MSGEvent(MPManager.GetUserName(), "DOORL", (++DoorLeftCount) % 2)).ToString());
and the analogous one for the right doors. May I ask why you eliminated those lines?

cesarBLG:
It seems I forgot to delete those variables, which are now unused. I moved the MP message to the SetDoors function. From my experience sometimes door state wasn't handled properly in MP mode.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Door.cs
cesarBLG: So do I revert this to Closed->Closing->Opening->Open for approval? I prefer this order too, but then some work has to be done for correct display at Train Driving Window

Sharpe49: Yes

cesarBLG: OK, done

Sharpe49 (approved):

PR #710, opened 2022-09-11
Fix GetDrawState when using upper characters in the name parameter

Sharpe49:
Draw state names are stored in lower case.
The GetDrawState function must convert its name parameter to lower case to work properly.


Csantucci (approved):
No need of other words, approved.

PR #712, opened 2022-09-17
Search for NORMAL signals if no argument is provided

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/36522-problem-with-signal-script
http://www.elvastower.com/forums/index.php?/topic/36544-testing-mode-crashes/

### DIFF
--- a/Source/Orts.Formats.Msts/SignalScripts.cs
+++ b/Source/Orts.Formats.Msts/SignalScripts.cs
@@ -2513,33 +2513,7 @@ namespace Orts.Formats.Msts
                 {
                     PartType = TypeIn;
                     PartParameter = IntIn;
-                    switch(IntIn)
-                    {
-                        case 0:
-                            SignalFunction = SignalFunction.NORMAL;
-                            break;
-                        case 1:
-                            SignalFunction = SignalFunction.DISTANCE;
-                            break;
-                        case 2:
-                            SignalFunction = SignalFunction.REPEATER;
-                            break;
-                        case 3:
-                            SignalFunction = SignalFunction.SHUNTING;
-                            break;
-                        case 4:
-                            SignalFunction = SignalFunction.INFO;
-                            break;
-                        case 5:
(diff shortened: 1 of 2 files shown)


Csantucci:
How did OR behave before the changes leading to crash were introduced? Did it default to NORMAL?

cesarBLG:
The default value for SIGSCR parameters is 0, which corresponded to the NORMAL signal type.

Csantucci:
> The default value for SIGSCR parameters is 0, which corresponded to the NORMAL signal type.

Then the fix is OK to me, and IMHO it should be introduced also in 1.5.

Csantucci (approved):

cesarBLG:
The problem was introduced during the conversion of the signal function from enum to class in #537, which is not present in 1.5

cesarBLG:
There are more errors related to this, so converting this to draft until they are solved.

cesarBLG:
@Sharpe49 It seems some scripts are really bad written, with usages like next_sig_lr() _[no argument]_ or next_sig_lr(5) _[number instead of SIGFN]_. My fix should allow this kind of weird syntax, but it's not elegant. I think it would be better to restore the previous code, keeping a Dictionary<int, SignalFunction>.

Any suggestions?

Sharpe49:
How about assigning values to the elements of the Enum (0, 1, 2, etc.) in order for them to always have the same value.
Then, during conversion, check if the integer value exists in the Enum (with Enum.IsDefined), then cast the value to SignalFunction (or set default value if it is undefined).

cesarBLG:
The values are already assigned to a enum (namely the MstsSignalFunction enum). The problem is from the conversion from that enum to a SignalFunction. I wasn't sure if using
```csharp
new SignalFunction((MstsSignalFunction)parameter_value);

was acceptable during the creation of the SCRParameter, due to the object allocation. I think it might be OK provided that SCRParameters are constructed during loading time and then never modified.

Csantucci (approved):

PR #715, opened 2022-09-27
Bug fix for //bugs.launchpad.net/or/+bug/1990930 Default snow terrtex not appearing

Csantucci:
…

### DIFF
--- a/Source/RunActivity/Viewer3D/Materials.cs
+++ b/Source/RunActivity/Viewer3D/Materials.cs
@@ -213,7 +213,9 @@ namespace Orts.Viewer3D
         {
             foreach (var path in TextureMarks.Where(kvp => !kvp.Value).Select(kvp => kvp.Key))
             {
-                Textures[path].Dispose();
+                // Sort of hack to avoid default snow terrtex not appearing
+                if (Textures[path] != SharedMaterialManager.DefaultSnowTexture && Textures[path] != SharedMaterialManager.DefaultDMSnowTexture) 
+                    Textures[path].Dispose();
                 Textures.Remove(path);
             }
         }


Sharpe49 (changes requested):
Nope, you should use the Mark(Texture2D texture) to mark the default texture for it not to be disposed.

Csantucci:
> Nope, you should use the SharedTextureManager.Mark(Texture2D texture) function to mark the default texture for it not to be disposed. It should be called in the Mark() function of the SharedMaterialManager. (BTW, it is weird that those textures are referenced in the SharedMaterialManager)

I'm not convinced. I'm sure you are taking into accont that there are many Textures[key, value] entries that have as value the same default snow texture. If you don't mark them all, the default texture will be disposed anyhow (isn't it so?).. And if you mark them all, you will have many unneeded entries in Textures[].,  With my fix all unneeded Textires[ ] entries will be removed, but the default texture won't be disposed.

Sharpe49:
If the default texture appears multiple times in the Textures list, that's an issue.
I think that's due to line 84 of Materials.cs.

Instead of:`else texture = defaultTexture;`
we should have `else return defaultTexture;`

cjakeman:
> If the default texture appears multiple times in the Textures list, that's an issue. I think that's due to line 84 of Materials.cs.
> 
> Instead of:`else texture = defaultTexture;` we should have `else return defaultTexture;`
> 
> Fixed in #720
> 
> With this fix, it is ensured that the default textures appear only once in the Textures list, and so, there will be only one Mark to do for each default texture.

Hi Cedric and Carlo,

Great that we have a fix for the bad default snow - many thanks.

With PR723 in place, do we still need this PR as well?
If we do, should it be changed now that we have PR 723?

Csantucci:
> > If the default texture appears multiple times in the Textures list, that's an issue. I think that's due to line 84 of Materials.cs.
> > Instead of:`else texture = defaultTexture;` we should have `else return defaultTexture;`
> > Fixed in #720
> > With this fix, it is ensured that the default textures appear only once in the Textures list, and so, there will be only one Mark to do for each default texture.
> 
> Hi Cedric and Carlo,
> 
> Great that we have a fix for the bad default snow - many thanks.
> 
> With PR723 in place, do we still need this PR as well? If we do, should it be changed now that we have PR 723?

Chris,
I've tested the release/1.5 branch, that includes Cédric's PR 723 but not my PR 720. It seems that Cédric's PR solves also the default slow issue. So PR 720 is no more needed. As soon as PR 723 will find its way also in the master/unstable release, I'll withdraw PR 720.

cjakeman:
> > Hi Cedric and Carlo,
> > Great that we have a fix for the bad default snow - many thanks.
> > With PR723 in place, do we still need this PR as well? If we do, should it be changed now that we have PR 723?
> 
> Chris, I've tested the release/1.5 branch, that includes Cédric's PR 723 but not my PR 720. It seems that Cédric's PR solves also the default slow issue. So PR 720 is no more needed. As soon as PR 723 will find its way also in the master/unstable release, I'll withdraw PR 720. This to avoid a transition time where the unstable release shows the default snow problem.

Hi Carlo,

That's a very satisfactory outcome.

Thanks for looking at this so quickly and for offering a solution in the first place.

Best wishes,

PR #720, opened 2022-10-03
Fix for default texture being added multiple times in Textures list

Sharpe49:
In the case a DDS texture doesn't exist, it is replaced by an ACE texture. If the ACE texture doesn't exist, the default texture is added in its place in the Textures list (which shouldn't happen).


Sharpe49:
Reopening it for inclusion in release/1.5 branch instead of master

PR #722, opened 2022-10-03
Fix Windows Forms deprecations in ActivityEditor

Sharpe49:
In preparation for the switch to .Net 5/6


pzgulyas (approved):
I don't see problems with this.

pzgulyas (approved):
I don't see why this couldn't be merged.

PR #726, opened 2022-10-08
Fix brake initialization problems

cesarBLG:
See http://www.elvastower.com/forums/index.php?/topic/36615-ep-brakes-misbehaving-in-unstable-release/page__pid__289427#entry289427 and http://www.elvastower.com/forums/index.php?/topic/36334-open-rails-15-is-coming-soon/page__view__findpost__p__289472


Csantucci (approved):
Thanks for that, César.

PR #736, opened 2022-10-24
Replaced truncated brake token table with lists

cjakeman:
See http://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/page__view__findpost__p__288675

Couldn't get a table to flow across multiple pages reliably, so replaced it with multiple lists.


cesarBLG (approved):
Thanks for solving this, Chris

PR #737, opened 2022-10-25
Fixed brake controller script not being loaded in timetable mode

Sharpe49:
BTW, the brake controller scripts that I have supports AI control, so it's not an issue to keep scripts active for a train in AI mode.
And I think I am probably the only that have written brake controller scripts.


cesarBLG (approved):
I don't think there should be problems either.

Sharpe49:
I think I should have selected the release 1.5 branch instead of master...

PR #740, opened 2022-10-26
Refactored the circuit breaker and the traction cut-off relay in order to use the same design pattern as the C# signal scripts (compatible with current scripts)

Sharpe49:
https://blueprints.launchpad.net/or/+spec/scripts-refactoring


cesarBLG (approved):
I don't see any issues, waiting for blueprint approval

PR #745, opened 2022-11-10
Fix digital displays in rear 3D cab

cesarBLG:
https://bugs.launchpad.net/or/+bug/1996158

http://www.elvastower.com/forums/index.php?/topic/24040-3d-cabs/page__view__findpost__p__290486


cesarBLG:
Taking out of draft this. The bug is solved for multiple unit trains with a cab in each side of the train, but not for locomotives with two cabs.
For locomotives more changes will be needed because currently controls for the rear and front cab are indistinguishable.

Csantucci (approved):
César, thanks for the fix

PR #749, opened 2022-11-19
OpenRailway Map

sweiland-openrails:
Feature request: https://www.elvastower.com/forums/index.php?/topic/36711-openrailway-map-added-to-open-rail-web-interface/

For the Open Rail - Web Interface I've added a Map entry. It shows an OpenRailway map with a locomotive icon on the current position of the train. Of course only useful for simulation of real railways. I use an extra monitor for displaying this map.

I've tested this for the Bernina Pass, the BNSF Scenic SUB and the Edingburgh - Glasgow route. When zooming in the position of the train is not always 100% accurate, but at least it gives you a nice idea where you are and heading for in the real world. The idea is from Flightgear, an open source flight simulator.


Source/RunActivity/Viewer3D/WebServices/Web/Map/index.js
```diff
@@ -0,0 +1,78 @@
...
+
+    hr.open("GET", "/API/MAP/", true);
+    hr.send();
+    hr.onreadystatechange = function () {
+        if (this.readyState == xmlHttpRequestCodeDone && this.status == httpCodeSuccess) {
+            var responseText = JSON.parse(hr.responseText);
+            if (responseText.length > 0) {
+                latLong = responseText.split(" ");
+                if (typeof locomotivMarker !== 'undefined') {
+                    if ((latLong[0] != latLongPrev[0]) && (latLong[1] != latLongPrev[1])) {
```
cjakeman, line 51: Should this not be || instead of && ?

sweiland-openrails: yes, you are right.

sweiland-openrails:
Should I have created the pull request against the unstable branch instead of the master branch?

cjakeman (commented):
Hi Siebren,

I tried your map using Demo Model 1 at Edinburgh. The map moves smoothly, but the loco icon stays at the centre but jitters. I guess that's because it moves with the map and then gets re-instated at the centre.

If there's no official way to avoid that, I guess we could use HTML to position the icon overlaid on top of the map and not use .addTo(map) at all.

You've used the recommended 500ms period - good. I didn't find any advantage in using a shorter period.

Thanks for using JsonParse as the other WebApis also do. I would prefer that an object was sent which is parsed by Json.Parse, rather than a string which needs parsing using .split(" "). I can see that the map library needs a string array, so maybe that's the object that should be passed.

Rather than calculate the Lat and Lon in the Web API, could we give the TrainCar class the ability to calculate its own Lat and Lon? Then the Web API becomes merely:

public LatLon LatLon() => Viewer.Simulator.PlayerLocomotive.GetLatLon();

and later on we can use .GetLatLon() for any locomotive.
I was surprised to find that there is no LatLon class in Open Rails. I would add one and append it to the "Orts.Simulation\Common\WorldLatLon.cs". 

I may have found a bug (see comments in the index.js code)

Could we have "locomotive" instead of "locomotiv"?
Also "LatLon" instead of "LatLong" (because "LatLon" is already used elsewhere in Open Rails)?

Hope that's helpful and look forward to your thoughts.

Best wishes,

Chris

sweiland-openrails:
hi Chris,

regarding the jitter you see. I've created a movie how I see it in the
browser. Do you see it differently?

regards, Siebren
 OR_ Map - Google Chrome 2022-11-21 12-12-21.mp4
<https://drive.google.com/file/d/1QGQHHZ3Hj6Pkb9VnaaZcD8-tVrRkro09/view?usp=drive_web>

Op zo 20 nov. 2022 om 20:27 schreef Chris Jakeman ***@***.***

cjakeman:
> Should I have created the pull request against the unstable branch instead of the master branch?

No, you did the right thing.

The Unstable Branch is populated by a script that runs every 15 mins and looks for suitable PRs from which to build a new Unstable version. It uses all PRs that are not in "Draft" and (have the label "for-unstable" or are submitted by a trusted developer) and do not conflict with any other PRs.

The resulting executables are available as a Zip archive from James' website.

Once approved and merged into the Master branch, this is published as a Testing Version on Friday evenings so the changes can be tried out more widely.

cjakeman:
> hi Chris, regarding the jitter you see. I've created a movie how I see it in the browser. Do you see it differently? 

Hi Siebren,

Thanks for the video. 

No, I see the same behaviour. Not a show-stopper, but would be nice to find a solution.

Best wishes,

Chris

cjakeman:
> for now this is the last commit. I think I've followed all your recommendations.

That's great, thanks.

I'll be free to have a look on Thursday.

It's usual to leave a PR in the Unstable Version for a couple of weeks before approving it and merging it into the master branch (i.e. Testing Version).

Best wishes,

Chris

cjakeman:
Hi Siebren,

> I'll be free to have a look on Thursday.
> 
> It's usual to leave a PR in the Unstable Version for a couple of weeks before approving it and merging it into the master branch (i.e. Testing Version).

Looks very good. Thanks.

I'll post a message on the forum to get some feedback.

Chris

cjakeman (approved):
Hi Siebren,

All looks good.

Approved with thanks and merged,

Chris

sweiland-openrails:
hi Chris,
for my understanding, what is the result of this merging? According to https://github.com/openrails/openrails/blob/6cf0e64619bf6b8735a66274ae4a19fc6fb0784d/Docs/Contributing.md this merging would make the change available in Testing. However after downloading the Testing Version of 30 december I do not find the Map entry on the OR website. Neither in the directory C:\OpenRails-Testing\Content\Web. What has gone wrong or am I misunderstanding this process? Or has it to do with the above "Some checks haven’t completed yet"?
regards, Siebren.

cjakeman:
Converted to Draft temporarily to see if changing it back will overcome the message "Waiting for status to be reported"

cjakeman:
> What has gone wrong or am I misunderstanding this process? Or has it to do with the below "Some checks haven’t completed yet"? 

No, you understand the process correctly. It's the message "Some checks haven't completed yet" that indicates a problem. I don't see why it should be a problem and I think I shall have to get help with this from James.

Best wishes,

Chris

PR #750, opened 2022-11-22
Bug fix for https://bugs.launchpad.net/or/+bug/1997482. Explore in Activity Mode, is not restored.

mbm-OR:
When the selected activity is "Explore in Activity Mode", this setting is not restored and "Explore Route" is displayed.
More info: https://bugs.launchpad.net/or/+bug/1997482


Csantucci (approved):
That's OK to me, thanks for that.

mbm-OR:
> That's OK to me, thanks for that.

Thanks, Carlo.

PR #751, opened 2022-11-24
Web interface to control cab controls with external hardware

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/34945-web-server/page__view__findpost__p__290210
https://blueprints.launchpad.net/or/+spec/webserver-cabcontrols

Uses #717 as base, so number of changes will be huge until that is merged.

PR #757, opened 2022-12-16
Unify RailDriver code implementations

cesarBLG:
https://blueprints.launchpad.net/or/+spec/raildriver-buttons
https://trello.com/c/zma8vdQA/228-support-button-remapping-for-raildriver

This introduces @perpetualKid implementation of RailDriver, which allows customizable buttons.


Csantucci:
To solve this http://www.elvastower.com/forums/index.php?/topic/24040-3d-cabs/page__view__findpost__p__294295
here is a suggested fix:
`diff --git a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
index 59159c92f..c81f35a0c 100644
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -246,7 +246,7 @@ public override void HandleUserInput(ElapsedTime elapsedTime)
                 foreach (var kvp in external.CabControls)
                 {
 
-                    if (_CabRenderer == null) break;
+                    if (_CabRenderer == null && !_has3DCabRenderer) break;
                     if (!kvp.Value.Changed) continue;
                     float val = kvp.Value.Value;
                     switch (kvp.Key.Item1.Type)
-- `

cesarBLG:
Thanks, Carlo. I uploaded the fix at #751 which is the first part of this PR, where this bug was introduced. I had to modify your fix because otherwise you would get a crash when using the web server.

Csantucci (approved):
This is now long time here and no one has complained; functional test was OK, and at a short glance the code is OK, so I approve it.

cesarBLG:
Thanks for review, Carlo!

PR #758, opened 2022-12-21
Steam wheel slip#1

peternewell:
This leverages off some changes introduced by cesarbl, and changes steam locomotive adhesion to use the same model as diesel locomotives.

Discussion - http://www.elvastower.com/forums/index.php?/topic/36775-steam-locomotive-adhesion/

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -2022,13 +2022,22 @@ of a relevant wagon (including diesel, steam or electric locomotives.
 
 OR supports the following special visual effects in a steam locomotive:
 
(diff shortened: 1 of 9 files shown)


cjakeman (approved):
Hi Peter,

It's great to see the informative comments you've included and find significant improvements as indicated by this line being deleted:

if (!(this is MSTSSteamLocomotive))

This is not a forum thread that I have been following, so I was surprised to find you've made so many improvements. 

I think you would find it helpful to split up your work into multiple PRs each with it's own focus, but that's a discussion we've had before, I think.

Best wishes,

Chris

PR #759, opened 2023-01-02
fix: Improve sky dome distortion causing bug #1471416

twpol:
Fixes https://bugs.launchpad.net/or/+bug/1471416

### DIFF
--- a/Source/RunActivity/Viewer3D/Cameras.cs
+++ b/Source/RunActivity/Viewer3D/Cameras.cs
@@ -171,7 +171,7 @@ namespace Orts.Viewer3D
         public void ScreenChanged()
         {
             var aspectRatio = (float)Viewer.DisplaySize.X / Viewer.DisplaySize.Y;
-            var farPlaneDistance = SkyConstants.skyRadius + 100;  // so far the sky is the biggest object in view
+            var farPlaneDistance = SkyPrimitive.SkyRadius + 100;  // so far the sky is the biggest object in view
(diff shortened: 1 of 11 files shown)


cjakeman (commented):
Hi James,

Interesting to see how you have re-factored this with a new SkyInterpolation class and enums and applied our style standards.
I also spotted that you have corrected errors for calculation of fog colour and the wvp matrix, so this PR is quite an improvement.

In your forum post, you reviewed the scheme as:

    Sky is formed of 3 layers (back to front):
        Cloud-less sky and night sky textures, blended according to time of day, and with sun effect added in (in the shader)
        Moon textures (phase is random currently)
        Clouds blended by overcast factor and animated by wind speed and direction

    Sky dome is 6000m away (but this doesn't matter since it is fixed to the camera and always drawn behind everything else)
    Sky dome is split into 32 segments (sides) around the center
    Sky dome is split into 16 segments (steps) from top to horizon and 1 more segment beyond horizon
    Moon is 5700m away
    Cloud dome is 5400m away
	
Would it be helpful to add that information into the comments?

Best wishes,
Chris

twpol:
> Would it be helpful to add that information into the comments?

I have added the relevant parts to the comment for `SkyPrimitive.TuneDomeComponentDegrees` which was already explaining some of it

The specific radii have changed and are all in constants just above the comment, so I didn't feel the need to repeat them inside the comment

cjakeman (approved):
Thanks for the extra comments, James.

Approved

PR #761, opened 2023-01-04
fix: Allow 100% cloud-free skys by making 0% overcast = 0% clouds

twpol:
Quick fix for being unable to do 100% cloud-free skys in http://www.elvastower.com/forums/index.php?/topic/34984-build-a-better-skydome-terragen-2/page__view__findpost__p__292401

Further work on the whole environmental effects will replace this in the future


Csantucci (approved):
That's OK for me, thanks!

PR #763, opened 2023-01-05
build: Remove use of SolutionDir which is not supported everywhere

twpol:
Certain tooling, especially that focused on .NET 5+, does not understand $SolutionDir - and since it is not actually necessary, by removing it more tools will work seamlessly with the codebase

### DIFF
--- a/Source/ContentChecker/ContentChecker.csproj
+++ b/Source/ContentChecker/ContentChecker.csproj
@@ -14,7 +14,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="MonoGame.Framework">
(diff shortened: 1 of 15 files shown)


cjakeman:
Marked as DRAFT temproarily to see if that restores the build of Unstable.

cjakeman:
James,

I playes with DRAFT and then not-for-unstable to find out if it's this PR that is blocking the build. Turns out it was PR 719 (Monogame).

Best wishes,
Chris

cjakeman (approved):
Hi James,

No problems reported and all looks fine.

Approved and merged.

Thanks,

Chris

PR #764, opened 2023-01-06
fix: Improve sky dome distortion causing bug #1471416

twpol:
Fixes https://bugs.launchpad.net/or/+bug/1471416

(PR #759 should have been a draft but since it is now merged, continuing the same bug fix here)


cjakeman (approved):
Hi James,

This is better with calculations in place of constants and the units of measure added.

Thanks,

Chris

PR #768, opened 2023-01-15
OpenRailway Map enhancements

sweiland-openrails:
Enhancements for OpenRailway Map:

- added web browser tab Open Rails icon (favicon)
- changed locomotive icon, now three icons: steam, electric, diesel
- added locomotive direction arrow
- added three OpenRailway layers to choose from: maximum speed, guage and electrification. Choice stored in registry and used the next opening of the map.
- added three layers with Open Rails data: track, named and rest. Choice also stored.

Don't know if these layers are usefull, anyway they can be switched on and off again.

Names of the layers are not translateble, is that a problem? Would be hard to implement.

For some routes the track layer takes too much performance from the web browser because of tracks size (for instance the MECoast line). But it gives you an idea of the layout of the track in the real world.

On the Ruebelandbahn for the "RLB Schneepflug Fahrbar" the locomotive arrow on the map is in the wrong direction (180 degrees). I can't find out why. Not seen this for any other locomotive.


cjakeman (changes requested):
Hi Siebren,

You've made some nice enhancements here and I have added the for-unstable label to this PR so your work is now included in the Unstable Version.

I'm happy that you have extended the ApiMap() to return the direction as well as the LatLon. Of course, once published, APIs cannot change as that would break the code based on them, but we haven't reached that point yet.

There are some changes I would like you to consider, if you wouldn't mind.

1. Could we have some images and text added to the Manual (section 19.7.1), please? If this is difficult for you to add, let me know and I can help or do it for you.

2. Storing settings is good practice and you've gone to some trouble to store these in the Windows Registry as part of the Open Rails settings.

I don't think this is a good idea because the Sample Web Pages are intended to be an extension to Open Rails rather than an integral part. The aim is to provide a robust API and some sample pages which users can tinker with and customise as they wish.

Would you please try saving the settings in the browser instead, using the local storage provided by the [HTML Web Storage API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API) instead?

3. C~ uses differnt capitalisation conventions from JavaScript and I notice that you have used camelCase for a number of methods and class properties. Please would you change them to PascalCase. We follow the Microsoft C# Coding Conventions (see link at [https://github.com/openrails/openrails/blob/master/Docs/Contributing.md](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md))

I hope this is not asking too much of you. I tried it out using the Northeast Corridor, as that matches LatLon better than Demo Model 1 and it's very effective.

sweiland-openrails:
hi Chris,

Done the changes you suggested. 

1. added some text and image to the manual, see a PM I send to you in Elvastower. Was wondering where I could see the result of this.
2. Web Storage API is better and not so difficult. Hop[e I've removed all the registry code I added.
3. I knew about PascalCase and camelCase for C#. However I thought variables where supposed to be named in camelCase. Please let me know if anything is still wrong of changed wrongly. No problem.

If you have more remarks, please let me know.

cjakeman:
>     1. added some text and image to the manual, see a PM I send to you in Elvastower. Was wondering where I could see the result of this.

I replied to your PR but, for other people reading this, once changes are committed, the HTML version of the Manual for Unstable is available at [https://open-rails.r...atform.html#id2](https://open-rails.readthedocs.io/en/unstable/software-platform.html#id2)


>     3. I knew about PascalCase and camelCase for C#. However I thought variables were supposed to be named in camelCase. Please let me know if anything is still wrong of changed wrongly. No problem.

In C# camelCase is used for method parameters. The Microsoft document says nothing about variables local to a method, but you are right that they are commonly camelCase too.

The class properties are now all in PascalCase - thanks. Would you also be able to change the class methods to PascalCase and the method parameters to camelCase as well, please? E.g.:

`private void initTypeOfLocomotive(string PowerSupplyName)
`

Thanks for this,
Chris

cjakeman (approved):
Hi Siebren,

Many thanks for doing that.

Approved.

Chris

PR #771, opened 2023-01-25
Action on Hold Signal is no longer invalidated if train is not stoppe…

Roeterdink:
…d in related station


cjakeman (approved):
Hi Rob,

Approved with thanks,

Chris

PR #775, opened 2023-01-25
Update turntable pools

Roeterdink:
Bug: https://bugs.launchpad.net/or/+bug/1929876


cjakeman (approved):
Hi Rob,

Approved with thanks,

Chris

PR #776, opened 2023-01-25
Update Receivers in Viewer.cs

Roeterdink:
Bug: https://bugs.launchpad.net/or/+bug/1929877


cjakeman (approved):
Hi Rob,

Approved with thanks,

Chris

PR #785, opened 2023-01-25
Incorrect calculation of distance to train ahead

Roeterdink:


cjakeman (approved):
Hi Rob,

Approved with thanks,

Chris

PR #802, opened 2023-02-21
Added support for activity location events to the TrackViewer

sweiland-openrails:
For the Contributed Track Viewer I've added support for Activity Location Events: the events that fire on a certain track location. Nothing fancy, just view, no editing. I wanted a better overview of these events on the tracks of a route.

Both events and event names selectable via menu "Track items". Event names in status bar selectable in menu "Statusbar".  Right mouse click on highlighted event gives pop message with the contents. Initially all the activities are shown after loading the route. If that's too much one of the activities can be selected via menu "File".


cjakeman (approved):
Hi Siebren,

I don't think we go through the formal approval process for supplmenetary programs like TrackViewer, so I am happy to approve this PR without references to a Trello card or a forum discussion.

Approved with thtanks,

Chris

PR #804, opened 2023-02-26
Bug #2008011 Bad key listings in manual

sweiland-openrails:
See http://www.elvastower.com/forums/index.php?/topic/36937-open-rail-manual/page__view__findpost__p__294566
and https://bugs.launchpad.net/or/+bug/2008011 

Open Rails Manual Release 1.5.1.263 Paragraph 13.7 (twice) and 13.8 (once) say that <Shift- E> is used to switch / gain control over own train. I believe this should be <Alt-E> instead.


cjakeman (approved):
Hi Siebren,

Thanks for sorting this out.

Approved and merged.

Chris

PR #805, opened 2023-02-26
bug 2008614 bad http link in manual

sweiland-openrails:
In: "3.3. Does Open Rails Require You to Have MSTS Installed?" link to www.burrinjuck.coalstonewcastle.com.au not clickable


cjakeman (approved):
Hi Siebren,

Thanks for fixing this.

Approved and merged.

Chris

PR #570, opened 2022-01-18
glTF 2.0 support with PBR lighting

pzgulyas:
glTF support with PBR lighting.
[Forum discussion thread](http://www.elvastower.com/forums/index.php?/topic/27639-orts-new-shape-format/page__view__getlastpost)
[Trello card](https://trello.com/c/QRIfnqhm/137-new-3d-model-format), and [another Trello card](https://trello.com/c/4oQrHy8o)
[Blueprint](https://blueprints.launchpad.net/or/+spec/or-3d-format)

**The visual test in assistance to it can be started by:**

1. `git clone https://github.com/KhronosGroup/glTF-Sample-Assets.git` into the `TRAINS/TRAINSET/glTF-Sample-Assets/` folder
2. Add VS debug options command line parameters e.g.: 
`-start -explorer "C:\Devel\MSTS\ROUTES\USA2\PATHS\tut6path.pat" "glTF-Sample-Assets" 12:00 1 0`
`-start -explorer "C:\Devel\MSTS\ROUTES\USA2\PATHS\tut6path.pat" "glTF-Sample-Assets&ABeautifulGame" 12:00 1 0`
`-start -explorer "C:\Devel\MSTS\ROUTES\USA2\PATHS\tut6path.pat" "glTF-Sample-Assets#3" 12:00 1 0`
3. When running this test, the animations can be switched on-off by [V], [Shift+,] and [Shift+.], go to the next/previous model with [Alt+PgUp], [Alt+PgDn].

**Tasks before this PR may be marked as ready to review.**

- [x] Solve the terrible terrain-flipping bug with certain gltf models, probably a render state not set or reset correctly.
- [x] Fix the various glitches in some models. Not all of them needs to be fixed, because some extensions they use are not necessarily supported.
  - [x] ABeautifulGame: corrupted geometry
  - [x] DragonAttenuation: corrupted geometry
  - [x] MetalRoughSpheresNoTextures doesn't show up
  - [x] ToyCar: corrupted geometry
  - [x] TextureTransformMultiTest: partial usage of interleaved buffers, separate vertex buffers and declarations would be needed, but XNA/MonoGame has the limitation that the declaration is part of the buffer, thus could be solved only by buffer duplications. Won't fix.
  - [x] Robert's test locomotive steam particles are sometimes covered up by something
- [x] Clean up the code so that only the vs_4_0 and ps_4_0 variants remain.
- [x] Turn down the spot light intensity a little.
- [x] Directional light – disable.
[...]

PR #807, opened 2023-03-02
Stop z-fighting by pushing world/view/projection multiplications onto the GPU

twpol:
Bug report: https://www.elvastower.com/forums/index.php?/topic/36606-z-fighting-issue/

This undoes some code we added ages ago to pre-multiply some of the shader matrices, I believe to reduce the amount of data being sent but turns out makes for some extra instability in the floating-point errors – and thus z-fighting is more likely.

Really, though, we should be removing the need for modellers to put objects mere millimetres apart, because this is simply not something real-time graphics pipelines are good at.

### DIFF
--- a/Source/RunActivity/Content/SceneryShader.fx
+++ b/Source/RunActivity/Content/SceneryShader.fx
@@ -24,7 +24,8 @@
 ////////////////////    G L O B A L   V A L U E S    ///////////////////////////
 
 float4x4 World;         // model -> world
-float4x4 WorldViewProjection;  // model -> world -> view -> projection
+float4x4 View;          // world -> view
+float4x4 Projection;    // view -> projection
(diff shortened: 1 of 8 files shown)


pzgulyas:
LGTM

pzgulyas:
Could this possibly be merged, so that I could proceed with merging this with #570 ?

twpol:
> Could this possibly be merged, so that I could proceed with merging this with #570 ?

Yes, sorry, I was planning to wait a week to check nothing had been broken but forgot about it. Please feel free to approve and merge this PR so you can put #570 back into unstable.

cjakeman (approved):
Hi James,

Looks good and so is feedback.

Approved with thanks,

Chris

PR #809, opened 2023-03-06
Some on-screen messages not suppressed, Bug #2008012

sweiland-openrails:
Warning and error messages can be suppressed now also. 

Options Checkbox in tab System has been changed in drop down checkbox "Messages Suppressed" with the following options:

- None: no messages suppressed (default)
- Information: informational messages and control confirmations suppressed
- Warning: warning messages and the ones mentioned above suppressed
- Error: error messages and the ones mentioned above suppressed

Ctrl+Alt+F10 cycles between these four.

Documentation "6.8.6. Control confirmations" has been changed accordingly.

Don't know if I have to do something concerning translations.

Name of the Settings variable SuppressConfirmations has not been changed. Name more or less is still correct. Changing this name gives many changes (I already programmed them, but removed them). That's because this name also makes it into the registry or .ini file. And the save play file. Old boolean value gives compatable behaviour.


cjakeman:
Hi Siebren,

Thanks for tackling this. I've added the label "For Unstable" so that your work is automatically included in the Unstable version.

Best wishes,

Chris

cjakeman (approved):
Hi Seibren,

Tidy code and I agree with the decision you've made here.

Approved with thanks,

Chris

PR #811, opened 2023-03-09
Bug fix for https://bugs.launchpad.net/or/+bug/2009842 DPU: wrong BP and MR values of helpers in display

Csantucci:
See also http://www.elvastower.com/forums/index.php?/topic/26249-remote-control-dpu-units/page__view__findpost__p__294519


pzgulyas (approved):
I see no problems with it.

Csantucci:
Thanks again, Peter.

PR #813, opened 2023-03-22
Refactored garbage generators

pzgulyas:
I've run the VS built-in profiler to see which classes generate the most garbage. This garbage causes micro-stuttering when they are collected by the garbage collector.

@Sharpe49 Cédric, could you please review these changes, because it seems many of them is your code originally. It would be good if you validated that my changes do exactly the same that you intended with the original.

### DIFF
--- a/Source/ORTS.Common/Filter.cs
+++ b/Source/ORTS.Common/Filter.cs
@@ -19,7 +19,6 @@ using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
-using System.Text;
 
 namespace ORTS.Common
 {
@@ -48,6 +47,12 @@ namespace ORTS.Common
     /// </summary>
     public class IIRFilter
     {
+        int NCoef;
+        List<float> ACoef;
+        List<float> BCoef;
+        float[] x;
+        float[] y;
+
         public IIRFilter()
         {
             /**************************************************************
@@ -73,25 +78,24 @@ namespace ORTS.Common
             z = 0.599839 + j -0.394883
             z = 0.599839 + j 0.394883
             ***************************************************************/
-            ACoef = new ArrayList();
-            ACoef.Add(0.00023973435363423468);
-            ACoef.Add(0.00047946870726846936);
-            ACoef.Add(0.00023973435363423468);
+            ACoef = new List<float>
+            {
+                0.00023973435363423468f,
+                0.00047946870726846936f,
+                0.00023973435363423468f
+            };
 
-            BCoef = new ArrayList();
-            BCoef.Add(1.00000000000000000000);
-            BCoef.Add(-1.94607498611971570000);
-            BCoef.Add(0.94703573071858904000);
+            BCoef = new List<float>
+            {
+                1.00000000000000000000f,
+                -1.94607498611971570000f,
+                0.94703573071858904000f
+            };
 
             NCoef = A.Count - 1;
 
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = new float[NCoef + 1];
+            y = new float[NCoef + 1];
 
             FilterType = FilterTypes.Bessel;
         }
@@ -102,19 +106,14 @@ namespace ORTS.Common
         /// <param name="a">A coefficients of the filter</param>
         /// <param name="b">B coefficients of the filter</param>
         /// <param name="type">Filter type</param>
-        public IIRFilter(ArrayList a, ArrayList b, FilterTypes type)
+        public IIRFilter(List<float> a, List<float> b, FilterTypes type)
         {
             FilterType = type;
             NCoef = a.Count - 1;
             ACoef = a;
             BCoef = b;
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = new float[NCoef + 1];
+            y = new float[NCoef + 1];
         }
 
         /// <summary>
@@ -126,11 +125,10 @@ namespace ORTS.Common
         /// <param name="samplingPeriod">Filter sampling period</param>
         public IIRFilter(FilterTypes type, int order, float cutoffFrequency, float samplingPeriod)
         {
-            NCoef = order;
-            A = new ArrayList();
-            B = new ArrayList();
-
             FilterType = type;
+            NCoef = order;
+            A = new List<float>();
+            B = new List<float>();
 
             switch (type)
             {
@@ -147,39 +145,25 @@ namespace ORTS.Common
             NCoef = A.Count - 1;
             ACoef = A;
             BCoef = B;
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = new float[NCoef + 1];
+            y = new float[NCoef + 1];
         }
-
-        int NCoef;
-        ArrayList ACoef;
-        ArrayList BCoef;
 
         /// <summary>
         /// A coefficients of the filter
         /// </summary>
-        public ArrayList A
+        public List<float> A
         {
             set
             {
-                if(NCoef <= 0)
+                if (NCoef <= 0)
                     NCoef = value.Count - 1;
-                x = new ArrayList();
-                y = new ArrayList();
-                for (int i = 0; i <= NCoef; i++)
-                {
-                    x.Add(0.0);
-                    y.Add(0.0);
-                }
+                x = new float[NCoef + 1];
+                y = new float[NCoef + 1];
                 if (ACoef == null)
-                    ACoef = new ArrayList();
+                    ACoef = new List<float>();
                 ACoef.Clear();
-                foreach (object obj in value)
+                foreach (var obj in value)
                 {
                     ACoef.Add(obj);
                 }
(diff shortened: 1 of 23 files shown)


pzgulyas:
In Explorer mode the most garbage comes now from the TrackCircuitSection.Reserve() function. I'm not sure exactly where, because I have never checked or modified this part of the program. One of my suspicions is the removeFromQueue() method:

https://github.com/openrails/openrails/blob/c9abcd22b114767ee17c60c3b7e3ab737189d62d/Source/Orts.Simulation/Simulation/Signalling/TrackCircuitSection.cs#L1313-L1316

Why is it needed to create a new List and a new Queue here? Can't be the existing thisQueue just cleared and reused? And if a helper List is needed, then it should be created in advance, before calling this method, and the same single object reused every time. According to the git blame it is also your code @Sharpe49 Cédric, but maybe Rob @Roeterdink has also some thoughts about this one.

Sharpe49:
Most of it is not my code. I just did a refactor.

pzgulyas:
Hi @cesarBLG, that line actually causes 2 allocations: one DieselEnum, and via the ToList() a new List. Actually I don't understand this whole code part. Why is it need to be so complicated? In my latest commit 6728bd5 I just replaced the foreach loop that tries to allocate this custom Enumerator with a for loop, which doesn't allocate anything. That is the place where the majority of the allocations are done because of this GetEnumerator(). If all the other places this used were also modified to use the for loop, this whole thing could be just deleted...

cesarBLG:
This is what I did for the Axles enumeration:
https://github.com/cesarBLG/openrails/blob/18fd0511d5bcd8fe85f67c657560e6caac234b23/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/Axle.cs#L256

It allows doing "foreach(var axle in Axles)" but, unless I'm missing something, it shouldn't allocate anything. If it's OK then we can do the same for Diesel Engines

pzgulyas:
First I wanted to blindly answer to you that, of course it doesn't allocate, because it just forwards the List's built-in enumerator. But then I had a second thought, and run the profiler on your patch. And it turns out, it still allocates! Just look at the results:

image

Then as a counter-test, I changed one of the functions in MSTSLocomotive as follows:

`        protected virtual void UpdateAxleDriveForce()
        {
            //foreach (var axle in LocomotiveAxles)
            for (var i = 0; i < LocomotiveAxles.Count; i++)
            {
                var axle = LocomotiveAxles[i];
                if (axle.DriveType == AxleDriveType.ForceDriven)
                {
                    axle.DriveForceN = TractiveForceN / LocomotiveAxles.Count;
                    if (SlipControlSystem == SlipControlType.Full)
                    {
                        // Simple slip control
                        // Motive force is reduced to the maximum adhesive force
                        // In wheelslip situations, motive force is set to zero
                        axle.DriveForceN = Math.Sign(axle.DriveForceN) * Math.Min(axle.AdhesionLimit * axle.AxleWeightN, Math.Abs(axle.DriveForceN));
                        if (axle.IsWheelSlip) axle.DriveForceN = 0;
                    }
                }
            }
        }
`
And that function disappeared from the allocators:

image

pzgulyas:
I think the reason is that your function still returns an IEnumerator, and the interfaces still don't have built in enumerators, a new one is getting allocated each time it is used. Unlike the Lists, that have "magic" built-in enumerators, thus don't need to allocate a new one.

So I'm sorry, but no, your implementation is still wrong for the hot path.

pzgulyas:
You can run these tests yourself to check, it is in VS Debug, Performance profiler menü.
image

cesarBLG:
Thanks for looking at it. I always try to be careful with explicit allocations in Update methods (things like new() , Clone() or string manipulation), but I didn't stop to think about these fancy functions.

For this particular thing about enumerators, I'm now curious because List<T>.Enumerator is a value type (i.e. is defined as a struct) and in theory it does not cause allocations, so maybe the casting from List<T>.Enumerator to IEnumerator is the one causing the allocation. I'll try to do some experiments with the tool you mention.

pzgulyas:
Yes, this seems to do the trick! I tested and all the related allocations disappeared. So we can say that actually it is a mistake for a class to implement the IEnumerable interface when it wants to minimize the allocations. Instead do how the MSDN says.

Source/ORTS.Common/Filter.cs
```diff
@@ -281,19 +257,13 @@ public enum FilterTypes
...
-            for (int n = NCoef; n > 0; n--)
-            {
-                x[n] = x[n - 1];
-                y[n] = y[n - 1];
-            }
             //Calculate the new output
-            x[0] = NewSample;
-            y[0] = (float)Convert.ToDouble(ACoef[0]) * (float)Convert.ToDouble(x[0]);
+            x.Insert(0, NewSample);
+            y.Insert(0, ACoef[0] * x[0]);
```
cesarBLG, line 262: I don't understand this modification. You are adding new elements to the list, which will make it grow unlimited, while the previous implementation was just shifting values and discarding the oldest ones, so it had fixed size.

I'd suggest using plain arrays instead of lists. The filter will always have constant size (actually only NCoef=1 is implemented). Also, ComputeButterworth has a terrible implementation, with the list being cleared and rebuilt again. So this has to be refactored in a way such that the allocations only happen at the constructor.

Source/ORTS.Common/Filter.cs
```diff
@@ -73,25 +78,24 @@ public IIRFilter()
...
 
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = Enumerable.Repeat(0f, NCoef).ToList();
+            y = Enumerable.Repeat(0f, NCoef).ToList();
```
cesarBLG, line 98: Previous implementation has dimension NCoef+1, yours has NCoef

Source/ORTS.Common/Filter.cs
```diff
@@ -102,19 +106,14 @@ public IIRFilter()
...
             BCoef = b;
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = Enumerable.Repeat(0f, NCoef).ToList();
+            y = Enumerable.Repeat(0f, NCoef).ToList();
```
cesarBLG, line 116: Same as above

Source/ORTS.Common/Filter.cs
```diff
@@ -147,39 +145,25 @@ public IIRFilter(FilterTypes type, int order, float cutoffFrequency, float sampl
...
             BCoef = B;
-            x = new ArrayList();
-            y = new ArrayList();
-            for (int i = 0; i <= NCoef; i++)
-            {
-                x.Add(0.0);
-                y.Add(0.0);
-            }
+            x = Enumerable.Repeat(0f, NCoef).ToList();
+            y = Enumerable.Repeat(0f, NCoef).ToList();
```
cesarBLG, line 149: Same as above

Source/ORTS.Common/Filter.cs
```diff
@@ -147,39 +145,25 @@ public IIRFilter(FilterTypes type, int order, float cutoffFrequency, float sampl
...
                     NCoef = value.Count - 1;
-                x = new ArrayList();
-                y = new ArrayList();
-                for (int i = 0; i <= NCoef; i++)
-                {
-                    x.Add(0.0);
-                    y.Add(0.0);
-                }
+                x = Enumerable.Repeat(0f, NCoef).ToList();
+                y = Enumerable.Repeat(0f, NCoef).ToList();
```
cesarBLG, line 162: Same as above

Source/ORTS.Common/Filter.cs
```diff
@@ -193,23 +177,18 @@ public ArrayList A
...
```
cesarBLG, line 187: Same as above

Source/ORTS.Common/Filter.cs
```diff
@@ -321,21 +291,12 @@ public float Filter(float NewSample, float samplingPeriod)
...
-            for (int n = NCoef; n > 0; n--)
-            {
-                x[n] = x[n - 1];
-                y[n] = y[n - 1];
-            }
-            //Calculate the new output
-            x[0] = NewSample;
-            y[0] = (float)ACoef[0] * (float)x[0];
+            x.Insert(0, NewSample);
+            y.Insert(0, ACoef[0] * x[0]);
```
cesarBLG, line 295: Same comment

cesarBLG (changes requested):
Everything looks OK to me, except for the IIRFilter class. It had a terrible implementation in the first place, but I don't think your changes are doing any better. Please see my comments.

Source/ORTS.Common/Filter.cs
pzgulyas: César, you are right in these! I've fixed my mistakes by your suggestion.

Source/Orts.Simulation/Simulation/Signalling/TrainQueue.cs
```diff
@@ -40,5 +40,21 @@ public bool ContainsTrain(Train.TrainRouted thisTrain)
...
+        /// </summary>
+        /// <param name="thisTrain"></param>
+        public void RemoveTrain(Train.TrainRouted thisTrain)
+        {
+            if (thisTrain == null) return;
+            var queueCount = Count; // loop through the original count to cover all trains in the queue
+            for (var i = 0; i < queueCount; i++)
+            {
+                var queueTrain = Dequeue();
+                if (queueTrain != null & queueTrain.Train != thisTrain.Train)
```
cesarBLG, line 55:
```suggestion
                if (queueTrain != null && queueTrain.Train != thisTrain.Train)
```

PR #818, opened 2023-04-11
Allow independent drive axles for locomotives

cesarBLG:

This is the first part of a set of PRs which will allow the existence of multiple engines, each one connected to a separate wheelset. Each Axle is independent, so it's possible to get wheelslip in one axle but not in another one.

Further changes to be done include:
- Move steam engine physics to a separate class, and allow multiple steam engines (WIP by @peternewell)
- Create a new block for the .eng file which allows the definition of multiple axles (each axle must define Wheel Radius, engines attached to it (or none if not driven) and which animations correspond to each axle).
- Link all existing engine types (diesel-mechanical, AC, steam) to different axles, and calculate tractive force for every engine separately.
- Link wheel animations to axles to get visual slip effects (partially implemented by me, but conflicts with #570 so won't be included here at the moment)

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AITrain.cs
@@ -6346,7 +6346,7 @@ namespace Orts.Simulation.AIs
                 if (car is MSTSLocomotive)
                 {
                     var loco = car as MSTSLocomotive;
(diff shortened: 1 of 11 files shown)


peternewell (approved):
HI Cesar,

This has been in the Unstable version for a while, so I think that it is safe to merge it.

Thanks for your work on it.

Peter

cesarBLG:
Thanks for approving

PR #819, opened 2023-04-12
Headlight switch makes sound when not appropriate

sweiland-openrails:
If trying to switch the headlight beyond or below the last position, for certain locomotives who have this defined, a click sound is to be heard. In real live the switch would not make such a sound.

Just a small fix.


pzgulyas (approved):
Looks good!

PR #822, opened 2023-04-23
battery switch two buttons option in combination with a delay

sweiland-openrails:
When using the battery switch two buttons option in combination with a delay:

Engine (
    ORTSBattery (
    Mode ( PushButtons )
    Delay ( 2s )
    DefaultOn ( 1 )
)

battery does not get switched on/off.

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -4315,7 +4315,8 @@ Three behaviours are available:
 
 In real life, the battery switch may not
 close instantly, so you can add a delay with the optional parameter
-``ORTSBattery( Delay ( ) )`` (by default in seconds).
+``ORTSBattery( Delay ( ) )`` (by default in seconds). When delay is used in combination with push buttons 
+you have to keep pressing the button until the battery is (dis)connected.  
 
 It is possible for the battery switch to be switched on at the start of the simulation.
 To activate this behaviour, you can add the optional parameter ``ORTSBattery( DefaultOn ( 1 ) )``


cesarBLG (changes requested):
I don't think this is 100% correct. This will cause the physical button on the cab to return to 0 while being pressed, as well as triggering the sound before the button has been released. I'm analyzing the original code and I don't see any obvious issue, so it might be related to your CVF. I'll try to test it.

Okay, so I tested and it works correctly on my side, so I think this is not an OpenRails bug, but something in your CVF (probably you are using Style ( ONOFF ) instead of Style ( PRESSED ))

sweiland-openrails:
> I don't think this is 100% correct. This will cause the physical button on the cab to return to 0 while being pressed, as well as triggering the sound before the button has been released. I'm analyzing the original code and I don't see any obvious issue, so it might be related to your CVF. I'll try to test it.
> 
> Okay, so I tested and it works correctly on my side, so I think this is not an OpenRails bug, but something in your CVF (probably you are using Style ( ONOFF ) instead of Style ( PRESSED ))

Ok, I'll give it another look. I'm rather new to Open Rails, where does the abbreviation CVF stand for?

cesarBLG:
Don't worry. CVF is the CabView File, which is usually located inside the CabView folder of a locomotive. The syntax of CVF files allows several types of buttons: Style ( ONOFF ) means that when you push a button it'll remain active until you push it again, while Style ( PRESSED ) means that it goes back to the original position when you release it.

I tested the bug by adding the block you suggest:
```
Engine (
ORTSBattery (
Mode ( PushButtons )
Delay ( 2s )
DefaultOn ( 1 )
)
```
And added the following to the .cvf file of the locomotive:
```
TwoState (
	Type ( ORTS_BATTERY_SWITCH_COMMAND_BUTTON_CLOSE TWO_STATE )
	Position ( 100 420 24 12 )
	Graphic ( "battery_on.ace" )
	NumFrames ( 2 2 1 )
	Style ( PRESSED )
	MouseControl ( 1 )
)
TwoState (
	Type ( ORTS_BATTERY_SWITCH_COMMAND_BUTTON_OPEN TWO_STATE )
	Position ( 115 420 24 12 )
	Graphic ( "battery_off.ace" )
	NumFrames ( 2 2 1 )
	Style ( PRESSED )
	MouseControl ( 1 )
)
```
And I'm getting the correct behaviour, so I couldn't reproduce the bug.
It works for me also using the keyboard (Insert for closing, Ctrl+Insert for opening).

sweiland-openrails:
Bought a secondhand SDD, so some more diskspace to experiment.

Still I see the same problem. I have no battery statements in my .cvf file. I'm testing with the keys <Insert> and <Ctrl+Insert>. Works correctly without the delay or a delay of 0 seconds. But with a delay the Battery switch in the Train Driving Info (F5) does not move. I'll have another look in the code...

Thanks for explaining me about the .cvf file.

cesarBLG:
Ah, I now see your thinking. The original code requires you to keep the button pressed _until_ the battery is connected, so you have to keep <Ctrl+Insert> pressed for two seconds. This is the correct behaviour for other electrical switches as the circuit breaker, but I don't know about the battery.

You could post a question at Elvas Tower to see if the button has to kept pushed until switch is closed in real trains.

sweiland-openrails:
Ah, thanks. I thought I tested keeping the button pressed. Tested once again and it works as you describe, keeping the button pressed. Seems logical to me, pressing the button for two seconds. Can I Undo this Pull Request by pressing the "Close with comment" button?

cesarBLG (dismissed):
Yeah, I think adding it to the documentation is a good idea.

sweiland-openrails:
If I now merge this change I not only have a changed .rst documentation file, but also the  BatterySwitch.cs file. That's because this .cs file contains 3 extra bytes EF BB BF (which is some UTF-8 byte order mark). I don't want to pollute the history of this BatterySwitch.cs with an extra non existent change. Any idea how I can get rid of this change? Git does not see this as a change, so I cannot commit the file without 3 extra bytes.
Or can I just close the Pull Request without the merge and redo the fix only in the .rst file?

cesarBLG:
I fixed it with a simple trick

cesarBLG (approved):

sweiland-openrails:
thanks...

PR #826, opened 2023-05-01
Bug fix for Crash when switching cab https://bugs.launchpad.net/or/+bug/2018230

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/35640-inserting-cabview-controls-in-side-views-of-2d-cabs/page__view__findpost__p__296512


cesarBLG (approved):
It seems correct

Csantucci:
> It seems correct

Thanks, César.

PR #732, opened 2022-10-23
Improvements for air brakes

cesarBLG:
- New EP system with release/apply wires that locally modify BP pressure at each vehicle
- Implement control reservoir for brake distributors
- Add keyboard command for BP overcharge
- Improve dynamic brake blending

http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__292763
http://www.elvastower.com/forums/index.php?/topic/36791-emergency-brake-application-bc-pressure-limit/
http://www.elvastower.com/forums/index.php?/topic/37000-air-brakes-emergency-function-not-working-in-recent-releases/
http://www.elvastower.com/forums/index.php?/topic/36817-stupid-independent-brakesshowtopic=36817/

https://blueprints.launchpad.net/or/+spec/braking-enhancement

PR #828, opened 2023-05-03
Harmless part of "Improvements for braking systems"

cesarBLG:
Divided #732 in two parts: a simple one and the biggest one which cannot be splitted since almost everything is connected.

https://blueprints.launchpad.net/or/+spec/braking-enhancement


Csantucci (approved):
Yes, this part is simple and already tested in the Unstable release within the former version of PR #732, so it's OK for me.

cesarBLG:
Thanks, Carlo

PR #830, opened 2023-05-07
Electric locomotive hot start

cesarBLG:
User-configurable hot start for electric locomotives:
- Battery connected
- Master key on
- Pantograph up
- Circuit breaker closed

https://trello.com/c/Rp0vQlVc/564-electric-locomotive-hot-start

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/BatterySwitch.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/BatterySwitch.cs
@@ -343,7 +343,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
                     {
                         case ModeType.Switch:
                             CommandSwitch = false;
-                            Wagon.SignalEvent(Event.BatterySwitchCommandOn);
+                            Wagon.SignalEvent(Event.BatterySwitchCommandOff);
                             break;
                         case ModeType.PushButtons:
                             CommandButtonOff = true;
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/BatterySwitch.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/BatterySwitch.cs
@@ -43,6 +43,8 @@ namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
         public bool CommandButtonOn { get; protected set; } = false;
         public bool CommandButtonOff { get; protected set; } = false;
         public bool On { get; protected set; } = false;
+        protected bool QuickPowerOn;
+        protected bool QuickPowerOff;
 
         public BatterySwitch(MSTSWagon wagon)
         {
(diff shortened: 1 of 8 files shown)


Csantucci (commented):
I have two notes:
1) Timetable mode is not covered by this; the player locomotive always starts initialized
2) Battery state is not covered by this: if the battery is off at startup, it must be set on by the player.
WDYT about covering also these two points?

cesarBLG:
1. I'm not sure about whether this should also affect timetable mode. Timetabled trains are previously started in sidings, in general. Maybe it would be better to individually select it per train? I can ask about it at the forum.
2. This is a bug in the power supply classes. The battery only reacts to "QuickPowerOn" if it is of Switch type, but not for Push Buttons. I'll look at it.

Csantucci:
In timetable mode the player train starts with pantos up (differently from activity mode). Therefore with your option, that has pantos up as default, nothing would change, but a new possibility would be available for players wanting to start with pantos down.
Of course it would be desirable to have the possibility to set within each timetable (and also within each activity) whether the player locomotive starts with pantos up or down, but this is an additional feature.

cesarBLG:
Looking at the code, diesel locomotives also start with their engines running in timetable mode regardless of the setting, so if we change that, it should be changed for diesels too.

Csantucci:
Yes, I agree.
As you suggested, it may be a good idea to shortly mention this evolution in the forum topic, simply stating that such option checkboxes apply to all operating modes (explorer, explore in activity mode, activity, timetable).

cesarBLG:
I think I won't change timetable mode. See http://www.elvastower.com/forums/index.php?/topic/36927-timetable-mode-switch-off-power-and-sound/page__view__findpost__p__294414
Rob says
> Note that the setting of the general option in the menu "start player without power" no longer will affect player trains in timetable mode.

Csantucci:
OK, I see.  So, let's not touch timetable mode.

Csantucci:
I think that it remains valid that you write a post in the forum specifying that what you are doing has no effect to timetable mode, and you can add the link to Rob's post. Also in the manual it should be mentioned that those options haven't effect to timetable mode.

cesarBLG:
Done as suggested

Csantucci (approved):
That's OK now for me, thanks!

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/BatterySwitch.cs
```diff
@@ -302,6 +322,37 @@ public virtual void HandleEvent(PowerSupplyEvent evt)
...
+                case PowerSupplyEvent.QuickPowerOff:
+                    switch (Mode)
+                    {
+                        case ModeType.Switch:
+                            CommandSwitch = false;
+                            Wagon.SignalEvent(Event.BatterySwitchCommandOn);
+                            break;
+                        case ModeType.PushButtons:
+                            CommandButtonOff = true;
+                            Wagon.SignalEvent(Event.BatterySwitchCommandOn);
```
pzgulyas, line 350: Are these all right? Shouldn't it be Event.BatterySwitchCommandOff?

Csantucci (approved):
After insertion of gpz' suggestion, It should be OK now.

cesarBLG:
Thanks, Carlo

PR #832, opened 2023-05-12
Fix Cruise Control together with Train Brake

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/33841-cruise-control/page__view__findpost__p__296962


Csantucci (approved):
That's OK to me, thanks!

Csantucci (approved):
After the check I confirm my OK, but please wait 3-4 days before merging.

cesarBLG:
Thanks Carlo. Let's wait some time just in case.
There are other places in the CC code where operator precedence is not clear, so I'd expect more bugs like this in the future.
It's strange that for some programming languages (a || b && c) is evaluated as ((a || b) && c) while in C# it's evaluated as (a || (b && c)) (which is more sensible, by the way). So in this case I agree with CodeFactor that parenthesis are needed to make logic explicit.

PR #835, opened 2023-05-19
Correct Brake Shoe Force Calculation

peternewell:
Adjust Brake Force to use Brake Shoe Force - 

Fixes an error in calculating the brake force applied to the wagon.

Bug report: https://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__297184


cjakeman (approved):
Hi Peter,

Thanks for working to unravel this complicated situation and for providing documented support for legacy arrangements and a better way of specifying brake force.

Approved.

Chris

PR #838, opened 2023-05-27
Code Cleanup

Looky1173:
I ran Visual Studio's _Code Cleanup_ function and it fixed some style errors. It should not have modified any functionality.


twpol:
I would advise _against_ doing things like this PR (I have tried it before); it will conflict with almost every other open PR _and_ every developer's local development of features.

IMHO we should instead encourage people to improve the formatting of the files _they are editing_ so that things progressively get better.

If you must format lots of files at once, I strongly recommend doing specific small chunks of files (by directory) in each PR. It'll still cause a lot of problems for people writing code, though.

PR #840, opened 2023-06-07
Bug fix for sanding max speed not working

SteelFill:
Baby's first pull request. Probably something weird looking or wrong in here (yes there are 7 commits, 6 of those were me trying to un-break something), just let me know if that's the case.

Fix for the issue noted here [http://www.elvastower.com/forums/index.php?/topic/37178-ineffective-train-brakes/page__view__findpost__p__297807](url) and here [https://bugs.launchpad.net/or/+bug/2023083](url)
- when above the speed defined by Sanding(), use of sand no longer increases adhesion
- sand is no longer consumed if sanding is blocked
- prevented sanding from _reducing_ adhesion, which was possible in some cases
- general clean-up and simplification of sand friction code

Even when sanding is blocked, sanding sounds and cab sand indicator lights will still function. It's not realistic, but fixing that would require a rework of the way sand events are handled and I'm not interested in that right now.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -2827,6 +2827,7 @@ public List<CabView> CabViewList = new List<CabView>();
  
             float max0 = DrvWheelWeightKg * 9.81f * uMax;  //Ahesion limit in [N]
             float max1;
+            float SandingFrictionFactor = 1;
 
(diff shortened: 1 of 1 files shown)


cesarBLG:
Thanks for the fix! The code is now much simpler than it was. I tagged the PR as "for-unstable" so that it appears in next unstable release, so users can test it.

cesarBLG (approved):
It has been in the unstable release for a while without any complains. It is OK to me.
I did some cleanup to revert undesired commits.

SteelFill:
Thanks for figuring out how to clean up those commits, that was really driving me crazy and I didn't come across any non-intimidating ways to get it to ignore the commits I reverted.

PR #843, opened 2023-06-17
Fix for https://bugs.launchpad.net/or/+bug/2024295  Freight weight only added for first wagon

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/37200-loading-not-calculated-correctly-for-train/


cesarBLG (approved):
Thanks for the simple fix

Csantucci:
> Thanks for the simple fix

Thanks to you, César

PR #845, opened 2023-06-21
Save axle integrator status

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/37210-engine-hp-and-load-is-zero-on-resuming-saved-activity/page__pid__298201#entry298201

Save some integrator data to get a more stable adhesion resume


Csantucci (approved):
This seems reasonable. That's OK to me.

PR #849, opened 2023-07-01
Contributed projects and forking clarity

twpol:
The contributed projects have been missing from this document but the intended policy has not changed; they should be accepted with little beyond code review, like documentation, manual, and translations.

The forking and branching recommendations have been clarified.


cjakeman (approved):
Hi James,

All looks fine. Approved with thanks,

Chris

PR #850, opened 2023-07-01
Correct output of sound system info

twpol:
All the other output in this method goes to `output`, just this one line was overlooked.


Csantucci (approved):
No objections to this.

PR #853, opened 2023-07-15
Notify out of focus

sweiland-openrails:
When this option is checked, on the corners of the Open Rails main window a red rectangle is shown when the window is not in focus. Just a reminder that the main window currently does not react on keystrokes.

see discusssion on Elvas Tower: http://www.elvastower.com/forums/index.php?/topic/37312-main-window-out-of-focus/page__p__298601#entry298601

trello card: [https://trello.com/c...ow-out-of-focus](https://trello.com/c/JTnLvcMj/568-get-alerted-when-main-window-out-of-focus)


cjakeman (commented):
Hi Siebren,

Well done with this. 

Would you mind changing over the default so that the checkbox is checked? That way, users of the Unstable Version might discover it even if they weren't expecting it.

I've talked it over with James Ross and, provided that there are no serious objections during the trial period, we would prefer to release it without the checkbox and so always in operation.

Best wishes,

Chris

sweiland-openrails:
Done, setting the default of this option to true.

However this red square might be annoying to people who are recording video. It might be an idea to use the option "SuppressConfirmations" ([http://www.elvastower.com/forums/index.php?/topic/36936-comfirmation-supression-for-all-messages/page__p__294784__hl__siebren__fromsearch__1#entry294784](url)) to switch this option off also. When set to suppress all messages switch the "out of foucus" option off also.

sweiland-openrails:
hi Chris,

this "Notify out of focus" option has a default True in the latest unstable version (Open Rails U2023.07.23-0837 2023-07-23 09-51-30). User is still able to switch it off in the options.

regards, Siebren.

cjakeman:
Hi Siebren,

> Done, setting the default of this option to true.

Thanks for that.
 

> However this red square might be annoying to people who are recording video. It might be an idea to use the option "SuppressConfirmations" ([http://www.elvastower.com/forums/index.php?/topic/36936-comfirmation-supression-for-all-messages/page__p__294784__hl__siebren__fromsearch__1#entry294784](url)) to switch this option off also. When set to suppress all messages switch the "out of foucus" option off also.

Good point, which we hadn't thought of, and I like this proposal to tie the behaviour to the SuppressConfirmations option.

Chris

cjakeman:
Thanks, Phillip,

> Hey there, I've been playtesting some of my own changes but in the process I believe I've encountered some errors related to this feature and saved games.
> 
> If I start an activity with this option unchecked, save the activity, check this option, then attempt to load the save, I get this error:

Thanks for reporting this. I can confirm that it happens to me too.

 
> I'm surprised a UI toggle would cause loading a save to fail, but it's the only option I changed between saving and loading these sessions.

The Save and Resume mechanism is a little complicated. There was a time when any change invalidated all saves from earlier version. This ensured that incompatible saves were always blocked, but was very inconvenient as Testing versions can change weekly.

We now have a warning, if Open Rails sees that the version has changed, and also a check value in case the length of the saved file does not match what Resume is expecting.

That's some background, but I haven't found the cause. I hope to look some more, when I have time.

Best wishes,

Chris

sweiland-openrails:
hi Phillip, Chris,

This problem is not that important, as I'm planning to remove this option from the Options dialog. And ty the behavior to the "SuppressConfirmations" option.

But it's curious, I see the problem also. I think it does not have to do with different versions, as both Philip and I see it happen on one and the same version. Had a look in the code, but still no clue. The fact that the .save file is a binary file does not help. But I do nog give up, I'l have another look.

regards, Siebren.

cjakeman:
Hi Siebren, Phillip,

> The fact that the .save file is a binary file does not help. But I do not give up, I'll have another look.

I find that the Save file contains an extra 10 bytes (of value 0) towards the end when this option is checked.

cjakeman:
Hi Siebren, Phillip,
 
> I find that the Save file contains an extra 10 bytes (of value 0) towards the end when this option is checked.

And this is due to WindowManager.Save() method which saves the window Orts.Viewer3D.Popups.OutOfFocusWindow when the option is checked and not when it isn't.

I don't think this issue will disappear when the option is merged with the Messages option. Instead, it might be necessary to filter out just this window from the WindowManager.Save() method (with a comment explaining why).

A simpler solution might be to create the window even when it's not used.

sweiland-openrails:
hi Chris,

Now I understand, thanks for looking into this. This extra "window" wants to save it's position, just like the other ones. Not necessary, but that's how the code works. It's a bit of misuse of this code, but it did work so nicely.

I always try to minimize the code running for switchable enhancements. Hence the not creating of the window when not in this option is not in use. Your suggestion will work. However saving this position is not necessary so I'll have a look if suppressing this saving in this case might be doable.

Such a small enhancement ...

regards, Siebren.

sweiland-openrails:
Chris, Phillip,

Problem with starting from save has been fixed. I suppressed both the save and the restore of the out of focus "window". Fix to be found in the unstable build https://james-ross.co.uk/projects/or/builds/2023-08-01_13-43-13_U2023.08.01-1237/Open%20Rails%20U2023.08.01-1237%202023-08-01%2013-43-13.zip.

Next I'm gonna replace the "Notify out-of-focus" option by the "Messages Suppressed" option as per your suggestion Chris.

regards, Siebren.

sweiland-openrails:
Final version of this enhancement to be found at https://james-ross.co.uk/projects/or/builds/2023-08-01_20-43-12_U2023.08.01-1937/Open%20Rails%20U2023.08.01-1937%202023-08-01%2020-43-12.zip.

regards, Siebren.

Looky1173:
Hi, I found a bug: when starting the game in windowed mode, then switching to full screen with `Alt + Enter`, the new dimensions of the window won't be taken into account when it goes out of focus.
image

sweiland-openrails:
hi Looky,

fixed the problem you found. My fault, never tested this scenario. Thanks for rising this.

regards, Siebren.

sweiland-openrails:
hi Chris,

Do you have some time to review this red square enhancement? I think the functionality is now ok and two weeks ago I fixed a bug for this PR. After ending my PR's I could tidy up my local disk a bit. But no hurry.

regards, Siebren.

cjakeman (approved):
Hi Siebren,

You've ironed out all the issues.

Thanks for this. Approved and merged.

Chris

sweiland-openrails:
thanks Chris. Never heard the expression "to iron out", but I understood it at one. Nice.

regards, Siebren.

PR #857, opened 2023-07-30
Adding Air Flow Meters

SteelFill:
This pull request covers the work I've done to add air flow meters for the air braking system. This type of gauge measures the volumetric flow rate of air out of the main res into the brake pipe, useful for gauging train brake behavior during recharge and release (though does nothing during applications).

Discuss observations about this change [here on E-T.](http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__299318)
[Trello card](https://trello.com/c/hSZFdMLl)

- Added air flow meter to all relevant HUD components (Alt + F5 HUD, F5 HUD, DPU HUD, control confirmations). Shows cubic feet per minute (cfm) in imperial, liters per second in metric. This will appear on the HUD for any air-braked locomotive with an equalizing res, even if the real locomotive doesn't actually have a flow meter.
- Added "ORTS_AIR_FLOW_METER" cabview control type. Default units are cubic meters per second (this is what's used under the hood) but custom units are CUBIC_FT_MIN for cubic feet per minute, LITERS_MIN for liters per minute, and LITERS_S for liters per second.
- Changed behavior of equalizing reservoir and brake pipe exhausting/recharging to have some smoothing, trying to replicate some behaviors I observed in test reports and make the behavior of the airflow meter more convincing. This will affect all air braked content, even MSTS original, but I did some testing and found it works quite well across a bunch of different routes, new and old.

Air flow meters are particularly common in North American designs but appear rarely elsewhere, so I'm curious to know if there are meters out there that use units other than cfm and L/s. Also, word of warning, the air flow rates show what is actually going on in the air brake physics, but this flow is limited by the setting of `ORTSBrakePipeChargingRate`. The flow meter does not lie, so if you are seeing flows different than what's expected, that means the brake pipe charging rate isn't right.

### DIFF
--- a/Source/Documentation/Manual/appendices.rst
+++ b/Source/Documentation/Manual/appendices.rst
@@ -61,6 +61,7 @@ Mass Flow                                           g/h
 Volumetric Flow         m^3/s           air flow    m^3/s
                                         meters
 \                                                   ft^3/min
+\                                                   L/min
 \                                                   L/s
 Speed                   m/s             other       m/s         m/s             meter per second
 \                                                   km/h
--- a/Source/Documentation/Manual/appendices.rst
+++ b/Source/Documentation/Manual/appendices.rst
@@ -58,6 +58,10 @@ Voltage                 volt                        V
 Mass Flow                                           g/h
 \                                                   kg/h
 \                       lb/h                        lb/h        lb/h
+Volumetric Flow         m^3/s           air flow    m^3/s
+                                        meters
+\                                                   ft^3/min
+\                                                   L/s
 Speed                   m/s             other       m/s         m/s             meter per second
 \                                                   km/h
 \                                                   kph         kph             kilometer per hour
--- a/Source/Documentation/Manual/cabs.rst
+++ b/Source/Documentation/Manual/cabs.rst
@@ -679,6 +679,37 @@ An example of implementation of the above controls in a .cvf file follows::
 		)
 
 
+Air Flow Meter
+--------------
+
+.. index::
+   single:  ORTS_AIR_FLOW_METER
+
+This cabview control is used on some locomotives, particularly in North America, to show the
+volumetric flow rate of air moving from the main res to the brake pipe during release/recharge.
+Such an indication can be used to determine when brake pipe charging is complete,
+measure the amount of brake pipe leakage, and so on.
+The control will only function on locomotives with air brakes.
+
+Here is an example implementation of ORTS_AIR_FLOW_METER as an analog dial::
+
+
+		Dial (
+			Type ( ORTS_AIR_FLOW_METER DIAL )
+			Position ( 258 271 1 32 )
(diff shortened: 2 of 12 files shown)


cjakeman (changes requested):
Hi Phillip,

Thanks for the link to the forum post. To follow our approval process, would you please add a brief Trello road-map card (see [Contributing.md](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)). It would be helpful to have a link in the PR to this Trello card.

Thanks,
Chris

SteelFill:
> Hi Phillip,
> 
> Thanks for the link to the forum post. To follow our approval process, would you please add a brief Trello road-map card (see [Contributing.md](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)). It would be helpful to have a link in the PR to this Trello card.
> 
> Thanks, Chris

I sent in a trello card a few days ago, it's on the roadmap now, just needs to be moved from the unsorted category.

cjakeman (changes requested):
Hi Phillip,

Could you also add some entries to the Units of Measure table in the appendices.rst file, please?

Thanks,
Chris

cjakeman:
Hi Phillip,

> Default units are cubic meters per second (this is what's used under the hood) but custom units are CUBIC_FT_MIN for cubic feet per minute and LITRES_S for liters per second.

I'm not comfortable with a default unit that the user cannot specify. That arrangement doesn't apply to any other UoM. How about adding cubic metres per second along with the other two?

Chris

SteelFill:
Hey Chris,

Got those updates in. As for the cab view units, it looks like most cab view controls are set up so that they don't even check for the unit setting if it's the same as what's under the hood (setting a fuel gauge to LITRES is the same as setting no units at all). In my case, I did go for an explicit 'CUBIC_M_S' case as I'm contemplating the possibility of setting the default case to something other than the under-the-hood m^3/s (have yet to find evidence cubic meters per second is even reasonable for a cab gauge)...might be unneeded though.

SteelFill:
@cjakeman could you review this PR? I haven't made any changes in a few weeks and folks seem pleased with the result.
I was interested in waiting for suggestions for new units of measure but since I haven't gotten any suggestions apart from liters/minute I'm ready to close the case on this PR and leave any future changes for different PRs. I was also hoping to have this PR merged before I began investigating air system leakage as the air flow meter is a very useful tool for testing air system leaks.

cjakeman (approved):
Hi Phillip,

Sorry you had to wait for this approval. It got lost in my stream of emails.

Approved and merged now.

Thanks,

Chris

SteelFill:
Thanks Chris, with all my current PRs taken care of I feel more comfortable about starting something new-less chance I conflict with myself, and my attention isn't as split that way

PR #844, opened 2023-06-19
Animate axles separately

cesarBLG:
https://blueprints.launchpad.net/or/+spec/enhance-adhesion

This feature links each Axle to its animated parts, to get visual wheelslips only on some axles.

PR #858, opened 2023-08-02
Fix wheel animation problems

cesarBLG:
Fix problems with wheel animations introduced in #844
Reported at http://www.elvastower.com/forums/index.php?/topic/37367-weird-graphics-problem/


Csantucci:
César,
I see that you have added the line
`AdvancedAdhesionModel = false;
for traintypes AI and AI_PLAYERHOSTING, while traintype REMOTE is managed together with traintypes PLAYER and PLAYERDRIVEN. Have you considered how to manage the REMOTE trains referring to Adhesion?`

cesarBLG:
AdvancedAdhesionModel is only a flag which tells other parts of the code whether advanced adhesion is being used or not. Whenever AdvancedAdhesion is called, AdvancedAdhesionModel shall be set to true, and otherwise it shall be false. I'm not changing anything regarding physics with this fix, I'm just properly indicating which model is being used for AI trains.

Remote trains have their physics updated in a different computer, so in my opinion they shouldn't be updated there, but I prefer not to change that.

Csantucci (approved):
Changes OK to me.

PR #859, opened 2023-08-06
Steam adhesion bug#1

peternewell:
Fix a bug in steam adhesion code which produces too much wheel force when throttle set to zero.

http://www.elvastower.com/forums/index.php?/topic/37327-issues-with-advanced-adhesion/

https://bugs.launchpad.net/or/+bug/2030305

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
@@ -142,6 +142,8 @@ namespace Orts.Simulation.RollingStocks
         bool AIFireOverride = false; // Flag to show ai fire has has been overriden
         bool InjectorLockedOut = false; // Flag to lock injectors from changing within a fixed period of time
 
+        float DebugTimerS;
+
         // Aux Tender Parameters
         public bool AuxTenderMoveFlag = false; // Flag to indicate whether train has moved
         bool SteamIsAuxTenderCoupled = false;
@@ -2105,29 +2107,7 @@ namespace Orts.Simulation.RollingStocks
                 for (int i = 0; i < NumCylinders; i++)
                 {
 
-//                    Trace.TraceInformation("NumCyl - {0} i {1}", NumCylinders, i);
-                    
-                    // float realCrankAngleRad = (float)(LocomotiveAxle.AxlePositionRad + i * WheelCrankAngleDiffRad[i]);
-                    float realCrankAngleRad = (float)(LocomotiveAxles[0].AxlePositionRad);
-                    float normalisedCrankAngleRad = 0;
-
-                    realCrankAngleRad = (float)(MathHelper.WrapAngle(realCrankAngleRad));
-
-                    if (realCrankAngleRad < 0)
-                    {
-                        realCrankAngleRad = (float)(2.0f * Math.PI + realCrankAngleRad); // angle must be maintained in a +ve range, ie 0 - 360
-                    }
-
-                    // Normalise crank angle so that it is a value between 0 and 360 starting at the real crank angle difference
-                    if (realCrankAngleRad >= WheelCrankAngleDiffRad[i])
-                    {
-                        normalisedCrankAngleRad = realCrankAngleRad - WheelCrankAngleDiffRad[i];
-                    }
-                    else
-                    {
-                        float diff = WheelCrankAngleDiffRad[i] - realCrankAngleRad;
-                        normalisedCrankAngleRad = (float)(2.0f * Math.PI - diff);
-                    }
+                    float normalisedCrankAngleRad = NormalisedCrankAngle(i);
 
                     // Exhaust crank angle
                     float exhaustCrankAngleRad = 0;
@@ -5085,7 +5065,7 @@ namespace Orts.Simulation.RollingStocks
 
                 for (int i = 0; i < NumCylinders; i++)
                 {
-                    float crankAngleRad = (float)(LocomotiveAxles[0].AxlePositionRad + i * WheelCrankAngleDiffRad[i]);
+                    float crankAngleRad = (float)(LocomotiveAxles[0].AxlePositionRad + WheelCrankAngleDiffRad[i]);
 
                     testCrankAngle = crankAngleRad;
 
@@ -5122,9 +5102,11 @@ namespace Orts.Simulation.RollingStocks
                         backwardCylinderPosition = crankCylinderPosition;
                     }
 
-/*
-                    Trace.TraceInformation("Cyl {0} Position {1} testcrankAng {2} crankAngle {3} CrankPosition {4} forwardCrankPosition {5} backwardCrankPosition {6}", i+1, LocomotiveAxle.AxlePositionRad, MathHelper.ToDegrees(testCrankAngle), MathHelper.ToDegrees(crankAngleRad), crankCylinderPosition, forwardCylinderPosition, backwardCylinderPosition);
-*/
+                    /*
+                                        Trace.TraceInformation("Cyl {0} Position {1} testcrankAng {2} crankAngle {3} CrankPosition {4} forwardCrankPosition {5} backwardCrankPosition {6}", i+1, LocomotiveAxle.AxlePositionRad, MathHelper.ToDegrees(testCrankAngle), MathHelper.ToDegrees(crankAngleRad), crankCylinderPosition, forwardCylinderPosition, backwardCylinderPosition);
+                    */
+
+                    float normalisedCrankAngleRad = NormalisedCrankAngle( i );
 
                     // Crank angles
                     float sin = (float)Math.Sin(crankAngleRad);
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -5402,6 +5386,46 @@ protected override void UpdateTractiveForce(float elapsedClockSeconds, float t,
         }
 
 
+        /// <summary>
+        /// Normalise crank angle so that it is a value between 0 and 360 starting at the real crank angle difference
+        /// </summary>
+        private float NormalisedCrankAngle( int cylinderNumber)
```
cesarBLG, line 5392: This whole function can be simplified to
```csharp
float crankPositionRad = LocomotiveAxles[0].AxlePositionRad + WheelCrankAngleDiffRad[cylinderNumber];
return (float)(MathHelper.WrapAngle(crankPositionRad)+Math.PI) // Range angle from 0 to 360º
```

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -5223,7 +5194,7 @@ protected override void UpdateTractiveForce(float elapsedClockSeconds, float t,
                     float connectRodInertiaForcelbf = inertiaSpeedCorrectionFactor * connectRodInertiaAngleFactor * ConnectingRodWeightLb;
 
                     // Account for the position of the crosshead position. In other words it depends upon whether the Rods and Reciporating gear is above or below the axle.
-                    if (crankAngleRad > 0 && crankAngleRad < Math.PI)
+                    if (crankAngleRad > 0 && crankAngleRad < Math.PI && normalisedCrankAngleRad < Math.PI)
```
cesarBLG, line 5197: I don't understand why you are using crankAngleRad and normalisedCrankAngleRad. You should use only normalisedCrankAngleRad, which I'd suggest that you rename to crankAngleRad, i.e.
```csharp
float crankAngleRad = NormalisedCrankAngle(i);
```

cesarBLG (changes requested):
I'm glad to see that the problem is now solved. I have a couple of comments which should make the code more clear.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: For force calculation a reference angle of 0 -180 - 0 angle is used, this lines up with the result tables in the reference book that I got the feature description from (and is also what was used in the confirmation spreadsheet that we have been circulating).

To confirm whether the locomotive cylinder is on the forward or backward stroke the normalised crank angle is used. It is also used to calculate the opening and closing points for the cylinder exhausts.

Hence my preference is to maintain both of them.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I have tried the suggested changes and it doesn't line up with the current values that I am calculating.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -5402,6 +5386,46 @@ protected override void UpdateTractiveForce(float elapsedClockSeconds, float t,
...
+
+            realCrankAngleRad = (float)(MathHelper.WrapAngle(realCrankAngleRad));
+
+            if (realCrankAngleRad < 0)
+            {
+                realCrankAngleRad = (float)(2.0f * Math.PI + realCrankAngleRad); // angle must be maintained in a +ve range, ie 0 - 360
+            }
+
+            
+            if (cylinderNumber == 0) // initial cylinder
```
cesarBLG, line 5405: I don't see why the first cylinder should be different to the rest. Wouldn't be better to set WheelCrankAngleDiffRad[0]=0 ?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
cesarBLG: Obviously not, sorry. You can keep your code

cesarBLG: Obviously not, sorry. This is my proposal, to reduce number of operations:
```csharp
float normalisedCrankAngleRad = (float)MathHelper.WrapAngle(LocomotiveAxles[0].AxlePositionRad+WheelCrankAngleDiffRad[cylinderNumber]);
if (normalisedCrankAngleRad < 0) normalisedCrankAngleRad += (float)(2*Math.PI);
return normalisedCrankAngleRad;
```

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
cesarBLG: I find really confusing that some of the formulas range from -180º to 180º and others go from 0º to 360º. I strongly suggest that you rewrite everything to use the same convention. Otherwise, other people won't understand what you are doing, and it's going to be bug prone.

In particular, the following line is terribly confusing, and has to be removed.
```csharp
if (crankAngleRad > 0 && crankAngleRad < Math.PI && normalisedCrankAngleRad < Math.PI)
```
It would be enough to test if (crankAngleRad > 0), since the rest of conditions are equivalent to that.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: It was already set = 0, so I have removed this irrelevant bit of code.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I think that the crank angle range should remain in the range 0 - 180 - 0 so that it matches the reference data tables and makes it easier to compare to actual results in OR. This makes debugging easier.

I have added some additional notes to try and make it easier to understand what the different angle ranges are used for, etc.

I have simplified some of the code based upon your suggested feedback.

Hopefully this is an agreeable compromise.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -5223,7 +5177,8 @@ protected override void UpdateTractiveForce(float elapsedClockSeconds, float t,
                     float connectRodInertiaForcelbf = inertiaSpeedCorrectionFactor * connectRodInertiaAngleFactor * ConnectingRodWeightLb;
 
                     // Account for the position of the crosshead position. In other words it depends upon whether the Rods and Reciporating gear is above or below the axle.
-                    if (crankAngleRad > 0 && crankAngleRad < Math.PI)
+                    // The crosshead will be -ve for normalised angles between 0 - 180, and +ve for normalised angles between 180 - 360
+                    if (normalisedCrankAngleRad > 0 && normalisedCrankAngleRad < Math.PI )
```
cesarBLG, line 5181: I still don't know why normalisedCrankAngleRad is used, when testing against crankAngleRad > 0 has the same effect

peternewell: crankAngleRad operates in the range 0-180-0 and therefore it will always be positive (ie it never goes negative, and hence the conditions in the if statement will always be met) in every cycle, and within every cycle there are two points that are the same. Hence it is not possible to differentiate between the up or down parts of the cycle.

normalisedCranAngleRad operates in the range 0-360 so it is possible to determine the up and down part of the cycle.

I have spent some time trying to get the best solution, and testing it to make sure that it works, so I would like to leave it as it is.

cesarBLG: I'm sorry, but I cannot approve it yet as it is really confusing. You say:
> crankAngleRad operates in the range 0-180-0 and therefore it will always be positive

but that's not true. In the code, crankAngleRad operates between -180º->0º->180º, so it can be positive and negative. I don't see a need to fix that, I'm OK with it. If that is working, then there is no need to use normalisedCrankAngle too in UpdateTractiveForce() (you can use it in UpdateFX if you wish). Condition 0<normalisedCrankAngle<180, is equivalent to 0<crankAngleRad

cesarBLG (approved):
Thanks for the extensive tests in order to fix this. I approve

peternewell:
Thanks for approving it.

PR #862, opened 2023-08-13
Fix engine leak problem

cesarBLG:
http://www.elvastower.com/forums/index.php?/topic/37388-acela-hhp-8-engineering/page__view__findpost__p__299632


SteelFill (approved):
Fix makes sense to me, haven't noticed any problems in playtesting.

Approved

cesarBLG:
Thanks for the approval

PR #829, opened 2023-05-03
Improvements for air brakes #3 - Emergency valves

cesarBLG:
Divided #732 in three parts

This covers emergency dump valves and related issues

https://blueprints.launchpad.net/or/+spec/braking-enhancement

PR #864, opened 2023-08-17
Fixes for Undesired Emergency Applications

SteelFill:
Some fixes/enhancements for emergency braking to reduce the chance of emergency braking occurring when not intended, and increase the change of emergency braking occurring when intended.

Expands upon PR #829
Related to https://blueprints.launchpad.net/or/+spec/braking-enhancement
Bug report: https://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__299828

- When an anglecock is opened, rather than opening instantly it is (automatically) gradually opened over 30 seconds to limit flow rate between the two segments of brake pipe. This happens whether the anglecock was opened manually with the F9 menu or automatically with brake initialization/connect brake hose keybinds. This simulates real life switching practice, where the conductor will open an anglecock only partially for some moments after coupling (with some deviations to accommodate inaccuracies in the brake pipe flow model). In my testing, this is enough to prevent emergency braking from occurring when coupling to a train with an empty brake pipe without interfering with regular operations. A partially open anglecock will be indicated on the brake hud with "/" instead of the usual "+" and "-".
- Rather than using instantaneous change in brake pipe pressure to determine when to make an emergency application, the change is averaged over a quarter of a second. This prevents small reductions in brake pipe over short periods of time from triggering an emergency application. Instead, the reduction must be sustained, which is a sign that the emergency application should actually be happening. Similar has been done in real brake valve designs to intentionally reduce emergency sensitivity.
- Change to emergency detection to more reliably get consists that have been uncoupled from the player train to go into emergency (unless air has been bottled) without using CPU power to fully simulate the static consist.
- Slightly off-topic change to accelerated application so it too uses the averaged brake pipe change. This significantly improves the stability of accelerated application, reducing the chance that accelerated application causes an emergency application (also reduces the chance of accelerated application venting brake pipe below the selected pressure).
- Tweak to locomotive brake valve code to allow emergency braking to bypass the equalizing reservoir. In real locomotives, emergency braking opens up a hole in the brake pipe that lets air out regardless of the equalizing reservoir. (Previously, emergency braking would not vent air below the ER pressure, and would even recharge the brake pipe during emergency braking if the brake pipe pressure managed to drop below the ER, this had no bearing on real life.) For most engine files, this will only slightly accelerate emergency applications, but every bit helps.
- Restore the pre-version 1.5 emergency brake behavior for train cars which haven't been updated with new emergency brake features, to ensure those train cars go to emergency when expected (ie: when brake pipe pressure drops too low).

No changes require any file editing, but this will impact all air braked locomotives even back to MSTS. Recommend testing switching scenarios, coupling to cars and charging the brake pipe by setting anglecocks rather than using the initialize brakes cheat, and seeing if triple valves go to emergency due to coupling (shouldn't happen ever). Also try uncoupling from a train without closing the anglecocks, then recouple and verify that the brake pipe pressure dropped and triple valves changed to emergency in the static consist. Apply and release brakes on a variety of content, and check that triple valves don't go to emergency unless an emergency application is commanded from the locomotive or opening in brake pipe.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -670,7 +670,9 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                 }
                 if (BrakeLine1PressurePSI - dpPipe < 0)
                 {
+                    // Prevent pipe pressure from going negative, also reset quick service to prevent runaway condition
                     dpPipe = BrakeLine1PressurePSI;
+                    QuickServiceActive = false;
                 }
 
                 if (TripleValveState != ValveState.Emergency && BrakeLine1PressurePSI < AuxResPressurePSI + 1)
@@ -694,7 +696,9 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                 AutoCylPressurePSI += dp;
                 BrakeLine1PressurePSI -= dpPipe;
 
-                if (QuickServiceActive && AutoCylPressurePSI > QuickServiceLimitPSI) // Reset quick service if brake cylinder is above limiting valve setting
+                // Reset quick service if brake cylinder is above limiting valve setting
+                // Also reset quick service if cylinders manage to equalize to prevent runaway condition
+                if (QuickServiceActive && (AutoCylPressurePSI > QuickServiceLimitPSI || AutoCylPressurePSI >= AuxResPressurePSI)) 
                     QuickServiceActive = false;
 
                 if (TripleValveState == ValveState.Emergency)
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -91,6 +91,9 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
         protected float prevBrakePipePressurePSI = 0f;
         protected float prevBrakePipePressurePSI_sound = 0f;
 
+        protected float BrakePipeChangePSIpS;
+        protected SmoothedData SmoothedBrakePipeChangePSIpS;
+
 
         /// <summary>
         /// EP brake holding valve. Needs to be closed (Lap) in case of brake application or holding.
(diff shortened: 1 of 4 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -455,12 +473,15 @@ public void UpdateTripleValveState(float elapsedClockSeconds)
             var prevState = TripleValveState;
             var valveType = (Car as MSTSWagon).BrakeValve;
             bool disableGradient = !(Car.Train.LeadLocomotive is MSTSLocomotive) && Car.Train.TrainType != Orts.Simulation.Physics.Train.TRAINTYPE.STATIC;
+            // Small workaround to allow trains to more reliably go into emergency after uncoupling
+            bool emergencyTripped = (Car.Train.TrainType == Orts.Simulation.Physics.Train.TRAINTYPE.STATIC) ?
+                BrakeLine1PressurePSI <= 0.75f * EmergResPressurePSI * AuxCylVolumeRatio / (AuxCylVolumeRatio + 1) : Math.Max(-SmoothedBrakePipeChangePSIpS.SmoothedValue, 0) > EmergencyValveActuationRatePSIpS;
```
cesarBLG, line 478: I wonder if we shall use the pressure condition to trigger the emergency not only for static trains, but also if EmergencyValveActuationRate hasn't been defined in the .wag file. Currently, if EmergencyValveActuationRate==0, it is set to a default value at Initialize(). Switching to a pressure threshold by default would restore backwards compatibility for old stock.

SteelFill: Some users have pointed out that emergency braking doesn't work properly on old content since 1.5, so backwards compatibility definitely needs work. While reinstating the pressure threshold would be the least disruptive way to fix old content, it would be more realistic if vent valves were added where missing. "Where missing" is pretty loaded though, I doubt a 20 year old .wag file will have enough information to accurately decide if it should have vent valves or not.

cesarBLG: The problem was that I initially set EmergencyValveApplicationRate to zero when I implemented the feature. Then I changed it to a nonzero value as default, due to complains. Neither of the solutions has been acceptable, so the only think that ensures that old, badly configured content still works is the pressure threshold I think.

cesarBLG (approved):
Thanks for solving the issue. Changes look great

PR #867, opened 2023-08-23
Code quality improvements

Looky1173:
After realising in hindsight how foolish of an idea #838 was, I wrote a Node.js script that compiles a list of the changed files in all open pull request. This enables me to tidy up files independently, one at a time, that are currently not being worked on. I've excluded the "not-for-unstable" and PRs #570 & #757 because they had not seen activity in a _very_ long time. Therefore, this PR should not conflict with any active pull requests. If the OR management team approves of such PRs, I would be happy to regularly submit more of them.

This PR has not modified any functionality.

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AIPath.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AIPath.cs
@@ -22,14 +22,14 @@
  * Could this be used for player trains also?
  * 
  */
-using Orts.Formats.Msts;
-using Orts.Formats.OR;
-using ORTS.Common;
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using Orts.Formats.Msts;
+using Orts.Formats.OR;
+using ORTS.Common;
 
 namespace Orts.Simulation.AIs
 {
@@ -42,10 +42,10 @@ namespace Orts.Simulation.AIs
 #if ACTIVITY_EDITOR
         public ORRouteConfig orRouteConfig { get; protected set; }
 #endif
-        public AIPathNode FirstNode;    // path starting node
-        //public AIPathNode LastVisitedNode; not used anymore
+        public AIPathNode FirstNode; // Path starting node
+        //public AIPathNode LastVisitedNode; not used anymore --- TODO: Remove?
         public List<AIPathNode> Nodes = new List<AIPathNode>();
-        public string pathName; //name of the path to be able to print it.
+        public string pathName; // Name of the path to be able to print it.
 
         /// <summary>
         /// Creates an AIPath from PAT file information.
@@ -84,11 +84,11 @@ namespace Orts.Simulation.AIs
                 node.Index = i;
                 TrPathNode tpn = patFile.TrPathNodes[i];
 
-                // find TVNindex to next main node.
+                // Find TVNindex to next main node.
                 if (tpn.HasNextMainNode)
                 {
                     node.NextMainNode = Nodes[(int)tpn.nextMainNode];
-                    node.NextMainTVNIndex = node.FindTVNIndex(node.NextMainNode, TDB, tsectiondat, i == 0 ? -1 : Nodes[i-1].NextMainTVNIndex );
+                    node.NextMainTVNIndex = node.FindTVNIndex(node.NextMainNode, TDB, tsectiondat, i == 0 ? -1 : Nodes[i - 1].NextMainTVNIndex);
                     if (node.JunctionIndex >= 0)
                         node.IsFacingPoint = TestFacingPoint(node.JunctionIndex, node.NextMainTVNIndex);
                     if (node.NextMainTVNIndex < 0)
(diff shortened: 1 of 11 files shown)


Looky1173:
This PR is now ready for review & merging. I have not touched any game logics, just cleaned up the style of the code, so in theory this PR does not modify game behaviour at all.

cjakeman:
Hi Looky,

> script that compiles a list of the changed files in all open pull request. This enables me to tidy up files independently, one at a time, that are currently not being worked on.

That's a really neat solution to the issue. 

I believe James wanted to check your "tidy up" rules against the ones already in place for Visual Studio before approving this PR (hope I've understood that correctly). I've prompted him in an email about something else, so hopefully he will be able to review this soon.

Best wishes,

Chris

Looky1173:
Hello,

The script can be found on [Replit](https://replit.com/@Looky1173/ORTS-PRs-Changed-Files) for reference.

Here are my code cleanup options as well:
image

Thank you for contacting James about this,
Looky

Source/Orts.Simulation/Simulation/AIs/AIPath.cs
```diff
@@ -516,7 +509,7 @@ public static int FindJunctionOrEndIndex(WorldLocation location, TrackDB trackDB
                 float dz = tn.UiD.Z - location.Location.Z;
                 dz += (tn.UiD.TileZ - location.TileZ) * 2048;
                 float dy = tn.UiD.Y - location.Location.Y;
-                float d = dx * dx + dy * dy + dz * dz;
+                float d = (dx * dx) + (dy * dy) + (dz * dz);
```
cesarBLG, line 512: Are this kind of unneeded parenthesis addition going to be enforced? It may be because I come from the Math/Physics fields instead of Computer Science, but I find the new expression more difficult to read. It's not a common thing there to put parenthesis around multiplications, everyone knows that they go first. When I see a parenthesis, my mind thinks "there is a different operator precedence than usual", to then see (in this situation) that the operator precedence is the standard one, effectively loosing more time than required when analyzing a formula.

To me, adding parenthesis to
```csharp
scalar = (dx*dx) + (dy*dy)
```
is the same as
```csharp
float a = (3+5);
```
i.e. both add unnecessary parenthesis that make the code harder to read. Why is it assumed that you know = goes after +, but not that addition goes after multiplication? To me this is completely inconsistent, and makes the code harder to read. My opinion is don't add parenthesis to mathematical operations unless they are required.

I'll still use parenthesis only when needed, unless this new rule is enforced. I would be very happy also if equations used for Physics calculations are kept as they are. Some of them are very complex, and having to waste time to match unnecessary parenthesis does no help.

Maybe there's a good reason to flood the code with parenthesis, but at the moment I can only see its drawbacks.

twpol: We have had a default set of style rules in the repository for some time, with the goal of slowing bringing all the code style back into a consistent shape. This is, I think, the first large-scale formatting to occur that could be accepted (myself and others have trialled formatting the whole repo, but the changes were unmanageable).

Nothing enforces any of the rules.

Visual Studio will apply the rules to snippets, completions, refactors, and other places where it generates code.

The relevant rules here are all set to the defaults and have not been customised. That means that this is both implicitly the recommended style and is, or will become, the most common style of C# code.

I am generally open to changing the rules if there are strong reasons, and you can see our customisations (which match existing Open Rails style) here: https://github.com/openrails/openrails/blob/d3944727386daae3fa7fdce2910280479865ac0b/Source/.editorconfig#L137-L160

However, I don't find "everyone knows" a convincing argument for change. And while "difficult to read" is a better argument, it is subjective. I find the parentheses beneficial, for example.

The parenthesis rules are here: https://github.com/openrails/openrails/blob/d3944727386daae3fa7fdce2910280479865ac0b/Source/.editorconfig#L38-L42

The defaults are from here: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/code-style-rule-options#example-editorconfig-file

twpol (approved):
To check that the changes had improved the overall conformance with the rules in `.editorconfig`, I ran `dotnet format` on the original code and the proposed code.

- Original: `378 files changed, 6195 insertions(+), 5967 deletions(-)`
- Proposed: `370 files changed, 6032 insertions(+), 5805 deletions(-)`

That's 325 fewer lines added or removed (and 8 files no longer in need of any changes), so is indeed an improvement in conformance.

I do think there are some less obviously okay changes, though. For example, comments seem to have gotten some amount of reformatting which doesn't always look good in the diff.

And there is the potential issue that people strongly disagree with some of the default rules, as @cesarBLG has shown.

Overall, I am happy to have this merged, but we may want to discuss the defaults before applying to any more code.

Source/Orts.Simulation/Simulation/AIs/AIPath.cs
cesarBLG: Thanks for your message. I'm still not convinced about it (I don't think "Microsoft sets this as the default" is a strong enough argument to overrule standard math conventions), but I'll bear with it. It's good to have some global consistency in the code.

My opinion is that this shouldn't be automatically refactored, but leave it to the coder's choice. The original author may add parenthesis if he thinks they are necessary for clarity. But as you say this is subjective.

PR #868, opened 2023-08-31
Bug  fix for Exhaust of non-lead diesels is not correct https://bugs.launchpad.net/or/+bug/2033621

Csantucci:
See http://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__299968


cesarBLG (approved):
Thanks, Carlo

PR #870, opened 2023-09-05
Fix water restore

cesarBLG:
Bug report: https://www.elvastower.com/forums/index.php?/topic/37430-water-quantity-increased-after-save/

Remove workaround introduced due to previously broken Restore procedure


Csantucci (approved):
That's OK to me, it's also a code cleanup. Thank you!

PR #874, opened 2023-09-14
Dynamic brake controller refactoring

cesarBLG:
- Add an option to not lock dynamic brake at setup position
- Better handling of train blending and local blending
- Add an option for train brakes which trigger dynamic brakes
https://blueprints.launchpad.net/or/+spec/dynamic-brake-controller-refactor
http://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
@@ -313,8 +313,20 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Controllers
                             float value = stf.ReadFloat(STFReader.UNITS.None, null);
                             int smooth = stf.ReadInt(null);
                             string type = stf.ReadString();
-                            Notches.Add(new MSTSNotch(value, smooth, type, stf));
-                            if (type != ")") stf.SkipRestOfBlock();
+                            string name = null;
+                            while(type != ")" && !stf.EndOfBlock())
+                            {
+                                switch (stf.ReadItem().ToLower())
+                                {
+                                    case "(":
+                                        stf.SkipRestOfBlock();
+                                        break;
+                                    case "ortslabel":
+                                        name = stf.ReadStringBlock(null);
+                                        break;
+                                }
+                            }
+                            Notches.Add(new MSTSNotch(value, smooth, type, name, stf));
                         }),
                     });
                     break;
@@ -484,7 +496,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Controllers
         {
             if (Script != null)
             {
-                string state = ControllerStateDictionary.Dict[Script.GetState()];
+                string state = Script.GetStateName();
                 string fraction = GetStateFractionScripted();
 
                 if (String.IsNullOrEmpty(state) && String.IsNullOrEmpty(fraction))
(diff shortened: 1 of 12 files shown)


SteelFill:
These changes are much appreciated, I'm already feeling the difference thanks to the disabling of dynamic brake locking. As I mentioned on Elvas Tower, this addition feels like a good time to investigate other issues with dynamic brake control locking. Specifically, some missing control interlocks that are typical on most locomotives:
- Dynamic brake handle cannot be moved out of the off position with the reverser in neutral.
- The reverser cannot be moved unless the dynamic brake handle is in the off position. (This was implemented for the throttle but it seems the dynamic brake was forgotten.)

Do you think you could investigate those interlocks as part of this PR?

cesarBLG:
Controller Increase/Decrease functions are quite a mess, but I'll try

cjakeman:
Hi Cesar,

Is this the right link to the forum? [http://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features](http://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features).

Best wishes,

Chris

cesarBLG:
Yes, I updated the description

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Controllers/BrakeController.cs
```diff
@@ -313,8 +313,20 @@ public void Parse(string lowercasetoken, STFReader stf)
...
+                            string name = null;
+                            while(type != ")" && !stf.EndOfBlock())
+                            {
+                                switch (stf.ReadItem().ToLower())
+                                {
+                                    case "(":
+                                        stf.SkipRestOfBlock();
+                                        break;
+                                    case "ortslabel":
+                                        name = stf.ReadStringBlock(null);
```
SteelFill, line 325: This is not a bug, but worth noting, stf.ReadStringBlock will read whitespace delimited, so only the first word of a label will be picked up unless the label is enclosed in quotation marks. (Eg: ORTSLabel ( Initial Reduction ) will produce name = "Initial", while ORTSLabel ( "Initial Reduction" ) will produce name = "Initial Reduction".)

The solution here is to just encourage users to use quotation marks in their labels, else they will only capture the first word. Also worth noting that [this post](http://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features/page__view__findpost__p__300693) would not function as intended. People have a habit of finding old posts in Elvas Tower and assuming any code shown is actually how things work, so it may be best to edit this post to use quotation marks so it's a better example.

cesarBLG: Thanks, I inserted quotation marks

SteelFill:
I noticed a possible savegame issue introduced here. I had saved a session with the dynamic brake applied, but upon loading in, the dynamic brake was not enabled. Pressing the dynamic brake apply keybind caused the dynamic brake to suddenly shoot up to the position it was in before the save, and then everything functions like normal.

Maybe more dynamic brake variables need to be saved to properly preserve the dynamic brake state.

cesarBLG:
Thanks, I added a couple of variables to the saved data

peternewell (approved):
Hi Cesar,

Given the fact that I can't see any specific issues, and that it has been in the Unstable version for a long time, I think that it can be merged.

cesarBLG:
Thanks, Peter!

PR #875, opened 2023-09-17
Bug fix for https://bugs.launchpad.net/or/+bug/2036346 Player train switching doesn't work with 3D cabs

Csantucci:


cesarBLG (approved):
Fix looks sensible to me. This has been on Unstable for a while, so I assume it's not introducing undesired bugs

Csantucci:
Thanks, César

PR #877, opened 2023-09-25
Initail build of duplex steam 

peternewell:
Initial build of duplex steam - addition of control commands only

https://blueprints.launchpad.net/or/+spec/duplex-steam-locomotives

It is desired to merge this so that control keys are "reserved".

Additional functionality supporting this will be added later under a separate PR

### DIFF
--- a/Source/Orts.Simulation/Common/Commands.cs
+++ b/Source/Orts.Simulation/Common/Commands.cs
@@ -1573,6 +1573,30 @@ namespace Orts.Common
 
 
     // Steam controls
+
+    // Steam booster command
+    [Serializable()]
+    public sealed class ContinuousSteamBoosterCommand : ContinuousCommand
+    {
+        public static MSTSSteamLocomotive Receiver { get; set; }
+
+        public ContinuousSteamBoosterCommand(CommandLog log, int injector, bool toState, float? target, double startTime)
+            : base(log, toState, target, startTime)
+        {
+            Redo();
+        }
+
+        public override void Redo()
+        {
+            if (Receiver == null) return;
+            {
+                Receiver.SteamBoosterChangeTo(ToState, Target);
+            }
+            // Report();
+        }
+    }
+
+    // Steam heat command
     [Serializable()]
     public sealed class ContinuousSteamHeatCommand : ContinuousCommand
     {
--- a/Source/Orts.Simulation/Common/Commands.cs
+++ b/Source/Orts.Simulation/Common/Commands.cs
@@ -1589,9 +1589,7 @@ namespace Orts.Common
         public override void Redo()
         {
             if (Receiver == null) return;
-            {
                 Receiver.SteamBoosterChangeTo(ToState, Target);
-            }
             // Report();
         }
     }
--- a/Source/Orts.Simulation/Common/Commands.cs
+++ b/Source/Orts.Simulation/Common/Commands.cs
@@ -1589,7 +1589,7 @@ namespace Orts.Common
         public override void Redo()
         {
             if (Receiver == null) return;
-                Receiver.SteamBoosterChangeTo(ToState, Target);
+            Receiver.SteamBoosterChangeTo(ToState, Target);
             // Report();
         }
     }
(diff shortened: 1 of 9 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -820,6 +823,7 @@ public override void Parse(string lowercasetoken, STFReader stf)
                 case "engine(enginecontrollers(dampersfront": DamperController.Parse(stf); break;
                 case "engine(enginecontrollers(shovel": FiringRateController.Parse(stf); break;
                 case "engine(enginecontrollers(firedoor": FireboxDoorController.Parse(stf); break;
+                case "engine(enginecontrollers(steambooster": SteamBoosterController.Parse(stf); SteamBoosterControllerFitted = true; break;
```
cesarBLG, line 826:
```suggestion
                case "engine(enginecontrollers(ortssteambooster": SteamBoosterController.Parse(stf); SteamBoosterControllerFitted = true; break;
```
I think we need an ORTS prefix there

peternewell: Thanks for pointing that out. I have now added the ORTS prefix.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -7175,7 +7195,71 @@ public void SteamStopGearBoxDecrease()
...
         }
 
-        //Small Ejector Controller
+        #region Steam booster controller
+
+        public void StartSteamBoosterIncrease(float? target)
+        {
+            SteamBoosterController.CommandStartTime = Simulator.ClockTime;
+            if (IsPlayerTrain)
+                Simulator.Confirmer.ConfirmWithPerCent(CabControl.SteamBooster, CabSetting.Increase, SteamBoosterController.CurrentValue* 100);
```
cesarBLG, line 7204:
```suggestion
                Simulator.Confirmer.ConfirmWithPerCent(CabControl.SteamBooster, CabSetting.Increase, SteamBoosterController.CurrentValue * 100);
```

Source/Orts.Simulation/Common/Commands.cs
```diff
@@ -1573,6 +1573,30 @@ public override void Redo()
...
+        {
+            Redo();
+        }
+
+        public override void Redo()
+        {
+            if (Receiver == null) return;
+            {
+                Receiver.SteamBoosterChangeTo(ToState, Target);
+            }
```
cesarBLG, line 1594:
```suggestion
            Receiver?.SteamBoosterChangeTo(ToState, Target);
```

cesarBLG (changes requested):
Just two style fixes, sorry for not having spotted them before

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: Thanks, I have made this adjustment, and also to a number of other similar entries.

Source/Orts.Simulation/Common/Commands.cs
cesarBLG: I'm sorry but I won't approve it as is. It looks confusing to me. In particular, at first glance your code seems to do
```csharp
if (Receiver == null)
{
    Receiver.SteamBoosterChangeTo(ToState, Target);
}
```
until you notice the "return" statement. The curly braces are completely misleading.

The ?. operator is my preferred choice in this case, and is also recommended by ORMT (see http://www.elvastower.com/forums/index.php?/topic/35509-cs-null-conditional-operator/). It is also used elsewhere in Commands.cs, so your argument about consistency is not really valid. In my opinion, new code should use it whenever possible.

However, my intention is not to force you to use that syntax. It was just my suggestion. I have no problem if you prefer to use "standard" syntax, as long as the misleading braces are removed. Other acceptable solutions for me are:
```csharp
if (Receiver == null) return;
Receiver.SteamBoosterChangeTo(ToState, Target);
```
or
```csharp
if (Receiver != null)
{
    Receiver.SteamBoosterChangeTo(ToState, Target);
}
```
or
```csharp
if (Receiver != null) Receiver.SteamBoosterChangeTo(ToState, Target);
```

peternewell: Ok, I have adopted one of your suggested solutions.

cesarBLG: I'm sorry for being so nitpicking, but the indentation of `Receiver.SteamBoosterChangeTo(ToState, Target);` is still incorrect. I think it is important that we follow the coding guidelines to make code more readable for others. I could have corrected it myself, but you disabled "allow edits from maintainers".

peternewell: I have moved it across now. Thanks for checking.

cesarBLG (approved):
Thanks, and sorry for the inconvenience of the multiple style suggestions.

peternewell:
Thanks Cesar.

PR #880, opened 2023-10-09
fix: Refined comment summarising property SData1

cjakeman:
As reported in [Launchpad.net/or/+bug/2038756](https://www.Launchpad.net/or/+bug/2038756) by David Webb


Csantucci (approved):
That's OK to me, thanks.

PR #884, opened 2023-10-21
build: Additional settings for dotnet build

twpol:


twpol:
Looks like it might be breaking at runtime, however.

PR #886, opened 2023-11-13
Scene viewer extension to TrackViewer

pzgulyas:


cjakeman:
Hi Peter,

It will be interesting to see the response from the community to your initiative.

When you have a moment, would you add a blueprint or a Trello Card setting out the scope of this project, please?

Thanks,

Chris

twpol:
This is a hugely exciting development, even as a "contributed" project!

Looking through the changes, it seems like there is a fair amount of fighting with the architecture - in particular, too much useful code is in `RunActivity` instead of a shared library. I'd probably be against all the additional `public` decorators in there, for example.

Instead, would you be interested in using this work to direct refactoring to get the viewer and related code out of `RunActivity` and into `Orts.Viewer` as [proposed here](https://github.com/openrails/openrails/blob/d7daf626f035de6b976cd847e2fae43cbee44374/Docs/Architecture.md#player-application-model)?

It would have to be a separate PR for review purposes, and obviously I would not expect you to refactoring any bits which aren't needed for this viewer/editor (although refactors are always appreciated).

---

The other thing which stands out is the code in `Source/Orts.Formats.Msts/WorldFile.cs` to write out new MSTS-format files. I'm not sure if we'd be happy with the placement, so we should discuss this in an architectural thread.

twpol:
I am removing this PR from Unstable, as it's someone's experimental draft (which is fine) but looks like it is blocking #891 which is not draft, and we'd like to get included.

PR #892, opened 2023-11-16
Signal Function OPP_SIG_ID_TRAINPATH

Roeterdink:
Add new signal function OPP_SIG_ID_TRAINPATH

Feature request: https://www.elvastower.com/forums/index.php?/topic/37572-proposed-new-signal-function-opp-sig-id-trainpath/page__view__findpost__p__302318


pzgulyas (approved):
I see no problems with this code.

PR #893, opened 2023-11-16
Signal errors

Roeterdink:
Correct errors in signal scripting, see threads:

Bug report: https://www.elvastower.com/forums/index.php?/topic/37570-debug-code/page__view__findpost__p__302315

Bug report: https://www.elvastower.com/forums/index.php?/topic/37571-undeclared-variables-in-signal-scripts/page__view__findpost__p__302317

### DIFF
--- a/Source/Orts.Formats.Msts/SignalScripts.cs
+++ b/Source/Orts.Formats.Msts/SignalScripts.cs
@@ -2115,6 +2115,7 @@ namespace Orts.Formats.Msts
                     // if only 1 part, it is a single function call without assignment
 
                     AssignType = SCRTermType.Invalid;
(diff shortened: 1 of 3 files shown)


Csantucci:
Probably I haven't understood well your check for undeclared variables, but what occurs if in the script the variable is declared with value = -1?

Roeterdink:
The variable AssignParameter is the index of the float as used in the script, in the list of (declared) floats. It is not the value of that float.
As things are, if a float is not declared, AssignParameter = 0. But this is the index of another float, so the instead of setting or getting the value of the undeclared float, the value of the other float is set or used.
This check will give proper warning of such an error in the script.

Csantucci (approved):
The bug fix is OK to me. Thank you.

PR #894, opened 2023-11-16
Correct Decrease Colour

Roeterdink:
Correct code for _Decrease Color setting in Cabview - Digital Speedometer_, see bug report: [`#2043690`](https://bugs.launchpad.net/or/+bug/2043690)

### DIFF
--- a/Source/Orts.Formats.Msts/CabViewFile.cs
+++ b/Source/Orts.Formats.Msts/CabViewFile.cs
@@ -876,11 +876,19 @@ namespace Orts.Formats.Msts
                     }),
                 new STFReader.TokenProcessor("decreasecolour", ()=>{
                     stf.MustMatch("(");
-                    stf.ReadInt(0);
+                    int NumColor = stf.ReadInt(0);
                     if(stf.EndOfBlock() == false)
                     {
-                        stf.ParseBlock(new STFReader.TokenProcessor[] {
-                            new STFReader.TokenProcessor("controlcolour", ()=>{ DecreaseColor = ParseControlColor(stf); }) });
+                        if (NumColor == 0)
+                        {
+                            // no. of colors is set as 0 and therefor no color is defined, but if set anyway, it must be skipped
+                            stf.SkipRestOfBlock();
+                        }
+                        else
+                        {
+                            stf.ParseBlock(new STFReader.TokenProcessor[] {
+                                new STFReader.TokenProcessor("controlcolour", ()=>{ DecreaseColor = ParseControlColor(stf); }) });
+                        }
                     }
                 }),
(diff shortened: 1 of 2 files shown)


Csantucci (dismissed):
This seems OK, thanks. By the way GetDigits(out Color DrawColor) seems to be not used.

Roeterdink:
Thanks.
As for GetDigits: it looks to be mostly the same code as in PrepareFrame.
It looks to me as if it was the intention to replace PrepareFrame by GetDigits for 2D and Get3DDigits for 3D cabs, but somehow this change was never worked through completely.

Csantucci:
Hi Rob,
please read this https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__302666 , which occurs also in Unstable release.
It's a block like this one
`		Digital (
			Type ( ACCELEROMETER DIGITAL )
			Position ( 483 261 15 10 )
			ScaleRange ( -99 99 )
			Accuracy ( 0 )
			AccuracySwitch ( 0 )
			LeadingZeros ( 0 )
			Justification ( 1 )
			PositiveColour ( 1
				ControlColour ( 0 255 0 )
			)
			NegativeColour ( 1
				ControlColour ( 255 255 0 )
			)
			DecreaseColour ( 0
				ControlColour ( 0 0 0 )
			)
			Units ( MILES/HOUR/MIN )
			ORTSFont ( 3 0 "Courier New" )
		)`
that shows black as decrease colour, while it should not consider such decrease colour, because the first line of the block is 
`DecreaseColour ( 0`
whic means zero decrease colours.
I fixed this on ORNYMG with following patch. 
[cabdigitalsflicker.zip](https://github.com/openrails/openrails/files/13480227/cabdigitalsflicker.zip)
If you agree with correcting this, you might apply the above patch also to your PR.

Roeterdink:
Carlo,
why not sort this issue immediately where the file is read?
The change is smaller and only locally : 
[Dif_cabview_NoDecColor.zip](https://github.com/openrails/openrails/files/13481008/Dif_cabview_NoDecColor.zip)

Csantucci:
> Carlo, why not sort this issue immediately where the file is read? The change is smaller and only locally : [Dif_cabview_NoDecColor.zip](https://github.com/openrails/openrails/files/13481008/Dif_cabview_NoDecColor.zip)

Yes, your solution is better. If you apply your patch to the PR, I'll reapprove it. Maybe instead of reading the dummy colors a SkipRestOfBlock may be used?

Roeterdink:
This pull request has been updated to include corrections for an error in the code which reads the decrease color information in CabView,cs.
This error was not caused by the original changes. This second error was hidden as the result, a setting for decrease color, was never processed due to the first error. Correcting the first error therefor revealed this second error.

PR #898, opened 2023-12-02
Extra line with all the arguments for debugging purposes in logfile

sweiland-openrails:
When debugging in Visual Studio all the parameters in one line are needed as Debug Property. Parameters can be found in the logfile, however in separate lines and without quotes. Now one line to be used in Visual Studio can be found with the keyword "Arguments". Only when the Runactivity is compiled in debug mode (#if DEBUG).

Trello card https://trello.com/c/VkoqrKbH/575-extra-line-with-all-the-arguments-for-debugging-purposes-in-logfile

### DIFF
--- a/Source/RunActivity/Viewer3D/Processes/GameStateRunActivity.cs
+++ b/Source/RunActivity/Viewer3D/Processes/GameStateRunActivity.cs
@@ -231,7 +231,7 @@ namespace Orts.Viewer3D.Processes
                     default:
                         MessageBox.Show("To start " + Application.ProductName + ", please run 'OpenRails.exe'.\n\n"
                                 + "If you are attempting to debug this component, please run 'OpenRails.exe' and execute the scenario you are interested in. "
-                                + "In the log file, the command-line arguments used will be listed at the top. "
+                                + "In the log file, a line with the command-line arguments used will be listed at the top. "
                                 + "You should then configure your debug environment to execute this component with those command-line arguments.",
                                 Application.ProductName + " " + VersionInfo.VersionOrBuild);
                         Game.Exit();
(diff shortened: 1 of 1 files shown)


sweiland-openrails:
In the menu code I found this:

                    if ((Control.ModifierKeys & Keys.Alt) == Keys.Alt)
                    {
                        Clipboard.SetText(joinedParameters);
                        MessageBox.Show(
                            "RunActivity.exe arguments have been copied to
the clipboard:" +
                            $"\n\n{joinedParameters}\n\n" +
                            "This is a debugging aid. If you wanted to
start the simulator instead, select Start without holding down the Alt
key.");
                        continue;
                    }

So if you press alt together with start, options are copied to the
clipboard. Makes the extra logline not that necessary. But I think I'll
still continue with this fix and use your suggestion to always add the line.

regards, Siebren

Op di 12 dec 2023 om 11:13 schreef Csantucci ***@***.***>:

Csantucci:
Yes, I remembered this but I didn't remember the key combination. I agree with you however that your extra logline can still be useful.

sweiland-openrails:
Removed the #debug.

And the alt key does not work in combination with the resume. Could be fixed, but this extra logline always works. So I leave it as it is now.

Csantucci (commented):
That's OK now, but maybe you should move the two curly brackets of the if statement into separate lines, or use a ?: syntax. Sorry for this, but I see that now there is big attention about these formal aspects.

sweiland-openrails:
thanks, changed. I normally have the { } on separate lines. That's what Visual Studio mostly takes care of.

Yes, there was some heavy discussion about the use of ?: versus an if statement. To my opinion it's not that important...

Csantucci (approved):
That's OK to me now, thanks!

PR #900, opened 2023-12-05
DMI in 3D cab + two more dials

pzgulyas:
Feature request: https://www.elvastower.com/forums/index.php?/topic/34577-planning-area-of-the-etcs-dmi

### DIFF
--- a/Source/Documentation/Manual/cabs.rst
+++ b/Source/Documentation/Manual/cabs.rst
@@ -20,46 +20,65 @@ next paragraphs.
 ETCS circular speed gauge
 -------------------------
 
-You can add to the cabview a
-circular speed gauge accordingly to the European standard train control
-system ETCS.
+A circular speed gauge accordingly to the standard European Train Control System 
+(ETCS) can be added to the cabview. The information displayed in the Driver 
+Machine Interface (DMI) is controlled via the TCS script. For more details,
+see :ref:`C# engine scripting - Train Control System <features-scripting-tcs>`.
+
 
 .. image:: images/options-etcs.png
    :scale: 60 %
    :align: center
 
 
-.. admonition:: For content developers
+The gauge is added by the insertion of a block like the following
+into the .cvf file::
 
-    The gauge is added by the insertion of a block like the following
-    into the .cvf file::
-
-        Digital (
-            Type ( SPEEDOMETER DIGITAL )
-            Style ( NEEDLE )
-            Position ( 160 255 56 56 )
-            ScaleRange ( 0 250 )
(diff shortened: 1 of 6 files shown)


cjakeman:
Hi Peter,

Is this work ready to come out of DRAFT and be approved and merged?

I don't see a link to a Trello card, but perhaps this is covered by a previous card or blueprint.

Thanks,

Chris

pzgulyas:
@cjakeman 
> Is this work ready to come out of DRAFT and be approved and merged?

Yes, it is now ready to be merged.

> I don't see a link to a Trello card, but perhaps this is covered by a previous card or blueprint.
Let's say it is a prerequisite for [this Trello card](https://trello.com/c/xShE5iVJ)

Source/Documentation/Manual/cabs.rst
```diff
@@ -1364,6 +1364,43 @@ Customizations for such file are possible following these rules:
 Except for the first column, fields 
 in the 3D distributed power display are always with center alignment.
 
+ETCS circular speed gauge and the Driver Machine Interface
```
cesarBLG, line 1367: There is already an entry describing the ETCS display. Could you merge it with this?

pzgulyas: Yeah, it's better that way. Merged.

pzgulyas:
> I don't know anything about 3D, but is there a reason for which we need to specify a texture to render the ETCS, but we don't need that to render e.g. a digital speedometer?

The digits implemented in a way that a new shape with a new texture is created for each digit. Please see the constructor of the ThreeDimCabDigit class in MSTSLocomotiveViewer.cs. The same approach is used for the more complex distributed power screen, where the screen is compiled from many small shapes for each indication bar, text, etc. First of all this results in a decrease in performance, since each subelement is a distinct shape shown in a separate drawcall. Besides this may lead to z-fighting problems, because the original texture is left under it, the elelement are only offset by a small distance. (See the v.Z = 0.01f lines in the constructor above.) You may see a picture of it created by James in [this thread](https://www.elvastower.com/forums/index.php?/topic/36606-z-fighting-issue/page__st__10__p__294932).

To avoid this all, I went with the path James also suggested at the last line of his above post, "Off-screen rendering for digital displays (which can happen in 2D, without any depth issues, and then be rendered in the correct 3D location)". This is a superior method to that done till now, because it gets rid of the original texture underneath, no z-fighting issues can emerge.

cesarBLG (approved):
I see, so the texture method is more efficient. Then everything seems fine for me

PR #901, opened 2023-12-09
Remove unnecessary spaces and incorrect usage of spaces across the codebase (`Orts.Simulation`)

Looky1173:
This PR removes unnecessary or confusing usage of spaces from `Orts.Simulation`. A large proportion of CodeFactor issues are simply about such incorrect usage of spaces, so addressing them will give more visibility to the "real" issues. It helps with code readability as well.

To be clear, I did not use find and replace to do this. I went through each line break of each file, and manually deleted (or added) spaces where necessary. Therefore, the integrity of the code could not have been compromised. I've also reviewed the complete diff just to make sure the PR is safe to merge.


twpol:
Thank you for your efforts to improve the code style. However, at this time:

- CodeFactor and Codacy are for curiosity only
- Have not been configured nor validated in their configuration
- There have been ongoing discussions about code style in the forums

As such, I don't think it's appropriate to approve and merge this PR right now - and we made a mistake doing so with the previous one.

Instead, I would like any code style work to focus on configuring the code quality and code style rules in https://github.com/openrails/openrails/blob/1ab6b9003239afd7bcac7ef23e6c131aa60b2832/Source/.editorconfig#L137-L160 to match the development team's expectations in the forums. These rules are used by `dotnet build`, `dotnet format`, and Visual Studio's analysis and formatting behaviours.

We can return to actually formatting the code later.

PR #904, opened 2023-12-21
fix: adds downgrade to low precision

cjakeman:
Bug report: https://www.elvastower.com/forums/index.php?/topic/37671-advanced-adhesion-model-switched-to-low-performance-option-due-to-low-frame-rate

This is a fix to merged PR 878.
It switches the Advanced Adhesion down from Polach to Pacha's if the framerate falls below 30 fps.
Also it only logs a single statement.

A further PR is following to extend this one with support for switching back to Polach provided transitions are not too frequent.


peternewell (changes requested):
I am happy to approve this.

Is it possible to add an "in game" indication as to what mode that it is in, besides the Log File message.

Perhaps, in the FORCES INFORMATION HuD, the text "(Advanced Adhesion Model)" could be changed to yellow text if it is in the Pacha model mode?

Thanks

cjakeman:
Thanks. I'll see what I can do.

peternewell (approved):
Chris,

Thanks for working this through.

PR #905, opened 2023-12-21
feat: supports switching adhesion precisions

cjakeman:
Forum: [https://www.elvastower.com/forums/index.php?/topic/37671-advanced-adhesion-model-switched-to-low-performance-option-due-to-low-frame-rate](url)
Extends PR 904 to switch back adhesion back to high precision Polach method from low precision Pacha's method, provided switching is not too frequent.

Currently switches to high precision if framerate exceeds 40 fps.
If switching to low precision happens more quickly than a 5 minute interval, then the switch is made permanent.

This is still DRAFT as lacks an entry for the Manual.


cesarBLG:
Hi Chris,
Thanks for looking at this. I think that, in order to avoid the "jumps" I mentioned in the previous PR, it could be useful to call Axle.InitializeMoving() when there is a transition between models. This would reset wheel speed to train speed, so the transition doesn't cause a wheelslip.

cjakeman:
Replaced by 908 as it contained conflicts

PR #908, opened 2023-12-26
feat: supports switching adhesion precisions

cjakeman:
Feature request: https://www.elvastower.com/forums/index.php?/topic/37671-advanced-adhesion-model-switched-to-low-performance-option-due-to-low-frame-rate

Extends PR 904 to switch back adhesion back to high precision Polach method from low precision Pacha's method, provided switching is not too frequent.

Currently switches to high precision if framerate exceeds 40 fps.
If switching to low precision happens more quickly than a 5 minute interval, then the switch is made permanent.

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -259,12 +259,17 @@ wheel adhesion. The first model is based upon an algorithm by Pacha, whilst the 
 uses an algorithm developed by Polach. The Polach algorithm provides 
 a more accurate outcome and facilitates the future inclusion of track conditions. 
 However due to the number of algorithm steps required to calculate the wheel adhesion 
-value, it is more CPU load intensive then the Pacha one. This can produce low 
-frame rates for the screen display in machines with low performance specifications. 
+value, it is more CPU load-intensive then the Pacha one. On low performance PCs, this would lower the 
+frame rate for the screen display to an unacceptable degree. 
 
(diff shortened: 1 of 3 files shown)


cjakeman:
> Hi Chris, Thanks for looking at this. I think that, in order to avoid the "jumps" I mentioned in the previous PR, it could be useful to call Axle.InitializeMoving() when there is a transition between models. This would reset wheel speed to train speed, so the transition doesn't cause a wheelslip.

Hi Cesar,

I've done that, but not called InitializeMoving() directly as looks as though that does more than is needed. Simply set the axles speed same as the train speed just as InitializeMoving does.

I moved the FPS check from Axle to Axles as it seems overkill to be checking it for every axle.

Best wishes

cesarBLG:
Thanks Chris,
InitializeMoving() also resets motor parameters for the DC series motor, but since that is unimplemented your solution is OK.

I agree with using the same model in all axles, and setting it globally at Axles.

cjakeman:
> Thanks Chris, InitializeMoving() also resets motor parameters for the DC series motor, but since that is unimplemented your solution is OK.
> 
> I agree with using the same model in all axles, and setting it globally at Axles.

Good, Thanks for your help with this, Cesar.

I'll update the Manual to include this behaviour and then wait a week or two in case there's feedback from anyone else.

peternewell (approved):
Hi Chris,

This appears ok, thanks for implementing it.

PR #910, opened 2024-01-13
Allow building code using .NET 6 (Windows)

twpol:
### We are not upgrading to .NET 6 yet! Windows 7 is still supported!

The changes we will need to make to upgrade from .NET Framework to .NET 5+ are potentially wide-ranging, but unknown to us at present. Many of the changes will also be portability issues, i.e. things which affect non-Windows compatibility.

This PR allows us to get started on making things work in .NET 6 and non-Windows systems **without** changing our system requirements for official builds.

**All build methods** (Visual Studio, `Build.cmd`, `MSBuild`, `dotnet build`, etc.) will continue to target .NET Framework 4.7.2 by default.

**Adding the parameter `-p:BuildDotNet=true`** to `MSBuild` or `dotnet build` will target .NET 6 (Windows) instead.

At present, this will compile successfully, with ~600 warnings, but not run correctly.

When addressing issues with this .NET 6 (Windows) build, for the moment you can use conditions like in `Source/RunActivity/Viewer3D/Processes/WatchdogProcess.cs`:

```
#if NET5_0_OR_GREATER
#else
#endif
```

I may also provide shims for e.g. `OperatingSystem.IsWindows()` in a subsequent PR so that less conditional compilation is needed.

### DIFF
--- a/Source/ContentChecker/ContentChecker.csproj
+++ b/Source/ContentChecker/ContentChecker.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net472</TargetFramework>
+    <TargetFrameworks Condition="'$(BuildDotNet)' == 'true'">net6-windows</TargetFrameworks>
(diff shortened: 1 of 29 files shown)


pzgulyas (approved):
Looks good.

PR #911, opened 2024-01-20
docs: Add refactoring as a special type of PR

twpol:
Currently, there is no documented path for refactoring PRs, for which we have a label and a history of accepting. This clarifies the requirements.


cjakeman (approved):
Hi James,

Looks better. Sorry it's been waiting so long.

Approved and merged.

Thanks,
Chris

PR #851, opened 2023-07-07
New Triple Valve Features

SteelFill:
My first enhancement for OR, this adds multiple new features to the air brake code to allow for simulating some features of more modern brake valve types (eg: ABD, ABDW), and also refactors/reorganizes some brake system behaviors to accommodate the new features better.

[(Start of) Discussion on ElvasTower](http://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__298101)
[Trello Card](https://trello.com/c/el72p3b7/569-new-features-for-american-triple-valves)

- Refactor of triple valve state updates
- Smoothing added to final stages of application/release to hopefully prevent triple valve from rapidly toggling states
- Changes to aux/emerg res charging behavior to better accommodate graduated release systems
- Quick release: Emerg res air can be piped into the brake pipe during release to speed up release. Enable with **`ORTSEmergencyResQuickRelease ( 1 )`** [default is 0]
- Uniform charging: Aux res charging can be slowed down when the brake pipe pressure is substantially higher than the aux res pressure to allow faster releases. Set the difference between BP and AUX to activate uniform charging with **`ORTSUniformChargingThreshold ( 3psi )`** [default is 3 psi], and the factor by which the rate of charging is reduced with **`ORTSUniformChargingRatio ( 2 )`** [default is 0, disabling the feature] {setting of 2 halves the rate, 0.5 would double the rate}.
- Quick service: When changing from release to apply, the brake valve will vent some air from the brake pipe to speed up initial applications until the brake cylinder pressure has reached a certain level. Set the target brake cylinder feed line pressure using **`ORTSQuickServiceLimit ( 10psi )`** [default is 0, disabling the feature], set the rate of brake pipe venting with **`ORTSQuickServiceVentRate ( 2.5psi/s )`** [default is 0] or leave it at 0 psi/s to simulate systems which don't vent the brake pipe, and optionally allow for brake cylinder feed pressure to increase faster than usual using **`ORTSQuickServiceApplicationRate ( 15psi/s )`** [default is 0].
[...]

PR #912, opened 2024-01-27
New Triple Valve Features Vol. 2

SteelFill:
I'm back! This is another round of triple valve improvements requested by the community. Can be considered a continuation of #851 
[Associated Trello Card](https://trello.com/c/z1ELrAB5)
[Discuss on ElvasTower](https://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__305027)

- Fixed an issue that could cause quick release to get stuck if retainers were used, preventing future brake applications (reported on TrainSimCommunity Discord)
- Fixed an easy to miss issue that would result in dynamic brake blending failing to activate on copied locomotives
- Fixed [this issue](https://www.elvastower.com/forums/index.php?/topic/37861-vehicles-causing-the-simulator-to-close/) that was giving a division by 0 error if the max brake cylinder pressure was set to 0
- Fixed `TrainPipeLeakRate` token. Previously the leak would only occur on the lead locomotive, now the leak will be applied to the whole train (I want to do better leak simulation eventually, but not yet).
- Added `psi/min` and `bar/min` as valid pressure units in the .eng/.wag file reader. This should help users define very slow rates of pressure change (eg: leakage, overcharge elimination).
- Changed `ORTSEmergencyResQuickRelease` to accept "2" as an input. If the value 2 is given, the emergency res will behave like an accelerated release res; it will not be used to enhance brake cylinder pressure in emergency applications but it will be used to provide accelerated release action.
- Rewrote quick service implementation to be more accurate to reality, uses the same user inputs as previously added.
- Changed the way the "threshold" (ie: target BC pressure) is calculated to accommodate new features, also refactored some application/release code to take advantage of threshold
- Added empty and loaded RelayValveRatio and Inshot for freight animations. If the freight is more than 25% loaded, the loaded setting will be used, otherwise the empty setting. Syntax is `EmptyBrakeRelayValveRatio`, `EmptyBrakeRelayValveInshot`, `FullBrakeRelayValveRatio`, and `FullBrakeRelayValveInshot` (this syntax will only work in the `ORTSFreightAnims` block).
- **Two stage brake pressures**: In some systems [not all, this is disabled by default], two distinct brake cylinder pressure limits may be used depending on speed. `ORTSTwoStageLowPressure ( x bar )` sets the pressure limit at low speeds (at high speeds, the pressure is limited in the usual ways). Or if appropriate, `ORTSTwoStageRelayValveRatio ( x )` can be used to set a different relay valve ratio at low speed, with the high speed relay valve ratio set as usual. `ORTSTwoStageIncreasingSpeed ( x kmh )` is the speed at which the lower pressure limit is removed when accelerating, while `ORTSTwoStageDecreasingSpeed ( x kmh )` sets the speed where the lower pressure limit is activated when decelerating. This is disabled during emergency (unsure if this is accurate).
- **High speed reducing valve**: A peculiar feature present only on some old American designs. It acts as a safety valve venting excess brake cylinder pressure during service applications, but during emergency applications intentionally allows brake cylinder pressure to exceed this limit. The brake pressure will then gradually return to normal. The activation pressure can be set with `ORTSHighSpeedReducingPressure ( 60 psi )` [60 is typical, but this is disabled by default]. I've tried to automatically determine exactly how quickly the valve will vent excess pressure, but it may need tuning.
- **Uniform release**: Similar to uniform charging in the previous PR, this is a feature that activates when the brake pipe is modestly higher than the aux res pressure [3 psi is the standard, and the default value]. The user sets this offset with `ORTSUniformReleaseThreshold ( 3 psi )`. Unlike uniform charging which reduced aux res charging speed, uniform release reduces brake cylinder release rate, which is done to prevent the brakes coming off entirely on the front of the train while they are still fully applied at the rear. The reduction in release rate is set with `ORTSUniformReleaseRatio ( x )`, where the release rate is divided by x [ORTSUniformReleaseRatio ( 2 ) would halve release rate, default is 0 which disables this feature.]
- **Separate brake application rates for service and emergency**: To implement distinct application rates, I added a new way to limit the application rate during service applications specifically using `ORTSMaxServiceApplicationRate ( x psi/s )`. During emergency applications, the brake application rate will be the limit set by `MaxApplicationRate ( y psi/s )`. If no service application rate is set, it is assumed to be the same as the maximum application rate.
- **Quick action**: Another feature from older American triple valves. This one allows for higher brake cylinder pressures during emergency not by having an emergency reservior but by throwing air from the brake pipe into the brake cylinder. `ORTSEmergencyQuickAction ( 1 )` will enable the feature, though it is disabled by default. Rate of brake cylinder filling is determined by MaxApplicationRate, which must be set to a fairly large number for this to work as intended. Remember ORTSMaxServiceApplicationRate can be used to keep the application rate under control for service applications.
- Did some testing with EP brakes this time to ensure EP brakes respect the service application rate, service brake cylinder pressure, and two stage braking. Unfortunately the current EP setup doesn't differentiate between full service and emergency so emergency pressures can only be achieved by the automatic air brake. This may justify some extra attention.
- **Supply reservoirs**: Option for an additional air volume which, if equipped, will be used to supply air for the brake cylinders (via the relay valve) instead of the aux res. Additionally, if applied to a locomotive/cab car, the supply res will be used to charge the brake pipe (and the supply res itself will draw from the main res). This is common on passenger stock and allows for a much smaller aux res to be used regardless of the size of brake cylinders. `Supply_Reservoir` must be added to `BrakeEquipmentType` to use a supply res. `ORTSSupplyResCapacity` sets the volume of the supply reservoir, and `ORTSSupplyResChargingRate` sets the rate at which the supply res pressure increases when draining air from the brake pipe. If on a twin pipe system, the supply res will charge from the MR pipe instead, unless `ORTSMainResPipeAuxResCharging ( 0 )` is set.
- **In-shot**: Something I noticed on many relay valves (at least in the US) with a ratio < 1 is that a small amount of demanded BC pressure is added at a 1:1 ratio to the output of the relay valve. This gives a boost in output pressure when input pressure is low, ensuring that the brake cylinders extend fully even with light applications. (Eg: A relay valve with 0.6 ratio and 5 psi of in-shot will produce 8 psi output with 5 psi input, instead of just 3 psi.) `ORTSBrakeRelayValveInshot ( x psi )` sets the in-shot pressure for automatic brake applications, and `ORTSEngineBrakeRelayValveInshot ( x psi )` does the same for the engine brake. I also found that some types of in-shot override the demanded pressure (until the demanded pressure is greater than inshot) rather than add to the demanded pressure. If that is desired, simply enter a _negative_ number as the inshot pressure.
- **A bunch of new brake cylinder tokens** that will allow for better simulation of the brake cylinder, `ORTSBrakeCylinderDiameter ( x in )` allows for defining the brake cylinder diameter. `ORTSBrakeCylinderPistonTravel ( x in )` gives the length of extension between the released and applied states of the brake cylinder, which is 8 inches by default. These two are used to determine the volume of the brake cylinder, and establish how the brake cylinder volume changes depending on the amount of pressure. `ORTSBrakeCylinderPipingVolume ( x in^3 )` allows for adding an additional fixed volume to the brake cylinder, simulating the volume of piping between the triple valve and brake cylinder. Some brake systems intentionally increase this volume to produce the expected brake cylinder pressures. By default, this volume is automatically calculated to produce the expected brake cylinder pressures. `ORTSNumberBrakeCylinders ( x )` is used to set the number of brake cylinders on the car, which will multiply the calculated volumes. On a related note, `ORTSAuxiliaryResCapacity ( x in^3 )` can now be used to define the aux res volume directly, rather than indirectly with ratios.
- If `ORTSBrakeCylinderDiameter` is defined, then a more **advanced brake cylinder simulation** is activated that calculates and considers the effects of the brake cylinder(s) changing in volume. This WILL lead to different results than the simplified simulation, and forces the user to abandon some incorrect assumptions that have gone around the train simulation circle. Still, the setup will attempt to do its best given user inputs, and will produce warnings if the brake system parameters don't make sense.
- **Animated brake rigging**: Thanks to the introduction of the new brake cylinder tokens, I figured it would make sense to use the simulated brake cylinder to drive animations for the brake cylinder and brake rigging. There are 3 new animation types: `ORTSBRAKECYLINDER`, `ORTSHANDBRAKE`, and `ORTSBRAKERIGGING` (specifically, any animation matrix that begins with one of these names will be used for the brake animation). For creators: there is some nuance to setting this up correctly. I'm not sure if it's intuitive, but I don't have room here to explain (have you seen how huge this wall of text has gotten?), so check the [post on ET](https://www.elvastower.com/forums/index.php?/topic/34527-wishes-for-improvement-of-braking-systems/page__view__findpost__p__308633) for an explanation.
- New token `ORTSBrakeForceReferencePressure ( x psi )` to solve some frustration other users and I had. When this token is provided, the forces entered in `MaxBrakeForce` or `ORTSMaxBrakeShoeForce` will be assumed to be the force at the given pressure, rather than at `BrakeCylinderPressureForMaxBrakeBrakeForce`. This should help data entry for cases where the brake force has been calculated at an intermediate pressure, rather than at the absolute maximum pressure.
- **Unlimited brake cylinder pressure**: If a 0, negative, or simply no value is defined for `BrakeCylinderPressureForMaxBrakeBrakeForce`, OR will completely remove the limit on brake cylinder pressure. The pressure will be limited by the parameters of the brake system, rather than this value. This will be accurate for some systems where there is no safety valve or pressure limiting valve, but is certainly not accurate for all situations.


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/EPBrakeSystem.cs
```diff
@@ -97,20 +97,22 @@ public override void Update(float elapsedClockSeconds)
                 }
                 else
                 {
-                    demandedAutoCylPressurePSI = Math.Min(Math.Max(Car.Train.BrakeLine4, 0), 1) * MaxTripleValveCylPressurePSI;
+                    demandedAutoCylPressurePSI = Math.Min(Math.Max(Car.Train.BrakeLine4, 0), 1) * ServiceMaxCylPressurePSI;
```
cesarBLG, line 100: This breaks EP brakes if ORTSMaxServiceCylinderPressure is not defined in the .eng file, since ServiceMaxCylPressurePSI is initialized to 0.

SteelFill: Seems I forgot I didn't initialize that. Added a check to set the service max cylinder pressure equal to the (actual) max cylinder pressure if the user doesn't specify otherwise

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -479,8 +488,12 @@ public override void Initialize()
             if (Car.Simulator.Settings.SimpleControlPhysics && EmergResVolumeM3 > 2.0)
                 EmergResVolumeM3 = 0.7f;
 
-            if (MaxTripleValveCylPressurePSI == 0) MaxTripleValveCylPressurePSI = MaxCylPressurePSI / RelayValveRatio;
-            if (EngineRelayValveRatio == 0) EngineRelayValveRatio = RelayValveRatio;
+            if (ServiceMaxCylPressurePSI == 0)
+                ServiceMaxCylPressurePSI = MaxTripleValveCylPressurePSI;
+            if (MaxTripleValveCylPressurePSI == 0)
+                MaxTripleValveCylPressurePSI = MaxCylPressurePSI / RelayValveRatio;
```
cesarBLG, line 494: Sorry to bother you again, but this should be initialized the other way around

SteelFill: Yeah set an uninitialized variable equal to another uninitialized variable, that'll do lots of good...fixed in the next commit.

SteelFill:
Temporarily removing this from unstable builds to allow #916 to build for purposes of review. Once #916 is merged, the conflict between this PR and that PR can be cleared in the system and this will return to the unstable version.

SteelFill:
Returning to unstable now that the conflict with the DPU changes in #916 are resolved.

Source/Documentation/Manual/physics.rst
```diff
@@ -3200,10 +3216,27 @@ MaxAuxilaryChargingRate and EmergencyResChargingRate.
...
+- ``Wagon(ORTSAuxiliaryResCapacity`` -- Volume of the car's auxiliary
+  reservoir. Normally determined automatically given the emergency res
+  volume, but can be entered manually if the car has no emergency res.
+- ``Wagon(ORTSBrakeCylinderVolume`` -- Volume of each brake cylinder on
+  the car. This allows specifying the brake cylinder volume independently
+  of triple valve ratio. This is useful when the cylinder is not directly
+  attached to a triple valve, e. g. when a relay valve exists.
+- ``Wagon(ORTSBrakeCylinderSize`` -- If brake cylinder dimensions are
+  available, this can be used in place of ``ORTSBrakeCylinderVolume`` to
+  set the size of the brake cylinder(s).
```
cesarBLG, line 3228: Please specify that "size" means the diameter, which is not obvious. 

Another option to consider is to rename the variable from ORTSBrakeCylinderSize to ORTSBrakeCylinderDiameter, and keep ORTSBrakeCylinderSize as an unpublicized legacy parameter (if required)

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -619,7 +776,47 @@ public void UpdateAngleCockState(bool AngleCockOpen, ref float AngleCockOpenAmou
         public override void Update(float elapsedClockSeconds)
         {
             var valveType = (Car as MSTSWagon).BrakeValve;
-            float threshold = valveType == MSTSWagon.BrakeValveType.Distributor ? Math.Max((ControlResPressurePSI - BrakeLine1PressurePSI) * AuxCylVolumeRatio, 0) : 0;
+
+            // Two stage braking: higher brake force is allowed at higher speeds
```
cesarBLG, line 780: For European high speed trains this is the opposite. When speed exceeds 160-200km/h the brake force is reduced, not increased. Could you rewrite the implementation not assuming that LowSpeedPressure < HighSpeedPressure, but allow both regimes?

It doesn't make sense to me that higher brake forces are allowed at higher speeds since the adhesion will be lower (is this to achieve a constant brake force taking into account the reduction in brake shoe friction?)

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -752,6 +948,26 @@ public override void Update(float elapsedClockSeconds)
...
                         }
                     }
+                    else if ((Car as MSTSWagon).SupplyReservoirPresent && BrakeLine1PressurePSI < 15.0f)
+                    {
+                        // Supply res air directed to BC feed line to ensure full emergency force on cars with no emergency res
+                        // Only activated with brake pipe pressure lower than 10-18 psi (assuming 15 is the design point)
+
+                        // Ratio of displacement (dummy brake cylinder) volume to supply res volume
+                        // We are supplying pressure to the brake cylinder line, not the brake cylinder itself
+                        float dispalcementSupplyVolumeRatio = AuxResVolumeM3 / AuxCylVolumeRatio / SupplyResVolumeM3;
```
cesarBLG, line 958: Typo: displacement

This feature of the supply reservoir giving air during emergency seems not so common to me, maybe it could be optional. This looks like we are just duplicating the emergency reservoir.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -798,29 +1024,70 @@ public override void Update(float elapsedClockSeconds)
...
+                AutoCylPressurePSI -= dp;
+            }
+
             // Manage emergency res charging
             if ((Car as MSTSWagon).EmergencyReservoirPresent)
             {
                 if (TripleValveState == ValveState.Release && EmergResPressurePSI > BrakeLine1PressurePSI)
                 {
-                    if (EmergResQuickRelease && AutoCylPressurePSI > 5) // Quick release: Emergency res charges brake pipe during release
+                    if (EmergResQuickRelease && QuickReleaseActive) // Quick release: Emergency res charges brake pipe during release
```
cesarBLG, line 1090: UIC distributors have another feature that is also called quick release. I'd suggest renaming QuickReleaseActive to EmergQuickRelease active to avoid confusion.

UIC quick release is not implemented yet, IIRC the idea is that BP is overcharged to trigger a valve that fully releases all brakes, and then BP pressure quickly returns to normal

Source/Documentation/Manual/physics.rst
SteelFill: Vacuum brakes already had an "ORTSBrakeCylinderSize" token, so the goal is to make vacuum brakes and air brakes a little more consistent with eachother. But yes, a 'diameter' token would be more obvious for both types of brake system.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
SteelFill: The systems I was looking at were indeed concerned about low brake shoe friction at speed. I'll look into allowing the relationship to be reversed, but using the two stage relay ratio feature should already allow for lower or higher braking force depending on how you set it up. And I think the relay ratio approach is a bit more accurate than the flat pressure limit for most systems.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
SteelFill: Oops on the typo. As for the supply res in emergency, this is specifically based on equipment that has no emergency reservoir such as 26C brake valves. Note that this section of code won't even run if there is an emergency reservoir.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: Ah, if using a higher relay ratio for low speeds work, then it's only necessary to update the feature description (i.e. manual and code comments)

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: I was thinking in systems with a supply reservoir for the relay valve, but without this emergency feature. I don't see an obvious way to disable this while keeping the supply reservoir for other uses

Csantucci:
Lines 632-633:
            if (CylVolumeM3 <= 0 && AuxCylVolumeRatio >= 0)
                CylVolumeM3 = AuxResVolumeM3 / AuxCylVolumeRatio / CylCount;
If AuxCylVolumeRatio is  = 0 and CylVolumeM3 is <=0, the second line causes CylVolumeM3 to become infinity.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
SteelFill: Next version will only enable this if the max triple valve pressure is higher than the max triple valve service pressure (as is the case on the rolling stock the feature is based on). This should disable the feature on all systems where the emergency pressure is not higher than the service pressure.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
SteelFill: I did some testing in the current unstable build and this token seems to work just fine if you use ORTSMaxServiceCylinderPressure to set the higher speed pressure lower than ORTSTwoStageLowPressure for the low speed pressure. I'll make sure the documentation reflects that this works fine whether set lower or higher than the service max pressure.

SteelFill:
With the addition of load/empty RelayValveRatio and Inshot I believe this PR is now feature complete (if I'm honest, this has gone well beyond scope creep at this point). I am ready for reviews once more.

As much as I'd like to see someone try out the brake animation feature, I suspect it will be some time before anyone will have models with animated brakes and I do not want the other useful features held back due to one very niche (but cool!) feature.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -454,20 +610,183 @@ public override void Initialize()
...
+                    airPV[0] = AdvancedBrakeCylinderAir(0, true);
+                    // 0 cylinder travel when pressure is insufficient to budge spring
+                    cylTravel[1] = 0;
+                    airPV[1] = AdvancedBrakeCylinderAir(BrakeCylinderSpringPressurePSI / 2.0f, true);
+                    // 80% cylinder travel when spring pressure is fully overcome
+                    cylTravel[2] = CylStrokeM * 0.8f;
+                    airPV[2] = AdvancedBrakeCylinderAir(BrakeCylinderSpringPressurePSI, true);
+                    // 100% cylinder travel at 50 psi, with linear change in travel between 50 psi and max pressure
+                    // This will work whether the max cylinder pressure is above or below 50 psi
+                    float strokePerPsi = (CylStrokeM - cylTravel[2]) / (50.0f - BrakeCylinderSpringPressurePSI);
```
cesarBLG, line 689: I'd like to see more information describing these constants (80%, 50psi, etc.). Are they arbitrary or based in any principle? Using constants may not be adequate for all braking systems

cesarBLG (commented):
Thanks for your work! New features are well documented and code is understandable. Everything looks backwards-compatible, so I hope we don't see many regressions.

There are some non-important things that I wanted to mention:
-Two of them are about OR policies regarding coding and variable names. I don't really know what to do with those, so it's up to you (maybe asking to ORMT?)
-One is likely a bug in the Copy() function
-Some suggestions about using enums instead of plain integers
-The rest are just some thoughts

Maybe we should switch to an exponential-based propagation model for reservoirs (i.e. the higher the pressure diff, the higher the equalizing rate). The biggest problem is to keep backwards-compatibility with release rates (maybe defining release rates at 1bar pressure diff or something like that?)

Regarding EP brakes during emergency, the EP line could be set to 2 to denote emergency (i.e. -1 inactive 0 release .. 1 full brake 2 emergency, in this way you could achieve higher emergency pressures. Brake controllers also have to be adapted because some systems disconnect EP brakes during emergency, others simply trigger full service and others may request a higher pressure.

Could you link your trello card to the description?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -997,6 +1065,12 @@ public virtual void LoadFromWagFile(string wagFilePath)
                     // Update brake parameters
                     MaxBrakeForceN = ((LoadFullMaxBrakeForceN - LoadEmptyMaxBrakeForceN) * TempMassDiffRatio) + LoadEmptyMaxBrakeForceN;
                     MaxHandbrakeForceN = ((LoadFullMaxHandbrakeForceN - LoadEmptyMaxHandbrakeForceN) * TempMassDiffRatio) + LoadEmptyMaxHandbrakeForceN;
+                    // Not sensible to vary the relay valve ratio continouously; instead, it changes to loaded if more than 25% cargo is present
```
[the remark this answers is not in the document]
SteelFill: Of course, this is for "freight" animations after all. Passengers, and load proportioning there, will need a different setup entirely. And for the equipment where brake force does continuously change with load using some other mechanism, the original Load/Empty brake force parameters will make more sense.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -230,6 +296,15 @@ public override float GetCylVolumeM3()
         {
             return CylVolumeM3;
         }
+        public override float GetTotalCylVolumeM3()
```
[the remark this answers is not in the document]
SteelFill: The way BrakeSystem data is stored and retrieved is pretty inconsistent, I mostly went this route because there were some existing calls to GetCylVolumeM3 that were easier to replace with another similar function call than to change to inspect a property directly.

Also, if TotalCylVolumeM3 were added as a BrakeSystem property then the other brake cylinder parameters would need to be moved as well and that starts to add clutter. Brake systems like manual braking don't need to inherit parameters that won't be used (that said, for some reason there are vacuum brake only parameters in BrakeSystem).

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -274,6 +349,7 @@ public override void Parse(string lowercasetoken, STFReader stf)
         {
             switch (lowercasetoken)
             {
+                case "wagon(brakecylinderpressureformaxbrakeforce":
```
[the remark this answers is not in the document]
SteelFill: This is a mistake I made once that made me question the original syntax. I'm not sure if anyone has released rolling stock with this 'mistake' in it (nothing I own seems to make this error) so it may be over protective of the userbase to allow it.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
SteelFill: This is all based on a characteristic curve established by Pennsylvania Railroad in [this document](https://ia600906.us.archive.org/24/items/braketestsreport00penn/braketestsreport00penn.pdf) (URL is also in the comments in the source code). Never found any other reference for brake cylinder travel vs pressure, so this was really the only thing to base an implementation off of. Obviously, the report was for one specific size of brake cylinder, so to adjust for other sizes I assume the relationship is the same, just scaled up or down. The one constant is that the 'nominal' cylinder travel is defined at 50 psi, which is the standard pressure used for brake cylinder measurements.

Obviously these assumptions could be overwritten with user input hypothetically, but that adds a lot of complexity (the goal here is for someone to be able to add just a few tokens to an old file) and relies on getting input data that simply seems to not exist. If brake cylinder travel vs pressure becomes a popular field of study, that can change.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -58,31 +66,61 @@ public class AirSinglePipe : MSTSBrakeSystem
...
+        protected bool EmergResQuickReleaseActive;
         protected float UniformChargingThresholdPSI = 3.0f;
         protected float UniformChargingRatio;
         protected bool UniformChargingActive;
-        protected bool QuickServiceActive;
+        protected float UniformReleaseThresholdPSI = 3.0f;
+        protected float UniformReleaseRatio;
+        protected bool UniformReleaseActive;
+        // 0: quick service inactive, 1: first stage of quick service active, 2: second stage active
+        protected int QuickServiceActive;
```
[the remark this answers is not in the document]
SteelFill: These started out as bools and turned into more complex parameters...enums would make more sense at this point, I'll look at that.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -36,17 +36,25 @@ public class AirSinglePipe : MSTSBrakeSystem
         readonly static float OneAtmospherePSI = 14.696f;
         protected float HandbrakePercent;
         protected float CylPressurePSI = 64;
+        protected float CylAirPV;
```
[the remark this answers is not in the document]
SteelFill: Fair point, usually we do put the units in the variable names. Just normally they aren't so chaotic (at least the ideal gas law doesn't care much about units).

Worrying about bar and the more sensible bar * m^3 is a problem for the future. There's some serious cleanup that could be accomplished, and conversion to bar achieved at the same time, but that is well out of scope for today.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -137,12 +175,25 @@ public override void InitializeFromCopy(BrakeSystem copy)
...
+            SupplyResVolumeM3 = thiscopy.SupplyResVolumeM3;
             BrakePipeVolumeM3 = thiscopy.BrakePipeVolumeM3;
             CylVolumeM3 = thiscopy.CylVolumeM3;
+            TotalCylVolumeM3 = thiscopy.TotalCylVolumeM3;
+            CylPipeVolumeM3 = thiscopy.CylPipeVolumeM3;
+            CylDiameterM = thiscopy.CylDiameterM;
+            CylAreaM2 = thiscopy.CylAreaM2;
+            CylStrokeM = thiscopy.CylStrokeM;
+            CylCount = thiscopy.CylCount;
+            CylTravelTab = CylTravelTab == null ? null : new Interpolator(thiscopy.CylTravelTab);
```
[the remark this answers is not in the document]
SteelFill: Ah good catch, this would cause the initialization to waste time recalculating the brake cylinder travel on every single instance of a train car. Not fatal, but wasteful.

SteelFill:
> Maybe we should switch to an exponential-based propagation model for reservoirs (i.e. the higher the pressure diff, the higher the equalizing rate). The biggest problem is to keep backwards-compatibility with release rates (maybe defining release rates at 1bar pressure diff or something like that?)
So I experimented with this a bit on brake release. I observed in test reports that brake cylinder release had obvious exponential behavior (other things are more linear once the pressure difference is high enough, fluid dynamics in action I guess?)
To guarantee backwards compatibility, the nonlinearity is introduced only for 'advanced' brake cylinders. But then to allow users to enter the same MaxReleaseRate as before, the release rate is assumed to apply at the reference pressure then is multiplied by 2.5. 2.5 is roughly the 100%-10% time constant for exponential decay, meaning that the release rate the user entered is basically the _average_ release rate from full service to nearly full release.

To implement such a thing on a wider scale, we'd need a more unified system of handling air transfer, or manually go through every case and tweak the behavior to have realistic nonlinearities. The unified system is much more appealing and can fit into a larger refactor of the brakes. Maybe next time.

> Regarding EP brakes during emergency, the EP line could be set to 2 to denote emergency (i.e. -1 inactive 0 release .. 1 full brake 2 emergency, in this way you could achieve higher emergency pressures. Brake controllers also have to be adapted because some systems disconnect EP brakes during emergency, others simply trigger full service and others may request a higher pressure.
I can try that. Still don't have much if any EP equipment in my collection so more practical testing will be needed in any case.

> Could you link your trello card to the description?
Added, but it looks like Outlook got clever and decided to make a url embed as part of the email (I thought it was just a graphical thing, not content of the message itself, ugh!) which the bot didn't like so the card is a little mangled...oh well.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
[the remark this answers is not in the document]
cesarBLG: Would be interesting for the future if we use moles of air instead of P*V (i.e. we would have to divide by the Boltzmann constant and the temperature), or do your formulas explicitly require P*V?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: Sorry, I didn't see the URL. It's fine then

cesarBLG:
What I've seen regarding EP brakes and emergency is just a 3-way valve that connects BC to triple valve when energized, and to the supply res otherwise, what achieves a high-pressure and fast application during emergency. We can leave that for another PR, this one is already big enough.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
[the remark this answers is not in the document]
SteelFill: I don't strictly need the pressure * volume, that was just the easiest way to measure the "mass" of air in the brake cylinder while maintaining compatibility with the other reservoirs that only track pressure. Everything brake cylinder could be converted to track kgs or moles of air but that would introduce more conversion factors when moving air in and out of the brake cylinder since everything else is PSI and M^3.

For a large-scale refactoring, it would certainly be better to use a standardized (class or struct) air volume for _everything_, not just the brake cylinder, that is manipulated by mass flow instead of the present 'pressure flow' which doesn't make physical sense.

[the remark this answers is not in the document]
twpol: Just to clarify the policy on variable names: yes, we prefer units to be included, and we prefer them to be SI or at least metric, but it's not a hard rule.

If the code clarity benefits from no units, simplified units, or non-metric units that's fine.

It should still be clear to those reading the code how to figure out the units.

cesarBLG:
Hi Chris @cjakeman, this has been on unstable for a while. I'm waiting for the Trello card to be sorted before approving, code review is finished from my side. Could you look at that?

SteelFill:
As a heads up, I am out of town and will soon be without reliable/any internet for the next two weeks, so I won’t be making any changes soon.

I’m ready for this PR to be merged and off my plate, every feature has gotten some successful testing (including brake cylinder animations) so any changes would be additions of new features which I’d much rather put in a subsequent PR.

All that considered, I will be unavailable to merge the PR, so feel free to merge it without waiting for me to get back.

cjakeman:
> Hi Chris @cjakeman, this has been on unstable for a while. I'm waiting for the Trello card to be sorted before approving, code review is finished from my side. Could you look at that?

Thanks Cesar.

Trello card now moved into v1.6, so PR can be reviewed, approved and merged.

cesarBLG (approved):

PR #913, opened 2024-01-27
chore: Add explict Read the Docs configuration file

twpol:
Let's find out if this works at all.


cjakeman (approved):
Thanks for finding and fixing this, James.

PR #917, opened 2024-02-11
Lighting Configuration Enhancements

SteelFill:
A bunch of new features for loco and wagon lights which have been requested by the community and allow for significantly more complicated lighting configurations than previously possible.

(There are a lot of feature requests for lighting this addresses)
[The trello card](https://trello.com/c/djPB4R7K/561-new-lighting-conditions)
[Additional trello card](https://trello.com/c/v8CeqTvE)
[Another related trello card](https://trello.com/c/8rPMJLa0)
[Elvas Tower thread](https://www.elvastower.com/forums/index.php?/topic/37030-proposal-for-new-lighting-conditions/page__view__findpost__p__305488)

A bunch of new lighting conditions were added, so some significant refactoring was done in LightCollection.cs and Lights.cs to better facilitate the new conditions, as well as making it easier to add even more conditions in the future:
- Split management of light activation conditions into a new `LightCondition` class in LightCollection.cs
- Multiple sets of light conditions are now allowed in order to increase flexibility-a single Light in the .eng or .wag file can now have more than one `Conditions ( )` block and if _any_ of the sets of conditions are satisfied, the light will be enabled.
- Rewrote `UpdateState(LightViewer)` in Lights.cs to be easier to read (switch statements instead of endless if else if else chains), support multiple conditions, and to support the new conditions
- Rewrote `UpdateState()` in Lights.cs to detect activation of the new light conditions, and added an optimization to ignore changes in any light conditions that won't affect the current train car (with the increase in number of conditions, it's significantly likely that time would be wasted updating light states for a conditions every single light ignored)

The main focus is the new lighting conditions:
- `Brake` lets lights be enabled or disabled by friction brake applications
- `Reverser` can enable or disable lights based on the vehicle's direction of travel
- `Doors` ties into the passenger doors state to indicate when left/right doors are open
- `Horn` gives the ability to make automatic flashing ditch lights a reality, which is often requested
- `Bell` works the same as Horn, but is triggered by the bell instead
- `MU` customizes light behavior depending if the attached vehicle is the lead locomotive, connected to the lead locomotive, or remote to the lead locomotive
Exact usage for each condition is in the documentation and the linked E-T thread.

Also added is the ability to attach lights to parts of the shape file other than the main body. This is achieved by using the `ShapeHierarchy` token in the Light block. As a string, enter the name of the shape sub-object you want to attach the light to as the name appears in a program like Shape Viewer. `Light ( ShapeHierarchy ( "BOGIE1" )` would attach this light to the first bogie, for example. Any sub-object can be used, even nonsensical things like wheels or wiper blades.

The addition of a second headlights switch (rear headlights switch, aux headlights switch, light mode selector, or other similar switch) is out of scope for this PR but will be easier to implement in the future.

### DIFF
--- a/Source/Documentation/Manual/features-rollingstock.rst
+++ b/Source/Documentation/Manual/features-rollingstock.rst
@@ -9,7 +9,272 @@ For a full list of parameters, see :ref:`Developing OR Content - Parameters and 
 Train Engine Lights
 ===================
 
-OR supports the whole set of lights accepted by MSTS.
+OR supports the whole set of lights accepted by MSTS, MSTS-bin, and adds many new
+options to enhance the variety and complexity of lighting systems that can be recreated.
+
+Lights with multiple conditions
+-------------------------------
+
+In the original MSTS light implementation, each light could only have one set of
+activation conditions. If the same light were to be activated in multiple situations,
+(for example, a light which should turn on for both the front and rear units)
+the entire light would need to be included twice, just with different conditions.
+
+.. index::
+   single: Conditions
+
+Open Rails now allows for a single light to have multiple ``Conditions ()`` blocks.
+If *any* one set of conditions is fulfilled, the light will be enabled. If no conditions
+are specified, the light will be assumed to be on always. An example of how this can
+be used to simplify ``Lights`` implementation is included below::
+
+		Light	(
+			comment( Nose light bright )
+			Conditions	(
+				Headlight ( 3 )
+				Unit ( 2 )
+			)
+			FadeIn	( 0.5 )
+			FadeOut	( 0.5 )
+			States	(	1
+				State	(
+					LightColour ( FFffffe6 )
+					Radius ( 0.6 )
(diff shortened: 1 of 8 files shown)


cjakeman (dismissed):
Hi Phillip,

You've done a lot of work here. And it's good to see someone picking up and implementing a Trello Card request. Thanks.

Would you add a matching entry in news.rst also, please?

Chris

SteelFill:
@peternewell Peter, I know this pull request has been of interest to you, feel free to review at this point. While there are more ideas I have for lighting, I have captured the most requested improvements in this PR, so those more niche ideas can wait until after this set of improvements is in the testing version.

peternewell (approved):
Hi Phillip,

These changes look good, and they appear to be working ok from the testing that has been done.

Thanks for doing this.

peternewell (approved):
Thanks for your work on this.

Approved.

PR #919, opened 2024-02-23
Added mouse wheel support for controls which can be moved by pressing t…

sweiland-openrails:
Added mouse wheel support for controls which can be moved by pressing the left mouse button.

See trello car https://trello.com/c/V6pIZGjM/580-add-mousewheel-support.

And discussion: https://www.elvastower.com/forums/index.php?/topic/37779-handling-throttle-brakes-etc-with-mouse-scrollwheel/page__p__305122__hl__%2Bmouse+%2Bwheel__fromsearch__1#entry305122

### DIFF
--- a/Source/RunActivity/Viewer3D/Viewer.cs
+++ b/Source/RunActivity/Viewer3D/Viewer.cs
@@ -103,6 +103,8 @@ namespace Orts.Viewer3D
         public TrainListWindow TrainListWindow { get; private set; } // for switching driven train
         public TTDetachWindow TTDetachWindow { get; private set; } // for detaching player train in timetable mode
         public EOTListWindow EOTListWindow { get; private set; } // to select EOT
+        public ControlRectangle ControlRectangle { get; private set; } // to display the control rectangles
+
         private OutOfFocusWindow OutOfFocusWindow; // to show colored rectangle around the main window when not in focus
 
         // Route Information
@@ -501,6 +503,7 @@ namespace Orts.Viewer3D
             TrainListWindow = new TrainListWindow(WindowManager);
             TTDetachWindow = new TTDetachWindow(WindowManager);
             EOTListWindow = new EOTListWindow(WindowManager);
+            ControlRectangle = new ControlRectangle(WindowManager, this);
             if (Settings.SuppressConfirmations < (int)ConfirmLevel.Error)
                 // confirm level Error might be set to suppressed when taking a movie
                 // do not show the out of focus red square in that case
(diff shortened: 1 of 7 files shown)


Source/RunActivity/Viewer3D/Viewer.cs
```diff
@@ -1364,29 +1368,38 @@ void HandleUserInput(ElapsedTime elapsedTime)
...
-                                break;
-                            }
-                            else
-                            {
-                                foreach (var screen in controlRenderer.Control.Screens)
+                                var UserCommandControlTypes = (PlayerLocomotiveViewer as MSTSLocomotiveViewer).UserCommandControlTypes;
+                                if (UserCommandControlTypes.ContainsKey(controlRenderer.Control.ControlType.Type) && UserInput.IsMouseWheelChanged)
                                 {
-                                    if (cabRenderer.ActiveScreen[controlRenderer.Control.Display] == screen)
+                                    continue;
```
SteelFill, line 1383: I found that by changing this line of code out for
`MouseChangingControl = mouseRenderer; break;`
and reverting the changes to MSTSLocomotiveViewer.HandleUserInput line 339, the mouse scrolling became much more responsive and predictable. Because this method uses the same behavior as moving the mouse around, it doesn't need any of the input buffering, it doesn't freak out when trying to move multiple controls, spinning the mousewheel faster causes the control to move faster. Overall a vastly better experience that doesn't need to rely on workarounds.

I also think the resulting scroll direction behavior is more intuitive, as scrolling down moves controls down/left and scrolling up moves controls up/right, regardless of which direction is increasing/decreasing the control. While this might seem less consistent (sometimes scrolling up increases the brakes, sometimes it decreases the brakes, depends on the cab view) it makes for a better connection between the motion of my hand and the motion of the levers in game.

The only problem with this strategy that makes it worse than the current simulated button press strategy is that notched controls may take more than one click of the scroll wheel to advance by one notch.

SteelFill (changes requested):
This feature is very nice to have in theory, but the current implementation of sending the same commands as button presses makes some controls sluggish to respond to mouse wheel input, hard to move precisely, and has some bugs produced by the fact that the input must be sent over a finite period of time. There are ways to fix some of that and improve efficiency, but I think the alternate approach of treating scrolling like mouse movement has a lot more potential.

The control box highlighting is a very useful feature though and has worked just fine on the handful of cabs I tried.

Source/RunActivity/Viewer3D/Viewer.cs
sweiland-openrails: hi Phillip, thanks for your comments. I wasn't already not that satisfied with my code changes. This should have been implemented right from the start of MSTS/OR. But it might be there was no scroll wheel back then. 
I'll have a look at your comments once I'm finished with the review comments for my download PR.

sweiland-openrails: hi Philip, 
I'm afraid this is all too difficult for me. As I said I was already not that satisfied with my changes. 
I'm thinking of undoing these changes (except the ctrl-F5). Or would you like to take over this PR? Because it's still a nice thing to have, mouse wheel support. It's not that I'm dissatisfied with your comments, on the contrary they look good to me.
Please let me know what you think,
regards, Siebren.

SteelFill: I think it's a good idea to split the Ctrl + F5 and mouse wheel into two different PRs. They are both useful but aren't so closely connected that they need to be developed at the same time. Since the control hitboxes are already implemented, that PR could be quickly approved without the distraction of the mouse wheel feature.

I can investigate the scroll wheel feature in my own time, since I think I know some good ways to pull that off.

sweiland-openrails:
@SteelFill Hi Phillip,
I would like to cleanup my pull requests. Do you need still need this one or shall I remove it without merging? 
regards, Siebren.

sweiland-openrails:
closed without merge

SteelFill:
> closed without merge

Heads up, I still want to re-implement this feature my way, but that will likely come after #954 is done. Don't worry about closing the PR, I still have the information I need to rebuild it.

PR #920, opened 2024-02-24
Update RailDriver in Manual

cjakeman:
Adds detail for operation, calibration and switch assignment


Csantucci (approved):
Thank you for the time you devote to OR documentation improvements.

PR #923, opened 2024-03-05
Add curve squeal to route

peternewell:
Initial work to add curve squeal to route

Blueprint - https://blueprints.launchpad.net/or/+spec/add-curve-squeal-route

ET Thread - https://www.elvastower.com/forums/index.php?/topic/37783-curve-wheel-squeal/


cjakeman (changes requested):
Hi Peter,

I see we have changes to the HUD and also a new parameter "ortscurvesquealsmsnumber".

Does that mean that we also need some updates to the manual?

Best wishes,

Chris

peternewell:
> I see we have changes to the HUD and also a new parameter "ortscurvesquealsmsnumber".
> 
> Does that mean that we also need some updates to the manual?
>

This is still a WiP, so I have marked it as a draft for the time being.

Thanks for following up.

peternewell:
Closed due to repository rebuild, will reopen as a different PR.

PR #925, opened 2024-03-11
Fix brakeshoe force bug

peternewell:
Correct issue identified here - 

https://bugs.launchpad.net/or/+bug/2056722


cjakeman (commented):
Hi Peter,

Why do we have > 21 in the test but "exceeded {20}" in the text? Should they not be the same value, perhaps a named constant? Also should there be a note explaining where this value comes from?

Best wishes,

Chris

peternewell:
> 
> Why do we have > 21 in the test but "exceeded {20}" in the text? Should they not be the same value, perhaps a named constant? Also should there be a note explaining where this value comes from?
> 

Point taken, I have now changed it all to 20. Thanks for pointing this out.

cjakeman (approved):
Hi Peter,

Thanks for this change.

Approved,

Chris

peternewell:
Thanks Chris

PR #926, opened 2024-03-16
Lights Bugfixes

SteelFill:
Follow up to PR #917 to fix some edge cases that could result in null references.

Bug was reported [here on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/37900-crash-in-lightscs-code/)

- Add check if lead locomotive is null, which may happen on AI trains where there's no locomotive at the front
- Add check if Train is null to reverser condition, I'm not sure why this would happen but I did get it to happen
- And made some comparisons use nullable references so they won't die upon a null reference, but return false instead


Csantucci (changes requested):
You should check about Train == null also lines computing newCarIsFirst and newCarIsLast

SteelFill:
> You should check about Train == null also lines computing newCarIsFirst and newCarIsLast

That looks weird but still works because of operator precedence and short-circuiting. I don't know why the original code did that though, definitely not intuitive.

Csantucci (approved):
That's OK now for me

PR #927, opened 2024-03-20
AI Train Power Supply/Lights Fix

SteelFill:
AI trains that use the low voltage power supply feature of OR would behave as if the power supply was off all the time. This would cause said AI trains to turn off their lights (if the setting to control headlights with the master key was enabled).

[Reported on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__306793)

Added some additional ways for the power supply to be activated, which should guarantee that all AI trains turn on the power supply.

### DIFF
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -926,18 +926,11 @@ namespace Orts.Simulation.AIs
                             mstsSteamLocomotive.TenderCoalMassKG = mstsSteamLocomotive.MaxTenderCoalMassKG * Simulator.Activity.Tr_Activity.Tr_Activity_Header.FuelCoal / 100.0f;
                         }
 
-                        if (train.InitialSpeed != 0)
-                        {
-                            car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
-
-                            if (car is MSTSLocomotive loco)
-                                loco.SetPower(true);
-                        }
+                        if (train.InitialSpeed != 0 && car is MSTSLocomotive loco)
+                            loco.SetPower(true);
                     }
                     else
                     {
-                        car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
-
                         if (car is MSTSLocomotive loco)
                             loco.SetPower(true);
 
--- a/Source/Orts.Simulation/Simulation/AIs/AI.cs
+++ b/Source/Orts.Simulation/Simulation/AIs/AI.cs
@@ -927,11 +927,20 @@ namespace Orts.Simulation.AIs
                         }
 
                         if (train.InitialSpeed != 0)
+                        {
                             car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
+
+                            if (car is MSTSLocomotive loco)
+                                loco.SetPower(true);
+                        }
                     }
                     else
                     {
                         car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
+
+                        if (car is MSTSLocomotive loco)
+                            loco.SetPower(true);
+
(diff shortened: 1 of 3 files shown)


cesarBLG:
Wouldn't be wiser that AI trains activate the master key? PowerSupply scripts can be used to use e.g. different pantographs depending on the situation. In the future, if multiple voltages are allowed, the script would select the right pantograph, which has to be done also for AI trains.

SteelFill:
> Wouldn't be wiser that AI trains activate the master key? PowerSupply scripts can be used to use e.g. different pantographs depending on the situation. In the future, if multiple voltages are allowed, the script would select the right pantograph, which has to be done also for AI trains.

That would be interesting but I don't actually know how to do that. When scanning through things I couldn't even tell how the player train was supposed to change the master key state as everything connected to it is set private (presumably only the scripting API can decide, but I have no clue how that works).

cesarBLG:
I'll investigate the issue, but at the moment I cannot agree with this fix. Disabling power supply scripts will introduce new bugs in the future.

The master key is activated by sending the event PowerSupplyEvent.TurnOnMasterKey, which for the player train is sent by the ToggleMasterKeyCommand. For AI trains, all the power on sequence is handled by PowerSupplyEvent.QuickPowerOn which raises pantograph, closes circuit breaker and battery switch, and activates master key.

SteelFill:
Using MSTSLocomotive.SetPower(true) as an alias for quick power on seems to work. Going to try this method and let people see if that makes AI train lights work the way they should.

Source/Orts.Simulation/Simulation/AIs/AI.cs
```diff
@@ -927,11 +927,20 @@ public AITrain CreateAITrainDetail(Service_Definition sd, Traffic_Service_Defini
                         }
 
                         if (train.InitialSpeed != 0)
+                        {
                             car.SignalEvent(PowerSupplyEvent.RaisePantograph, 1);
```
cesarBLG, line 931: I assume it is no more required to raise the pantograph, since that is handled by QuickPowerOn?

cesarBLG:
> Using MSTSLocomotive.SetPower(true) as an alias for quick power on seems to work. Going to try this method and let people see if that makes AI train lights work the way they should.

This looks much better to me, thanks. Let's see if it works

Source/Orts.Simulation/Simulation/AIs/AI.cs
SteelFill: Probably not, though technically this runs for every single car while SetPower only works on locomotives. I'll check if it's even possible to have an edge case where a non-locomotive can raise a pantograph.

SteelFill: Sorry for the delay, I was stuck working on some engine file stuff rather than code. Looks to me like there probably won't be unintended consequences by removing the raise pantograph event so I've updated that for more testing.
A locomotive's pantographs are defined in MSTSWagon.cs but most useful pantograph behavior seems to be exclusively contained in MSTSElectricLocomotive.cs or assumes the pantograph is attached to a locomotive.

cesarBLG: Now that I think about it, it is common for EMUs to have pantographs in unpowered cars (i.e. in .wag files), so in principle this could cause those to not be raised. However, the PowerSupplyScript of the lead locomotive orders that pantographs are raised all along the train, so anyway they will be raised and it is correct to remove it as you did.

Having the RaisePantograph handled by the PowerSupply allows more flexibility over which pantographs are raised (e.g. in high speed trains usually only the pantograph on the rear locomotive is raised), so I think everything should be OK now.

SteelFill:
Been a couple of weeks and nobody's mentioned any problems about AI headlights or pantographs, so it seems to be working just fine.

PR #916, opened 2024-02-04
Distributed Power Air Brake Synchronization

SteelFill:
This PR adds a sorely missing feature for distributed power: the ability for remote locomotives to apply and release train/loco brakes in response to what the player locomotive is doing. This can dramatically improve brake response times on very long trains, the American creators out there will no doubt appreciate this.

[Relevant roadmap card](https://trello.com/c/urWcGqkA)
[Discussion thread on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/26249-remote-control-dpu-units/page__view__findpost__p__305271)

Changes:
- Changes to air brake propagation code to allow locomotives other than the player locomotive to apply/release brakes.
- Complete rewrite of engine brake code to allow groups of locomotives other than the player group to apply/release engine brakes. Also added a fix that allows steam locomotives to bail off the locomotive brake application on their tender.
- New token `engine(ORTSDPBrakeSynchronization` to allow users to decide which brake behavior should be transmitted from the lead locomotive to remote locomotives. `ORTSDPBrakeSynchronization ( "Apply, Release, Emergency, Independent" )` will fully enable brake synchronization. `"Apply"` means the remotes can apply the train brake, `"Release"` means remotes can release the train brakes, `"Emergency"` for remote emergency brake applications, and `"Independent"` allows remote units to change the locomotive brake to match the lead unit's brake cylinder pressure (including bail-off). Any combination of settings can be used. If the `ORTSDPBrakeSynchronization` token is missing, then the only setting active will be `"Independent"` since this somewhat replicates the default behavior from previous versions. Note that a setting will only have effect if BOTH the player locomotive and the distributed power locomotive have that setting.
- New cab view control `ORTS_TRAIN_AIR_FLOW_METER` which behaves like `ORTS_AIR_FLOW_METER` but will show the air flow rate for all locomotives in the train, rather than just the player locomotive. This distinction only matters if the train has distributed power units that can release the train brakes.
- Changed F5 HUD to show the total flow rate of all locomotives instead of just the player locomotive flow rate.
- Small rewrite to compressor activation logic to allow compressors to synchronize in groups of locomotives other than the lead locomotive. Also improved accuracy of compressor synchronization; if one locomotive still needs air, connected locomotives will be allowed to overcharge their MR by up to 5 psi in order to provide the demanded air.
- Bail-off will now use the engine brake release speed instead of the train brake release speed, should make bail-off take a lot less time.

[...]

PR #928, opened 2024-03-24
Fix Missing Initializations From Copy in MSTSLocomotive

SteelFill:
I noticed in some testing that locomotives initialized from a copy were not synchronizing DP brake commands (PR #916) as expected. Failed to notice this earlier as I had done all of my testing using consists formed of unique locomotives, never locomotives that were duplicates.

In the process of adding DP brake synchronization to the Copy() function, I noticed many other user inputs would not be copied to duplicated locomotives. This would result in those parameters producing the intended value only for the very first locomotive loaded, all other copies would use the default value.

I believe this PR covers all the missing copy commands in MSTSLocomotive.cs:

Player and trail loco variables: (These affect the operation of both the player locomotive and other locomotives in their train. As I found out, it can impact gameplay if these are not copied properly.)
- BrakePipeChargingRatePSIorInHgpS
- BrakePipeQuickChargingRatePSIpS
- EngineBrakeApplyRatePSIpS
- EngineBrakeReleaseRatePSIpS
- DPSyncTrainApplication
- DPSyncTrainRelease
- DPSyncEmergency
- DPSyncIndependent

Player loco only variables: (These only affect the player loco, so usually it would be harmless if these initialized to defaults, unless the player decided to change cabs. Less visible, but still added to Copy() for robustness.)
- BrakePipeTimeFactorS
- BrakeServiceTimeFactorPSIpS
- BrakeEmergencyTimeFactorPSIpS
- DoesVacuumBrakeCutPower
- DoesBrakeCutPower
- BrakeCutsPowerAtBrakeCylinderPressurePSI
- BrakeCutsPowerAtBrakePipePressurePSI
- BrakeRestoresPowerAtBrakePipePressurePSI
- EngineOperatingProcedures
- VigilanceMonitor (I don't actually know if this does anything)
- OnLineCabRadio
- OnLineCabRadioURL

I can only advise others and myself to ensure the Copy() initializer is **updated with every new user input added**, and to remember that the Parse() initializer is skipped entirely for duplicate rail vehicles.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -1243,7 +1243,7 @@ namespace Orts.Simulation.RollingStocks
             DynamicBrakeBlendingEnabled = locoCopy.DynamicBrakeBlendingEnabled;
             DynamicBrakeAvailable = locoCopy.DynamicBrakeAvailable;
             airPipeSystem = locoCopy.airPipeSystem;
-            DoesVacuumBrakeCutPower = locoCopy.DoesBrakeCutPower;
+            DoesVacuumBrakeCutPower = locoCopy.DoesVacuumBrakeCutPower;
             DoesBrakeCutPower = locoCopy.DoesBrakeCutPower;
             BrakeCutsPowerAtBrakeCylinderPressurePSI = locoCopy.BrakeCutsPowerAtBrakeCylinderPressurePSI;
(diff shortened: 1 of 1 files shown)


cesarBLG (dismissed):

SteelFill:
Sorry @cesarBLG, when doing my own checks before merging I realized I copy-pasted a bit too liberally and had `DoesVacuumBrakeCutPower = locoCopy.DoesBrakeCutPower` which isn't quite correct. Fixing that broke the review.

cesarBLG (approved):

PR #929, opened 2024-03-29
Add support for common circuit breaker variants

cesarBLG:
There are some common circuit breaker variants which are widely used in different stock. Including them into the OR source allows to easily include new features into them, without user intervention.

The following variants are added:
- PushButtons: dedicated buttons for opening and closing. Behaviour is similar to the Italian script. This variant is the most common in modern vehicles.
- TwoStage: one switch for driver closing authorization, and one button for driver closing order. Behaviour is similar to the French script. This variant is used in TGV trains, as well as in older electric locomotives.

Partly implements https://trello.com/c/xShE5iVJ/583-built-in-scripts-for-common-systems
Discussion: https://www.elvastower.com/forums/index.php?/topic/28645-circuit-breaker-for-electric-locomotives/page__view__findpost__p__307197

### DIFF
--- a/Source/Documentation/Manual/features-rollingstock.rst
+++ b/Source/Documentation/Manual/features-rollingstock.rst
@@ -1912,8 +1912,18 @@ Use the following .eng parameter to load a circuit breaker script::
   )
 
 ``ORTSCircuitBreaker`` refers to the circuit breaker script in the engine's ``Script`` 
-subfolder. For this field, the .cs extension is optional. "Automatic" and "Manual" load the generic OR 
-circuit breaker implementation, so do `not` use these names for your own script.
+subfolder. For this field, the .cs extension is optional. Alternatively, there are several
+built-in OR circuit breaker implementations:
+
+- "Automatic": no driver intervention required, circuit breaker is closed when conditions are met.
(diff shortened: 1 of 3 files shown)


cjakeman:
Hi Cesar,

Thanks for this. Very pleased to see you have some time for enhancing Open Rails.

Do you have an Trello Card or a Blueprint which I can approve?

Thanks,

Chris

cesarBLG:
Thanks, Chris! Sure, I updated the PR description

cesarBLG:
Hi Chris,
"Automatic" and "Manual" circuit breaker variants already have a built-in implementation. I'm not sure about whether we should flood the log file with unnecessary messages (i.e. it is more likely that the built-in implementation is what the content creator wants). In CircuitBreaker.cs the existence of a .cs file with the same name is not known, so this situation will always be logged.

The typical use case for scripts is to store them in a common.script folder, and in that situation the built-in behaviour is never invoked. Also, if you specify "PushButtons.cs" instead of "PushButtons", the custom script will be loaded instead of the built-in one. Maybe this can be clarified in the manual.

cjakeman (approved):
Hi Cesar,

Approved and merged.

Thanks,

Chris

PR #933, opened 2024-04-12
Fix AI DM crashing

peternewell:
Correct bug identified in this Bug Report

https://bugs.launchpad.net/or/+bug/2061047

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -2015,9 +2015,6 @@ namespace Orts.Simulation.RollingStocks
 
             // Pass Gearbox commands
 
-
-
-
             // Note - at the moment there is only one GearBox Controller created, but a gearbox for each diesel engine is created. 
             // This code keeps all gearboxes in the locomotive aligned with the first engine and gearbox.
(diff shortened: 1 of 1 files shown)


cjakeman (approved):
Hi Peter,

Looks very straightforward.

Approved with thanks,

Chris

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2156,8 +2153,13 @@ public override void Update(float elapsedClockSeconds)
                                     if (de.State != DieselEngineState.Running)
                                         de.Initialize();
                                 }
+
+                                // if train is a geared locomotive then set it to automatic operation as AI driver can't operate manual gearboxes
                                 if (de.GearBox != null)
                                     de.GearBox.GearBoxOperation = GearBoxOperation.Automatic;
+
+                                // Set gear to "low gear" at start.
+                                de.GearBox.currentGearIndex = de.GearBox.NumOfGears - 1;
```
SteelFill, line 2162: Appears this change is causing a new crash on diesel electric locomotives https://www.elvastower.com/forums/index.php?/topic/37993-940-stable-build-autopilot-immediate-ctd/

This line should probably be added underneath the if statement in line 2158 so that it doesn't try to set low gear if there is no gearbox.

peternewell: Ok, I have raised another PR to adjust as you have suggested.

Thanks

PR #934, opened 2024-04-16
Fix bug 1944070: ContentManager search exception.

rwf-rr:
ContentManager search crashed when one of the following letters was typed into the search box: A,C,E,G,M,N,O,R,T. These are the letters in "ContentManager", the name of the root element in the content tree. The root element was being added to the search results, which then searches the element for a directory separator (/), and caused an System.ArgumentOutOfRangeException.

See https://bugs.launchpad.net/or/+bug/1944070.

The fix is to not compare the element name tothe search string when the element is the root.

### DIFF
--- a/Source/Contrib/ContentManager/ContentManagerGUI.cs
+++ b/Source/Contrib/ContentManager/ContentManagerGUI.cs
@@ -270,7 +270,8 @@ namespace ORTS.ContentManager
                     var content = pending[path];
                     pending.Remove(path);
 
-                    if (content.Name.ToLowerInvariant().Contains(search))
+                    // do not add root element to search results, see bug 1944070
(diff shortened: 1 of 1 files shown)


rwf-rr:
I am disinclined to fix the Codacy Static Code Analysis issue. Adding curly braces for my change would be inconistent with the rest of the file.

rwf-rr:
@twpol I believe you were the original; author of the ContentManager. Could you please review this simple fix? Thanks.

twpol:
> I am disinclined to fix the Codacy Static Code Analysis issue. Adding curly braces for my change would be inconistent with the rest of the file.

Don't worry about the two analysis checks, we are not enforcing anything they say - they're just for curiosity.

twpol (approved):

PR #935, opened 2024-04-20
Correct bug created in diesel electric locomotive

peternewell:
Correct bug identified in this Bug Report

https://bugs.launchpad.net/or/+bug/2061047

Subsequent bug reported - 

Appears this change is causing a new crash on diesel electric locomotives https://www.elvastower.com/forums/index.php?/topic/37993-940-stable-build-autopilot-immediate-ctd/

This line should probably be added underneath the if statement in line 2158 so that it doesn't try to set low gear if there is no gearbox.


cjakeman (approved):
Hi Peter,

Approved and thanks for quick response,

Chris

peternewell:
Thanks Chris

PR #936, opened 2024-04-21
ContentManager Enhancement: show more info for consists

rwf-rr:
In the **ContentManager**, show additional properties for **consists**, such as: engine and car counts, total length, total weight, horsepower-per-ton, tons-per-operative-brake, etc. The latter two are North America specific - I don't know if there are equivalents for pther parts of the world.

### Update Nov 2024
- Added lowest coupler's strength to the consist info (and also to the car info).
- Added special cases for units when using the system's locale (UK, Canada).

## Details
The most significant change is that the consist view now has to read and parse each car (.wag or .eng file). The search function is negatively impacted, taking about 40% longer (in my case 2:40 min vs 2 min). 

The units are based on the ORTS options. Since the ContentManager does not have a route context for consists, the system units are used in that case.

Impacted files that are not ContentManager specific:
- Source/ORTS.Common/Conversions.cs: add methods to format HPT, TOPB and the weight bar-graph.
- Source/Orts.Formats.Msts/ActivityFile.cs: added IsWagon flag.
- Source/Orts.Formats.Msts/EngineFile.cs: read more properties from the .eng files.
- Source/Orts.Formats.Msts/WagonFile.cs: read more properties from the .wag files.

Tested on: Train Simulations Scenic, Mullan Pass and Selgman packages; base MSTS packages; some Swiss packages.

### Consist
The additional properties. HPT and TPOB are shown only when US Imperial units are used.
image

Also added a bar-graph to the consist, so that loads and empties are identifyable. Each asterisk represents 20 metric tons.
image

### Cars
Since some of the above fields are aggregated, they might as well be shown for cars.

Engines:
image

Wagons:
image

### Other
Disabled word-wrapping in the Text box, in favour of horizontal scrolling. The bar-graph is not very readable with line-wraps.


Source/Contrib/ContentManager/ContentManagerGUI.Designer.cs
rwf-rr, line 1: Besides the WordWrap change, also corrected the size of two panels. Does not have any impact, as they were sized based on the parent container.

Source/Orts.Formats.Msts/WagonFile.cs
```diff
@@ -27,7 +27,32 @@ namespace Orts.Formats.Msts
...
+            public float CarHeight;
+            public float CarLength;
+
+            public CarSize(STFReader stf)
+            {
+                stf.MustMatch("(");
+                CarWidth = stf.ReadFloat(STFReader.UNITS.Distance, null);
+                CarHeight = stf.ReadFloat(STFReader.UNITS.Distance, null);
+                CarLength = stf.ReadFloat(STFReader.UNITS.Distance, null);
+                stf.SkipRestOfBlock(); // or should this be stf.MustMatch(")")
```
rwf-rr, line 42: Which would be better here, SkipRestOfBlock or MustMatch?

twpol: If we have no reason to expect more data in the block, I'd lean slightly towards `MustMatch()` simply so that it'll print a warning message if it **isn't** the end of the block.

Source/Orts.Formats.Msts/WagonFile.cs
```diff
@@ -38,6 +63,10 @@ public WagonFile(string filePath)
                         stf.ReadString();
                         stf.ParseBlock(new STFReader.TokenProcessor[] {
                             new STFReader.TokenProcessor("name", ()=>{ Name = stf.ReadStringBlock(null); }),
+                            new STFReader.TokenProcessor("type", ()=>{ WagonType = stf.ReadStringBlock(null); }),
+                            new STFReader.TokenProcessor("mass", ()=>{ MassKG = stf.ReadFloatBlock(STFReader.UNITS.Mass, null); }),
+                            new STFReader.TokenProcessor("size", ()=>{ LengthM = new CarSize( stf).CarLength; }),
```
twpol, line 68: Wouldn't it make more sense to put the `CarSize` as a whole into a field?

Source/Orts.Formats.Msts/ActivityFile.cs
```diff
@@ -1331,7 +1332,7 @@ public class Wagon {
                 new STFReader.TokenProcessor("uid", ()=>{ UiD = stf.ReadIntBlock(null); }),
                 new STFReader.TokenProcessor("flip", ()=>{ stf.MustMatch("("); stf.MustMatch(")"); Flip = true; }),
                 new STFReader.TokenProcessor("enginedata", ()=>{ stf.MustMatch("("); Name = stf.ReadString(); Folder = stf.ReadString(); stf.MustMatch(")"); IsEngine = true; }),
-                new STFReader.TokenProcessor("wagondata", ()=>{ stf.MustMatch("("); Name = stf.ReadString(); Folder = stf.ReadString(); stf.MustMatch(")"); }),
+                new STFReader.TokenProcessor("wagondata", ()=>{ stf.MustMatch("("); Name = stf.ReadString(); Folder = stf.ReadString(); stf.MustMatch(")"); IsWagon = true; }),
```
twpol, line 1335: I don't see anything that uses `IsWagon`, but more importantly it is simply the opposite of `IsEngine` logically. There's only the two possibilities here.

Can we remove this change?

Source/ORTS.Common/Conversions.cs
```diff
@@ -720,6 +720,25 @@ public static string FormatLargeMass(float massKg, bool isMetric, bool isUK)
...
+                case 3: massBar = "***"; break;
+                case 4: massBar = "****"; break;
+                case 5: massBar = "*****"; break;
+                case 6: massBar = "******"; break;
+                case 7: massBar = "*******"; break;
+                case 8: massBar = "********"; break;
+                default: massBar = massKg < 1.0 ? "?" : "********>"; break;
+            }
+            return massBar;
+        }
```
twpol, line 740: I don't think this is appropriate in this shared area, as it's not a generic conversion that could have use anywhere else; let's move it into the ContentManager project.

This should also be simplified with the knowledge that `new string('*', count)` will give a string containing `count` `*` characters.

Source/ORTS.Common/Conversions.cs
```diff
@@ -913,5 +932,33 @@ public static string FormatApproximateTime(double clockTimeSeconds)
...
+        /// </summary>
+        /// <param name="consistMassKG"></param>
+        /// <param name="consistNumOpBrakes"></param>
+        /// <returns>tons-per-operative-brake formated to an integer.</returns>
+        //TODO: implement UK and metric version
+        public static string FormatTPOB(float consistMassKG, float consistNumOpBrakes)
+        {
+            var tpob = consistNumOpBrakes > 0 ? Kg.ToTUS(consistMassKG) / consistNumOpBrakes : 0;
+            return string.Format("{0:0}", tpob);
+        }
```
twpol, line 962: I would prefer we skip these two formats; they haven't been needed anywhere else to date, so I'm not sure on the merits of adding them just for an unofficial project (all `Contrib` code is ~unofficial).

The lack of metric/imperial/etc. implementation is also an issue.

Let's move them into the Content Manager project for now, and they can always be moved here in the future if we need them.

Source/Contrib/ContentManager/Models/Car.cs
```diff
@@ -33,25 +33,39 @@ public enum CarType
...
+        public readonly float MaxBarkeForceN;
+        public readonly float MaxPowerW;
+        public readonly float MaxForceN;
+        public readonly float MaxSpeedMps;
 
         public Car(Content content)
         {
             Debug.Assert(content.Type == ContentType.Car);
+
+            // .eng files also have a wagon block
```
twpol, line 50: Nice refactor here!

twpol (changes requested):
This is a very solid set of changes and – even though I am surprised that people are interested in the contributed Content Manager project – I am happy for it to go in.

There are just a few minor things I'd like changed, which are hopefully easy.

Source/Orts.Formats.Msts/WagonFile.cs
rwf-rr: Thanks for the feedback. Changed to use MustMatch.

Source/ORTS.Common/Conversions.cs
rwf-rr: Moved to ContentInfo.cs (it already had a Format function). Also simplified as suggested.

Source/ORTS.Common/Conversions.cs
rwf-rr: Done.

Source/Orts.Formats.Msts/ActivityFile.cs
rwf-rr: Done.

Source/Orts.Formats.Msts/WagonFile.cs
rwf-rr: Yes. Done.

Source/Contrib/ContentManager/Models/Car.cs
rwf-rr: Thanks for the feedback.

twpol (approved):
Thanks very much for the updates and apologies for the time it has taken me to get back and review this!

PR #938, opened 2024-04-26
Fix Missing Initializations From Copy in MSTSWagon.cs

SteelFill:
Follow up to PR #928, similar process but for MSTSWagon.cs. Again while testing, I noticed that the first instance of a given wagon had different standstill friction than any other instances of the wagon, indicating to me that something was missing when initializing from a copy.

Luckily it seems most things are set correctly, I only noticed missing initialization from copy for:
- WagonNumAxles (this is what was causing the discrepancy in friction calculations)
- HasInsideView (unsure if this was actually causing problems, but it might help for trains with lots of passenger/3D cabviews)

I also checked the save and restore functions, some variables were being saved even though it was impossible for them to change during runtime:
- IsDavisFriction
- IsRollerBearing
- IsLowTorqueRollerBearing
- IsFrictionBearing
- StandstillFrictionN
- MergeSpeedMpS

As always, make sure any variables calculated during initialization are copied, and don't bother to save/restore a variable unless that variable changes during gameplay.


cesarBLG (approved):
Thanks, it looks great

PR #939, opened 2024-05-01
Add KM/HOUR/MIN to Cabview Units + Custom Units

SteelFill:
At the request of multiple people, kilometers per hour per minute has been added as a cabview unit. This seems to be common on American built locomotives sent internationally as a metric replacement for the usual mph/min display.

[Trello Card](https://trello.com/c/pOk3XiFK)
[Elvas Tower Discussion](https://www.elvastower.com/forums/index.php?/topic/24040-3d-cabs/page__st__790__p__308227#entry308227)

Realizing it is entirely impractical to make a code change for every single unit people want to add, I also added some tokens which can be added to the control block in the .cvf file to allow for custom units.
- `ORTSUnitsExponent ( x )` - Raises the value shown by the cab view control to the power `x`. There aren't many cases where this is useful for units conversion, but it is useful for the various types of nonlinear gauges present in locomotives. Can be set to any decimal number (negative exponents to invert the relationship, fractional exponents to do square roots, 0...all should work). If not specified, the exponent is 1 (ie: no change).
- `ORTSUnitsScaleFactor ( y )` - After the exponent, multiplies the value shown by the cab view control by a factor of `y`. For example, a cab view control with units of MILES_PER_HOUR can be converted to feet per second (a unit ORTS does not support) using ORTSUnitsScaleFactor ( 1.467 ). Can be any decimal number (negative, fractional, 0...all should work). If not specified, the factor is 1 (ie: no change).
- `ORTSUnitsOffset ( z )` - After the scale factor is applied, adds `z` to the value shown by the cab view control. For example, a cab view control with units of BAR actually shows gauge pressure specifically. This can be converted to absolute pressure in BAR with ORTSUnitsOffset ( 0.987 ). This can also be any decimal number, but negative numbers will reduce the value shown by the cab view control. If not specified, the offset is 0 (ie: no change).

### DIFF
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2147,7 +2147,7 @@ namespace Orts.Viewer3D.RollingStock
                         {
                             if (Locomotive.CruiseControl != null)
                             {
-                                if ((Locomotive.CruiseControl.SpeedRegMode == Simulation.RollingStocks.SubSystems.CruiseControl.SpeedRegulatorMode.Auto && !Locomotive.CruiseControl.DynamicBrakePriority) || Locomotive.DynamicBrakeIntervention > 0)
+                                if (Locomotive.CruiseControl.SpeedRegMode == Simulation.RollingStocks.SubSystems.CruiseControl.SpeedRegulatorMode.Auto && !Locomotive.CruiseControl.DynamicBrakePriority)
                                 {
                                     index = 0;
                                 }
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2147,7 +2147,7 @@ namespace Orts.Viewer3D.RollingStock
                         {
                             if (Locomotive.CruiseControl != null)
                             {
-                                if (Locomotive.CruiseControl.SpeedRegMode == Simulation.RollingStocks.SubSystems.CruiseControl.SpeedRegulatorMode.Auto && !Locomotive.CruiseControl.DynamicBrakePriority)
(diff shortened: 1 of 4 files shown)


Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2147,7 +2147,7 @@ public virtual int GetDrawIndex()
                         {
                             if (Locomotive.CruiseControl != null)
                             {
-                                if ((Locomotive.CruiseControl.SpeedRegMode == Simulation.RollingStocks.SubSystems.CruiseControl.SpeedRegulatorMode.Auto && !Locomotive.CruiseControl.DynamicBrakePriority) || Locomotive.DynamicBrakeIntervention > 0)
+                                if (Locomotive.CruiseControl.SpeedRegMode == Simulation.RollingStocks.SubSystems.CruiseControl.SpeedRegulatorMode.Auto && !Locomotive.CruiseControl.DynamicBrakePriority)
```
cjakeman, line 2150: Hi Phillip,

Wondering why is there a change in the logic here. Surely this has nothing to do with units of measure?

If it's a bug-fix, please move it to a separate PR.

Thanks,

Chris

cjakeman (changes requested):
Hi Phillip,

Scale and offset is a nice extension.

Only issue is the missing "Locomotive.DynamicBrakeIntervention > 0" and the reason for it.

Chris

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
SteelFill: Noticed while testing cab views that anything with cruise control would not show the dynamic brake setting at all due to a change in dynamic brake code made after cruise control was added (basically, dynamic brake intervention used to always be 0 unless cruise control was on, now dynamic brake intervention will change from 0 whenever the dynamic brake is used).

This can be done as a separate mini PR, got a bit I want to change with the units anyway as some more feedback has pointed out how non-linear gauges are effectively impossible to calibrate-an option for exponential units will be needed.

cjakeman (approved):
Hi Phillip,

All good.
Love this bit:
"To avoid potential imaginary numbers, use data's absolute value"

Approved and merged.

Chris

PR #940, opened 2024-05-04
Browser HUD column alignment improvements

rwf-rr:
A limited effort to improve the alignment of columns in the browser-based head up display (HUD). The whole HUD could benefit from some refactoring to improve the presentation, but that is beyond my intent. Some challenges:
- The non-common part is represented by a single table (two-dimensional array of strings), but it contains data in multiple formats:
   - preformated data spanning the full row, eg "Wind speed" in the forces desplay;
   - keyword-value pairs;
   - sub-tables with the label in the first column;
   - sub-tables with the label in the first row.
- The browser auto-sizes the columns. When a value changes magnitude, the cell-size may change and the subsequent columns may shift.
- The table cells are left-aligned. The decimal separator shifts as the magnitude changes, or the value goes negative.
- The unit may change, possibly causing the cell-size (and thus column size) to change.

Unfortunately I got stuck after the first change, fixing general alignment problems with the Forces HUD. I was not able to fix the Brake and Loco HUD alignment (without putting too much semantic details into index.js). And trying to right-align cells quickly became complex and ugly, and would also affect the in-display presentation.

### Force Info

Only the browser presentation is changed, the in-display presentation is unaffected.

Before:
image

After:
image

I verified that it also works with the first set of data reported for each axle, even though there does not seem to be any case where multiple axles are reported (I had to fake it).

### Brake Info

There is misalignment between the player locomotives and the rest of the locomotives.

image

Adjusting colspan did not fix it, or cause other problems.

Also , the following looks like a bug:

```
else if (colspan3 || BrakeColSpanTo2.indexOf(obj.extraTable.values[next]) !== -1){
    Str += "<td colspan='3' >" + obj.extraTable.values[next] + "</td>";
}
```

``BrakeColSpanTo2`` is different from ``colspan3``. But splitting this up into two blocks, one for colspan 2 and one for colspan 3, does not help. The "PlayerLoco" and "Loco" rows are still misaligned.

### DIFF
--- a/Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
+++ b/Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
@@ -86,8 +86,9 @@ function ApiHeadUpDisplay() {
 						DebugWidthTofix = ['primitives','Camera'],
 					// Force information
 						ForceColSpanTo15 = ['Wind Speed'],
-						ForceColSpanTo4 = ['(Simple adhesion model)','Dynamic brake'],
-						ForceColSpanTo3 = ['Axle out force','Loco Adhesion','Wagon Adhesion'],
+						ForceColSpanTo4 = ['(Advanced adhesion model)', '(Simple adhesion model)'],
+						ForceColSpanTo3 = ['Wheel slip (Thres)', 'Conditions', 'Axle drive force', 'Axle brake force', 'Number of substeps', 'Wheel Adh. (Max)',
+							'Axle out force', 'Comp Axle out force', 'Wheel speed (Slip)', 'Wheel ang. pos.', 'Loco Adhesion', 'Wagon Adhesion'],
 					// Brake information
 						BrakeColSpanTo2 = ['PlayerLoco','Brake Sys Vol'];
 
@@ -96,6 +97,7 @@ function ApiHeadUpDisplay() {
 						var colspanmax = false,
 							colspan10 = false,
 							colspan3 = false,
+							colspan2 = false,
 							fixwidth = false,
 							locomotivetype = false;
 						
@@ -126,6 +128,10 @@ function ApiHeadUpDisplay() {
 								locomotivetype = true;
 								Str += "<td colspan='2' >" + obj.extraTable.values[next] + "</td>";
 							}
+							else if (colspan2)
+							{
+								Str += "<td class='td_nowrap' colspan='2' >" + obj.extraTable.values[next] + "</td>";
+							}
 							else if (colspan3 || BrakeColSpanTo2.indexOf(obj.extraTable.values[next]) !== -1){
 								Str += "<td colspan='3' >" + obj.extraTable.values[next] + "</td>";
 							}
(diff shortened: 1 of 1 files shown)


Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
```diff
@@ -86,8 +86,9 @@ function ApiHeadUpDisplay() {
 						DebugWidthTofix = ['primitives','Camera'],
 					// Force information
 						ForceColSpanTo15 = ['Wind Speed'],
-						ForceColSpanTo4 = ['(Simple adhesion model)','Dynamic brake'],
-						ForceColSpanTo3 = ['Axle out force','Loco Adhesion','Wagon Adhesion'],
+						ForceColSpanTo4 = ['(Advanced adhesion model)', '(Simple adhesion model)'],
+						ForceColSpanTo3 = ['Wheel slip (Thres)', 'Conditions', 'Axle drive force', 'Axle brake force', 'Number of substeps', 'Wheel Adh. (Max)',
+							'Axle out force', 'Comp Axle out force', 'Wheel speed (Slip)', 'Wheel ang. pos.', 'Loco Adhesion', 'Wagon Adhesion'],
 					// Brake information
```
rwf-rr, line 92: - If simple adhesion is present, advanced adhesion should also be there.
- "Dynamic brake" is not present in the forces HUD. It is present in the Brakes HUD (part of a pre-formated string) and in the common part (uses different presenation code).
- The existing colspan-3 fields may all have multiple values (per axle), although that does not seem to ever happen. The fields I added also have per-axle values.

Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
```diff
@@ -126,6 +128,10 @@ function ApiHeadUpDisplay() {
 								locomotivetype = true;
 								Str += "<td colspan='2' >" + obj.extraTable.values[next] + "</td>";
 							}
+							else if (colspan2)
+							{
+								Str += "<td class='td_nowrap' colspan='2' >" + obj.extraTable.values[next] + "</td>";
+							}
```
rwf-rr, line 134: For the rows that have per-axle values (in the Forces table).

Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
```diff
@@ -144,6 +150,7 @@ function ApiHeadUpDisplay() {
 							}
 							else if (ForceColSpanTo3.indexOf(obj.extraTable.values[next]) !== -1 ){
 								Str += "<td colspan='3' >" + obj.extraTable.values[next] + "</td>";
+								colspan2 = true; // data cells in these row span two columns
```
rwf-rr, line 153: For the rows that have per-axle values (in the Forces table).

cjakeman (approved):
Hi Roger, 

Thanks for the before and after screenshots. This looks much better.

No quibbles with your code.

Approved with thanks,

Chris

Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
cjakeman: Agreed

Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
cjakeman: Agreed

Source/RunActivity/Viewer3D/WebServices/Web/HUD/index.js
cjakeman: Agreed

PR #941, opened 2024-05-05
AI Penalty Lights Fix

SteelFill:
Fix for [this bug](https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__308409). The AI train brake simulation is not detailed enough to correctly determine if an AI train should be showing emergency lights or not. In this case, the SNCF Z 50000 unit spawns in with the brake handle in emergency, which triggers the emergency lights on every AI train even if they are not braking.

Since AI trains should never use the emergency brake anyway, we are reverting to the old behavior which simply ignores the emergency lights on AI trains.


cjakeman (approved):
Hi Phillip,

As you say "Quite an edge case"!

Approved and merged. 

Thanks,

Chris

PR #942, opened 2024-05-11
Fix wrong case in filename

cesarBLG:
This eases in a case-sensitive filesystem (Linux)


cjakeman (approved):
Hi Cesar,

Good catch.

Approved and merged.

Chris

PR #943, opened 2024-05-14
Dynamic Brakes Behavior & Display Improvements

SteelFill:
Due to some relatively recent improvements in dynamic brake behavior, particularly PR #874, parts of the dynamic brake control and display code have become bloated with workarounds that are no longer needed, which are resulting in some unrealistic and unintended behaviors when combined with the updated behaviors. This PR reworks much of the dynamic brake handling and cab view display to closer match reality, simplify the code, and provide more robust handling of user input.

- Dynamic brakes are enabled if the dynamic brake lever is set above 0%, disabled if the lever is at 0%; it is no longer possible to enter the confusing state where the dynamic brake lever is set to off but the dynamic brakes remain active
- Side effect: Except in some edge cases, it will no longer be possible to have the dynamic brakes enabled and producing no braking force-this is per reality
- Tracking of whether or not the dynamic brakes should be enabled or disabled is now done directly based on the position of the dynamic brake handle
- Removed significant redundant code, variables, and functions related to determining whether or not the dynamic brake is enabled or disabled
- Reworked some of the combined handle code to more reliably handle transition between power and braking, especially in mouse controls
- Removed now-unneeded checks related to cruise control in the dynamic brake cab view code (this was preventing dynamic brake displays from working correctly)
- Improved consistency of dynamic brake interlocking (with reverser and with throttle) in all control styles (combined and split) and control modes (keyboard, mouse...unsure of raildriver since that behaves differently); should be much harder to produce a situation where the throttle and dynamic brakes are applied simultaneously
- Refactored how control animation frames are selected to improve tolerance to user errors, especially when control values are entered in the wrong order, (including some errors in original MSTS cab views) and more reliably produce the indications expected for the situation at hand (specifically combined controls)

https://bugs.launchpad.net/or/+bug/2066456

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -2106,10 +2106,10 @@ namespace Orts.Simulation.RollingStocks
                     if (DynamicBrakeCommandStartTime + DynamicBrakeDelayS < Simulator.ClockTime)
                     {
                         DynamicBrake = true; // Engage
-                        if (IsLeadLocomotive() && DynamicBrakeController.CurrentValue > 0)
+                        if (IsLeadLocomotive() && DynamicBrakeController?.CurrentValue > 0)
                             Simulator.Confirmer.ConfirmWithPerCent(CabControl.DynamicBrake, DynamicBrakeController.CurrentValue * 100);
                     }
-                    else if (IsLeadLocomotive() && DynamicBrakeController.CurrentValue > 0)
+                    else if (IsLeadLocomotive() && DynamicBrakeController?.CurrentValue > 0)
                         Simulator.Confirmer.Confirm(CabControl.DynamicBrake, CabSetting.On); // Keeping status string on screen so user knows what's happening
                 }
             }
@@ -2394,24 +2394,31 @@ namespace Orts.Simulation.RollingStocks
             {
                 LocalDynamicBrakePercent = TrainBrakeController.TrainDynamicBrakeIntervention * 100;
             }
-            if (DynamicBrakeController != null && DynamicBrakeController.CurrentValue > 0)
+            if (DynamicBrakeController != null && (DynamicBrakeController.CurrentValue > 0 || DynamicBrakeController.UpdateValue > 0))
             {
                 float prevValue = DynamicBrakeController.CurrentValue;
+
                 if (DynamicBrake || !DynamicBrakeControllerSetupLock)
                     DynamicBrakeController.Update(elapsedClockSeconds);
-                // Dynamic brake is about to be disabled, pause motion of controller to prevent accidental shut-off
-                if (DynamicBrakeController.CurrentValue <= 0)
+
+                // Dynamic brake is about to be enabled or disabled, pause motion of controller to prevent accidental overshooting
+                if ((DynamicBrakeController.CurrentValue <= 0 && prevValue > 0) || (DynamicBrakeController.CurrentValue > 0 && prevValue <= 0))
                 {
                     if (!DynamicBrakePause)
                     {
                         DynamicBrakePause = true;
-                        StopDynamicBrakeDecrease();
-                        DynamicBrakeController.CurrentValue = prevValue;
+                        if (DynamicBrakeController.UpdateValue < 0)
+                        {
+                            StopDynamicBrakeDecrease();
+                            DynamicBrakeController.CurrentValue = prevValue;
+                        }
+                        else if (DynamicBrakeController.UpdateValue > 0)
+                            StopDynamicBrakeIncrease();
                     }
                     else
                         DynamicBrakePause = false;
                 }
-                else if (DynamicBrakeController.CurrentValue > prevValue)
+                else if (DynamicBrakeController.UpdateValue != 0)
                     DynamicBrakePause = false;
 
                 LocalDynamicBrakePercent = Math.Max(DynamicBrakeController.CurrentValue * 100, LocalDynamicBrakePercent);
@@ -4098,7 +4105,7 @@ namespace Orts.Simulation.RollingStocks
                 ThrottleController.IntermediateValue = 0;
             }
             // Return to throttling
-            else if ((DynamicBrakeController.CurrentValue <= 0 && value < CombinedControlSplitPosition) ||
+            else if ((DynamicBrakeController?.CurrentValue ?? 0) <= 0 ||
                 (CruiseControl != null && !CruiseControl.DynamicBrakePriority && CruiseControl.SpeedRegMode == CruiseControl.SpeedRegulatorMode.Auto))
             {
                 SetThrottleValue(1 - MathHelper.Clamp(value, 0, CombinedControlSplitPosition) / CombinedControlSplitPosition);
@@ -4118,9 +4125,9 @@ namespace Orts.Simulation.RollingStocks
         /// <returns>Combined position into 0-1 range, where arrangement is [[1--throttle--0]split[0--dynamic|airbrake--1]]</returns>
         public float GetCombinedHandleValue(bool intermediateValue)
         {
-            float throttleValue = intermediateValue ? ThrottleController.IntermediateValue : ThrottleController.CurrentValue;
-            float dynamicsValue = intermediateValue ? DynamicBrakeController.IntermediateValue : DynamicBrakeController.CurrentValue;
-            float brakesValue = intermediateValue ? TrainBrakeController.IntermediateValue : TrainBrakeController.CurrentValue;
+            var throttleValue = intermediateValue ? ThrottleController?.IntermediateValue : ThrottleController?.CurrentValue;
+            var dynamicsValue = intermediateValue ? DynamicBrakeController?.IntermediateValue : DynamicBrakeController?.CurrentValue;
+            var brakesValue = intermediateValue ? TrainBrakeController?.IntermediateValue : TrainBrakeController?.CurrentValue;
 
             if (CruiseControl?.SpeedRegMode == CruiseControl.SpeedRegulatorMode.Auto)
             {
@@ -4142,14 +4149,14 @@ namespace Orts.Simulation.RollingStocks
                 if (CruiseControl != null && CruiseControl.SkipThrottleDisplay && !CruiseControl.DynamicBrakeCommandHasPriorityOverCruiseControl)
                     return CombinedControlSplitPosition;
                 else
-                    return CombinedControlSplitPosition + (1 - CombinedControlSplitPosition) * dynamicsValue;
+                    return CombinedControlSplitPosition + (1 - CombinedControlSplitPosition) * (dynamicsValue ?? 0);
             }
             else if (CombinedControlType == CombinedControl.ThrottleAir && throttleValue <= 0 && brakesValue > 0)
-                return CombinedControlSplitPosition + (1 - CombinedControlSplitPosition) * brakesValue;
+                return CombinedControlSplitPosition + (1 - CombinedControlSplitPosition) * (brakesValue ?? 0);
             else if (CruiseControl == null)
-                return CombinedControlSplitPosition * (1 - throttleValue);
+                return CombinedControlSplitPosition * (1 - (throttleValue ?? 0));
             else if (CruiseControl.SpeedRegMode == CruiseControl.SpeedRegulatorMode.Manual)
-                return CombinedControlSplitPosition * (1 - throttleValue);
+                return CombinedControlSplitPosition * (1 - (throttleValue ?? 0));
             else if (CruiseControl.UseThrottleAsSpeedSelector)
                 return CombinedControlSplitPosition * (1 - (CruiseControl.SelectedSpeedMpS / MaxSpeedMpS));
             else if (CruiseControl.UseThrottleAsForceSelector && CruiseControl.UseThrottleInCombinedControl)
@@ -4726,6 +4733,8 @@ namespace Orts.Simulation.RollingStocks
             // Only allow increasing dynamic brake if dynamic braking is off, there's no setup lock, or dynamic braking is already active
             if (DynamicBrake || !DynamicBrakeControllerSetupLock || DynamicBrakeController.CurrentValue <= 0)
             {
+                float prevValue = DynamicBrakeController.CurrentValue;
+
                 DynamicBrakeController.StartIncrease(target);
 
                 AlerterReset(TCSEvent.DynamicBrakeChanged);
@@ -4734,6 +4743,12 @@ namespace Orts.Simulation.RollingStocks
                 {
                     StopDynamicBrakeIncrease();
                     Simulator.Confirmer.ConfirmWithPerCent(CabControl.DynamicBrake, DynamicBrakeController.CurrentValue * 100);
+                }
+                // If this is the first application of dynamics, pause movement to prevent accidental over-application
+                else if (DynamicBrakeController.CurrentValue > 0 && prevValue <= 0)
+                {
+                    DynamicBrakePause = true;
+                    StopDynamicBrakeIncrease();
                 }
             }
         }
@@ -4862,7 +4877,8 @@ namespace Orts.Simulation.RollingStocks
 
         public bool CheckDisableDynamicBrake()
         {
-            if (DynamicBrakeController.CurrentValue <= 0)
+            // Only disable the dynamic brake if the lever is in the off position and the controller isn't trying to increase
+            if (DynamicBrakeController.CurrentValue <= 0 && DynamicBrakeController.UpdateValue <= 0)
             {
                 StopDynamicBrakeDecrease();
                 Simulator.Confirmer.Confirm(CabControl.DynamicBrake, CabSetting.Off);
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -4731,7 +4731,7 @@ namespace Orts.Simulation.RollingStocks
                 return;
 
             // Only allow increasing dynamic brake if dynamic braking is off, there's no setup lock, or dynamic braking is already active
-            if (DynamicBrake || !DynamicBrakeControllerSetupLock || DynamicBrakeController.CurrentValue <= 0)
+            if (DynamicBrakeController != null && (DynamicBrake || !DynamicBrakeControllerSetupLock || DynamicBrakeController.CurrentValue <= 0))
             {
                 float prevValue = DynamicBrakeController.CurrentValue;
 
@@ -4767,7 +4767,7 @@ namespace Orts.Simulation.RollingStocks
         {
             // Only allow decreasing dynamic brake if dynamic brake isn't off
             // Note: For safety, this will allow decreasing the dynamic brake even if a mechanical interlock should have locked dynamic brake in off position
-            if (DynamicBrakeController.CurrentValue > 0)
+            if (DynamicBrakeController?.CurrentValue > 0)
             {
                 DynamicBrakeController.StartDecrease(target);
 
@@ -4878,7 +4878,7 @@ namespace Orts.Simulation.RollingStocks
         public bool CheckDisableDynamicBrake()
         {
             // Only disable the dynamic brake if the lever is in the off position and the controller isn't trying to increase
-            if (DynamicBrakeController.CurrentValue <= 0 && DynamicBrakeController.UpdateValue <= 0)
+            if (DynamicBrakeController?.CurrentValue <= 0 && DynamicBrakeController?.UpdateValue <= 0)
             {
                 StopDynamicBrakeDecrease();
                 Simulator.Confirmer.Confirm(CabControl.DynamicBrake, CabSetting.Off);
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -478,6 +478,9 @@ namespace Orts.Simulation.RollingStocks
         public float ThrottleIntervention = -1;
         public float DynamicBrakeIntervention = -1;
 
+        // Has dynamic brake movement been paused to give user a chance to react?
+        private bool DynamicBrakePause = false;
+
         public enum TractionMotorTypes
         {
             DC,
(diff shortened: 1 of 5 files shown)


cesarBLG:
Separating driver controllers from the real setting transmitted to the engine was a mess, I hope my refactoring helped to fix that. However, this code block still looks quite buggy to me.

I wonder if we could use Locomotive.DynamicBrakeController.CurrentValue instead of LocalDynamicBrakePercent. It looks more sensible to me, at least for CABViewControlTypes.DYNAMIC_BRAKE (not sure about the DYNAMIC_BRAKE_DISPLAY, and how it should act with respect to blending).

Also I think we can remove all references to CruiseControl here. Previously, it worked by physically moving the DynamicBrakeController handle, but now it acts separately, so there is no need to "hide" the real value when CC is active.

It's just a suggestion, but obviously that can be done later in another PR.

cesarBLG:
The "setup" handling for the dynamic brake controller is horrible. Maybe using float.Epsilon for "setup" would work, I don't know how that will behave with respect to old stock.

However, I don't see the problem here, as it seems like -1 and 0 values are both mapped to index = 0 (which indeed may be a bug, because the GP38-2 uses index=1 as a setup notch). The problem is that it's not all dynamic brake controllers have a setup notch

SteelFill:
> The "setup" handling for the dynamic brake controller is horrible. Maybe using float.Epsilon for "setup" would work, I don't know how that will behave with respect to old stock.
> 
> However, I don't see the problem here, as it seems like -1 and 0 values are both mapped to index = 0 (which indeed may be a bug, because the GP38-2 uses index=1 as a setup notch). The problem is that it's not all dynamic brake controllers have a setup notch

Didn't have the time to finish yesterday (combined controls are a mess), but I'm making progress refactoring the dynamic brake controller so that 0% is off, and anything greater than 0% triggers set-up, completely removing the ability to have the superposition where the dynamic brake lever hasn't been moved from 0% but the dynamic brake is somehow enabled.
The previous implementation has left me quite confused as it seemed to go out of its way to ensure that the first application of dynamic brakes would begin set-up, but also discard any movement of the brake handle (allowing for the quantum state where the dynamic brake handle is in the off position but the dynamic brakes are on). Similarly, it was specifically coded to only turn off the dynamic brake if the handle was already at 0% and you tried to decrease further than 0%. That's not intuitive at all, and it doesn't seem to be replicating what MSTS did either. Real mystery why it's like that.

Will have to see how people respond once I've sorted it out, as this does change train handling a bit. Since the dynamic brake handle will need to be above 0% to set up, it will no longer be possible to have the dynamic brakes enabled and producing 0 force (the force will still be quite low though). Technically this is more realistic as most things will produce a small amount of braking even in the set-up position, and anything without a set-up position will still work as intended, producing a small amount of force in the first non-zero position.

That'll also make MSTS locos and cab views behave more like the original (no more dynamic brake lever saying 'off' while you set up the dynamic brakes, like in the GP38), since MSTS didn't allow this 'dynamic brakes are on but off' state to happen. But for Open Rails content, it's possible someone has taken advantage of the dubious implementation...though I'm not sure what that would even look like.

cesarBLG:
> allowing for the quantum state where the dynamic brake handle is in the off position but the dynamic brakes are on

Yes, probably it doesn't make sense. I looked at some manuals and set-up position already applies a little force as you say, so it's basically a notch with a small % of braking. The motivation for the old behaviour was to have a position where brakes are ready to be applied, but with no actual braking force applied.

I'm not sure about whether the timing to engage the brakes starts when the lever is at set-up position, or when the throttle is at idle. I remember reading somewhere that one (probably electric) locomotive prepared the motors for braking as soon as the throttle was set to 0, but this is probably not a standard thing.

With your proposal, DynamicBrakeIntervention is no longer needed, since the only remaining function of that is to distinguish between set-up and off, and basically dynamic brake is active when Controller.CurrentValue > 0, so this is simplified.

SteelFill:
> I'm not sure about whether the timing to engage the brakes starts when the lever is at set-up position, or when the throttle is at idle. I remember reading somewhere that one (probably electric) locomotive prepared the motors for braking as soon as the throttle was set to 0, but this is probably not a standard thing.

So, the answer is a bit of both. Many locomotives will have a delay between moving the dynamic brake to set-up and actually getting effort because they remain in power mode while idling and need time to change the circuitry for braking. But there could also simultaneously be extra systems to add a delay after going to idle, not because the wiring needs to change but because it won't immediately be safe to change to dynamic brake mode (residual generator excitation, wrong engine RPM, anything that could cause the wrong dynamic brake effort) but the simulation right now isn't detailed enough for that to matter.
OR really only simulates that first type of delay with any success, and realistically that should only be a couple of seconds at most. Yet still, you see people set it to 10 seconds because "the rules say you need to be in idle for 10 seconds before moving the dynamic brake to set up!" which is entirely the wrong approach because MSTS/OR can only add that 10 second delay _after_ set-up, not before! (And in reality, there usually is no timed delay-the locomotive just prevents changing to dynamic brake mode until all the conditions are met, regardless of how long that actually takes.)

I don't think there's a reason to add this idle-delay until the traction simulation is more sophisticated to the point that waiting in idle actually matters.

> With your proposal, DynamicBrakeIntervention is no longer needed, since the only remaining function of that is to distinguish between set-up and off, and basically dynamic brake is active when Controller.CurrentValue > 0, so this is simplified.

I'll have to double check but I think DynamicBrakeIntervention technically still has a use case because it remains at 0 until the dynamic brakes are set up. That said, there are other ways to handle that case which would involve less juggling of different variables that all mean the same thing.

SteelFill:
Ok, reworked dynamic brake controller is in. I was able to remove DynamicBrakeIntervention completely, it was entirely redundant now that the dynamic brake state could be easily determined from the dynamic brake handle setting. Some rework of combined controllers was required to properly change from powering to braking while accounting for the revised dynamic brake behavior. I also found a subtle disadvantage of the new setup; notchless dynamic brakes could be disabled by just holding down the dynamic brake release key, so I added a feature to pause the dynamic brake before it fully releases, forcing the player to release and press the button again to fully disable braking (ie: the same thing you had to do in previous builds).

Unfortunately, the changes are now conflicting with #922 due to a tiny change in MSTSLocomotiveViewer.cs. I am going to trust Carlo on that one and let his change take priority over mine. I think the results will be identical.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2134,29 +2134,15 @@ public virtual int GetDrawIndex()
...
-                                {
-                                    index = 0;
-                                }
-                                else
-                                    index = PercentToIndex(dynBrakePercent);
-                            }
-                            else
-                                index = PercentToIndex(dynBrakePercent);
-                        }
+                            index = PercentToIndex(dynBrakePercent);
```
cesarBLG, line 2145:
```suggestion
                            index = PercentToIndex(Locomotive.DynamicBrakeController.CurrentValue);
```

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2105,10 +2106,10 @@ public override void Update(float elapsedClockSeconds)
                     if (DynamicBrakeCommandStartTime + DynamicBrakeDelayS < Simulator.ClockTime)
                     {
                         DynamicBrake = true; // Engage
-                        if (IsLeadLocomotive() && DynamicBrakeIntervention >= 0)
+                        if (IsLeadLocomotive() && DynamicBrakeController.CurrentValue > 0)
```
cesarBLG, line 2109: I'm getting a null reference exception in trains with blended braking but no physical controller. You have to check if DynamicBrakeController is null

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -478,6 +478,9 @@ public float OdometerM
         public float ThrottleIntervention = -1;
         public float DynamicBrakeIntervention = -1;
 
+        // Has dynamic brake movement been paused to give user a chance to react?
+        private bool DynamicBrakePause = false;
```
cesarBLG, line 482: I wonder whether this is really needed. With this we get two notches to switch from braking to throttle in smooth combined controls. I would only expect a pause in throttle-0 and brake-0 (so two keypresses to switch from full throttle to full braking and vice-versa).

cesarBLG (changes requested):
Unfortunately I'm getting some crashes. Also, I cannot activate dynamic braking in a combined control.

Mouse control is also not working properly (throttle is reduced to ~2% and stays there until I click again), also cannot operate dynamic braking.

If I press the A key, the confirmer stays at Dynamic Brake: off and then I can't apply brakes, however in that situation the mouse starts working. So the only way to brake is to first use keyboard, then mouse

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: I tested with MSTS's HPP-8

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Well, this behavior is there to mitigate frustration and make the dynamic brakes behave the same as they did in versions from a few days ago. In both cases, you need to hold D until dynamic brakes reach 0, hit D again to disable the dynamic brakes, then hit D a second time to go into the throttle range. The first hit to D to disable the dynamic brakes is vital in making sure nobody accidentally un-sets-up the dynamic brakes and I prefer the second hit as well to make sure I don't start throttling on accident.

Furthermore, I realized I failed to replicate a similar behavior when moving into the dynamic braking range-on those older builds, a double press is required to engage dynamic brakes as well (one press to start set-up, a second press to actually increase the dynamic brake value) but now it's just a single press.

Maybe this can be disabled if there's no set-up notch (with a check for the number of notches on the dynamic brake controller) and/or disabled if the set-up time is small or 0 (no reason to worry about un-setting the dynamic brake if it can be restored immediately). But such a thing is another deviation from the previous standard that players could notice.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
SteelFill: This whole area was a bit of a mess when it came to null checks on DynamicBrakeController. Cleaned up in the next version.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Fixed now, anything that runs into a dynamic brake controller when it isn't assumed to be present is nullable/null-coalescing/null checked. I also found some situations where combined control was checking for dynamic brakes even when not present (like the MSTS series 7000) and those are better now.

SteelFill:
Ok, another build. Transition between power and braking was getting stuck with smooth controllers due to 'noise' in the system not setting controller values all the way to 0 and the fact that notchless dynamic brakes start at _exactly_ 0 when enabling, which was indistinguishable from the dynamic brake being off. Now the dynamic brakes won't turn off at 0 if a command is being sent to increase the dynamics.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -4738,29 +4765,23 @@ public void StopDynamicBrakeIncrease()
...
+            // Note: For safety, this will allow decreasing the dynamic brake even if a mechanical interlock should have locked dynamic brake in off position
+            if (DynamicBrakeController.CurrentValue > 0)
             {
-                Simulator.Confirmer.Warning(CabControl.DynamicBrake, CabSetting.Warn1);
-                return;
-            }
-            AlerterReset(TCSEvent.DynamicBrakeChanged);
-            if (!CanUseDynamicBrake())
-                return;
+                DynamicBrakeController.StartDecrease(target);
```
cesarBLG, line 4772: Getting a crash if I press the "," key in HHP-8 (no dynamic brake controller

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
cesarBLG: My point was something else. I think the current behaviour is not doing what it should for smooth controllers. The right code should be:
```csharp
if (DynamicBrakeController != null)
{
    if (HasSmoothStruct) index = PercentToIndex(DynamicBrakeController.CurrentValue);
    else index = DynamicBrakeController.CurrentNotch;
}
else
{
    index = PercentToIndex(dynBrakePercent);
}
```

That is, we should use the controller also for smooth controllers. Otherwise, if blending is applied the lever will show the blending value, which is incorrect.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Didn't even consider testing that. Fixed in next build.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
SteelFill: Seems like blended braking doesn't affect the LocalDynamicBrakePercent, so that's accounted for...maybe not foolproof though. However, I did notice combined controllers were worse and would show the first frame of dynamic braking during blended braking. Mitigated that as well.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2174,8 +2155,8 @@ public virtual int GetDrawIndex()
                         index = PercentToIndex(Locomotive.GetCombinedHandleValue(false));
                     break;
                 case CABViewControlTypes.CP_HANDLE:
-                    if (Locomotive.CombinedControlType == MSTSLocomotive.CombinedControl.ThrottleDynamic && Locomotive.DynamicBrakePercent >= 0
-                            || Locomotive.CombinedControlType == MSTSLocomotive.CombinedControl.ThrottleAir && Locomotive.TrainBrakeController.CurrentValue > 0)
+                    if (Locomotive.CombinedControlType == MSTSLocomotive.CombinedControl.ThrottleDynamic && Locomotive.DynamicBrakeController?.CurrentValue > 0
+                            || Locomotive.CombinedControlType == MSTSLocomotive.CombinedControl.ThrottleAir && Locomotive.TrainBrakeController?.CurrentValue > 0)
                             index = PercentToIndex(Locomotive.GetCombinedHandleValue(false));
                         else
                             index = PercentToIndex(Locomotive.GetCombinedHandleValue(false));
```
cesarBLG, line 2162:
```suggestion
                    index = PercentToIndex(Locomotive.GetCombinedHandleValue(false));
```
Probably we should remove the if clause. The body is the same for both branches.

SteelFill: Weird. This section is bothering me the more I look at it. Looks like this might have been set up to handle the case where dynamic brakes were activated but at 0%, but that's entirely pointless.

Then there's the whole `index = (int)((ControlDiscrete.FramesCount) * Locomotive.GetCombinedHandleValue(false))` case. CJ's comment justifying this implementation talks about 11.11% instead of 12.5%, but I don't think that's the problem here (or, maybe that was a problem but the changes to dynamic brakes in this PR fixed it). I found that if I unify all cases to use `index = PercentToIndex(Locomotive.GetCombinedHandleValue(false))` the _only_ problem with the MSTS Dash 9 is that dynamic braking notch 3 shows "B2" on the combined power display. And this is sort of a bug in the Dash 9 cab view, as the `SwitchVal` for the B3 frame is set to 0.73, when it should be 0.72222 (the actual combined control value when in B3).

Of course, it worked fine in MSTS, which implies `PercentToIndex` is not functioning the same as the equivalent code in MSTS. Looking at some MSTS videos, I have to guess that (in braking at least), MSTS would look for the index associated with the smallest value _greater_ than the percent, while the ORTS implementation is specifically looking for the largest value _below_ the given percent regardless of power or braking. Neither way of doing things is all that great, I'd like to try just matching for the closest index, higher or lower, and seeing what happens.

I'm going to unify the CPH cab view calculations and give reworking PercentToIndex a shot. It's a bit of a bold move, but I don't think it's too risky as the difference will be minor and difficult to notice. Like, the current implementation would show "B3" on the Dash 9 for an input of 0.73 to 0.77775, but a "closest index" implementation would show "B3" between 0.698 to 0.754, sliding back the range by just a few percent (and including 0.72222 as intended).

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2835,9 +2810,29 @@ protected int PercentToIndex(float percent)
...
-                    var val = ControlDiscrete.Values[0] <= ControlDiscrete.Values[ControlDiscrete.Values.Count - 1] ?
-                        ControlDiscrete.Values.Where(v => (float)v <= percent + 0.00001).Last() : ControlDiscrete.Values.Where(v => (float)v <= percent + 0.00001).First();
-                    index = ControlDiscrete.Values.IndexOf(val);
+                    // Binary search process to find the control value closest to percent
+                    List<double> vals = ControlDiscrete.Values;
+                    // Check if control values were defined in reverse, reverse them back for this calculation
+                    // This is less efficient, so creators should be encouraged to not do this
+                    bool reversed = ControlDiscrete.Values[0] > ControlDiscrete.Values[ControlDiscrete.Values.Count - 1];
+                    if (reversed)
+                        vals.Reverse();
```
cesarBLG, line 2819: This is reversing ControlDiscrete.Values, is this intentional? The problem is that cloning the list for this function is not good, since that would trigger unnecessary garbage collection. I think the best thing would be to reverse the list just once, maybe adding a Reversed variable to CVCWithFrames

SteelFill: I haven't noticed anything in my collection with reversed values but the previous (linear search) code supported it-still unsure this is needed at all. I'll look for a good place during initialization of the cab view control to see if it's sorted in the right direction.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
cesarBLG: Someone reported that indicators stopped working after the refactoring
```
CombinedControl (
Type ( CP_HANDLE COMBINED_CONTROL )
Position ( 426.875 397.03125 176.25 82.96875 )
Graphic ( "..\\..\\1CABVIEW\\S100\\CD_CP.ace" )
Style ( NONE )
MouseControl ( 1 )
NumFrames ( 3 3 1 )
NumPositions ( 3 0 1 2 )
NumValues ( 3 1 0.5 0 )
Orientation ( 1 )
DirIncrease ( 0 )
ScaleRange ( 0 1 )
)
Lever (
Type ( DYNAMIC_BRAKE LEVER )
Position ( 426.875 397.03125 176.25 82.96875 )
Graphic ( "..\\..\\1CABVIEW\\S100\\CD_dec.ACE" )
Style ( SPRUNG )
MouseControl ( 0 )
NumFrames ( 2 2 1 )
NumPositions ( 2 0 1  )
NumValues ( 2 0 0.01 )
Orientation ( 1 )
DirIncrease ( 1 )
ScaleRange ( 0 0.01 )
)
CombinedControl (
Type ( CP_HANDLE COMBINED_CONTROL )
Position ( 262.5 459.375 138.125 15.46875 )
Graphic ( "..\\..\\1CABVIEW\\S100\\flecha_intensidadPuff.ace" )
Style ( NONE )
MouseControl ( 1 )
NumFrames ( 54 2 27 )
NumPositions ( 0 )
NumValues ( 0 )
Orientation ( 1 )
DirIncrease ( 0 )
ScaleRange ( 0 1 )
)
CombinedControl (
Type ( CP_HANDLE COMBINED_CONTROL )
Position ( 169.375 265.78125 10 11.25 )
Graphic ( "..\\..\\1CABVIEW\\S100\\ind_inv.ace" )
Style ( NONE )
MouseControl ( 1 )
NumFrames ( 3 3 1 )
NumPositions ( 3 0 1 2 )
NumValues ( 3 1 0.5 0 )
Orientation ( 1 )
DirIncrease ( 0 )
ScaleRange ( 0 1 )
)
Lever (
Type ( DYNAMIC_BRAKE LEVER )
Position ( 169.375 265.78125 10 11.25 )
Graphic ( "..\\..\\1CABVIEW\\S100\\ind_MP2.ACE" )
Style ( SPRUNG )
MouseControl ( 0 )
NumFrames ( 2 2 1 )
NumPositions ( 2 0 1  )
NumValues ( 2 0 0.01 )
Orientation ( 1 )
DirIncrease ( 1 )
ScaleRange ( 0 0.01 )
)
```

Some of them are just boolean on/off lights that indicate whether traction and/or dynamic braking is active.

Also, there is another indicator (the one using flecha_intensidadPuff.ace), which works properly for braking, but it shows non-zero throttle when throttle is at zero.

[Textures.zip](https://github.com/openrails/openrails/files/15388801/Textures.zip)

SteelFill: I'll test those out, I can see these controls are a bit creative with the use of NumPositions and NumValues compared to what I've been able to test so far.

SteelFill: Looks like my consideration to force combined control displays to not show the 'idle' frame when even the slightest amount of dynamic braking was applied should have also been applied when the slightest amount of traction was applied-now the 'idle' frame will only be shown if the combined control is _exactly_ at the split position; that fixes the on/off light.

Unfortunately, the power/braking indicator isn't quite set up ideally to show exactly 0 throttle at idle (you need an odd number of frames for a switch value of exactly 0.5 to be produced, 54 frames produces two values equally spaced [neglecting floating point error] above and below 0.5), so I have added a bit of a bodge that I think will result in the same animation frame selected for 0 throttle as in old versions.

cesarBLG: I've been told that the indicators are working now, thanks!

SteelFill: Good to hear, I realized that my bodge for the idle indication wouldn't work exactly as expected (it would calculate the animation frame of the split position in an MSTS-compatible way, but that wouldn't guarantee that index would actually be used at idle throttle), so it's dumb luck that's working. I know what I missed so I'll get that in later and continue working through a variety of content to look for other things to fix.
It's baffling how MSTS just handled every unusual thing people threw at it.

cesarBLG (approved):
No more regressions have been reported after a couple of weeks, so I'm approving

PR #924, opened 2024-03-07
Default Asset Improvements

SteelFill:
This PR is all about replacing the programmer 'art' included in the Open Rails 'Content' folder. Many of these assets simply aren't pretty, and the community has since produced higher-quality alternates. After Erick Cantu of NAVS offered some replacements of his own, I asked around and found community members willing to contribute their art to make Open Rails just a bit prettier.

[Discuss on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/37832-sky-dome-and-cloud-layer-testing/page__view__findpost__p__306340)
[Trello Card for snowflake texture](https://trello.com/c/PczCjKAD) (I know the user who suggested this, they will be quite pleased)

The changed files are as follows:

- Clouds01.png: New file offered by Erick Cantu. Improves look of procedural clouds, making them appear more natural and fluffy.
- LightGlow.png: Also from Erick Cantu. This light glow was taken from a photograph, so it's a much more accurate representation of a light flare than the previous one.
- SignalLightGlow.png: Asset from BaldoBi that gives a 'lens flare' effect to signal lights, as opposed to being a blob of color. Hard to notice, but looks more like real photos.
- SkyDome1.png: Also Erick Cantu's work. Coloring is a bit more natural than the previous version, based on Microsoft Flight Simulator.
- Snowflake.png: A mod from 2018 made by Paul Peterson, makes the snowflake look like a snowflake rather than a white blob.
- StarDome1.bmp, Starmap_N.png, and Starmap_S.png: More assets from BaldoBi, his starmaps are based on NASA data and come in at much higher resolution than the default stars. Makes nighttime far more beautiful than before.
- WindowScrollbar.png: A super minor change from me, Phillip, to fix a bug in the scrollbar rendering. I doubt many people noticed the glitch but I figured now would be a good time to fix it.

Also added an entry in news.rst.

PR #944, opened 2024-05-20
Colorless Light Texture Replacement

SteelFill:
Follow up to #924, the community has decided that it would be best if the default light texture has no built in color so that it can be used for any sort of lighting without any residual color interfering with the intended appearance.

This PR replaces Erick's light with the same texture but changed to black and white (sorry Erick).

Longer-term, the preferred solution will be for .eng files to define custom lighting textures for each light so content creators can decide which light appearance they prefer. However, this light will be quite adequate for improving the looks of legacy content.


cjakeman:
Hi Phillip,

> Longer-term, the preferred solution will be for .eng files to define custom lighting textures for each light so content creators can decide which light appearance they prefer. 

Yes, I think so too.

cjakeman (approved):
Hi Phillip,

Thanks for providing this amendment. And, of course, for this and the other improved assets.

I see no reason why this can't be approved and merged straight away, so that's what I'm doing.

Chris

PR #945, opened 2024-05-22
Display tractive force instead of compensated out force

cesarBLG:
Bug report: https://www.elvastower.com/forums/index.php?/topic/38063-independent-brake-x-throttle/page__pid__308884#entry308884

MotiveForceN represents the actual force transmitted to the rails, after the adhesion calculations. This variable has the frictional force (i.e. brake force) substracted, so that TrainCar physics can handle starting friction accurately (namely, the train does not start until brake force is overcome). Due to this substraction, MotiveForce is most of the time similar to the tractive force (traction+dynamic braking) provided by the engine. Thus, it has been used to display engine-related displays (ammeter, traction-braking, etc.). However, there is another variable (TractiveForceN) which is the actual force developed by the engine, so it is more accurate to use it.

This replaces MotiveForceN with TractiveForceN wherever the engine force is required, and not the rail force.


SteelFill:
Looks like the Load % shown on the debug of diesel locomotives is still using the rail force rather than the tractive force. The calculation for DieselEngine.OutputPowerW is still using motive force, check DieselEngine.cs line 1046.

peternewell:
> 
> MotiveForceN represents the actual force transmitted to the rails, after the adhesion calculations. This variable has the frictional force (i.e. brake force) substracted, so that TrainCar physics can handle starting friction accurately (namely, the train does not start until brake force is overcome). Due to this substraction, MotiveForce is most of the time similar to the tractive force (traction+dynamic braking) provided by the engine. Thus, it has been used to display engine-related displays (ammeter, traction-braking, etc.). However, there is another variable (TractiveForceN) which is the actual force developed by the engine, so it is more accurate to use it.
> 

Can you add some of this information as developer notes to the code so that it is clear in regard to the difference between the two parameters, and how they are used. 

Perhaps as a "summary" item so that when you hover over the parameter you can see the notes.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2263,7 +2263,7 @@ public override void Update(float elapsedClockSeconds)
             UpdateHornAndBell(elapsedClockSeconds);
 
             UpdateSoundVariables(elapsedClockSeconds);
-            PrevMotiveForceN = MotiveForceN;
+            PrevTractiveForceN = MotiveForceN;
```
SteelFill, line 2266: Is this supposed to be `PrevTractiveForceN = TractiveForceN`?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -1039,12 +1039,12 @@ public void Update(float elapsedClockSeconds)
...
             if (Locomotive.Direction == Direction.Reverse)
-                Locomotive.PrevMotiveForceN *= -1f;
+                Locomotive.PrevTractiveForceN *= -1f;
 
             if ((State == DieselEngineState.Running) && (Locomotive.ThrottlePercent > 0))
             {
-                var abstempMotiveForce = Math.Abs(Locomotive.PrevMotiveForceN);
-                OutputPowerW = ( abstempMotiveForce > 0 ? abstempMotiveForce * Locomotive.AbsSpeedMpS : 0) / Locomotive.DieselEngines.NumOfActiveEngines;
+                var abstempTractiveForce = Math.Abs(Locomotive.PrevTractiveForceN);
+                OutputPowerW = ( abstempTractiveForce > 0 ? abstempTractiveForce * Locomotive.AbsSpeedMpS : 0) / Locomotive.DieselEngines.NumOfActiveEngines;
```
SteelFill, line 1047: Also as a general note, any horsepower calculation done using TractiveForceN should use WheelSpeedMpS/AbsWheelSpeedMpS. As we are considering the force going into the wheel, rather than the force going into the rails, the relevant speed is the wheel speed instead of the rail speed.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: Sure! I don't know how I missed it

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
cesarBLG: You're right, I changed it also for the HUD

SteelFill (approved):
The changes look good and in my testing have provided much more accurate and consistent ammeter/effort meter indications. Haven't noticed any more user-facing elements showing motive force so I believe this is good to go.

PR #948, opened 2024-05-29
Fix Cycling Lights

SteelFill:
Most content creators don't use the `Cycle ( 1 )` option in their lights. From what I understand, the design intent of `Cycle ( 1 )` is that the light will step through all of the `States` in forward order, but upon reaching the final state, instead of looping back to the first state, the light will begin to step through the states in reverse. This can be useful to reduce the number of `State` blocks entered in.

However, upon trying to implement a light with `Cycle ( 1 )` myself I encountered a crash immediately. As far as I can tell, this crash was guaranteed to happen regardless of the lighting configuration as there's a sort of 'off by one' error where the code would try to access one state beyond what is actually present. I suppose this crash was never noticed as `Cycle ( 1 )` seems to go completely unused.

Reported as a bug in [this old post from 2016](https://www.elvastower.com/forums/index.php?/topic/28312-what-do-you-do-with-your-lights-parameters/page__p__199121#entry199121).

Simply reducing the state count by one counteracts the off by one error, fixing the crash and allowing cycling lights to be used as intended.


SteelFill:
@Csantucci Carlo, going to bug you about this one too since you've been keeping up with lighting bugs. Should be simple to test-just change a light with more than two states from Cycle ( 0 ) to Cycle ( 1 ). Should crash in older versions (probably going back to the initial implementation of lights) but should not crash in unstable.

Csantucci (approved):
It's OK to me.

SteelFill:
Sorry @Csantucci I did some more testing just to double check things...and realized that light cones were still crashing-I only tested on light glows before and didn't think to try light cones since I rarely give those more than one state. Fixing that dismissed the review.

Csantucci (approved):

PR #951, opened 2024-06-06
fix: Fix watchdog process state name

twpol:
Noticed this copy-paste error working on other things :)


Csantucci (approved):
Removing bugs from OR is really a never ending story!
Thanks

PR #952, opened 2024-06-06
Investigation - Pulsing graphics

twpol:
While a lot more has been figured out, and plenty more can be improved here, this is a really tiny change which means that - for content which generates them - we no longer have to scan through loads of null signals in the array. `foundSignals` tracks how many signals are actually allocated in the array.

This makes things better, but lots more to come.

Bug report: https://www.elvastower.com/forums/index.php?/topic/37592-lags-by-mg-with-route-24-gb/

### DIFF
--- a/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
+++ b/Source/Orts.Simulation/Simulation/Signalling/Signals.cs
@@ -567,10 +567,8 @@ namespace Orts.Simulation.Signalling
 
             if (foundSignals > 0)
             {
-                // loop through all signals
-```            // loop through all the signals, but only one batch of signals with every call to this method.
-                // update required part
-```             // update one batch of signals. Batch ends when time taken exceeds 1/20th of time for all signals.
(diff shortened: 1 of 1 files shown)


SteelFill:
This has definitely improved frame time consistency on heavily signalled routes. Curious that the array of SignalObjects has (sometimes) thousands of trailing null values. Of course it works to simply ignore them, but considering the code goes out of the way to ensure all null elements are moved to the end of the array, it sure seems like it would be easier to truncate all the null elements from the array to save on memory allocation and reduce confusion.

cjakeman:
Hi James,

Is it time to approve and merge this PR and start a new one for any follow-on optimisations?

Chris

cjakeman:
Hi James,

Happy to approve, but feel that the algorithm would be better for some comment - see suggestions.

Chris

cjakeman (approved):
Hi James,

Thanks for those changes.

Approved and merged.

Chris

PR #953, opened 2024-06-09
Fix Lights Crash on Corrupt Shapes

SteelFill:
Bug report: https://www.elvastower.com/forums/index.php?/topic/38103-crash-in-lightscs-code-with-unstable-release/

Changes in lights from PR #917 have resulted in lights being pickier about engine/wagon shape files. A corrupt shape file would previously have no impact on lighting specifically, but now lights need to know where Indvidual parts of the shape file are to allow for attaching lights to the various shape components. This may not be possible if a shape file is corrupted in a particular way.

To solve this, when a poseable shape (includes engines and wagons but also other animated shapes too) is generated, if the XNAMatrices requires by the lights code are missing, a warning is produced and a default matrix is used. This should prevent crashes in lights and help users find more broken shape files.


cjakeman (approved):
Hi Phillip,

Surprised to find this useful fix has been unapproved for so long.

Approved and merged.

Thanks,

Chris

PR #946, opened 2024-05-23
Advanced track sounds

peternewell:
Initial work to add curve squeal to route

Blueprint - https://blueprints.launchpad.net/or/+spec/add-curve-squeal-route

ET Thread - https://www.elvastower.com/forums/index.php?/topic/37783-curve-wheel-squeal/

PR #954, opened 2024-06-11
Multiple Track Profiles & Superelevation Rewrite

SteelFill:
As some have noticed, the current superelevation system is lacking. The superelevation used by the physics system depends only on track curvature which does not reflect real superelevation design. A completely different superelevation system is in play for track profiles, which also has no basis in real superelevation design. And then when superelevation is used on routes that have multiple track types, the appearance of superelevated track will inevitably fail to match the appearance of some of the tracks on the route. The choice of track profile can be changed to match the appearance of other tracks, but fundamentally falls short as there can only ever be one TrProfile file regardless of the number of different track types present in the route.

[Discussion started on TrainSim](https://www.trainsim.com/forums/forum/open-rails/open-rails-modeling-and-physics/2302880-what-s-wrong-with-my-wp3-track-profile)
[Continue discussing on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38114-multiple-track-profiles/)
[Blueprint For Superelevation Updates](https://blueprints.launchpad.net/or/+spec/unified-superelevation)
[Wishlist Item For Track Profiles](https://bugs.launchpad.net/or/+bug/1286424)
[Old Bug That's Now Fixed](https://www.elvastower.com/forums/index.php?/topic/28751-jerky-superelevation-effect/)

We now support multiple track profiles. You can now add more than one track profile file to the "TRACKPROFILES" folder inside a route. Keep in mind some rules:
- The track profile named "TrProfile" will always be the track profile used by any dynamic track sections. It isn't feasible to have different profiles for different sections of dynamic track as there is no way to have different types of dynamic track.
- The "TrProfile" profile is also the profile that will be used by default if something goes wrong and a different profile can't be selected.
- If there is no "TrProfile" file then the default Kuju profile will be used for dynamic track (even if other profiles are present!)
- All other track profile files must begin with "TrProfile" but you can put anything afterward to differentiate between different profiles.
- I have never seen anyone use .xml track profiles, but OR will prefer to load an .xml track profile over a .stf one of the same name (eg: if you have "TrProfile_ScaleRail.xml" and "TrProfile_ScaleRail.stf" OR will ignore the .stf file and load the .xml file instead)
- _I have been unable to test XML profiles as a "TrProfile.xsd" file is needed which I cannot find anywhere._ This may jeopardize future support for XML track profiles.
- By default, the way the track profile is selected is entirely automatic and is chosen by finding the track profile that uses textures most similar to the track shape file textures. This matching is not case sensitive, and not file type sensitive (so, a track shape using "TRACK.DDS" will successfully match to a track profile using "track.ace").
- For most intents and purposes, matching based on the textures should be good enough. For more specific cases, I have added some new parameters for the TrProfile file that can be used to only allow a track profile to be used for superelevation when the original shape file meets certain conditions. If these custom parameters are used, the behavior described above is disabled.
- **IncludedTextures ( "first, second, third" )**: Only track shapes that use a texture named first, second, or third (note that for all these parameters, case is ignored and file extensions are ignored, so do not specify .ace or .dds) will be allowed to use this track profile. Only one of the textures used by the shape needs to match one of the filters, but if multiple shape textures match then the profile is more likely to be selected over other profiles. If none of the shape textures match, the shape will not be allowed to use this track profile. All custom filters can use the wildcards * and ? to allow a single filter to match to multiple file names.
- **ExcludedTextures ( "fourth, fifth, sixth" )**: Only track shapes that do NOT use a texture named fourth, fifth, or sixth will be allowed to use this track profile. If even one shape texture name matches one of the filters set here, the shape will not be allowed to use this track profile.
- **IncludedShapes ( "first, second, third" )**: Only track shapes with a shape file named first, second, or third will be allowed to use this track profile. The shape name only needs to match one of the filters (this is not possible in this situation, but is possible if wildcards are used), but if the name matches multiple filters then the profile is more likely to be selected over other profiles. If the shape name matches none of the filters then the shape will not be allowed to use this track profile.
- **ExcludedShapes ( "fourth, fifth, sixth" )**: Only track shapes that do NOT use a shape file named fourth, fifth, or sixth will be allowed to use this track profile. If the shape name matches even one of the filters set here, the shape will not be allowed to use this track profile.

And this also includes a near complete rewrite of the entire superelevation system and a few related systems.
- The **visual** superelevation and superelevation used by **physics** are now (mostly) **unified**. The only times where the visual amount of superelevation will not directly indicate the superelevation used by physics are when superelevation is disabled in the settings (this only disables the visual element, has no effect on the physics of superelevation) and when driving over specific track sections that cannot have visual superelevation generated (namely, tunnels and junctions).
- Old superelevation calculations using curve radius only have been replaced with more **true to life calculations** considering track speed limit, cant deficiency, runoff rate (rate of change of superelevation), and curve radius. Note that the better calculations are not used on any route with `ORTSTrackSuperElevation` in the .TRK file, delete that parameter if you want the better system. (There may be some railroads for which ORTSTrackSuperElevation is more accurate, but for many modern railroads that is very inaccurate.)
- Superelevation can now be applied to **straight sections** of track. This allows some superelevation to build up before reaching a curve, improving smoothness and passenger comfort. (The ideal solution to this is curves with easements but that is far outside the scope of this change.)
- Superelevation can now be applied to arbitrary track sections; previously the system would give up if encountering a track shape composed of both straight and curved sections. Now this is supported, rendering each section of the shape as an independent dynamic track segment.
- Rewrote the generation of dynamic track to better handle sections that aren't flat, especially ones at highly unusual angles. There should be no more discontinuities between dynamically generated sections (unless the discontinuity exists at the TDB level)
- **New tokens have been added to the .TRK** file to control superelevation using the same design constraints as real superelevation design, such as cant deficiency limits (separate for passenger and freight trains), minimum and maximum limits on cant, limits on runoff slope (change in superelevation over unit length of track) and runoff rate (change in superelevation over one second), finite precision on the level of superelevation, and a minimum speed for superelevation (not a feature on real designs, but useful to prevent superelevation from being added in places like yards with low speed limits). See the additions to the manual for description on how to actually use those.
- **New tokens also added to track profiles** to allow selecting the **style of superelevation** (as in, is superelevation applied to the inside rail, outside rail, both rails, or not applied at all?). The default has been changed from superelevation on both rails to superelevation on the outside rail only. This should reduce instances of the inside rail dipping so low as to clip into the terrain.
- There are also some slight provisions in place for other game systems to set the superelevation of a track section, which would bypass the automatic calculations. In theory, this could be used to allow the route editor to set the value of superelevation, but there's much to consider before that.
- **Fixed some bugs** in superelevation that were previously covered up with workarounds. This also meant **removing those workarounds** (eg: superelevation is now processed on static consists, there's no longer artificial smoothing added to the superelevation motion) which has made superelevation behavior more responsive and predictable.
- Basically redid the **bogie positioning code**, so bogie movement is now entirely dynamic based on the locations where wheels hit the rail. This means bogies now follow the track much more accurately and smoothly, including pitch and roll motions (where previously bogies would only yaw around corners). This makes superelevation and gradient changes more convincing in particular.
- Reworked the **passenger comfort** code, removing many wasteful calculations and relocating many other calculations to more appropriate locations. Also, because the physics superelevation now has some level of smoothing (as does the visual system, the two are connected), added a temporary tolerance factor so that passengers are less likely to complain about comfort when superelevation is being eased in.
- By coincidence, fixed [this bug](https://www.elvastower.com/forums/index.php?/topic/38066-rail-vehicles-shrink-as-gradient-increases/) that's been annoying me forever.


SteelFill:
Looks like this is conflicting with some other PR. Setting the for-unstable label so some testing can go in on these fairly major changes.

cjakeman (commented):
Hi Phillip,

It's great to see these 2 distinct systems coming together. Well done.

Would you add a Blueprint or a Trello card for it, please? No detail needed as your PR provides loads already.

Thanks,

Chris

SteelFill:
> Hi Phillip,
> 
> It's great to see these 2 distinct systems coming together. Well done.
> 
> Would you add a Blueprint or a Trello card for it, please? No detail needed as your PR provides loads already.
> 
> Thanks,
> 
> Chris

Since this is more technical and less of a shiny new feature the average Timmy would ask for, I added a blueprint [here](https://blueprints.launchpad.net/or/+spec/unified-superelevation).

SteelFill:
@cjakeman I am going to do some testing of these features across some assorted routes, but I'm not starting development on any new features at the moment. Now would be a good time to review the PR and clear up conflicts (eg: with #946 and #968) while I catch up on other tasks.

I do anticipate returning to the topic of superelevation somewhat soon, but not for anything related to track profiles or 3D dynamic track generation, so nothing that directly affects the features of this PR. The current version works on its own (ignoring any potential bugs), so it's safe to separate future features into a different PR.

cjakeman (dismissed):
Hi Phillip,

Nothing here that I should grumble about - all looks good.

I'm pleased to see you using LINQ and I notice that you have simplified the existing code in a few places. 
Also I spotted some new checks added which will hopefully make the code more robust.
Also some of the old code has been given more useful comments and variable names.

James started to re-factor Train.cs, so I will advise him to bring that work into sync with this PR.

I see there are still multiple occurences of "super elevation" still in Physics.rst - not your doing - and it would be neat to have these changed to "superelevation".

Now I've scanned every change, I can see what a thorough-going project this has been, though I can't see how you could have split it into smaller pieces.

Approved with thanks,

Chris

SteelFill:
> I see there are still multiple occurences of "super elevation" still in Physics.rst - not your doing - and it would be neat to have these changed to "superelevation".

Didn't think to check that part of the manual for superelevation since I was more concerned about the consequences of superelevation on the track rather than on the train, yet that section seems to be a more comprehensive explanation of superelevation than anywhere else! That's the trouble with the old system, duplicating separate implementations of superelevation for every part of the program means duplication of documentation too.

Of course, adjusting that kills the approval. Feel free to re-approve whenever.

cjakeman (dismissed):
Hi Phillip,

Approved with thanks,

Chris

cjakeman:
Hi Phillip,

Can't merge this yet. It seems the Appveyor build fails with the error:
"C:\projects\openrails\Source\ORTS.sln" (Clean;Build target) (1) ->
"C:\projects\openrails\Source\Menu\Menu.csproj" (default target) (26:2) ->
(CoreCompile target) -> 
  C:\projects\openrails\Source\Menu\Options.cs(927,21): error CS0103: The name 'baseUrl' does not exist in the current context [C:\projects\openrails\Source\Menu\Menu.csproj]

Strange that the Unstable builds OK.

Are you able to fix that?

Thanks,

Chris

SteelFill:
Looks like someone else's PR changed the name of "baseUrl" to "BaseDocumentationUrl" and my branch wasn't showing that as I hadn't synchronized the fork in a little while. "Works on my machine" doesn't mean a whole lot when my machine is running with old data! Though that does beg the question of why Unstable would function at all since Unstable should include that change...

I suppose this is exactly why we run a build test. I'm going to wing it and change the variable name through the website rather than VS (and throw out the review again) but I won't be able to check that things still work after synching until later.

cjakeman (approved):
Hi Phillip,

Thanks for your quick response.

Chris

PR #955, opened 2024-06-14
Copy dynamic brake speeds from other vehicles

cesarBLG:
DynamicBrakeSpeed1MpS is used for blending


Csantucci (approved):
Thank you, a never ending story that of updating the code to copy the loco parameters!

PR #956, opened 2024-06-18
Map settings saved

sweiland-openrails:
Form map settings (state checkboxes, size, etc.) saved and restored at startup


Source/RunActivity/Viewer3D/Map/MapDataProvider.cs
```diff
@@ -253,40 +253,42 @@ public float GetUnusedYLocation(float startX, float wantY, string name)
                 return noFreeSlotFound;
 
             var positionY = desiredPositionY;
-            while (positionY >= 0 && positionY < F.alignedTextY.Length)
-            {
-                // If the line contains no text yet, put it there
-                if (F.alignedTextNum[positionY] == 0)
-                    return SaveLabelLocation(startX, endX, positionY);
+            if (F.alignedTextY != null) {
```
sweiland-openrails, line 256: This extra if had to be added, otherwise during startup this code failed on F.alignedTextY being null. Tried to do the setting of the defaults at other place during start of the Form, but no luck.

Looky1173 (dismissed):
Looks good to me and works at my end! Thank you Siebren for taking this on!

sweiland-openrails:
Still one issue to fix: map window has focus after startup. Would like to switch back focus to main window. But that's a hard one to fix...

sweiland-openrails:
@Looky1173 Could you have a last look at the changes I've made? When the Map is opened focus returns to the main window. The Map now also is populated with the correct resolution, even when OR is paused.

sweiland-openrails:
@Looky1173 I came up with the idea when starting OR from a save. Most of the times you have the map window opened with the options you want. Now with the focus on the main window you can start OR right away. More or less the same when opening the map window from OR. And if you want to change something in the map window, it gets the focus anyway. I see no disadvantages.

sweiland-openrails:
@Looky1173 Thanks. It's good to wait a while to see if anything pops up...

Looky1173 (approved):
I haven't heard of any issues related to these changes so I'm approving them for merging. Thank you @sweiland-openrails for your contribution!

sweiland-openrails:
@Looky1173 You're welcome. I was not that much. But I like to startup with settings saved. Just a finishing touch.

PR #959, opened 2024-06-23
Fix TrackViewer crash on big zoom value

Fred-si:
Hi,

My son found a little bug in the TrackViewer program. With a really big value of zoom, the program crashes due to an unhandled ArgumentException

After investigating, I have found that with a really big zoom value, the width becomes 0.0, and the result of pass zero for the value of nearPlaneDistance parameter of the  Matrix.CreatePerspectiveFieldOfView method.

This PR proposed to solve this issue by returning without updating the camera when the nearPlaneDistance computed value is less or equals to zero.

### DIFF
--- a/Source/Contrib/TrackViewer/Drawing/DrawTerrain.cs
+++ b/Source/Contrib/TrackViewer/Drawing/DrawTerrain.cs
@@ -486,8 +486,14 @@ namespace ORTS.TrackViewer.Drawing
             float camHeight = width / device.Viewport.AspectRatio / 2;
             Vector3 cameraPosition = cameraTarget;
             cameraPosition.Y = -camHeight;
+            float nearPlaneDistance = camHeight / 2;
+            if (nearPlaneDistance <= 0)
+            {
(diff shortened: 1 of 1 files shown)


cjakeman:
Hi Fred,

Thanks for investigating this and submitting your fix.

I've added the label "for-unstable", which means that it will now be included in the automatic builds of the Unstable Version.

Your code change looks very clean, but let's leave it in Unstable for 2 weeks or so to check that there are no unwanted consequences. After that, I will approve and merge into the Testing Version ("master" branch).

Best wishes,

Chris

Fred-si:
Hi Chris,

Thanks for the review. I've taken note of the validation process.

Best wishes.

SteelFill (approved):
Been in testing for some time with no complaints. I was unable to crash the trackviewer by zooming in too far so the fix looks good to me.
Feel free to merge to add this to the official releases, but I recommend double checking everything first in case you think of any last changes (will need to be re-reviewed if any changes are made but that's not a problem).

PR #960, opened 2024-06-25
Fix draw state name in scripts

cesarBLG:
See https://www.elvastower.com/forums/index.php?/topic/38055-seranas-tcs-scripts-not-working-in-ornymg-after-rev-140/


Csantucci (approved):
That's OK for me, and it has already been tested.

PR #961, opened 2024-06-28
Improvements for Power Supplies

cesarBLG:
Includes some features by @Sharpe49:
- Voltage, pantograph and power limitation selectors
- Support for INI parameter files
- Refactoring of power supplies
- Minor variable additions for scripts

Additionally:
- Power limitation for traction and dynamic braking
- Default implementations for selectors
- Generic buttons for power supply scripts
- Customizable minimum RPM as required by train supply
- Backup dynamic brake for neutral sections
- Minor fixes
- Variable additions for scripts

This is a huge PR, probably the best way to review it is commit by commit (there are also a lot though)

- Implements https://blueprints.launchpad.net/or/+spec/electric-locomotive-selectors
- Further fixes https://bugs.launchpad.net/or/+bug/2071495
- Supersedes https://trello.com/c/OnORVZof/237-trigger-engine-power-on-off-sound-streams-in-all-cars by introducing sound triggers related to the power supply of passenger cars
- Implements https://blueprints.launchpad.net/or/+spec/power-supply-scripting-enhancements


peternewell:
Whilst you are working on this module, could you also review the power requirements for the Unpowered Control Car?

Some power operations should be able to be controlled from it as well.

Thanks

cesarBLG:
Sure, I'll look at it

cjakeman:
Hi Cesar,

Would you be able to add a link to a Blueprint or a Trello card for this PR?

Also, is it ready to be included in the Unstable Version yet?

Thanks,

Chris

SteelFill:
Scanned through all the commits and it certainly looks like a very useful set of features for those using the power supply scripts. Don't use those myself so I am not in a position to judge the functionality, but it all at least looks to be consistent with the existing functionality.

cesarBLG:
Hi Chris,

I was on holidays, but I'll add a blueprint/card as soon as possible.

Only documentation is missing before I can mark it as 'ready for review'. It was also low priority for me because this PR conflicts with a previous one (I don't remember which), so it couldn't be merged at the time I created it.

I'm working on this very slowly, and my intention was not to create the PR until it was fully ready, but I created it since it contained a bugfix. Eventually I created a separate PR for the bug and kept this as a draft due to the merge conflict.

Sharpe49:
To answer about the scriptable selector, yes it makes sense, because sometimes, you have to code interlockings between selectors which are specific to a type of train.
For example, for the TGV, the voltage selector can only be turned if the pantograph is selected down by the driver.
On other trains, you may not even need to lower the pantograph to change voltage (for example, on the Citadis Dualis).

cesarBLG:
Hi Cédric, thanks for your comment.

My reason for removing the scripting features is that I thought that the interlocking could be implemented centrally at the PowerSupply script. That is, you intercept the related events there, and if conditions are not met you don't propagate the event to the related controller. The motivation was that more information is available from the PowerSupply class, so you can have a richer condition set for the interlocking.

However, I can revert [90f2542](https://github.com/openrails/openrails/pull/961/commits/90f2542bbeb3b813a694af34b5d7d092b6420976) if you prefer so, I don't have a strong opinion against scripting the controllers.

Sharpe49:
I don't like the idea to do the interlocking in the PowerSupply class. First, that would mean it will be done differently than the brake controller (a class for the controller, and other classes for the system). Secondly, it will pollute the code of the PowerSupply. It has already many things to do, let's not add things which can be separated in a class. We must not put everything in the same box, that would be a mess in the long run.

cesarBLG:
Hi @cjakeman,
I finally found some time to update the docs, I'm removing the draft condition. I registered a new blueprint https://blueprints.launchpad.net/or/+spec/power-supply-scripting-enhancements . This also covers some other feature requests that I indicated in the first message.

pzgulyas (approved):
It would be good to merge this, I couldn't spot any outstanding problems with the code. But I don't know what this conflict reported by appveyor is about. I think that must be false positive, just don't know how to make appveyor to try to rebuild again, because the link for the build doesn't appear.

cesarBLG:
Thanks Peter, I think the problem is that another recent PR has been merged causing a conflict. I'll solve it and ask for blueprint approval.

cesarBLG:
Hi @pzgulyas the blueprint is now approved. I solved the merge conflicts (which cancelled your review). Could you re-approve it? Thanks

pzgulyas (approved):
I cannot spot any problems with these changes.

cesarBLG:
Thanks!

PR #962, opened 2024-06-28
Fix pantographs on unpowered cars

cesarBLG:
Raise and lower pantograph events are not sent to pantographs if car doesn't have a power supply

https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__309746

https://bugs.launchpad.net/or/+bug/2071495


SteelFill (dismissed):
No response from the original bug poster, but the changes make sense and no subsequent problems have been reported. Looks good to merge.

Also, I'm pretty sure I caused this bug with one of the changes in #927, I recall wondering why events were being set to non-locomotives and it seems the community figured out why.

cesarBLG:
Thanks, Phillip. Actually I think the fix only works for the player train, since AI trains don't have a LeadLocomotive I believe. I have a fix for that, but that will be included in #961 instead, because porting that here requires new functions which are not yet included.

I had to reset the review because the CI was failing for some reason.

PR #963, opened 2024-07-06
Fix dynamic brake force indicator

cesarBLG:
DynamicBrakeForceN is always positive, checking sign is incorrect


SteelFill (approved):
New code is much simpler than before and my testing has shown no unusual behavior from dynamic brake indicators.

PR #964, opened 2024-07-07
fix: Warning CA1417: Do not use the 'OutAttribute' for string parameter

twpol:
Reference: https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1417


cjakeman (approved):
Approved and merged.

Thanks,

Chris

PR #966, opened 2024-07-07
fix: Warning CS0618: 'string.Copy(string)' is obsolete

twpol:
Reference: https://learn.microsoft.com/en-us/dotnet/api/system.string.copy?view=net-8.0#remarks


cjakeman (approved):
Approved and merged.

Thanks,

Chris

PR #967, opened 2024-07-08
Fix AI Lights After Reversing Points

SteelFill:
Bug fix for [this bug on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38120-ai-train-lights-dont-commute-from-front-to-rear-and-vice-versa-at-reversal-points/)

When a train trips a reverse point, the order of the train cars is reversed and every train car is flipped. For AI trains, this should result in the lights treating the train as if the front is now the rear and the rear is now the front (ie: any red tail lights move to the opposite end of the train). However, the check to determine which locomotive in the AI train is the lead locomotive also gets reversed, thus cancelling out the reversing point and causing the lights not to respond to the reversing point.

Previous versions of lights simply ignored the lead locomotive of AI trains so we are going back to that behavior to maintain consistency with old versions.


Csantucci (approved):
OK, now AI trains lights work as intended and switch frront to rear when AI trains reverse.

PR #969, opened 2024-07-20
Bugfix refocus to the main window when opening  the map window

sweiland-openrails:
When opening the map window focus is returned to the main OR game window. This can cause confusion when running in full screen mode. In that case the map window will be hidden behind the game window. Therefore this bug fix does not refocus when (part) of the map window is overlapping the game window.

### DIFF
--- a/Source/RunActivity/Viewer3D/Map/MapForm.cs
+++ b/Source/RunActivity/Viewer3D/Map/MapForm.cs
@@ -1375,10 +1375,10 @@ namespace Orts.Viewer3D.Debugging
             int gameY1 = GameForm.Size.Height;
 
             return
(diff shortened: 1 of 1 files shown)


SteelFill:
I noticed this issue and was quite frustrated by it, all good now though!

sweiland-openrails:
@SteelFill hi Philip, yes I noticed this when I had to reinstall Windows, losing the OR registry settings. After the reinstall the map pops up after the main window losing focus. I have an extra monitor where the map normally pops up. Switching focus again to the main window is good inn that case. regards, Siebren.

sweiland-openrails:
@SteelFill hi Phillip, do you have time to review these two fixes? The first one you know already. The second one is described in my elvastower forum topic: [https://www.elvastower.com/forums/index.php?/topic/38177-map-window-resolution/page__p__310314#entry310314](url)
thanks, Siebren

SteelFill (approved):
Fix works as intended and code makes sense

PR #971, opened 2024-07-23
fix: Restore assembly info which allows platform compatibility checks

twpol:
This was the only project with this set, and I cannot see any reason for it. Having this set also causes a load of incorrect "CA1416: This call site is reachable on all platforms" warnings in the .NET 6 build.


cjakeman (approved):
Hi James,

No problem with this. Good catch.
Approved with thanks.

Chris

PR #972, opened 2024-07-30
On Map window color changed switch or signal is not changed

sweiland-openrails:
When status of a switch or signal is changed color of that switch or signal is not changed to reflect the status (probably when train is not moving, not tested)

### DIFF
--- a/Source/RunActivity/Viewer3D/Map/MapForm.cs
+++ b/Source/RunActivity/Viewer3D/Map/MapForm.cs
@@ -1596,6 +1596,7 @@ namespace Orts.Viewer3D.Debugging
                     break;
             }
 
+            mapCanvas.Invalidate(); // Triggers a re-paint
             UnHandleItemPick();
         }
 
@@ -1640,6 +1641,7 @@ namespace Orts.Viewer3D.Debugging
                         break;
                 }
             }
+            mapCanvas.Invalidate(); // Triggers a re-paint
             UnHandleItemPick();
         }
 


Csantucci:
I've made three tests with U2024.07.2-1838, but I see colours of signals and switches regularly changing when their state changes.

sweiland-openrails:
Strange behavior, I see this problem on the Bernina, the Ruebelandbahn, but not on the Italy28V3. I saw this when I was checking the problem Werter described in form topic [https://www.elvastower.com/forums/index.php?/topic/38177-map-window-resolution/page__p__310314#entry310314](url).

I tested this with latest U2024.07.27-1222. This version does not contain this fix. Looks like the build for unstable  [https://james-ross.co.uk/projects/or](url) is not running (holiday?).

sweiland-openrails:
See discussion on Elvastower: [https://www.elvastower.com/forums/index.php?/topic/38177-map-window-resolution/page__gopid__310474#entry310474](url)

cjakeman:
Hi Siebren,

As an approved developer, your PRs are automatically included in the Unstable Version, so they don't need an explicit label "for-unstable". Just a thought.

Best wishes,

sweiland-openrails:
hi Chris,

what do you think about this bugfix/enhancement? It's just two refreshes being added. The switch/signal can be changed and is changed: can be seen in the main OR window. However the map does not show this change.

regards, Siebren.

sweiland-openrails:
Hi Chris (@cjakeman ), what do you think about this fix? It's about the map not showing changes if switch/signal is changed by hand when OR is in pause mode. I use this often when on a complex layout, See also on Elvas Tower 
 [https://www.elvastower.com/forums/index.php?/topic/38177-map-window-resolution/page__gopid__310474#entry310474](url).

regards, Siebren

cjakeman (commented):
Hi Siebren,

Thanks for fixing this.

cjakeman (approved):

PR #973, opened 2024-07-31
fix: Using singular target framework to allow for different platforms across different projects

twpol:

This change allows for future changes where some projects are targeted at non-platform-specific TFMs, e.g. `net6`, instead of the current platform-specific TFMs, e.g. `net6-windows`

This seems to be something to do with using the plural target option forcing all referenced projects to build with the exact same TFM, which won't actually work if that TFM isn't also in the referenced project's list, due to nuget package data being tied to the TFM

This PR alone makes no change to .NET Framework builds nor .NET 5+ builds


cjakeman (approved):
Hi James,

Looks fine. Tricky problem to explain.
Approved with thanks.

Chris

PR #975, opened 2024-08-06
Bug fix for https://bugs.launchpad.net/or/+bug/2076133 Crash when AI train dissolves near camera

Csantucci:
See https://www.elvastower.com/forums/index.php?/topic/38196-crash-when-ai-train-dissolves/


SteelFill (approved):
I can see how that would cause a crash. Fix is simple and has been reported to work correctly.

PR #976, opened 2024-08-07
 Increased center of gravity range

albertosaurio65:
This range change is already implemented in the New Year version, but not in the other versions, which causes visual problems in some train compositions where the centre of gravity exceeds the +-1 range even if only by a little.


SteelFill (approved):
Change has been in testing for a couple of weeks and nobody reported anything unusual. I tried a CoG of -2 meters on some unusual articulated equipment in my collection and it worked correctly (the 'technically' correct CoG offset I need is -2.081 meters, so a limit of 2 is vastly closer to what I needed than the old limit of 1).

Looks good to merge in.

PR #977, opened 2024-08-18
adds cache for SD files

cjakeman:
SD files are short and simple, but the overhead of testing for them and reading them more than once is excessive. More than half of the file accesses at start are due to SD files for:
- 2176/4178  Demo Model 1 - Queens Street
- 1481/1844  BNSF Scenic Sub - Eastbound
- 2345/2873  Portugal 79 - Linha do Algarve

This PR adds a simple cache.

See [https://trello.com/c/8hcG5645/595-cache-sd-files-for-faster-loading](https://trello.com/c/8hcG5645/595-cache-sd-files-for-faster-loading)

### DIFF
--- a/Source/RunActivity/Viewer3D/Scenery.cs
+++ b/Source/RunActivity/Viewer3D/Scenery.cs
@@ -341,19 +341,23 @@ namespace Orts.Viewer3D
                     }
                 }
 
-                if (shapeFilePath != null && File.Exists(shapeFilePath + "d"))
+                var shapeDescriptorPath = shapeFilePath + "d";
+                if (shapeFilePath != null)
                 {
-                    var shape = new ShapeDescriptorFile(shapeFilePath + "d");
-                    if (shape.shape.ESD_Bounding_Box != null)
+                    var shape = new ShapeDescriptorFile(shapeDescriptorPath);
+                    if (shape != null)
                     {
-                        var min = shape.shape.ESD_Bounding_Box.Min;
-                        var max = shape.shape.ESD_Bounding_Box.Max;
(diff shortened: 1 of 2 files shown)


Source/RunActivity/Viewer3D/Scenery.cs
```diff
@@ -341,19 +341,23 @@ public WorldFile(Viewer viewer, int tileX, int tileZ, bool visible)
                     }
                 }
 
-                if (shapeFilePath != null && File.Exists(shapeFilePath + "d"))
+                var shapeDescriptorPath = shapeFilePath + "d";
+                if (shapeFilePath != null)
```
SteelFill, line 345: This is no longer checking if the SD file exists before loading it, which causes crashes for the many situations where someone's deleted SD files because they were told they weren't needed.

SteelFill (changes requested):
The change is causing unhandled crashes every time an SD file is missing, which was previously not a problem.

cjakeman:
Thanks, Phillip,

> The change is causing unhandled crashes every time an SD file is missing, which was previously not a problem.

I've marked it as not-for-unstable until I've fixed it.

Source/RunActivity/Viewer3D/Scenery.cs
cjakeman: I believe it's fixed now, so I've removed the label not-for-unstable.

SteelFill (approved):
Not causing trouble anymore, no adverse effects on the random routes I tried. Haven't benchmarked anything so I don't know how significant any performance changes are (big routes still take long to load, short routes don't) but it does appear to be a bit quicker loading now. Nevertheless, the process makes sense and caching the SD files is a good idea.

PR #978, opened 2024-08-19
fix: adds MilepostUnitsMiles to Manual

cjakeman:
also MilepostUnitsKilometers
See [https://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/page__view__findpost__p__310808](https://www.elvastower.com/forums/index.php?/topic/34208-missing-from-the-manual/page__view__findpost__p__310808)


SteelFill (approved):
It is worthwhile to document the occasional MSTS feature that OR still relies on, this one's good to know and clearly wasn't as well known as it should be. I've had to change this setting in the .trk file a few times myself.

PR #980, opened 2024-09-04
Downloading route content (Github, zip) second part 

sweiland-openrails:
Discussion: http://www.elvastower.com/forums/index.php?/topic/31918-downloading-content-for-open-rails
Roadmap: https://trello.com/c/1YXKK7N5/164-publishing-and-downloading-content


sweiland-openrails:
After adding code for Content tab "Manually Installed" still to be done:
- fix automatic route download, looks like it's hanging sometimes
- test the "Manually Installed" form with an empty registry/ini file
- translatability
- remove the original Content tab from Options
- documentation

cjakeman (approved):
Hi Siebren,

No issues in the last 2 weeks so I'm approving and merging.

Chris

Source/Menu/Options.cs
```diff
@@ -259,21 +233,6 @@ public OptionsForm(UserSettings settings, UpdateManager updateManager, bool init
...
-            if (initialContentSetup)
-            {
-                tabOptions.SelectedTab = tabPageContent;
-                buttonContentBrowse.Enabled = false; // Initial state because browsing a null path leads to an exception
-                try
-                {
-                    bindingSourceContent.Add(new ContentFolder() { Name = "Train Simulator", Path = MSTSPath.Base() });
-                }
-                catch { }
-            }
```
SteelFill, line 275: We certainly have an interesting community, apparently someone really liked that OR would attempt to find the MSTS installation on first set up and is very upset that this behavior was removed.

I guess this implies they are doing fresh installs regularly or something? Very odd, but it doesn't seem it was necessary to remove that behavior.

PR #981, opened 2024-09-09
Multiple type trainset lightglows

Csantucci:
Blueprint: https://blueprints.launchpad.net/or/+spec/multiple-type-lightglows
Discussion: https://www.elvastower.com/forums/index.php?/topic/38247-actual-lightglowpng-generates-too-small-lights-in-existing-trainsets/page__view__findpost__p__311139
Roadmap: https://trello.com/c/mhGYl3gu


Source/RunActivity/Viewer3D/Lights.cs
```diff
@@ -102,6 +103,8 @@ public LightViewer(Viewer viewer, TrainCar car, TrainCarViewer carViewer)
                     {
                         case LightType.Glow:
                             LightPrimitives.Add(new LightGlowPrimitive(this, Viewer.RenderProcess, light));
+                            if (light.Graphic != null)
+                                (LightPrimitives.Last() as LightGlowPrimitive).SpecificGlowMaterial = viewer.MaterialManager.Load("LightGlow", DefineFullTexturePath(light.Graphic));
```
SteelFill, line 107: Some other parts of the code could be simplified if `SpecificGlowMaterial` was set equal to the `LightGlowMaterial` if `light.Graphic` is null.

On line 235 and 237, a null check is performed every frame to select between the specific glow material or the general glow material, but that check could effectively be done one time on initialization by changing this.
```suggestion
                            if (light.Graphic != null)
                                (LightPrimitives.Last() as LightGlowPrimitive).SpecificGlowMaterial = viewer.MaterialManager.Load("LightGlow", DefineFullTexturePath(light.Graphic));
                            else
                                (LightPrimitives.Last() as LightGlowPrimitive).SpecificGlowMaterial = LightGlowMaterial;
```

Csantucci:
Thanks for the good suggestion, Phillip. I have now inserted it in the PR

Source/RunActivity/Viewer3D/Lights.cs
```diff
@@ -149,6 +154,15 @@ public LightViewer(Viewer viewer, TrainCar car, TrainCarViewer carViewer)
             UpdateActiveLightCone();
         }
 
+        string DefineFullTexturePath(string textureName)
+        {
+            if (File.Exists(Path.Combine(Path.GetDirectoryName(Car.WagFilePath), textureName)))
+                return Path.Combine(Path.GetDirectoryName(Car.WagFilePath), textureName);
+            if (File.Exists(Path.Combine(Viewer.ContentPath, textureName)))
+                return Path.Combine(Viewer.ContentPath, textureName);
+            return Path.Combine(Viewer.ContentPath, "LightGlow.png");
```
SteelFill, line 163: It may help if a warning message is added to the log when we reach this line, as it means the file defined couldn't be found. On my first attempt, I formatted the texture path slightly incorrectly and had to look at a debugger to see what exactly was wrong. But if I had an error message like "Could not find light graphic XYZ at 'wagon file path\\light path you entered'" it would have been much faster to see my mistake.

Source/Documentation/Manual/features-rollingstock.rst
```diff
@@ -318,6 +318,79 @@ for the purposes of calculation.
...
+may have different customized light glow textures.
+
+Detailed spec
+'''''''''''''
+
+
+1) In the ``Content`` folder there is the default ``LightGlow.png``, which is displayed if 
+   no changes are done to the .eng file.
+2) In such folder there is also an ``ORTSLightGlow.png``, which is maybe more realistic.
+3) adding a line within the .eng file it is possible to select either ORTSLightGlow.png or any other .png.
```
SteelFill, line 347: I tried using a .ace texture and that actually worked. However, _.dds does not work_. This seems to be a slight oversight in Materials.cs line 180, which is a backup case for loading .ace but was never updated to consider .dds support. We should probably support all the texture formats used in the MSOR ecosystem.

In any case, the manual should not imply that .png is the only allowed format. Currently, .png, .jpg, .gif, .bmp, and .ace are supported. That's pretty much every common type of image file (and the MSTS format, of course), so all file types should be listed out, or described shorthand "all standard image formats are supported".

Source/Documentation/Manual/features-rollingstock.rst
```diff
@@ -318,6 +318,79 @@ for the purposes of calculation.
...
+
+
+1) In the ``Content`` folder there is the default ``LightGlow.png``, which is displayed if 
+   no changes are done to the .eng file.
+2) In such folder there is also an ``ORTSLightGlow.png``, which is maybe more realistic.
+3) adding a line within the .eng file it is possible to select either ORTSLightGlow.png or any other .png. 
+   Here an example for the legacy Acela loco::
+
+    	Lights	( 17
+	        ORTSGraphic ( "ORTSLightGlow.png")
```
SteelFill, line 351: Multiple of these ORTSGraphic lines were typed with slightly inconsistent formatting that's bugging me, please add a space between the " and ) to match the space present between ( and "

SteelFill (changes requested):
Current implementation is functional, but I had some random observations in my comments and I really think that .dds should be added as a supported light texture format before this is merged since this goes against the rest of the program in that it supports .ace but not .dds.

I also think it would help (though certainly not required) to add a feature to allow a light to be rotated about its Z axis. We can rotate lights about their X and Y axes using Elevation and Azimuth (respectively), but not about Z (using a "Roll" parameter, I guess). Normally, this would not be very useful since most light textures were designed to be radially symmetrical...but with custom textures you might want to use something that's not symmetric (eg: a light glow showing light emanating downward only, rather than in all directions), in which case any rotation has to be baked into the texture...if the texture rotation needs to be changed (eg: making that beam of light point up instead of down) then an entire additional file is needed-that's wasteful.

Similarly, defining a light size by the 'Radius' parameter only makes sense for radial lights, but there are now situations where I'd like to have lights that aren't circular (eg: walkway lights on the older EMD locomotives), in which case it would help to define the light size based on "Width" and "Height", allowing for rectangular light primitives instead of the square primitives we are limited to right now.

I don't think either of those would be too outlandish to implement as it's just a matter of adding some additional manipulation to the vertex locations.

Csantucci:
> Current implementation is functional, but I had some random observations in my comments and I really think that .dds should be added as a supported light texture format before this is merged since this goes against the rest of the program in that it supports .ace but not .dds.
> 
> I also think it would help (though certainly not required) to add a feature to allow a light to be rotated about its Z axis. We can rotate lights about their X and Y axes using Elevation and Azimuth (respectively), but not about Z (using a "Roll" parameter, I guess). Normally, this would not be very useful since most light textures were designed to be radially symmetrical...but with custom textures you might want to use something that's not symmetric (eg: a light glow showing light emanating downward only, rather than in all directions), in which case any rotation has to be baked into the texture...if the texture rotation needs to be changed (eg: making that beam of light point up instead of down) then an entire additional file is needed-that's wasteful.
> 
> Similarly, defining a light size by the 'Radius' parameter only makes sense for radial lights, but there are now situations where I'd like to have lights that aren't circular (eg: walkway lights on the older EMD locomotives), in which case it would help to define the light size based on "Width" and "Height", allowing for rectangular light primitives instead of the square primitives we are limited to right now.
> 
> I don't think either of those would be too outlandish to implement as it's just a matter of adding some additional manipulation to the vertex locations.

Re .dds: this PR uses the default OR code method Texture2D Get(GraphicsDevice graphicsDevice, string path) which handles .gif, .jpg, .png and .bmp fomats, but not .dds, which returns an error, because Texture2D.FromStream seems unable to handle .dds. Such method is called in many places within the code. So I consider a modification of that code to add .dds support something that goes beyond the scope of this PR.

In my opinion also your other suggestions about rotation and rectangular lights go beyond the scope of this PR and could be inserted at a later time.

Source/RunActivity/Viewer3D/Lights.cs
Csantucci: I'll add something similar.

Source/Documentation/Manual/features-rollingstock.rst
Csantucci: OK, but it seems to me that OR uses the other typing, that is without parentheses. I'll modify the .rst file that way.

Csantucci:
> Current implementation is functional, but I had some random observations in my comments and I really think that .dds should be added as a supported light texture format before this is merged since this goes against the rest of the program in that it supports .ace but not .dds.
> 
> I also think it would help (though certainly not required) to add a feature to allow a light to be rotated about its Z axis. We can rotate lights about their X and Y axes using Elevation and Azimuth (respectively), but not about Z (using a "Roll" parameter, I guess). Normally, this would not be very useful since most light textures were designed to be radially symmetrical...but with custom textures you might want to use something that's not symmetric (eg: a light glow showing light emanating downward only, rather than in all directions), in which case any rotation has to be baked into the texture...if the texture rotation needs to be changed (eg: making that beam of light point up instead of down) then an entire additional file is needed-that's wasteful.
> 
> Similarly, defining a light size by the 'Radius' parameter only makes sense for radial lights, but there are now situations where I'd like to have lights that aren't circular (eg: walkway lights on the older EMD locomotives), in which case it would help to define the light size based on "Width" and "Height", allowing for rectangular light primitives instead of the square primitives we are limited to right now.
> 
> I don't think either of those would be too outlandish to implement as it's just a matter of adding some additional manipulation to the vertex locations.

Added support for .dds files. Added in OR manual file extensions supported.

Source/Documentation/Manual/features-rollingstock.rst
Csantucci: Done

Csantucci: I don't foresee to add light rotation and rectangular lights. This goes quite beyond the initial aim of this PR. Such additions could be object of a further PR.

SteelFill (approved):
Thanks Carlo, .dds works now and the log helps in finding any missing textures.

Other ideas can be considered later, this is a great feature to have already.

Csantucci:
> Thanks Carlo, .dds works now and the log helps in finding any missing textures.
> 
> Other ideas can be considered later, this is a great feature to have already.

Thanks Phillip, now only the approval of the blueprint is needed in order to merge.

cjakeman:
> Thanks Phillip, now only the approval of the blueprint is needed in order to merge.

Hi Carlo,

The blueprint was approved on 19-Oct-2024.

Merging your PR now.

Chris

Csantucci:
> > Thanks Phillip, now only the approval of the blueprint is needed in order to merge.
> 
> Hi Carlo,
> 
> The blueprint was approved on 19-Oct-2024.
> 
> Merging your PR now.
> 
> Chris

Sorry Chris, I oversaw it. Thanks for merging.

Carlo

PR #983, opened 2024-09-09
WEB based Switch Panel bug: DPU window

sweiland-openrails:
The DPU window opened from the WEB based Switch Panel enhancement does not cycle between it's variants

[https://bugs.launchpad.net/or/+bug/2080059](url)


mbm-OR (approved):
This bug has been fixed. 
Thank you.

PR #985, opened 2024-09-13
Correct bug with steam locomotive chuffing sound

peternewell:
Corrects a bug identified with no sound occurring for steam locomotive chuffing when advanced adhesion is not used by a train.

https://www.elvastower.com/forums/index.php?/topic/38284-no-chuffing-from-steam-locos/

https://bugs.launchpad.net/or/+bug/2080599

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
@@ -3505,14 +3505,7 @@ public readonly SmoothedData StackSteamVelocityMpS = new SmoothedData(2);
 
                 if (((Train.TrainType == Train.TRAINTYPE.PLAYER && !Train.Autopilot) || Train.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN) && (Simulator.UseAdvancedAdhesion && !Simulator.Settings.SimpleControlPhysics))
                 {
-                    if (SteamEngines[i].AttachedAxle.IsWheelSlip)
-                    {
-                        variable[i] = Math.Abs(SteamEngines[i].AttachedAxle.SlipSpeedMpS / SteamEngines[i].AttachedAxle.WheelRadiusM / MathHelper.Pi * 5);
-                    }
-                    else
-                    {
                         variable[i] = Math.Abs((float)SteamEngines[i].AttachedAxle.AxleSpeedMpS / SteamEngines[i].AttachedAxle.WheelRadiusM / MathHelper.Pi * 5);
-                    }
                 }
                 else 
                 // Axle code is not executed if it is an AI train, on Autopilot, or Simple adhesion or simple physics is selected. Hence must use wheelspeed in these instances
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -3503,13 +3503,26 @@ private void UpdateFX(float elapsedClockSeconds)
                 // Variable is proportional to angular speed, value of 10 means 1 rotation/second.
                 // If wheel is not slipping then use normal wheel speed, this reduces oscillations in variable1 which causes issues with sounds.
 
-                if (SteamEngines[i].AttachedAxle.IsWheelSlip)
+                if (((Train.TrainType == Train.TRAINTYPE.PLAYER && !Train.Autopilot) || Train.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN) && (Simulator.UseAdvancedAdhesion && !Simulator.Settings.SimpleControlPhysics))
```
cesarBLG, line 3506:
```suggestion
                if (AdvancedAdhesionModel)
```

Not sure if that works, but that would be a much simpler check

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -3503,13 +3503,26 @@ private void UpdateFX(float elapsedClockSeconds)
...
                 {
-                    variable[i] = Math.Abs(SteamEngines[i].AttachedAxle.SlipSpeedMpS / SteamEngines[i].AttachedAxle.WheelRadiusM / MathHelper.Pi * 5);
+                    if (SteamEngines[i].AttachedAxle.IsWheelSlip)
+                    {
+                        variable[i] = Math.Abs(SteamEngines[i].AttachedAxle.SlipSpeedMpS / SteamEngines[i].AttachedAxle.WheelRadiusM / MathHelper.Pi * 5);
+                    }
+                    else
+                    {
+                        variable[i] = Math.Abs((float)SteamEngines[i].AttachedAxle.AxleSpeedMpS / SteamEngines[i].AttachedAxle.WheelRadiusM / MathHelper.Pi * 5);
+                    }
```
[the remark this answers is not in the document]
peternewell: I agree, and have adopted your change.

Thanks for pointing it out.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: I am not comfortable with this change as I can't see where AdvancedAdhesionModel is set/rest by the Autopilot function. Thus I suspect that it could be possibly set to true whilst in Autopilot, which would not be desirable if Autopilot doesn't use the Advanced Adhesion(?).

cesarBLG: The idea for that variable was that whenever advanced adhesion is being used, this is set to true. For AI controlled trains or user-selected adhesion, it is set to false. The motivation was precisely to avoid all the checks for !SimpleControls&Settings.AdvancedAdhesion&&!AITrain that happen multiple times in the code, and use that variable instead. However, I'm not sure about whether that variable is correctly updated at all times, so let's keep it as is, as you suggest.

cesarBLG (approved):
Thanks for investigating and solving the bug!

peternewell:
Cesar, 

Thanks for approving this.

PR #986, opened 2024-09-30
Fix: The TrainCarOperations window does not resize correctly.

mbm-OR:
The TrainCarOperation window doesn't resize when the number of cars changes.

https://bugs.launchpad.net/or/+bug/2083270


Csantucci (approved):
That's OK for me

mbm-OR:
Thanks for the review.

PR #987, opened 2024-10-05
fix: Temporary workaround for building with NET 5+

twpol:

`SystemParameters` (from WPF I think) isn't available here in .NET 5+ and I suspect we can implement this better in modern .NET anyway, so for the moment replacing the code with a build warning - only affects .NET 5+ builds, no change to .NET Framework builds


cjakeman (approved):
Hi James,

All good.

Approved and merged. Thanks,

Chris

PR #993, opened 2024-10-14
Fix trainset windows documentation in OR manual

Csantucci:
See. https://www.elvastower.com/forums/index.php?/topic/37120-opening-and-closing-trainset-windows/page__view__findpost__p__306996


SteelFill (approved):
Documentation is now correct, the ORTS_ prefix is indeed required.

Csantucci:
> Documentation is now correct, the ORTS_ prefix is indeed required.

Thank you

PR #995, opened 2024-10-19
Fix: TrainCarOperations and TrainOperationsViewer windows overlap.

mbm-OR:
When the TrainCarOperations window is resized or moved, it may overlap the TrainOperationsViewer window.
The purpose of this is to ensure that the TrainOperationsViewer is always automatically placed around the TrainCarOperations window.

Bug [#2084957](https://bugs.launchpad.net/or/+bug/2084957)


sweiland-openrails (dismissed):
Code changes are ok.

However...  When moving around windows on the screen I like to be in control. My opinion is is that the Viewer window now is a bit jumpy. But if you want to keep it this way, no problem.

mbm-OR:
> My opinion is is that the Viewer window now is a bit jumpy.

The idea was to automatically keep the Viewer window around the TrainCarOperations (TCO) window.
Now it is a bit jumpy, if the TCO position is changed frequently. A way to avoid this is welcome. 

Thank you.

sweiland-openrails (approved):

PR #996, opened 2024-10-25
Replace dynamic brakes by engine brakes at low speeds

cesarBLG:
This is applicable to train brake controllers which also command dynamic braking application ('train blending'). At low speeds, dynamic brakes are unusable, so they are replaced by air brakes, but only in engines (the automatic braking, which also affects wagons, is handled independently).

Not to be confused with the 'local blending' where automatic air brakes are partially replaced by dynamic brakes.

https://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features/page__view__findpost__p__312406

https://blueprints.launchpad.net/or/+spec/braking-enhancement

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -1714,6 +1714,14 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
                                     }
                                 }
                             }
+                            if (loco.DynamicBrakeEngineBrakeReplacement && loco.RemoteControlGroup == 0 && loco.AbsTractionSpeedMpS < loco.DynamicBrakeSpeed2MpS && loco.Train.LeadLocomotive is MSTSLocomotive lead && lead.TrainBrakeController.TrainDynamicBrakeIntervention > 0)
+                            {
+                                var requiredBrakeForceN = loco.MaxDynamicBrakeForceN * lead.TrainBrakeController.TrainDynamicBrakeIntervention;
+                                var reverseBlendingPressurePSI = Math.Min(Math.Max((requiredBrakeForceN - loco.DynamicBrakeForceN) / Car.FrictionBrakeBlendingMaxForceN * ReferencePressurePSI
+                                + BrakeCylinderSpringPressurePSI, 0), MaxCylPressurePSI);
+                                reverseBlendingPressurePSI /= RelayValveRatio;
+                                if (demandedPressurePSI < reverseBlendingPressurePSI) demandedPressurePSI = reverseBlendingPressurePSI;
+                            }
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -1714,6 +1717,14 @@ public override void Update(float elapsedClockSeconds)
                                     }
                                 }
                             }
+                            if (loco.DynamicBrakeEngineBrakeReplacement && loco.RemoteControlGroup == 0 && loco.AbsTractionSpeedMpS < loco.DynamicBrakeSpeed2MpS && loco.Train.LeadLocomotive is MSTSLocomotive lead && lead.TrainBrakeController.TrainDynamicBrakeIntervention > 0)
```
SteelFill, line 1720: Feature works as coded, though I have some observations

- Relying on DynamicBrakeSpeed2MpS (`DynamicBrakesFadingSpeed`) should be documented as I could see a situation where someone hasn't set DynamicBrakesFadingSpeed (eg: because they used `ORTSDynamicBrakeForceCurves` in place of MSTS parameters) and then the local braking starts to apply at the wrong speed (30 km/h is the default, which won't be the correct speed for all cases). Even when using ORTSDynamicBrakeForceCurves, DynamicBrakesFadingSpeed can be used to change the speed this feature activates at without impacting dynamic brake performance.
- There is no way to control the limit on local brake pressure used, it'll always try to match the dynamic brake force up to the maximum cylinder pressure, which won't work for any units that might have a specifically lowered pressure on the holding brake.
- Because the _current_ value of TrainDynamicBrakeIntervention is used, any ORTSTrainDynamicBlendingTable set to give 0 dynamic braking at a complete stop will actually disable the holding brake as well. That's not entirely intuitive but can be worked around with different dynamic brake force curves and train dynamic blending settings.

cesarBLG: Thanks for your feedback!

- You're right. I think it's better to define a dedicated parameter instead of relying on DynamicBrakesFadingSpeed, I'll rewrite that. I found a different vehicle where dynamic brake force is always replaced by the air brake, regardless of speed, so it makes more sense to me to define a new token.

- Do you have an specific use case for that? In all the stock I found where this is implemented, the main goal is to provide a smooth transition between dynamic and air brakes, such that the total braking force is constant all the time, so there is no special limit on the brake pressure.

- This looks correct to me, the idea is that dynamic brake is replaced with air brakes, but if no dynamic brake is commanded then no engine brake is applied. If ORTSTrainDynamicBlendingTable is 0 at 0 speed, it is likely that dynamic brakes are replaced with air brakes in the whole train, not only in locomotives.

SteelFill (commented):
The new speed parameter works as intended. Only other thing I noticed is that the local brake pressure is applied incredibly quickly on locomotives that weren't built to use the relay valve feature, as there is no limit on the rate of pressure change unless a relay valve is fitted.
 I imagine the brake technology used here requires the use of a relay valve or similar advanced equipment, so perhaps the `RelayValveFitted` [initialization in AirSinglePipe.cs](https://github.com/openrails/openrails/blob/master/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs#L783) should have an additional OR condition added alongside the `loco.DynamicBrakeAutoBailOff || loco.DynamicBrakePartialBailOff` conditions to enable a relay valve if `loco.DynamicBrakeEngineBrakeReplacement` is true.

SteelFill (approved):
Looks good to go, should be a useful feature for many trains out there.

cesarBLG:
Thanks for the review!

PR #997, opened 2024-10-27
Disable independent brake syncing by default

cesarBLG:
Activating it by default breaks the behaviour of almost all UIC locomotives


SteelFill (approved):
This has been discussed and the removal of this default behavior should help in more cases than it may hurt.

cesarBLG:
Thanks for your understanding!

PR #1001, opened 2024-11-08
Telemetry collection and menu options #or-telemetry-system

twpol:
Roadmap: https://trello.com/c/Dj01ajw0/579-opt-in-anonymous-telemetry-for-installs
Discussion: https://www.elvastower.com/forums/index.php?/topic/37702-opt-in-anonymous-telemetry/


cjakeman:
Hi James,

I noticed that the (i) icon is there, but takes the user to the Manual page which is still to be updated. Guess that's already on your list somewhere.

Chris

twpol:
> I noticed that the (i) icon is there, but takes the user to the Manual page which is still to be updated. Guess that's already on your list somewhere.

Yeah, another reason this is still draft. Something needs to go there in the manual. :)

twpol:
> I noticed that the (i) icon is there, but takes the user to the Manual page which is still to be updated. Guess that's already on your list somewhere.

Manual updates are now included so no longer draft

cjakeman (approved):
I know that there is a problem with Notifications, which may be caused by this PR.
However, it's not a show-stopper, so this PR can be merged and we can start collecting useful data.

Approved and merged.

Thanks,

Chris

PR #1002, opened 2024-11-10
It locales for release 1.6

strawberryfield:
Updated it locales for version 1.6
I also updated the .pot files (a lot of strings missing)


cjakeman (approved):
Hi Roberto,

I am assuming that the Italian words are good. I've checked that they fit the spaces available for them and found only 5 issues with space in the F1 > Commands list.

I'll be making more space there for other languages, so it's not something for you to work on.

Many thanks for all your work.

Chris

PR #1004, opened 2024-11-24
Bugfix/fix train car operations brakehoses not sync coupling cars

mbm-OR:
Sometimes, when coupling cars, the brake hose of new car is not sync with the previous car.
More info [here](https://www.elvastower.com/forums/index.php?/topic/37062-proposal-f9-train-operations/page__view__findpost__p__312986).

Launchpad [bug](https://bugs.launchpad.net/or/+bug/2089491).


Source/Orts.Simulation/Simulation/Simulator.cs
sweiland-openrails, line 1: I think this fix is fixing the root cause of this problem. Uncoupling must set the decoupling of two brake hoses. Tested it and it looks like it's working.

sweiland-openrails (approved):

Source/Orts.Simulation/Simulation/Simulator.cs
mbm-OR: The cause of this bug is fixed by your fix. 
Thanks.

PR #1005, opened 2024-11-28
Bugfix: Content Form exception when there is manually installed content in a top level directory.Bugfix content in top dir

rwf-rr:
When manually installed content has a path of only a top level directory, eg. "H:\", then the Content form of the main Menu (Content button) will crash with an exception. The problem is in the code that attempts to find the most used content directory.

The fix is to ignore content that is installed in a directory without a parent directory.

The commit also includes fixing a simple typo in a variable name.

See also:
https://bugs.launchpad.net/or/+bug/2089801
https://www.elvastower.com/forums/index.php?/topic/38414-crash-with-content-manager/page__view__findpost__p__313090


rwf-rr:
Closing, as something went wrong with my commit history.

PR #1008, opened 2024-12-03
Fix: TrainCarOperations doesn't work as expected after resume.

mbm-OR:
The TrainCarOperations windows is not working as expected after resume.
More info here, [Bug #2090952](https://bugs.launchpad.net/or/+bug/2090952).


Csantucci (approved):
That's OK to me.

mbm-OR:
Thank you, Carlo.

PR #1009, opened 2024-12-03
Fix: TrainCarViewer overlaps when display size changed

mbm-OR:
The TrainCarViewer window has an overlap with the TrainCarOperations window when changing the screen size.
More info here, [Bug #2090960](https://bugs.launchpad.net/or/+bug/2090960).


sweiland-openrails (approved):
fix ok

PR #1010, opened 2024-12-03
Fix: TrainDrivingInfo window width not stable.

mbm-OR:
When using automatic mode, the width of the TrainDrivingInfo window may change frequently due to the data being displayed. This can be visually disturbing.
[Bug #2090962](https://bugs.launchpad.net/or/+bug/2090962).


SteelFill (approved):
Helpful feature, the window scaling is a bit annoying and this is a good start to making that better.

mbm-OR:
> Helpful feature, the window scaling is a bit annoying and this is a good start to making that better.
Thank you.

PR #1013, opened 2024-12-16
Tools for reverse engineering shape and ace files (converting)

Hirek193:
I've added two tools:
- ShapeToObj - Tool to convert shape file to Obj file (actually has some limitations but does it's job - can't convert matricies, not converting materials, provisional import of submodels)
- AceToTga - As name suggests this is tool for converting Ace files to Tga files

Those tools use Orts.Formats.Msts to perform reading operations and in case of ShapeToObj aspose.3d to export mesh as obj file.


twpol (changes requested):
Thanks for these two utilities!

I think these tools would both be better off added to the existing Data Converter (Contributed) program, as it already handles a few file format conversions, and there is no need to get community nor management approval for features in "contributed" programs.

SlavaArduino12345:
> how to use this

Either clone the submitter's repository, or checkout this pull request, and build the ShapeToObj executable inside Visual Studio 2022. Then just drag the shape into the ShapeToObj.exe! How simple is that?

PR #1014, opened 2024-12-16
Bug fix for https://bugs.launchpad.net/or/+bug/2091895 Train disappears from train list window

Csantucci:
See https://www.elvastower.com/forums/index.php?/topic/38454-changing-train-driven-by-the-player/

### DIFF
--- a/Source/Orts.Simulation/Simulation/Simulator.cs
+++ b/Source/Orts.Simulation/Simulation/Simulator.cs
@@ -1955,14 +1955,14 @@ namespace Orts.Simulation
                             TrainSwitcher.ClickedSelectedAsPlayer = false;
                             return;
                         }
-                        if (playerTrain.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN || !playerTrain.Autopilot)
+                        if (playerTrain.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN || TimetableMode && !playerTrain.Autopilot)
                         {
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/Simulator.cs
```diff
@@ -1955,14 +1955,14 @@ private void StartSwitchPlayerTrain()
...
                         }
-                        if (playerTrain.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN || !playerTrain.Autopilot)
+                        if (playerTrain.TrainType == Train.TRAINTYPE.AI_PLAYERDRIVEN || TimetableMode && !playerTrain.Autopilot)
                         {
                             // it must be autopiloted first
                             playerTrain.SwitchToAutopilotControl();
                         }
                         // and now switch!
                         playerTrain.TrainType = Train.TRAINTYPE.AI;
-                        playerTrain.Autopilot = false;
```
cesarBLG, line 1965: Is removing this intentional? The line is present in ORNYMG

Csantucci: Correcting an error and inserting another error.
Thanks!

cesarBLG (approved):
Thanks for the fix! Looks correct to me.

Csantucci:
> Thanks for the fix! Looks correct to me.

You have suggested it, thanks!

PR #1019, opened 2024-12-24
Rename files and folder to correct case

twpol:
Architecture: https://github.com/openrails/openrails/blob/master/Docs/Architecture.md#projects-assemblies-namespaces

This PR is mostly file renamed which only change the case of the file/folder. As such, people on Windows will not see most of the changes when pulling (`git` will update the case of files but not the folders).

To update the case locally without deleting everything, people can use File Explorer to manually make the same changes to the following directories:
- `Source\Orts.Common`
- `Source\Orts.Content`
- `Source\Orts.IO`
- `Source\Orts.Menu`
- `Source\Orts.Settings`
- `Source\Orts.Updater`


cjakeman (approved):
Hi James,

Tidiness just makes everything easier.

Approved with thanks,

Chris

PR #1020, opened 2024-12-25
Bug fix for https://bugs.launchpad.net/or/+bug/2092489 Containers not appearing in timetable mode

Csantucci:
See https://www.elvastower.com/forums/index.php?/topic/38462-container-freight-shapes-not-appearing-in-timetable-mode/


cesarBLG (approved):
Thanks for the fix!

PR #1021, opened 2024-12-27
Various traincar operations windows fixes

sweiland-openrails:


mbm-OR:
Hi Siebren,
When the Brake SystemType is Vacuum_single_pipe, the BrakeSystem.AngleCockAOpenAmount value will always be 1, so the AngleCocks will not work  as required.

Regards,
Mauricio

sweiland-openrails:
hi Mauricio,

I've messed things up, see above after [sweiland-openrails](https://github.com/sweiland-openrails) added 2 commits

I wanted to Commit and Push my fixes in the third commit. And it came with the extra commit Merging the branch from my fork. Does that undo your fix? If so do you know how to repair?

regards, Siebren.

mbm-OR:
The "Fix F9 reversal" commit, allows the TrainCarOperations window to be reversed.

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1080,7 +1080,7 @@ protected override void OnActivate(bool sameCamera)
             }
             else
             {
-                isVisibleTrainCarViewerOrWebpage = (Viewer.TrainCarOperationsWindow.Visible && !Viewer.TrainCarOperationsViewerWindow.Visible) || Viewer.TrainCarOperationsViewerWindow.Visible || (Viewer.TrainCarOperationsWebpage.Connections > 0 && Viewer.TrainCarOperationsWebpage.TrainCarSelected);
+                isVisibleTrainCarViewerOrWebpage = (Viewer.TrainCarOperationsWindow.Visible && !Viewer.TrainCarOperationsViewerWindow.Visible) || Viewer.TrainCarOperationsViewerWindow.Visible || (Viewer.TrainCarOperationsWebpage?.Connections > 0 && Viewer.TrainCarOperationsWebpage.TrainCarSelected);
```
sweiland-openrails, line 1083: Looks like this question mark is not that necessary. That's because the else belongs to the if condition "Viewer.TrainCarOperationsWebpage == null". So in the else the varioable TrainCarOperationsWebpage  is not null. It does not harm either.

mbm-OR:
This bug was discovered in the ORNYMG version when the web server was disabled. Carlo created this fix.
Are we sure that the web server will never go down in our environment?

sweiland-openrails:
hi Mauricio,
I did resolve my conversation with the following comment:

I did some testing. And the webserver can be in the process of starting when executing the above code. So for the three added question marks two of them are valid. The first one is not necessary as the if is testing for the field Viewer.TrainCarOperationsWebpage to be null. However it does not harm either.

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -1076,7 +1080,7 @@ protected override void OnActivate(bool sameCamera)
             }
             else
             {
-                isVisibleTrainCarViewerOrWebpage = Viewer.TrainCarOperationsViewerWindow.Visible || (Viewer.TrainCarOperationsWebpage.Connections > 0 && Viewer.TrainCarOperationsWebpage.TrainCarSelected);
+                isVisibleTrainCarViewerOrWebpage = (Viewer.TrainCarOperationsWindow.Visible && !Viewer.TrainCarOperationsViewerWindow.Visible) || Viewer.TrainCarOperationsViewerWindow.Visible || (Viewer.TrainCarOperationsWebpage?.Connections > 0 && Viewer.TrainCarOperationsWebpage.TrainCarSelected);
```
pzgulyas, line 1083: The && and the first || is interesting here, I don't think the expression is correct here. Could you please double check? Also the ? mark  and the same object after & & without ? is incorrect

pzgulyas (approved):
Don't see codewise problems.

PR #1024, opened 2024-12-30
Fix: Adds guard for missing notifications

cjakeman:
DRAFT as don't understand yet why this is needed.
MainForm_shown() is not being called, so Notifications are not being populated.
Thorough fix is needed, but this will prevent an exception and then mark the Notifications as "not available".


cjakeman:
No longer needed.

PR #1025, opened 2024-12-31
Fix for Multiplayer: lost players cannot be kicked out https://bugs.launchpad.net/or/+bug/2092767

Csantucci:
See https://www.elvastower.com/forums/index.php?/topic/38428-standalone-multiplayer-server/page__view__findpost__p__313633


cesarBLG (approved):
Thanks for the feature! This will definitely increase the stability during MP sessions.

Csantucci:
> Thanks for the feature! This will definitely increase the stability during MP sessions.

Thank you!

PR #1026, opened 2025-01-01
Fix for Multiplayer: remote trains chuff can't be heard https://bugs.launchpad.net/or/+bug/2092771

Csantucci:
See https://www.elvastower.com/forums/index.php?/topic/38353-other-engines-missing-sound-in-or-multiplayer/


pzgulyas (approved):
Seems right

Csantucci:
> Seems right

Thanks Peter!

PR #1027, opened 2025-01-03
docs: Update manual link to self-hosted version

twpol:


Csantucci (approved):
Thanks for that, James

PR #1029, opened 2025-01-06
Superelevation Follow Up Fixes

SteelFill:
Follow up to #954 

One lacking point of the superelevation rewrite that a few [users noticed](https://www.elvastower.com/forums/index.php?/topic/38114-multiple-track-profiles/page__view__findpost__p__313415) was the [lack of support for DynaTrax](https://www.trainsim.com/forums/forum/open-rails/open-rails-modeling-and-physics/2315220-dbtracks-with-superelevation?p=2315452#post2315452). Due to the unusual way DynaTrax is handled, the system would interpret a DynaTrax object as a missing section of track, resulting in no superelevation when some should be present. Dedicated detection for this edge case has been added, which will attempt a second method of checking the TSection if the usual method fails to find a match, which should catch a lot of DynaTrax and other custom track sections.

No changes should be required from the content side, though track profiles might not correctly match to DynaTrax sections if "IncludedShapes" or "ExcludedShapes" are used in the track profile as DynaTrax shapes tend to have very different naming schemes than typical track shapes.

Except in niche situations (eg: you want dynamic track with different appearances in different places, as opposed to the current dynamic track system which always produces the same appearance), regular old dynamic track plus TrProfile(s) will be more reliable and easier to work with than DynaTrax.

Also includes other small bugfixes

- [Rigid frame rolling stock will correctly have superelevation disabled when superelevation is disabled globally](https://www.elvastower.com/forums/index.php?/topic/38542-problem-with-superelevation-and-two-axle-rolling-stock/)
- [Dynamic track segments are now limited to 250 sections in order to lower risk of excessively complicated 3D models](https://www.elvastower.com/forums/index.php?/topic/38586-crash-with-track-profile/)
- [Bogie rotation calculations are now disabled while on a turntable as the superelevation method used doesn't work with turntables](https://bugs.launchpad.net/or/+bug/2101066)
- [Some unusual track sections are now accounted for, fixing some cases of missing superelevation](https://www.elvastower.com/forums/index.php?/topic/38699-super-elevation-anomoly/page__gopid__317131#entry317131)
- [Gravity calculations tweaked again to ensure the expected gradient is shown when in a rear cab](https://www.elvastower.com/forums/index.php?/topic/38497-incorrect-gradient-display-on-track-monitor/)

### DIFF
--- a/Source/RunActivity/Viewer3D/SuperElevation.cs
+++ b/Source/RunActivity/Viewer3D/SuperElevation.cs
@@ -59,39 +59,34 @@ namespace Orts.Viewer3D
             // each subsection.  The rotation component changes only in heading.  The translation 
             // component steps along the path to reflect the root of each subsection.
 
-            TrackShape shape;
-
             bool dontRender = false; // Should this shape be left as a static object?
+            bool removePhys = false; // Should superelevation physics be removed from this object?
             SectionIdx[] SectionIdxs;
 
-            try
+            // Using the track object, determine the track sections
+            // Most track sections can be recovered directly from a TrackShape object
+            if (viewer.Simulator.TSectionDat.TrackShapes.TryGetValue(trackObj.SectionIdx, out TrackShape shape))
             {
-                shape = viewer.Simulator.TSectionDat.TrackShapes.Get(trackObj.SectionIdx);
-
                 if (shape.RoadShape)
                     return false; // Roads don't use superelevation, no use in processing them.
 
                 // Can't render superelevation on tunnel shapes
                 dontRender = shape.TunnelShape;
+                // Can't render superelevation and shouldn't have physics superelevation on junctions and crossovers
+                dontRender |= removePhys = (shape.NumPaths > 1 && shape.ClearanceDistance != 0);
                 SectionIdxs = shape.SectionIdxs;
-            }
-            catch (Exception)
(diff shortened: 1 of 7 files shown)


Source/RunActivity/Viewer3D/SuperElevation.cs
```diff
@@ -61,23 +61,35 @@ public static bool DecomposeStaticSuperElevation(Viewer viewer, TrackObj trackOb
...
+
+                // Can't render superelevation on tunnel shapes
+                dontRender = shape.TunnelShape;
+                SectionIdxs = shape.SectionIdxs;
             }
             catch (Exception)
             {
-                return false; // Won't be able to render with superelevation
+                // Some route-specific shapes (DynaTrax) won't be populated in the TrackShapes list, check the TrackPaths list
+                if (viewer.Simulator.TSectionDat.TSectionIdx.TrackPaths.TryGetValue(trackObj.SectionIdx, out TrackPath path))
```
pzgulyas, line 81: Why not doing a similar check for the TrackShapes above inside the try block, so that the try-catch could be avoided altogether, instead of uprunning this case onto an exception?

SteelFill: That should work, I'll investigate next time I have a moment to code.

twpol (approved):

PR #1030, opened 2025-01-11
Refactor settings, in prep for settings exporter

rwf-rr:
This was intended to be a refactoring of the Settings, to eventually support exporting the settings from the registry to the `OpenRails.ini` file. Current OpenRails uses default settings when it sees an empty `OpenRails.ini` file (the current way to switch).

However, I ended up doing only minimal refactoring, to limit the impact to existing code  / functionality. Notably I did not attempt to use a single `SettingsStore` object for all Settings objects. This would have had a large impact, and could have negatively affect the UpdateManager / UpdateSettings, which use a different file (`Updater.ini`).

But I added all changes to existing code, needed to support the exporter. This will allow all impacting changes to be reviewed at once. The exporter itself will be a separate PR.

## Refactoring

The refactoring is limited to decouple the other Settings classes from the `UserSettings` class. The static variables `RegistryKey` and `SettingsFilePath` are moved from `UserSettings` to `SettingsBase`. The other Settings classes are update to reference `SettingsBase` instead of `UserSettings` for those statics.

Before:
```C#
    public class UserSettings : SettingsBase
    {
        public static readonly string RegistryKey;        // ie @"SOFTWARE\OpenRails\ORTS"
        public static readonly string SettingsFilePath;   // ie @"C:\Program Files\Open Rails\OpenRails.ini"
        static UserSettings()
        {
            // Only one of these is allowed; if the INI file exists, we use that, otherwise we use the registry.
            RegistryKey = "SOFTWARE\\OpenRails\\ORTS";
            SettingsFilePath = Path.Combine(ApplicationInfo.ProcessDirectory, "OpenRails.ini");
            if (File.Exists(SettingsFilePath))
                RegistryKey = null;
            else
                SettingsFilePath = null;
```
After:
```C#
    public abstract class SettingsBase
    {
        public const string DefaultRegistryKey = "SOFTWARE\\OpenRails\\ORTS";
        public const string DefaultSettingsFileName = "OpenRails.ini";

        public static string RegistryKey { get; private set; }        // ie @"SOFTWARE\OpenRails\ORTS"
        public static string SettingsFilePath { get; private set; }   // ie @"C:\Program Files\Open Rails\OpenRails.ini"

        static SettingsBase()
        {
            // Only one of these is allowed; if the INI file exists, we use that, otherwise we use the registry.
            RegistryKey = DefaultRegistryKey;
            SettingsFilePath = Path.Combine(ApplicationInfo.ProcessDirectory, DefaultSettingsFileName);
            if (File.Exists(SettingsFilePath))
                RegistryKey = null;
            else
                SettingsFilePath = null;
        }
```
## Additions

`OverrideSettingsLocations()` is added to `SettingsBase` to override the location before any objects are created. It just changes the static variables. It does not affect any existing Settings objects, only future objects.

`ChangeSettingsStore()` is added to `SettingsBase` to change the settings store after the settings were loaded.
It releases the existing store and creates a new one. But it does not change the static variables, as that would affect other future objects.


Source/Orts.Common/SettingsBase.cs
```diff
@@ -100,6 +148,33 @@ protected SettingsBase(SettingsStore settings)
...
+
+        /// <summary>
+        /// Change the settings store. Creates a new SettingsStore based on the provided parameters.
+        /// Only one of filePath and registry key should be specified. If both are provided, filePath
+        /// prevails. Does not change the static locations (as that would affect other objects).
+        /// See also SettingsStore.GetSettingStore() and OverrideSettingsLocations().
+        /// </summary>
+        /// <param name="filePath">The path to the INI file, or NULL if using the registry.</param>
+        /// <param name="registryKey">The registry key (name), or NULL if using an INI file. </param>
+        /// <param name="section">Optional, the name of the section / subkey.</param>
```
twpol, line 166: If we were to make `SettingsStore.Section` public, could we drop the `section` parameter here?

It seems suboptimal to have to keep old of the section names public, but it's not a big deal either way

twpol (approved):
I know this is a draft but @cjakeman and myself thought it worth providing feedback early

This looks really good and clean, with only the one comment, and nothing that'd stop it being accepted

Source/Orts.Common/SettingsBase.cs
rwf-rr: Thanks for the feedback. I don't see how the handling of `section` could be made simpler. `section` is a member of `SettingsStore`, and in the case of `ChangeSettingsStore()`, the settings store object is discarded.

How would you like to handle this? Do you want to merge this PR first? Or should I add the [SettingsExporter](https://github.com/rwf-rr/rwf-openrails/pull/5) and then remove the Draft flag?

twpol: I was thinking that the code in `ChangeSettingsStore()` could save the section into a local variable and use it for the `SettingsStore.GetSettingStore()` call, since it looks to me like the `SettingStore` won't ever be `null` there, even though the code checks for that (we shouldn't defend code against things which should not be possible, better to crash)

If you don't believe that would work, that's okay, leave the code as-is

We generally like to keep the [different kinds of change](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md#making-changes) in separate PRs because they have different requirements for reviewing and merging, so if the [refactoring of the main Open Rails code](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md#refactoring-process) is complete in this PR, I would recommend starting a separate one for the [contributed projects](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md#contributed-projects)

Since I've already approved this PR, unless you have changes to make, we should be able to un-draft it and merge it - but I've also marked it for inclusion in the Unstable Version, so that if we don't merge it straight away, we'll still have people testing the code indirectly

rwf-rr: I prefer not to make any more changes, as I have done extensive testing. Let's go with as it is.

cjakeman: Roger has indicated his preference, so I'm resolving the conversation and approving and merging.

cjakeman:
Roger,

Would you resolve the conflict that we have with Testing. Then I will approve and merge.

Thanks,

Chris

rwf-rr:
> Would you resolve the conflict that we have with Testing. Then I will approve and merge.

Done, and retested. I had to make a simple change in the TelemetrySettings.cs, so that the section name is externally available (for the settings exporter).

Once this is merged, I will create a PR for the settings exporter.

twpol (approved):
Apologies for the wait, this looks good to me!

PR #1032, opened 2025-01-16
fix: revert namespace changes to restore Notifications

cjakeman:
Don't yet know why these refactor commits broke the Notifications.
Until it gets figured out, these reverts will restore normal operation.


cjakeman:
No longer needed as James has provided a fix.

PR #1033, opened 2025-01-18
Multiple signal light textures per signal

cesarBLG:
https://trello.com/c/5KLJIbNL/603-multiple-light-textures-per-signal
https://www.elvastower.com/forums/index.php?/topic/38532-multiple-light-textures-per-signal/

### DIFF
--- a/Source/RunActivity/Viewer3D/Signals.cs
+++ b/Source/RunActivity/Viewer3D/Signals.cs
@@ -374,7 +374,7 @@ namespace Orts.Viewer3D
                     {
                         frame.AddPrimitive(material, slp, RenderPrimitiveGroup.Lights, ref xnaMatrix, ShapeFlags.None, state);
                     }
-                    renderEffect(SignalTypeData.Material);
+                    renderEffect(slp.Material);
                     if (Viewer.Settings.SignalLightGlow)
                         renderEffect(SignalTypeData.GlowMaterial);
                 }
@@ -465,7 +465,6 @@ namespace Orts.Viewer3D
     {
         readonly Viewer Viewer;
 
-        public readonly Material Material;
         public readonly Material GlowMaterial;
 #if DEBUG_SIGNAL_SHAPES
             public readonly SignalTypeDataType Type;
@@ -488,7 +487,6 @@ namespace Orts.Viewer3D
             if (!viewer.SIGCFG.LightTextures.ContainsKey(mstsSignalType.LightTextureName))
             {
                 Trace.TraceWarning("Skipped invalid light texture {1} for signal type {0}", mstsSignalType.Name, mstsSignalType.LightTextureName);
-                Material = viewer.MaterialManager.Load("missing-signal-light");
 #if DEBUG_SIGNAL_SHAPES
                     Type = SignalTypeDataType.Normal;
 #endif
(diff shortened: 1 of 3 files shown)


Source/RunActivity/Viewer3D/Signals.cs
```diff
@@ -535,7 +533,20 @@ public SignalTypeData(Viewer viewer, SignalType mstsSignalType)
                             continue;
                         }
                         var mstsLight = viewer.SIGCFG.LightsTable[mstsSignalLight.Name];
-                        Lights.Add(new SignalLightPrimitive(viewer, new Vector3(-mstsSignalLight.X, mstsSignalLight.Y, mstsSignalLight.Z), mstsSignalLight.Radius, new Color(mstsLight.r, mstsLight.g, mstsLight.b, mstsLight.a), glowDay, glowNight, mstsLightTexture.u0, mstsLightTexture.v0, mstsLightTexture.u1, mstsLightTexture.v1));
+                        var material = defaultMaterial;
+                        if (!string.IsNullOrEmpty(mstsSignalLight.LightTextureName))
+                        {
+                            if (!viewer.SIGCFG.LightTextures.ContainsKey(mstsSignalType.LightTextureName))
```
SteelFill, line 539: I believe this line is intended to be `if (!viewer.SIGCFG.LightTextures.ContainsKey(mstsSignalLight.LightTextureName))` as checking for `mstsSignalType.LightTextureName` allows line 545 to proceed even if the custom light texture is missing, which causes the signal to error out and not get rendered.

Replacing this line fixes the errors, causes the warnings to appear in the log, and results in signals appearing with no graphical issues.

```suggestion
                            if (!viewer.SIGCFG.LightTextures.ContainsKey(mstsSignalLight.LightTextureName))
```

Source/RunActivity/Viewer3D/Signals.cs
```diff
@@ -535,7 +533,20 @@ public SignalTypeData(Viewer viewer, SignalType mstsSignalType)
...
+                            {
+                                Trace.TraceWarning("Skipped invalid texture {1} for signal light {0}", mstsSignalLight.Name, mstsSignalLight.LightTextureName);
```
SteelFill, line 541: This warning could be a bit more helpful. Right now, telling the missing texture name and missing signal light name doesn't narrow it down all that much, as the `mstsSignalLight.Name` is usually something very generic like "Green Light". To really find which signal is causing the problem, this warning should additionally say the name of the signal type `mstsSignalType.Name` causing the error.

SteelFill (changes requested):
Very handy feature, however error handling is not set up correctly and any misconfigured textures are causing signals to disappear instead of reporting a warning.

Source/RunActivity/Viewer3D/Signals.cs
cesarBLG: You're right, I copy-pasted the error handling and forgot to change it.

Source/RunActivity/Viewer3D/Signals.cs
cesarBLG: Done

SteelFill (approved):
All looks good now, looking forward to what folks create with this new ability.

pzgulyas (approved):
Seems okay

PR #1034, opened 2025-01-18
Fix menu notifications

twpol:
This addresses two related issues:
- Deserializing notification JSON is sensitive to namespaces
- Non-web exceptions (such as deserializing errors!) are not handled


cjakeman (approved):
Hi James,

Works for me and no one has reported otherwise.

I'm approving and merging this now (i.e. before Friday) so that it gets into the next Testing Version.

Thanks,

Chris

PR #1038, opened 2025-01-31
Git ignore Visual Studio Properties/launchSettings.json in all folders.

rwf-rr:
The .gitignore only mentioned the RunActivity folder.


cjakeman (approved):
Hi Roger,

Good catch.

Approved and merged,

Chris

PR #1040, opened 2025-02-04
Make change of odometer count direction apply immediately.

rwf-rr:
Previously the change in count direction (Shift-Ctrl-Z) would not apply until a reset (Ctrl-Z) is done.

Now, the odometer will change immediately when its count direction is changed. 

For example, you pass a switch and reset (Ctrl-Z) the odometer. Then you notice it counts up. Now, when you hit toggle-direction (Shift-Ctrl-Z), the counter will jump to the length of the train minus the distance already traveled between when the reset was done and when the count-direction was changed. When the counter reaches zero, your train has passed the switch.

Feature request: https://www.elvastower.com/forums/index.php?/topic/38587-odometer-count-updown-not-taking-effect-until-reset/

I ended up doing some refactoring, as I had a hard time understanding the logic. Part of the complexity is that the odometer behaves different from a typical odometer. When doing a reverse move, the odometer can be used to count the distance as if it were a normal move.

Hence, the reset logic depends on the position of the reverser at the time of the reset. The reverser position at reset is preserved in the OdometerDirectionForward flag.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -360,7 +360,7 @@ namespace Orts.Simulation.RollingStocks
 
         protected float OdometerResetPositionM = 0;
         protected bool OdometerCountingUp = true;
-        protected bool OdometerCountingForwards = true;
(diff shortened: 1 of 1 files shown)


cjakeman (approved):
Hi Roger,

Your re-factor and extended comments will be very helpful.

Approved and merged.

Chris

PR #1041, opened 2025-02-08
Trans: updated Portuguese translations

cjakeman:


Source/Locales/Orts.Simulation/pt_PT.po
```diff
@@ -0,0 +1,2962 @@
+msgid ""
```
cesarBLG, line 1: Hi Chris,

I think this file should be named pt-PT.po instead

Source/Locales/Orts.Simulation/pt_PT.po
```diff
@@ -0,0 +1,2962 @@
...
+msgstr ""
+"Project-Id-Version: CPV v1.0\n"
+"POT-Creation-Date: 2023-12-03 15:16:52+0100\n"
+"PO-Revision-Date: 2025-01-05 01:38+0000\n"
+"Last-Translator: \n"
+"Language-Team: Rui Fonseca\n"
+"Language: pt_PT\n"
```
cesarBLG, line 8:
```suggestion
"Language: pt-PT\n"
```
Hi Chris,
I think this file should be named pt-PT.po (i.e. with - instead of _), by looking at the rest of locales.

cjakeman: Good catch, Cesar. 
I've corrected the filename and also the Language field which was wrong in several files.
Thanks,
Chris

cesarBLG (approved):
Thanks Chris!

PR #1042, opened 2025-02-09
Fix Emergency Brake Pressure Being Overridden

SteelFill:
[Bug on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38608-ortsmaxservicecylinderpressure/)

A small oversight in the implementation of the service cylinder pressure limit was causing emergency applications to be vented down to the service cylinder pressure all the time. While the service pressure limit was correctly disabled for the actual level of application, the venting is handled separately and needed an additional check to be disabled for emergency applications.


cesarBLG (approved):
Thanks for the fix!

PR #1045, opened 2025-02-14
Bugfix: Empty Start Time for Explore, and other issues loading from Menu Selection and Content Routes

rwf-rr:
Bug: [2097690](https://bugs.launchpad.net/or/+bug/2097690)
Discussion: [purpose-of-start-element-in-contentroutes-setting](https://www.elvastower.com/forums/index.php?/topic/38617-purpose-of-start-element-in-contentroutes-setting/)

This started with a simple issue: The Start Time was empty (unset) for Explore activities of (many) automatically installed routes. That caused RunActivity to fail. 

As I tried to understand the code, I discovered other issues where values from the settings (Menu Selection or Content Routes) were not handled correctly. For example:
- Some fields were loaded from the Menu Selection settings even thought the settings were from a different activity.
- When loading from the Content Route settings, the StartAt and HeadingTo fields were set to the same.

Eventually I decided to refactor some of the code, to make it easier to understand.

### DIFF
--- a/Source/Menu/MainForm.cs
+++ b/Source/Menu/MainForm.cs
@@ -1335,6 +1335,8 @@ namespace Menu
 
         void UpdateFromMenuSelection<T>(ComboBox comboBox, UserSettings.Menu_SelectionIndex index, Func<T, string> map, T defaultValue)
         {
+            // if folder field, or the selected folder matches the menu selection setting
+            // and there is a value in the menu selection setting
             if (((index == UserSettings.Menu_SelectionIndex.Folder) ||
                  ((comboBoxFolder.Items.Count > 0) && (SelectedFolder != null) &&
                   (Settings.Menu_Selection.Count() > 0) &&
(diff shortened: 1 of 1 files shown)


SteelFill:
Didn't realize this one wasn't on unstable, it is now. I reproduced the bug in a previous version and couldn't reproduce it after this was added, so it seems to work, but this'll have to stew in unstable for a moment to make sure it doesn't cause anything unusual for normal play.

SteelFill (approved):
Original issue is fixed, and no new issues have been noted. This PR is good to merge when you next have the opportunity.

rwf-rr:
There is a bug in this PR. Specifically the commit [a9316c718](https://github.com/openrails/openrails/pull/1045/commits/a9316c71810b3137a3540a1b86afa5445fb53f10) is incorrect, due to a misunderstanding.

The fix is [PR 1135](https://github.com/openrails/openrails/pull/1135).

PR #1050, opened 2025-03-01
refactor: TrainCar junction detection

twpol:
Before:

- TrainCar is-over-junction did not work at all
- Sound is-over-junction worked for switches but not crossovers

After:

- New is-over-junction works for switches and crossovers
- TrainCar and Sound both use same flag
- Is-over-water-trough check also migrated to new code for simplicity


cjakeman:
Hi James,

Isn't this called with every frame and every car in the train?
If so, it sounds a bit demanding for a lengthy train (but I've not made any measurements).

From the comments here:
// Scan through the track sections forwards from the REAR of the train
// stopping as soon as we've passed this car (`checkedM`) or run out of track
it looks as though we could do this once for the leading car and, if nothing is found, skip the remaining cars in the train.

Have I undestood the algorithm correctly?

Chris

peternewell (approved):
Hi James,

Thanks for sorting this out.

PR #1051, opened 2025-03-02
Mouse wheel support for cabview controls + refactor

pzgulyas:
Roadmap: https://trello.com/c/V6pIZGjM/580-add-mousewheel-support

### DIFF
--- a/Source/RunActivity/Viewer3D/Cameras.cs
+++ b/Source/RunActivity/Viewer3D/Cameras.cs
@@ -266,7 +266,7 @@ namespace Orts.Viewer3D
         {
             // Will not zoom-in-out when help windows is up.
             // TODO: Property input processing through WindowManager.
-            if (UserInput.IsMouseWheelChanged && !Viewer.HelpWindow.Visible)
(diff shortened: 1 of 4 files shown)


cjakeman (approved):
Hi Peter,

Refactoring seems very effective. Looks much cleaner and shorter.

Approved and merged.

Thanks,

Chris

Source/RunActivity/Viewer3D/Cameras.cs
```diff
@@ -266,7 +266,7 @@ protected void ZoomByMouseWheel(float speed)
         {
             // Will not zoom-in-out when help windows is up.
             // TODO: Property input processing through WindowManager.
-            if (UserInput.IsMouseWheelChanged && !Viewer.HelpWindow.Visible)
+            if (UserInput.IsMouseWheelChanged && !Viewer.HelpWindow.Visible && !Viewer.RenderProcess.IsMouseVisible)
```
SteelFill, line 269: This check seems to prevent camera zooming when people expect it to function (eg: in exterior view #2 and #3). Checking if the mouse cursor isn't visible is clearly not a good way to determine when zooming should be allowed, as the mouse cursor is regularly visible in exterior views when there aren't even any controls to manipulate. This must be changed to be more useful (perhaps ignore the mouse check unless the camera is in a 3D cab view).

PR #1052, opened 2025-03-03
Content Manager: Add axle count, and lowest derail force

rwf-rr:
Adding the axle count for the consist, as mentioned in this thread:

Feature request: https://www.elvastower.com/forums/index.php?/topic/38493-consist-data-in-content-manager/page__view__findpost__p__314665

Also adding the minimal derail force. There are some issues with cars derailing in the BNSF Starter Route (Scenic Sub) See [here ](https://www.trainsim.com/forums/forum/open-rails/open-rails-discussion/2316088-problems-with-trainsimulations-bnsf-scenic-subdivision-demo-route?p=2317033#post2317033)(TrainSIm.com). This helps identify trains at risk (and needing some fixing).

BTW, some rolling stock has incorrect axles/wheels config. Most are because of legacy-MSTS issues. But some also mix ORTS Axles and MSTS Wheels. That usually explains unexpected Axle counts.

**Update**: Added one more enhancement to this PR: Showing the path under the route category (after activities). 
I often run RunActivity directly, and this makes it easy to find a suitable path. Previously one had to mavigate through each service to view a path.

**Update 2**: Also  added TpEPA and TpEDBA, when units are imperial.
TpEPA = Tons per Equivalent Powered Axles.
TpEDBA = Tons per Equivalent Dynamic Brake Axles.

These metrics are more accurate than HP per Ton, as they are based on
tractive effort.

The calculations are based on Union Pacific practices. UP uses a
standard axle of 10'000 lbf tractive effort or dynamic brake force.
For example a GP40-2 is 5.0 EPA, an SD40-2 is 7.1 EPA. Both have the
same HP (and thus result in the same HP-per-Ton).

**Example of Path list**:

image

**Example of Consist info (before second update)**:

image

### DIFF
--- a/Source/Contrib/ContentManager/Models/Car.cs
+++ b/Source/Contrib/ContentManager/Models/Car.cs
@@ -15,12 +15,16 @@
 // You should have received a copy of the GNU General Public License
 // along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
 
+// MSTS Wheels and ORTS Axles
+// * ORTS uses unpowered and drive axles. Eg. 2 and 4 for an ES44C4
+// * MSTS uses a mix of wheels and axle. The Wagon section has the total,
+//   the Engine section the driven only.
+//   Values greater than 6 are wheels, less than 6 are axles. Some are
+//   intentionally wrong to work around MSTS issues.
+
 using Orts.Formats.Msts;
 using System;
-using System.Collections.Generic;
 using System.Diagnostics;
-using System.Linq;
-using System.Text;
 
 namespace ORTS.ContentManager.Models
 {
@@ -38,15 +42,20 @@ namespace ORTS.ContentManager.Models
         public readonly string Description;
         public readonly float MassKG;
         public readonly float LengthM;
-        public readonly float MaxBarkeForceN;
+        public readonly int NumDriveAxles;
+        public readonly int NumAllAxles;
+        public readonly float MaxBrakeForceN;
         public readonly float MaxPowerW;
         public readonly float MaxForceN;
         public readonly float MaxSpeedMps;
         public readonly float MinCouplerStrengthN;
+        public readonly float MinDerailForceN;
 
         public Car(Content content)
         {
             Debug.Assert(content.Type == ContentType.Car);
+
+            const float GravitationalAccelerationMpS2 = 9.80665f;
 
             // .eng files also have a wagon block
             var wagFile = new WagonFile(content.PathName);
@@ -55,7 +64,7 @@ namespace ORTS.ContentManager.Models
             Name = wagFile.Name;
             MassKG = wagFile.MassKG;
             LengthM = wagFile.WagonSize.LengthM;
-            MaxBarkeForceN = wagFile.MaxBrakeForceN;
+            MaxBrakeForceN = wagFile.MaxBrakeForceN;
             MinCouplerStrengthN = wagFile.MinCouplerStrengthN;
 
             if (System.IO.Path.GetExtension(content.PathName).Equals(".eng", StringComparison.OrdinalIgnoreCase))
@@ -68,7 +77,23 @@ namespace ORTS.ContentManager.Models
                 MaxForceN = engFile.MaxForceN;
                 MaxSpeedMps = engFile.MaxSpeedMps;
                 Description = engFile.Description;
+                // see MSTSLocomotive.Initialize()
+                if (engFile.NumDriveAxles > 0) { NumDriveAxles = engFile.NumDriveAxles; }
+                else if (engFile.NumEngWheels >= 7f) { NumDriveAxles = (int)(engFile.NumEngWheels / 2f); }
+                else if (engFile.NumEngWheels > 0f) { NumDriveAxles = (int)engFile.NumEngWheels; }
+                else { NumDriveAxles = 4; }
             }
+
+            // see MSTSWagon.LoadFromWagFile()
+            if (wagFile.NumWagAxles > 0) { NumAllAxles = wagFile.NumWagAxles + NumDriveAxles; }
+            else if (wagFile.NumWagWheels >= 7f) { NumAllAxles = (int)(wagFile.NumWagWheels / 2f); }
+            else if (wagFile.NumWagWheels > 0f) { NumAllAxles = (int)wagFile.NumWagWheels; }
+            else { NumAllAxles = 4; }
+            if (NumDriveAxles > NumAllAxles) { NumAllAxles = NumDriveAxles; }
+
+            // see TrainCar.UpdateTrainDerailmentRisk()
+            var numWheels = NumAllAxles * 2;
+            if (numWheels > 0) { MinDerailForceN = MassKG / numWheels * GravitationalAccelerationMpS2; }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Source/Orts.Formats.Msts/EngineFile.cs
```diff
@@ -47,12 +51,16 @@ public EngineFile(string filePath)
                             new STFReader.TokenProcessor("type", ()=>{ EngineType = stf.ReadStringBlock(null); }),
                             new STFReader.TokenProcessor("maxpower", ()=>{ MaxPowerW = stf.ReadFloatBlock( STFReader.UNITS.Power, null); }),
                             new STFReader.TokenProcessor("maxforce", ()=>{ MaxForceN = stf.ReadFloatBlock( STFReader.UNITS.Force, null); }),
```
SteelFill, line 53: As I mentioned on the code for the help menu consist display, when doing EPA calculations, you want to use the max continuous force (assuming the continuous force was actually entered in). An additional line here should be added to read MaxContinuousForce, and the consist tractive effort should use the continuous force in preference to the absolute max force.

Source/Contrib/ContentManager/Models/Consist.cs
```diff
@@ -75,18 +75,51 @@ public Consist(Content content)
...
 
-                        if (wag.IsEngine && engFile.MaxForceN > 25000)  // exclude legacy driving trailers / cab-cars
+                        if (wag.IsEngine)
                         {
-                            EngCount++;
+                            // see MSTSLocomotive.Initialize()
+                            ortsEngAxles = engFile.NumDriveAxles;
+                            if (ortsEngAxles >= 0) { numDriveAxles = ortsEngAxles; }
+                            else if (engFile.NumEngWheels > 7f) { numDriveAxles = (int)(engFile.NumEngWheels / 2f); }
+                            else if (engFile.NumEngWheels > 0) { numDriveAxles = (int)engFile.NumEngWheels; }
```
SteelFill, line 84: This is inconsistent with the behavior in the simulator. If the number of wheels is >= 7 (not strictly greater than!), then the number of axles is defined to be 4. There's no behavior that divides the number of wheels by 2, so this will produce numbers different from the actual sim.

```suggestion
                            else if (engFile.NumEngWheels != 0 && engFile.NumEngWheels < 7) { numDriveAxles = (int)engFile.NumEngWheels; }
```

That doesn't strictly make sense, but there are years worth of discussions about how useless 'NumWheels' is as a parameter and how huge swaths of content use NumWheels values that are neither correct as the number of wheels nor correct as the number of axles. It's more helpful to tell people what the program will do than to try and make sense of any NumWheels values.

Source/Contrib/ContentManager/Models/Consist.cs
```diff
@@ -75,18 +75,51 @@ public Consist(Content content)
...
                         }
-                    } catch (IOException e) // continue without details when eng/wag file does not exist
+
+                        // see MSTSWagon.LoadFromWagFile()
+                        if (ortsEngAxles >= 0 && wagonFile.NumWagAxles >= 0) { numAllAxles = ortsEngAxles + wagonFile.NumWagAxles; }
+                        else if (wagonFile.NumWagAxles >= 0) { numAllAxles =  wagonFile.NumWagAxles; }
+                        else if (wagonFile.NumWagWheels >= 7f) { numAllAxles = (int)(wagonFile.NumWagWheels / 2f); }
+                        else if (wagonFile.NumWagWheels >= 0f) { numAllAxles = (int)wagonFile.NumWagWheels; }
+                        else { numAllAxles = 4; }
+                        if (numDriveAxles > numAllAxles) { numAllAxles = numDriveAxles; }
```
SteelFill, line 107: This whole section is a bit clunky and also doesn't reflect what the simulator will do. Again, there's no case anywhere in the sim where NumWheels is divided by 2, and despite calculating 'num**All**Axles' most of the cases here will just give the wagon axles. Something like this would be more accurate and concise:

```suggestion
                        if (wagonFile.NumWagAxles != 0 || wag.IsEngine) { numAllAxles = wagonFile.NumWagAxles; }
                        else if (wagonFile.NumWagWheels != 0 && wagonFile.NumWagWheels < 6) { numAllAxles = (int)wagonFile.NumWagWheels; }
                        else { numAllAxles = 4; }
                        
                        numAllAxles += numDriveAxles;
```
In this approach, we are consistent with the (weird) implementation in the simulator, and more consistently produce a correct number of total axles by first calculating the unpowered axles, then adding on powered axles (if there were any).

Source/Contrib/ContentManager/Models/Car.cs
```diff
@@ -68,7 +79,26 @@ public Car(Content content)
...
+                else { NumDriveAxles = 4; }
             }
+
+            // see MSTSWagon.LoadFromWagFile()
+            if (ortsEngAxles >= 0 && wagFile.NumWagAxles >= 0) { NumAllAxles = ortsEngAxles + wagFile.NumWagAxles; }
+            else if (wagFile.NumWagAxles >= 0) { NumAllAxles = wagFile.NumWagAxles; }
+            else if (wagFile.NumWagWheels >= 7f) { NumAllAxles = (int)(wagFile.NumWagWheels / 2f); }
+            else if (wagFile.NumWagWheels > 0f) { NumAllAxles = (int)wagFile.NumWagWheels; }
+            else { NumAllAxles = 4; }
+            if (NumDriveAxles > NumAllAxles) { NumAllAxles = NumDriveAxles; }
```
SteelFill, line 97: Same comments here as in Consist.cs. This won't always produce the same number(s) of axles as the simulator will truly use.

Source/Contrib/ContentManager/Models/Consist.cs
rwf-rr: Thanks for your feedback. As a new contributor I appreciate the details and insights.

There is one additional difference that you have not commented on and that affects the requested changes. In the parsers, I initialize the ORTS axle counts with -1 (negative one). I did this to know when the ORTS axle counts were absent from the config file. The normal parser defaults to 0 (zero) which cannot be differentiated from a value of 0 (zero).

To accommodate this the checks need to be "> 0" instead of "!= 0".

Are you also concerned with the parser change, and recommend that I revert them?

Source/Contrib/ContentManager/Models/Consist.cs
SteelFill: Yes, it is different. Perhaps the thinking is that NumWheels( 6 ) on a wagon could mean a 6 wheel wagon (3 axle stock wasn't unusual in the UK) or a 6 axle wagon, both of which are reasonable, so it's ambiguous and we give up. Meanwhile on a locomotive NumWheels( 6 ) probably means 6 axles as 6 axle locomotives are more prevalent than 3 axle locomotives.

Of course, that is making the assumption NumWheels was set to be the wheels or the axles, and often times it's neither!

Source/Contrib/ContentManager/Models/Consist.cs
SteelFill: The NumEngWheels count will initialize to 0 according to the setup of EngineFile.cs, don't see anything that needs to be changed about that.

rwf-rr:
Changes as discussed here and in the Train Info window PR are now completed.

- Axle count now matches what the simulation uses.
- Using continuous tractive effort for the TpEPA calculation.
- Using trailing tons for all rating calculations.

twpol (approved):

PR #1054, opened 2025-03-09
Steam Cylinder Sound Bug

peternewell:
https://bugs.launchpad.net/or/+bug/2101725

Fix bug with steam locomotive steam cylinder sounds not playing when stationary.


pzgulyas (approved):
I don't see any bad with this

PR #1055, opened 2025-03-10
Route Based TTrack Sounds

peternewell:
This PR adds the ability to define track based sounds "within" the route, and thus removes need for individual stock to have track sounds associated with it.


cjakeman:
Hi Peter,

I was wondering what happens when a route-based sound and a traincar-based sound are both valid? Do we hear them both or does one take precedence? What is the mechanism to choose the precedence?

Best wishes,

Chris

peternewell:
> 
> I was wondering what happens when a route-based sound and a traincar-based sound are both valid? Do we hear them both or does one take precedence? What is the mechanism to choose the precedence?
> 
If there are track based sounds in the car then they will be heard.

I am unaware of how to disable car based track sounds as I don't believe that they have a specific trigger that could be "ignored"?

Hence it will probably require Users to remove track based sounds from the cars SMS file.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2009,26 +2009,69 @@ public virtual void Update(float elapsedClockSeconds, bool auxiliaryUpdate = tru
...
             // This is then reset every time a track joint is triggered, and positioned the same distance apart, hence reletative positions are maintained.
             // Only runs once at start up.
-            if (SoundSetupInitialise)
+            if (TrackJointSoundSetupInitialise)
             {
                 var trackjointdistanceM = (float)Simulator.TRK.Tr_RouteFile.DistanceBetweenTrackJointsM;
-                float trainLengthM = 0;
+                var trainLengthM = 0.0f;
+                var cummulativeTrackJointDistanceM = 0.0f;
+                var remainDistanceM = (float)Simulator.TRK.Tr_RouteFile.DistanceBetweenTrackJointsM;
```
rwf-rr, line 2017: When ```ortsdistancebetweentrackjoints``` is not defined, ```DistanceBetweenTrackJointsM``` ends up to be 0 (zero). This then causes an infinite loop below.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2009,26 +2009,69 @@ public virtual void Update(float elapsedClockSeconds, bool auxiliaryUpdate = tru
             // Initialise track joint trigger points. Sets the trigger point for the track joint reletative to other cars.
             // This is then reset every time a track joint is triggered, and positioned the same distance apart, hence reletative positions are maintained.
             // Only runs once at start up.
-            if (SoundSetupInitialise)
+            if (TrackJointSoundSetupInitialise)
             {
                 var trackjointdistanceM = (float)Simulator.TRK.Tr_RouteFile.DistanceBetweenTrackJointsM;
```
rwf-rr, line 2014: When ```ortsdistancebetweentrackjoints``` is not defined, ```DistanceBetweenTrackJointsM``` ends up to be 0 (zero). This then causes an infinite loop below.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2009,26 +2009,69 @@ public virtual void Update(float elapsedClockSeconds, bool auxiliaryUpdate = tru
...
+                        remainDistanceM -= car.CarLengthM;
+
+      //                  Trace.TraceInformation("Initialise Track Joints< - CarID {0} RealDistance {1} TrainLength {2} TRackJointDistance {3} CarLength {4} CumDistance {5} RemDistance {6}", car.CarID, car.realTimeTrackJointDistanceM, trainLengthM, trackjointdistanceM, car.CarLengthM, cummulativeTrackJointDistanceM, remainDistanceM);
+
+                        if (remainDistanceM < 0.0f)
+                        {
+
+                            while (Math.Abs(remainDistanceM) > trackjointdistanceM)
+                            {
+                                remainDistanceM += trackjointdistanceM;
```
rwf-rr, line 2061: This is an infinite loop when ```trackjointdistanceM``` is 0 (zero). See above, the presence and value of ORTSDistanceBetweenTrackJoints needs to be validated, and a suitable default used.

PR #1057, opened 2025-03-12
Switchable brake system

pzgulyas:
[Forum thread](https://www.elvastower.com/forums/index.php?/topic/38658-air-brakes-europe-uic)
[Trello card](https://trello.com/c/wBxnzpHS)


pzgulyas:
It is not ready for review, but seems like the "for-unstable" label no longer works on Draft PR-s. @twpol James, could you please check?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -640,23 +638,8 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
-            }
-            else
-            {
-                MaxBrakeForceN = InitialMaxBrakeForceN;
-
-                if (Simulator.Settings.VerboseConfigurationMessages)
-                {
-                    Trace.TraceInformation("Unknown BrakeShoeType set to OR default (Cast Iron) with a MaxBrakeForce of {0}", FormatStrings.FormatForce(MaxBrakeForceN, IsMetric));
-                }
-            }
```
SteelFill, line 659: This behavior was removed and never relocated anywhere else. This is breaking all of my content, which uses `ORTSMaxBrakeShoeForce` without using `MaxBrakeForce`. Without any lines to set `MaxBrakeForceN` to `MaxBrakeShoeForceN`, all my stuff is being set to a brake force of 0...not very useable.

pzgulyas: Could you please recheck it?

pzgulyas: Hold on, still not quite right.

pzgulyas: Should be okay now

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -409,8 +409,8 @@ public override string[] GetDebugStatus(Dictionary<BrakeSystemComponent, Pressur
                 FrontBrakeHoseConnected ? "I" : "T",
                 string.Format("A{0} B{1}", AngleCockAOpenAmount >= 1 ? "+" : AngleCockAOpenAmount <= 0 ? "-" : "/", AngleCockBOpenAmount >= 1 ? "+" : AngleCockBOpenAmount <= 0 ? "-" : "/"),
                 BleedOffValveOpen ? Simulator.Catalog.GetString("Open") : string.Empty,
+                string.Format("{0}t-{1}%", (int)Kg.ToTonne(BrakeMass), GetBrakePercent()),
```
cesarBLG, line 412: A couple of comments about the display (not about the feature itself):
- I'm not 100% convinced about including the braked mass in the brake HUD, nothing about forces is being displayed there. I think a better place is the 'consist information' tab, which also includes consist mass and length.
- Brake mass/brake percentage are arbitrary numbers specific to UIC, do we want to display them for all stock?
- I would include the brake percentage only for the train, not for individual cars (but still include the braked mass for them). It's uncommon to talk about brake percentage for wagons, because in the end the important number for drivers is the brake percentage. You can't combine the brake percentage of wagons to get the train brake percentage, you have to add all braked masses first.

pzgulyas: I'm also not sure that all info is needed in this HUD, but I needed to display it somewhere for debug, to see if the calculation is done correctly. After all it is called "debug" HUD for a reason, to help the debug.

The info is useful only for UIC stock, but the problem is we don't know which stock is UIC. :-)

For calculating the % for the train, the brake mass has to be back-calculated for all rolling stock that doesn't has this info from the wag (practically none yet). I'm trying to do this with the same formula Darwin provided, back from the brake force. And need to display the results on some HUD to see if the results make sense at all. And in the end the brake % of the train will be the average of the %-s of the individual wagons-engines. (Do I understand it correctly?)

SteelFill: The brake % can be incorporated in the help menu train info #1064 once that's done. There are already some American brake parameters listed in there, like tons per operative brake. These are hidden if the sim is in metric mode, so if it is in metric mode those stats could be replaced with the total brake weight and overall brake %.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: Appears friction brakes are working again with max brake shoe force.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
cesarBLG: > And in the end the brake % of the train will be the average of the %-s of the individual wagons-engines. (Do I understand it correctly?)

No, it doesn't work that way. That's why I say that displaying the % for individual cars is misleading. The train % is calculated as (total braked mass)/(total mass), which is a weighted average of the individual %s (i.e. heavy wagons are more important for the %). I've never seen brake % calculate for individual cars.

> The brake % can be incorporated in the help menu train info https://github.com/openrails/openrails/pull/1064 once that's done. There are already some American brake parameters listed in there, like tons per operative brake. These are hidden if the sim is in metric mode, so if it is in metric mode those stats could be replaced with the total brake weight and overall brake %.

That makes sense

pzgulyas: Seems you are right, I'll change it. Also the brake mass fits better into the consist HUD, at least geometrically, since that one is shorter. :-)

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1084,16 +1005,134 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
+                    LoadFullMaxHandbrakeForceN = brakeSystemNext.InitialMaxHandbrakeForceN;
+                    LoadFullMaxBrakeForceN = BrakeShoeType != BrakeShoeTypes.Unknown && brakeSystemNext.MaxBrakeShoeForceN != 0 ? brakeSystemNext.MaxBrakeShoeForceN : brakeSystemNext.InitialMaxBrakeForceN;
+                    LoadFullRelayValveRatio = (brakeSystemNext as AirSinglePipe)?.RelayValveRatio ?? 0;
+                    LoadFullInshotPSI = (brakeSystemNext as AirSinglePipe)?.RelayValveInshotPSI ?? 0;
+                }
+            }
+            else
+            {
+                // Set up the traditional brake config style without any brake modes and freight animations
+                LoadEmptyMaxHandbrakeForceN = LoadFullMaxHandbrakeForceN = BrakeSystem.InitialMaxHandbrakeForceN;
```
pzgulyas, line 1099: @SteelFill Yeah, seems I didn't realize that while I set the default values for the brake related freightanim variables here, I need to do the same for the Davis related ones and/or move the UpdateDavisLoadCompensation(TempMassDiffRatio) call at the end of the LoadFromWagFile into the freightanim block above it.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1084,16 +1005,134 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
+                LoadEmptyMaxBrakeForceN = LoadFullMaxBrakeForceN = BrakeShoeType != BrakeShoeTypes.Unknown && BrakeSystem.MaxBrakeShoeForceN != 0 ? BrakeSystem.MaxBrakeShoeForceN : BrakeSystem.InitialMaxBrakeForceN;
+                LoadEmptyRelayValveRatio = LoadFullRelayValveRatio = (BrakeSystem as AirSinglePipe)?.RelayValveRatio ?? 0;
+                LoadEmptyInshotPSI = LoadFullInshotPSI = (BrakeSystem as AirSinglePipe)?.RelayValveInshotPSI ?? 0;
```
pzgulyas, line 1102: @SteelFill Yeah, seems I didn't realize that while I set the default values for the brake related freightanim variables here, I need to do the same for the Davis related ones and/or move the UpdateDavisLoadCompensation(TempMassDiffRatio) call at the end of the LoadFromWagFile into the freightanim block above it.

SteelFill: Reworking the load empty/full davis values seems to have done the trick, it's a lot less messy too.

SteelFill:
Peter, I have noticed a curious new issue. With the change made in commit [254b6d4](254b6d45c984b9f030375a9745af6599a1152162) some values are being set erroneously to -9999. This is because [FreightAnimations.cs initializes a lot of values to -9999](https://github.com/pzgulyas/openrails-1/blob/254b6d45c984b9f030375a9745af6599a1152162/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs#L64) to indicate they are unassigned. The problem is, your `setIfNonZero` function doesn't realize this and sets the values to -9999 because -9999 is 'non zero' after all. The things that get set to -9999 aren't supposed to accept negative values and this breaks physics when it happens (consider negative aerodynamic drag, negative brake force, negative center of gravity...the physics system is very unhappy with this).

The previous implementation that checked explicitly for _positive_ values actually worked much better, and I don't get the physics glitchyness if I revert to [commit 707560c](https://github.com/openrails/openrails/pull/1057/commits/254b6d45c984b9f030375a9745af6599a1152162). The only parameter in the list that can reasonably accept a negative value is `LoadEmptyInshotPSI` and `LoadFullInshotPSI`, everything else should be set to only accept the value if it's positive (maybe add a second `setIfPositive` function, or change `setIfNonZero` to only set if positive, then add special behavior to allow negative inshot values.)

I wonder if [this issue](https://www.elvastower.com/forums/index.php?/topic/38771-open-rails-is-broken/) is related? Setting things negative when they shouldn't be could cause the problem in the original post (the problems later on seem unrelated).

pzgulyas:
Phillip, thanks for investigating this, I made the changes you suggested. My fault that I haven't properly checked, but honestly I didn't expect this. :-) And probably the -9999-s could also be removed now, defaulting them back to zero, but I leave that to another time, for someone else. I didn't even thought to be needed bothering with freightanims in this PR, but someone put brake settings to an inappropriate place.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -651,330 +641,118 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
+                setIfPositive(ref LoadFullORTSDavis_C, FreightAnimations.FullPhysicsStaticOne?.FullStaticORTSDavis_C);
+                setIfPositive(ref LoadFullDavisDragConstant, FreightAnimations.FullPhysicsStaticOne?.FullStaticORTSDavisDragConstant);
+                setIfPositive(ref LoadFullWagonFrontalAreaM2, FreightAnimations.FullPhysicsStaticOne?.FullStaticORTSWagonFrontalAreaM2);
+                setIfPositive(ref LoadFullMaxBrakeForceN, FreightAnimations.FullPhysicsStaticOne?.FullStaticMaxBrakeShoeForceN);
+                setIfPositive(ref LoadFullMaxBrakeForceN, FreightAnimations.FullPhysicsStaticOne?.FullStaticMaxBrakeForceN);
+                setIfPositive(ref LoadFullMaxHandbrakeForceN, FreightAnimations.FullPhysicsStaticOne?.FullStaticMaxHandbrakeForceN);
+                setIfPositive(ref LoadFullCentreOfGravityM_Y, FreightAnimations.FullPhysicsStaticOne?.FullStaticCentreOfGravityM_Y);
+                setIfPositive(ref LoadFullRelayValveRatio, FreightAnimations.FullPhysicsStaticOne?.FullStaticRelayValveRatio);
+                setIfPositive(ref LoadFullInshotPSI, FreightAnimations.FullPhysicsStaticOne?.FullStaticInshotPSI);
```
SteelFill, line 740: This one should be `setIfNonZero` as well fyi. (The other places where inshot is set are already fine.)
```suggestion
                setIfNonZero(ref LoadFullInshotPSI, FreightAnimations.FullPhysicsStaticOne?.FullStaticInshotPSI);
```

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1084,18 +798,134 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
+                            loStageAir.RelayValveRatio = LoadEmptyRelayValveRatio;
+                            hiStageAir.RelayValveRatio = LoadFullRelayValveRatio;
+                            loStageAir.RelayValveInshotPSI = LoadEmptyInshotPSI;
+                            hiStageAir.RelayValveInshotPSI = LoadFullInshotPSI;
+                            loStageAir.BrakeMassToShoeForce(null);
+                            hiStageAir.BrakeMassToShoeForce(null);
+                        }
+
+                        BrakeSystems.Add((BrakeSystem.BrakeMode, LoadEmptyMassKg), loStage);
+                        BrakeSystems.Add((BrakeSystem.BrakeMode, LoadFullMassKg), hiStage);
```
SteelFill, line 824: This is causing a bunch of content to error out if `LoadEmptyMassKg` is equal to `LoadFullMassKg`, which will happen on any rolling stock that has a freight animation but doesn't have variable cargo weight (eg: static freight animation). In this case, the key for the BrakeSystems list is the same for both the loStage and hiStage `(BrakeSystem.BrakeMode, LoadEmptyMassKg) == (BrakeSystem.BrakeMode, LoadFullMassKg)`, and dictionaries can't have two values with the same key.

You should probably check if `LoadEmptyMassKg != LoadFullMassKg` before trying to create separate load/empty brake systems; if the mass is the same it should need only one brake system.

pzgulyas: I've committed a fix. It is actually enough if the masses are different, otherwise there no basis for the load compensation, so can switch back to the original method. Thanks for the investigation, you are absolutely correct.

SteelFill:
Heads up Peter, #1067 is merged and the associated conflict can be solved. Looks like Cesar got unlucky and added some save and load commands on the same line as you did, so there's no real conflict, just need to tell github to accept both changes.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -3242,81 +3057,60 @@ private void UpdateTenderLoad()
...
-                        DavisDragConstant = ((LoadFullDavisDragConstant - LoadEmptyDavisDragConstant) * TempMassDiffRatio) + LoadEmptyDavisDragConstant;
-                    }
+        private void UpdateBrakeLoadCompensation(float massDiffRatio)
+        {
+            // The LoadFull... and LoadEmpty... variables are set by the SetBrakeSystemMode() method based on the actually active brake mode.
+            // The massDiffRatio is always 1 for manually switchable load stages, even for an empty wagon.
+            // Only wagons with automatic continuous load compensation can have a massDiffRatio below 1.
+            MaxBrakeForceN = ((LoadFullMaxBrakeForceN - LoadEmptyMaxBrakeForceN) * massDiffRatio) + LoadEmptyMaxBrakeForceN;
+            MaxHandbrakeForceN = ((LoadFullMaxHandbrakeForceN - LoadEmptyMaxHandbrakeForceN) * massDiffRatio) + LoadEmptyMaxHandbrakeForceN;
+            // Not sensible to vary the relay valve ratio continouously; instead, it changes to loaded if more than 25% cargo is present
```
cesarBLG, line 3230: Are you sure about this? I know this is just for freight trains, but for passenger trains with automatic load compensation, the triple valve ratio is continuously varied depending on the load.

SteelFill: This is from #912, not this PR. Just showing as additions because of the way things were moved around. If someone wanted to make a different load proportioning system than the freight animation one, that's what the rest of this PR is for.

cesarBLG:
Hello Peter, regarding the brake controller, wouldn't be better to expose the brake mode to the script and make it react accordingly? Creating multiple scripts may cause some issues if it has some 'memory' of previous states. I'd implement the switch using dedicated cabview controls, but the logic should be handled inside the script. For example, you could add a button to enable/disable EP braking, regenerative braking, etc. for the brake controller.

I think it's a different situation to the distributors which are handled by levers outside the wagons, for which it makes sense to have them in the F9 menu with other outside operations.

pzgulyas:
@cesarBLG Yes, it would make sense to make these things available for the scripts, I can add the hooks. What would be needed to be exposed in your opinion? Only the name to get, and a setter to switch between the modes?

pzgulyas:
@twpol James, could you please help me on this PR? For some reason it doesn't get merged into the unstable release. It seems not to conflict with the 'master', but when I try to manually merge it to the 'unstable' on my computer, I get a lot of strange conflict errors, which seems to show conflicts with itself (?) or some other commits that already got merged into master. I'm not sure what I am doing wrong here. Maybe I should just close this PR and reopen again as new as a single commit? I fear that would conflict too.

cesarBLG:
@pzgulyas yes, that would be great. My concerns are now that 1) the per-mode brake controllers are defined in the wagon section. I think they belong to the engine section, as the original controller. 2) You are creating a new brake controller for every mode. While this may happen for very specific stock, what we usually have is a single brake controller that behaves differently depending on the brake mode.

pzgulyas:
@cesarBLG I've added (2) specifically per Darwin's request, he needs this for configuring the UK dual-mode air-vacuum locomotives, where the notch controller notches differ. In (1) I would agree with you, but duplicating the same BrakeMode() block logic inside the Engine section for the sole purpose to be in that section feels counterproductive. That brings in further complexities in the parsing code along with additional needed logic for matching the similar named blocks in the Engine-Wagon section, what to do if there is one mode available in only one or the other, and so on... And it brings no value that I see. Would it be worth it, what's on the pro side?

cesarBLG:
I think it makes sense from an architectural point of view to distinguish between the actuator (wagon section)  and the controller (engine section). In some cases the brake mode of the controller may not be the same as the actuator (e.g. due to a mistake). But I agree with you that it adds duplicated code.

Regarding the duplication of the controller, I guess if someone wants to implement the controller logic inside the script then it would just skip the definition of per-brake mode controllers, and your approach does not restrict that, so I'm fine with it. It's also very simple to setup in the .eng file. Something that may make sense is that, even if you duplicate the ScriptedBrakeController class for every brake mode, all of the classes reference the same instance of the script.

pzgulyas:
> @twpol James, could you please help me on this PR?

@twpol Sorted out, now it merges. It was the 1b6b22e0 commit in PR #570 of merging 'master' (by you actually in last October :-) ) that added a couple of white space alignments into MSTSWagon.cs that caused merging conflicts with this PR. I reverted those in bb5ba8d and now it merges. (We may readd them later in a different PR)

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1162,33 +865,138 @@ public virtual void LoadFromWagFile(string wagFilePath)
...
-
-                // Replace any null davis values with zeros to prevent errors
-                if (!DavisAN.HasValue)
-                    DavisAN = 0.0f;
-                if (!DavisBNSpM.HasValue)
-                    DavisBNSpM = 0.0f;
-                if (!DavisCNSSpMM.HasValue)
-                    DavisCNSSpMM = 0.0f;
-                if (!DavisDragConstant.HasValue)
-                    DavisDragConstant = 0.0f;
```
SteelFill, line 1184: This section is vital to prevent crashes elsewhere in the program, removing it causes [this bug](https://www.elvastower.com/forums/index.php?/topic/39588-and-a-third-one-from-horseshoe/).

PR #1059, opened 2025-03-17
Fix F9 powersupply crashes

mbm-OR:
Fix a crash reported [here](https://bugs.launchpad.net/or/+bug/2102839).

Bug #2102839


Csantucci (commented):
Hi Mauricio,
may I understand why Cars[CarPosition -1] surely exists if Cars[CarPosition] does not exist?

rwf-rr (commented):
The change looks fine to me. I stumbled on this bug, made the same change, and it tested fine.

Csantucci (approved):
That's OK for me.

PR #1061, opened 2025-03-17
Fix F9 Web-page conflict.

mbm-OR:
An internal conflict was detected that caused an internal loop when the TrainCarOperations web page version was opened and F9 was not visible.

More info here:
[Bug #2103422](https://bugs.launchpad.net/or/+bug/2103422)


sweiland-openrails (approved):
correct

PR #1062, opened 2025-03-18
Train Forces popup Window.

rwf-rr:
This is a new popup window that shows the main forces within the train in graph form. See forum for the initial discussion. It is bound to Alt-F7.

Feature request: https://www.elvastower.com/forums/index.php?/topic/38056-proposal-for-train-forces-popup-display/

HighPull-DPU

The window scales with the train length, with a minimum and maximum size. Scrolling may be needed to see everything.

**Coupler Force (longitudinal):**

Shows the length-wise pull or push force at each coupling, as a colored bar graph. Up (positive) is pull, down (negative) is push. The scale is determined by the weakest coupler in the train. The steps  are non-linear, to provide more sensitivity near the breaking point (of the weakest coupler).

A red bar may not mean that the coupler is about to break, if it is not at the car with the weakest coupler.

**Derail Force (lateral):**

Shows the side-way push or pull at the wheels as a colored bar graph. Up (positive) is pull to the inside (stringline), down (negative) is push to the outside (jackknife). The scale is determined by the lowest axle-load (lowest vertical force). The steps are non-linear, to provide more sensitivity near the derailing point. But this is less effective for lateral forces, as the force is proportional to the curve radius, which changes in discrete steps.

A red bar may not mean that the car is about to derail, if it is not at the car with the lowest derail force.

**Brake Force:**

Shows the brake force of each car as a bar graph. The scale is determined by the car with the smallest brake force (smallest weight). The steps are non-linear, to provide more sensitivity near small brake applications. As the weight (and thus brake force varies greatly between cars (and especially engines), the graph can be quite jagged.

**Bar Graph:**

- +/- 9 bars; 4 green, 3 orange, 2 red  (except the brake force graph only has green)
-  blue middle-bar is an engine, white is a car
- front (head-end) is to the left

**Text Line:**

- **Max Coupler**: This is the highest coupler force in the train, and the car (count) at which the force is the highest.
- **Max Derail**: This is the highest derail (lateral) force in the train, and the car (count) at which the force is the highest.
- **Low Coupler**: This is the lowest break-force of all the couplers in the train. It is used to scale the graphs.
- **Low Derail**: This is the lowest derail force of all the cars in the train. It is used to scale the graphs.

The max and lowest brake forces are intentionally not included. They are less relevant for train-handling.

**Design Notes**:

- The graphs show the absolute force. I considered using relative forces (scaled for each car). But I found the graphs less useful that way.
- The information is similar to what is in the HUD (Forces Tab), but presented more concise.
- I find the window most useful for running long, heavy trains on mountain grades.

Note that this PR does **not** include the WebServer version. That requires a separate re-implementation, which is significant effort. It may be added later.

HighDerailDPU

BrakesApply

LongTrain-DPU
This train is too long and requires scrolling to see the tail end.


SteelFill:
Added the for-unstable label so this can start appearing in unstable versions for testing.

Source/RunActivity/Viewer3D/Popups/TrainForcesWindow.cs
```diff
@@ -0,0 +1,429 @@
...
+        }
+
+        /// <summary>
+        /// Update the brake force icon for a car. The image has 10 icons;
+        /// index 0 is neutral, 9 is max braking.
+        /// </summary>
+        private void UpdateBrakeForceImage(TrainCar car, int carPosition)
+        {
+            var idx = 0;  // neutral
+            var absForceN = car.BrakeForceN;
```
SteelFill, line 414: I think it would make the brake display a bit more useful if this also included the dynamic brake force on locomotives. It's a little hard to judge train forces when bailed-off locomotives in dynamic braking show 0 brake force.

rwf-rr: Good point. But I think I have to somehow differentiate (color?) between air and dynamic brakes.
The current version helped me to realize that the DPUs did not apply the independent in sync with the lead. Some modern North American locos seem to be missing `ORTSDPBrakeSynchronization ( "Apply, Release, Emergency, Independent" )` part.

SteelFill: Personally, I'd be fine even if dynamics and air have the same appearance. The total brake force is what affects coupler forces, so from a coupler stress standpoint it doesn't really matter if it's 50 klb of dynamics, 50 klb air, or 25klb air + 25klb dynamics.

rwf-rr:
Discussed changes completed. Dynamic brake force is now shown in blue.

Source/RunActivity/Viewer3D/Popups/TrainForcesWindow.cs
SteelFill: Dynamic brake force is shown now which is nice, but I'm not sure I agree with the choice to scale the bar to the higher of dynamic brake force or air brake force. While freight operations will only use one of the brake systems at a time, on passenger trains and in other parts of the world blended air and dynamic braking is very typical. The window won't give a full picture of the actual amount of braking the locomotive does because it isn't adding air and dynamic brake force together.

It's fine if the color change is driven by whichever is greater, but the total force is a more useful metric.

Source/RunActivity/Content/TrainForcesSprites.png
SteelFill, line 1: I don't know how feasible it is, but it really feels like this texture could be simpler.
There's a bit much empty space while simultaneously having a lot of redundancy. It doesn't feel like it should be necessary to have a white rectangle with a bunch of green rectangles above it, then another part of the texture showing a blue rectangle with a bunch of green rectangles above it-surely the blue/white rectangles could be stored without any green rectangles, then the green rectangles stored separately and stitched into place with rendering magic.
Similarly, it's quite redundant to store a stack of 9 green rectangles, then entirely separately store a stack of 8 green rectangles . Surely we could just store the stack of 9 green rectangles then cut ones off the top when they aren't needed. (I would understand storing all the stacks of rectangles separately if previous rectangles changed shape/color between stacks, but each stack simply adds more rectangles on top of the existing ones).

SteelFill:
image
It seems another area of improvement is these numeric readouts below the graphs. Even with the longest train I know of (the 230 car double empty grainer from Mullan Pass), the window reaches max width...but isn't wide enough to clearly show the low coupler strength and requires scrolling to see the low derail force at all. If there was less empty space between these readouts, I could probably read them all, but that might not be any help for shorter trains where the window (currently) is only wide enough for the max coupler force to be seen fully.

This is also different from the screenshots in the PR card where you can clearly see everything once the train is long enough, something may have changed on accident.

SteelFill:
Fixed merge conflict caused by merging #1029 since Roger is on vacation. There wasn't actually a conflict but both PRs added new code in the same location.

Source/RunActivity/Content/TrainForcesSprites.png
twpol: I expect this is mostly due to a limitation in the current `Image` control – it can only stretch the texture rectangle over the whole, exact area of the control

I would suggest adding a new property to `Orts.Viewer3D.Popups.Image` of `public Rectangle Padding` which could be used to adjust the `destinationRectangle` left/top/right/bottom inside `void Draw(SpriteBatch, Point)` (when not `Rectangle.Empty`)

I would make **positive** left/top/right/bottom values move the `destinationRectangle` **inwards** (just like you'd expect padding to work)

That should be just enough to allow the source texture to contain a maximum of 1 copy of each bar stack – because adjusting the top or bottom of both the `Source` and `Padding` by the same amount should cause the image to be cropped without shifting or stretching

twpol (dismissed):
This is a fantastic feature!

Although I am comfortable with the code as it is currently, there are some outstanding comments from @SteelFill and I would love to have the image rendering improved to simplify the sprite file (see comments on `Source/RunActivity/Content/TrainForcesSprites.png`)

cjakeman:
> This is a new popup window that shows the main forces within the train in graph form. See https://www.elvastower.com/forums/index.php?/topic/38056-proposal-for-train-forces-popup-display/ for the initial discussion. It is bound to Alt-F7.

Now that this work has stabilised, would you extend this PR to update the Manual document too, please?

Would you mind creating a Trello card for this PR in line with "[Contributing to Open Rails](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)". Thanks.

Source/RunActivity/Viewer3D/Popups/TrainForcesWindow.cs
rwf-rr: Done, now showing the sum of brake forces, and using the higher to determine the colour.

```
            var absForceN = Math.Abs(car.BrakeForceN);  // using Math.Abs() for safety and consistency
            var absDynForceN = Math.Abs(car.DynamicBrakeForceN);
            if (absDynForceN > absForceN) { isDynamicBrakes = true; }
            absForceN += absDynForceN;
```

Source/RunActivity/Content/TrainForcesSprites.png
rwf-rr: I would like to do this in a **separate PR**, as it is an optimization, and also applies to the Train Info PR.

I will also explore changing the Add(image) method, as it is an alternate place to specify an offset for the image.

rwf-rr:
> image It seems another area of improvement is these numeric readouts below the graphs. Even with the longest train I know of (the 230 car double empty grainer from Mullan Pass), the window reaches max width...but isn't wide enough to clearly show the low coupler strength and requires scrolling to see the low derail force at all. If there was less empty space between these readouts, I could probably read them all, but that might not be any help for shorter trains where the window (currently) is only wide enough for the max coupler force to be seen fully.
> 
> This is also different from the screenshots in the PR card where you can clearly see everything once the train is long enough, something may have changed on accident.

This has not changed recently (but was different in the very beginning). I added sufficient padding to accommodate translations, and I right-justified so that the position does not changes as the number of digits changes. Lastly, I did not want the window to be larger than necessary. The limits will also be in the Train Info tab.

For the labels, I could calculate the string length, and use that for the field size. That would shorten the line a bit.

rwf-rr:
I have added a section to the user manual.

Trello card at https://trello.com/c/MzuIbNah/608-add-train-forces-popup-window

Source/RunActivity/Content/TrainForcesSprites.png
SteelFill: That's fine, the current solution works and if you want to change the train info screen as well that would be much easier done in a since PR after these two are merged in.

Source/RunActivity/Viewer3D/Popups/TrainForcesWindow.cs
```diff
@@ -0,0 +1,436 @@
...
+                    }
+
+                    carPosition++;
+                }
+
+                if (MaxCouplerForceForTextBox != null)
+                {
+                    // TODO: smooth the downslope
+                    MaxCouplerForceForTextBox.Text = FormatStrings.FormatLargeForce(absMaxCouplerForceN * couplerForceSign, false) +
+                        string.Format(" ({0,3})", maxCouplerForceCarNum);
```
SteelFill, line 297: Check your coupler and derail force string formatting, you are passing an `isMetric = false` in multiple places, giving forces in klbf even if on a European route where kN would be more standard. The proper value of IsMetric can be found from the player locomotive.

rwf-rr: Oops, I meant to change that after the initial implementation, but then forgot. Will do now.

BTW, observation on IsMetric:

IsMetric (and IsUK) is defined in TrainCar.cs and set in MSTSWagon.cs (MSTSLocomotive is a subclass of MSTSWagon). Each car (and locomotive) has its own value, but they are all the same. Oddly, Train.cs does not have IsMetric, even though it would seem more appropriate to be there. But I guess there can be cars that are not part of a train.

Train.cs uses MilepostUnitsMetric, which is not necessarily the same as IsMetric.

```
IsMetric = Simulator.Settings.Units == "Metric" || (Simulator.Settings.Units == "Automatic" && System.Globalization.RegionInfo.CurrentRegion.IsMetric) ||
    (Simulator.Settings.Units == "Route" && Simulator.TRK.Tr_RouteFile.MilepostUnitsMetric);
IsUK = Simulator.Settings.Units == "UK";
```

But at this stage I will _not_ embark on any refactoring.

rwf-rr: Done. Using `PlayerTrain.LeadLocomotive.IsMetric`.

SteelFill: Better now. The `IsMetric` and `IsUK` system is indeed pretty confusing and dumb. Doesn't need to change here and now but a more unified and customizable system is definitely a good idea.

SteelFill (approved):
This change is good to go for now. With all the functionality complete, small tweaks like the sprite situation can be investigated later as that's not a blocker for getting this very useful feature out there.

PR #1063, opened 2025-03-23
Better Missing Track Section Error Message

SteelFill:
Feature request: https://www.elvastower.com/forums/index.php?/topic/38643-change-an-error-message/

This changes the error message displayed when a route needs a track shape missing from tsection.dat.

- Error message now declares which shape index couldn't be found (only the _first_ missing shape can be detected, as the program will stop before additional track sections are processed) which should make it vastly easier to determine what's missing.
- The error no longer mentions any specific track systems, as it is misleading to imply this error is related to any specific track system.
- Probable cause is now stated to be the wrong version of the standard tsection.dat or the route requiring a non-standard tsection.dat, as this is what will actually produce the error.


rwf-rr:
Looks fine to me, and it addresses the concern raised in the referenced thread.

cesarBLG (approved):
The new error message is more clear, thanks for the change

PR #1064, opened 2025-03-24
Add Train Info tab to Help window (F1)

rwf-rr:
Add a new tab to the Help Window (F1) that shows information about the train. Train Info is currently not very accessible. The HUD has some, but not all that is used by real train-crews.

TrainInfo-mid-dpu-manifest

The above shows a train with a mid-train DPU sets. Two engines in front and 4 in the middle.

Number of Engines shows the count for each set. It will be a single number for a conventional train.

Total Weight also shows the trailing weight of each group of cars. This is useful for calculating train forces.

Max Tractive Effort also shows the maximum for each set of engines. Again, this is useful for calculating train forces.

Horsepower per Ton (HPT) and Tons per Operative Brake (TPOB) are North American metrics. They are only shown when imperial metrics are used. There likely are similar metrics in European, but I am not familiar with them.

Tons per EPA (equivalent powered axles) and Tons per EDBA (equivalent dynamic brake axles) are modern replacements for HP-per-Ton. They use tractive effort and dynamic brake force respectively. The implementation is based on Union Pacific, which uses 10'000 lbf as a standard axle.

I added Lowest Coupler Strength and Lowest Derail Force to aid with train handling. Some cars have configuration that results in lower than realistic values. This can then be considered in the train handling.

The graph shows the weight of each car. The challenge is to find the right scale. I scaled it for North American freight cars. The graph is most useful for long, heavy trains in mountain grades.

Not shown are the number of loads and empties. There is no explicit config for loaded or empty, and using the mass is not reliable across a large variety of cars.

I had to make the window slightly wider, so that all the tab labels fit.

The info is similar to what has recently been added to the Content Manager, Consist view. However, there are minor differences. The Content Manager interprets config files, whereas the Train Info tab uses values from the simulation. A particular difference is the maximum speed, where the simulation also uses the route's maximum speed.

Lastly here is a view for a metric route:

TrainInfo-metric-freight


Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
```diff
@@ -1152,6 +1183,231 @@ private void writeline()
...
+
+                // wag and eng
+                if (wag != null)
+                {
+                    maxBrakeForceN += wag.MaxBrakeForceN;
+                    var couplerStrength = GetMinCouplerStrenght(wag);
+                    if (couplerStrength < lowestCouplerStrengthN) { lowestCouplerStrengthN = couplerStrength; }
+                    var derailForce = GetDerailForce(wag);
+                    if (derailForce < lowestDerailForceN) { lowestDerailForceN = derailForce; }
+                    if (wag.MaxBrakeForceN > 0) { numOperativeBrakes++; }  // simplistic: real world has cutout brakes, and no brakes on some articulated car-segments
```
SteelFill, line 1273: A better check may be to consider the wagon's brake equipment. If the wagon has any brake valve, count it as an operative brake.

```suggestion
                    if (wag.BrakeValve != MSTSWagon.BrakeValveType.None) { numOperativeBrakes++; }  // simplistic: real world has cutout brakes, and no brakes on some articulated car-segments
```

That's still not foolproof (some content creators will still put a triple valve on every section of an articulated unit), but closer to the idea of an operative brake.

Using brake force alone could mistakenly count something with handbrakes only as an operative brake, when that should not count according to the rulebooks.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
```diff
@@ -1152,6 +1183,231 @@ private void writeline()
...
+
+                if (car.IsMetric) { isMetric = true; }; if (car.IsUK) { isUK = true; }
+
+                if (isEng)
+                {
+                    engCount++;
+                    numAxles += eng.LocoNumDrvAxles + eng.GetWagonNumAxles();
+                    totPowerW += eng.MaxPowerW;
+                    totMaxTractiveEffortN += eng.MaxForceN;
+                    engMaxTractiveForceN += eng.MaxForceN;
```
SteelFill, line 1237: For the purposes of what you are doing here, you don't want the total _maximum_ tractive effort, you want the total _continuous_ tractive effort.

```suggestion
                    totMaxTractiveEffortN += eng.MaxContinuousForceN > 0 ? eng.MaxContinuousForceN : eng.MaxForceN;
                    engMaxTractiveForceN += eng.MaxContinuousForceN > 0 ? eng.MaxContinuousForceN : eng.MaxForceN;
```
(The > 0 check is needed as the max continuous force is not a required input and may be left at 0)

Knowing the maximum tractive effort isn't helpful if that level of effort would require frying the motors. Hence, all of UP's rated power axles are defined based on the continuous rating to ensure nobody makes a train that can only temporarily give enough tractive effort for the job at hand.

A good example of that is the current UP SSI giving data for their E9s. Because of the extreme gear ratio used on E units, the power rating for the E9 is just 3.5 despite having 4 updated motors and enough weight to get at least 55,000 pounds down without slipping. Anything more than around 35,000 pounds will overheat the motors as a lot of amperage is needed to achieve that traction.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
rwf-rr: This is a bit more involved. 

For the Content Manager, I would have to replicate the following code from MSTSWagon.cs:

```
                case "wagon(brakeequipmenttype":
                    foreach (var equipment in stf.ReadStringBlock("").ToLower().Replace(" ", "").Split(','))
                    {
                        switch (equipment)
                        {
                            case "triple_valve": BrakeValve = BrakeValveType.TripleValve; break;
                            case "distributor":
                            case "graduated_release_triple_valve": BrakeValve = BrakeValveType.Distributor; break;
                            case "distributing_valve": BrakeValve = BrakeValveType.DistributingValve; break;
                            case "emergency_brake_reservoir": EmergencyReservoirPresent = true; break;
                            case "handbrake": HandBrakePresent = true; break;
                            case "auxilary_reservoir": // MSTS legacy parameter - use is discouraged
                            case "auxiliary_reservoir":
                                AuxiliaryReservoirPresent = true;
                                break;
                            case "manual_brake": ManualBrakePresent = true; break;
                            case "retainer_3_position": RetainerPositions = 3; break;
                            case "retainer_4_position": RetainerPositions = 4; break;
                            case "supply_reservoir":
                                SupplyReservoirPresent = true;
                                break;
                        }
                    }
                    break;
```
Also, this code does not seem to cover all brake equipment. MSTS has cars with "EP_brake" and "Vacuum_brake" in the list. I am not sure how relevant these are.

Second, I found some modern equipment that has wrong values.
* The Amtrak cars in the BNSF Starter Route are missing BrakeEquipmentType, but have MaxBrakeForce . In the same package, EOT devices have Triple_valve in the equipment, but a MaxBrakeForce  of 0 (zero).

Regarding the hand brakes, there is a separate config, MaxHandbrakeForce. 

So, the question is what is the least wrong looking forward.

As to making the change for the Train Info window and Content Manager:

* I am fine to use `BrakeValveType.None` in the F1 Train Info window.
* I am reluctant to copy the above code into the Content Manager.
* I could refactor the above code into a method, and call the method from the Content Manager. I don't know if or how much that will affect the linking of the content manager.
* I could make an assumption that anything other than `Handbrake` probably means that the car has brakes. But that will not match the TrainInfo.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
rwf-rr: Done. I also renamed the variables to reflect the change.

I will do the same in the Content Manager.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
SteelFill: So the EP_brake and Vacuum_brake equipment types led me down a bit of a rabbit hole. Those equipment types, like any equipment type not on the list, don't actually do anything. This means that those trains have a BrakeValve of none, but work anyway. Didn't know that was an allowed state, but this means the BrakeValve isn't foolproof.

The next best thing would be to check the brake system type. If the brake system is SingleTransferPipe or ManualBraking, you know for sure that it won't be simulated with any sort of automatic brake valve. The other brake system types can count as operative. That could be done like this:
```suggestion
                    if (car.BrakeSystem != null && !((car.BrakeSystem is SingleTransferPipe) || (car.BrakeSystem is ManualBraking)) && wag.MaxBrakeForceN > 0) { numOperativeBrakes++; }  // simplistic: real world has cutout brakes, and no brakes on some articulated car-segments
```

Not sure what you mean by starter route wagons missing BrakeEquipmentType, I don't see anything missing there. Besides, if BrakeEquipmentType was missing the brakes wouldn't work at all, which you'd probably want to count as a non-functional brake valve. Similar story if BrakeSystemType was missing, then it gets set as SingleTransferPipe, which shouldn't count as functional either.

For the content manager, the brake system type might be easier, as you just need to read the one string in 'BrakeSystemType' and if it's (converted to lowercase) "manual_braking", "air_piped", or "vacuum_piped" then it definitely won't have any automatic brakes.

rwf-rr: > Not sure what you mean by starter route wagons missing BrakeEquipmentType, I don't see anything missing there. 

Example: `TS_STARTER_ROUTE.zip\TRAINS\trainset\SLI_AMT_EMPIRE_BUILDER\SUPERLINER1_P4_COACH_34001.wag`, 
 extracted from `https://ts-files.com/files/TS_STARTER_ROUTE.zip`. It matches what my ORTS instance has installed (via the content form). Here is the relevant part:
```
	BrakeSystemType( "Air_single_pipe" )
	MaxBrakeForce( 24.3kN )

	MaxHandbrakeForce( 57.2kN )
	NumberOfHandbrakeLeverSteps( 100 )

	EmergencyBrakeResMaxPressure( 90 )
	TripleValveRatio( 2.5 )
	MaxReleaseRate( 50 )
	MaxApplicationRate( 50 )
	MaxAuxilaryChargingRate( 1.684 )
	EmergencyResCapacity( 2.604 )
	EmergencyResChargingRate( 1.684 )
	BrakeCylinderPressureForMaxBrakeBrakeForce( 77 )
```
There is no `BrakeEquipmentType` (and there is no include).

I like your suggestion on using `BrakeSystemType`. I will go with that,

rwf-rr:
Changes as discussed here and in the Content Manager PR are now completed.

- Axle count now matches what the simulation uses.
- Using continuous tractive effort for the TpEPA calculation.
- Using trailing tons for all rating calculations.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
```diff
@@ -1152,6 +1184,239 @@ private void writeline()
...
+        }
+
+        /// <summary>
+        /// Get the lowest coupler strength for a car.
+        /// </summary>
+        private float GetMinCouplerStrenght(MSTSWagon wag)
+        {
+            float couplerStrength = 1e10f;  // default from TrainCar.GetCouplerBreak2N()
+            if (wag.Couplers.Count > 1 && wag.Couplers[1].Break2N < couplerStrength) { couplerStrength = wag.Couplers[1].Break2N; }
+            else if (wag.Couplers.Count > 0 && wag.Couplers[0].Break2N < couplerStrength) { couplerStrength = wag.Couplers[0].Break2N; }
```
SteelFill, line 1402: Is this really supposed to be an _else_ if? The way this has been set up, if a car was constructed with two couplers then the previous if statement will pass and this line will never run, leading to a possible case where Couplers[1] (rear) is stronger than Couplers[0] (front) (think of the A unit of an articulated car where the front end has a standard type E coupler but the rear end has a permanent articulated connection) but the code here will think the rear coupler is the weaker one as it never checks the front one.

Also, I'm wondering about the use of Break2N vs Break1N. While the couplers actually pop at the Break2N force, the Break1N value is considered the 'overload' limit and I could see a future where coupler strength is changed so there are consequences for exceeding Break1N. Perhaps people should be told to avoid exceeding the Break1N value.

SteelFill:
Fixed merge conflict caused by merging #1029 since Roger is on vacation. There wasn't actually a conflict but both PRs added new code in the same location.

Also the exact same code is present in #1062, be careful that doesn't get duplicated when it comes time to merge these.

cjakeman:
> Add a new tab to the Help Window (F1) that shows information about the train. Train Info is currently not very accessible. The HUD has some, but not all that is used by real train-crews.

The forum discussion is [here](https://www.elvastower.com/forums/index.php?/topic/38690-proposal-train-info-tab-in-help-window-f1)

Roger, now that this work has stabilised, would you extend this PR to update the Manual document too, please? (Do let me know if you need help with that.)

Would you mind creating a Trello card for this PR in line with "[Contributing to Open Rails](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)". Thanks.

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
rwf-rr: Corrected. It now considers both couplers and the Break1N value first.

```
            foreach (var coupler in wag.Couplers)
            {
                // ignore unrealistically low values; some cars have Break2N lower than Break1N
                if (coupler.Break1N > 99f && coupler.Break1N < couplerStrength) { couplerStrength = coupler.Break1N; }
                if (coupler.Break2N > 99f && coupler.Break2N < couplerStrength) { couplerStrength = coupler.Break2N; }
            }
```

Source/RunActivity/Viewer3D/Popups/HelpWindow.cs
```diff
@@ -1152,6 +1184,239 @@ private void writeline()
...
+                else { image.Source = new Rectangle(spriteIdx * spriteWidth, spriteHeight, spriteWidth, spriteHeight); }
+                image.Texture = TrainInfoSpriteSheet;
+                weightbox.Add(image);
+            }
+        }
+
+        /// <summary>
+        /// Get the lowest coupler strength for a car.
+        /// </summary>
+        private float GetMinCouplerStrenght(MSTSWagon wag)
```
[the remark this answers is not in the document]
rwf-rr: Corrected.

rwf-rr:
> Roger, now that this work has stabilised, would you extend this PR to update the Manual document too, please? (Do let me know if you need help with that.)
> 
> Would you mind creating a Trello card for this PR in line with "[Contributing to Open Rails](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)". Thanks.

User Manual updated. Trello card created (https://trello.com/c/fmh6lu75/607-add-train-info-tab-to-help-window-f1).

SteelFill (approved):
I think you are good to go at this moment. As the functionality is complete and useful as-is, non-functional changes like messing with sprites are fine to save for later.

PR #1065, opened 2025-03-28
Fix for PantographToggle sound event

cesarBLG:
See https://www.elvastower.com/forums/index.php?/topic/27839-trigger-47-in-sms-files/

### DIFF
--- a/Source/Documentation/Manual/sound.rst
+++ b/Source/Documentation/Manual/sound.rst
@@ -101,7 +101,7 @@ Trigger       Function
     44        SteamHeatChange
     45        Pantograph1Up
     46        Pantograph1Down
-    47        Pantograph1Toggle
+    47        PantographToggle
     48        VigilanceAlarmReset
     54        TrainBrakePressureDecrease 
     56        VigilanceAlarmOn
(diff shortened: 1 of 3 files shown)


rwf-rr:
I am a bit concerned that this change adds future risk.
Previously, the PantographToggle event was conditional to PowerSupplyEvent.RaisePantograph or PowerSupplyEvent.LowerPantograph.
Now, PantographToggle is handled on any event. 
There currently seem to be only the two above events, but if another event were added in the future, there could be a problem.

And a question: the PantographToggle event is now conditional to there being a soundEvent. Is it reasonable to assume that there always will be a sound event?

cesarBLG:
A few comments:
- In OR, Orts.Common.Event mostly means "sound event"
- Previously, the pantograph toggle sound was linked to one of the possible ways the pantograph can be raised/lowered, which is incorrect.
- Now, PantographToggle is handled when any pantograph-related event happens (that is, any pantograph is raised/lowered), which is the expected behaviour.
- The 'soundEvent' is just a local variable that sets the event name whenever a pantograph changes state. Thus, whenever a pantograph changes, we trigger the specific event associated to it (i.e. 'soundEvent') as well as the generic PantographToggle sound event.

The original fix is from @Csantucci and I agree with it. This is the way to go, the previous behaviour was incorrect.

rwf-rr:
Thanks for the explanation, Cesar.

Csantucci (approved):
OK for me

PR #1066, opened 2025-03-28
Log derailment, using TraceInformation.

rwf-rr:
Create a log-file entry when a car derails due to excessive lateral forces on curves.

Note that the simulation continues when a train car derails. There is no physical effect for the derailment, the train proceeds normally. The only indication of the derailment is a temporary notification (bottom left of main window, Train Driving Info window).

This PR adds an entry in the log file for each derailment.

The log uses the distance travelled to indicate the location of the derailment. This is not ideal, as the distance is based on the train (lead loco), and thus two  cars that derail at the same location will have a different distance. Better would be a milepost-based location, but such info is currently not easy to obtain for a car.

Example log:

`Information: Car 0 - 11 derailed (stringlined), on Right curve with radius 0.1mi, at speed 24.6mph, after traveling 1.6mi`

Other variants are "jackknifed" and "Left".

PS: I initially used TraceWarning, but then I saw that for a broken coupler, a TraceInformation log is used.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
@@ -1,4 +1,4 @@
-// COPYRIGHT 2009 - 2022 by the Open Rails project.
+﻿// COPYRIGHT 2009 - 2022 by the Open Rails project.
 // 
 // This file is part of Open Rails.
 // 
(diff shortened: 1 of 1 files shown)


SteelFill (approved):
Works in my testing, should be a helpful feature whenever people have concerns about derailments. Feel free to merge when you can.

pzgulyas (approved):
I think this can be merged.

PR #1067, opened 2025-03-29
Traction and dynamic brake retardation

cesarBLG:
https://trello.com/c/9r6fqJw4/605-traction-and-dynamic-brake-retardation
https://www.elvastower.com/forums/index.php?/topic/37445-dynamic-brake-features/page__view__findpost__p__312552

Include ramp up and ramp down rates, such that when 100% throttle is applied, tractive force is gradually incremented until it reaches 100%

Additionally, include an optional delay between a lever is moved and the setting is accepted by the control electronics.


SteelFill:
Looks like a very useful feature, definitely something I wanted to investigate for quite some time. I'll give it a try with my physics set soon, but one thing I suspect I might want is a limit on the rate of change of dynamic brake _power_. Most of the locomotives I work with have systems in place that effectively (though indirectly) limit both dynamic brake force _and_ power. At least with traction, the diesel engine settings can limit the power ramp up (not helpful for electrics though? maybe power ramping would help there as well).

cesarBLG:
I believe the main goal of force retardation is to reduce the stress to the couplings, and maybe for passenger comfort. Nevertheless, it makes sense that the control electronics also limit the power application rate, so I included that too.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2498,78 +2519,184 @@ protected virtual void UpdateControllers(float elapsedClockSeconds)
 			}
 #endif
         }
+        public void UpdateForceWithRamp(ref float forceN, float elapsedClockSeconds, float targetForceN, float maxForceN, float targetPowerW, float maxPowerW, float rampUpNpS=0, float rampDownNpS=0, float rampZeroNpS=0, float rampUpWpS=0, float rampDownWpS=0, float rampZeroWpS=0)
```
SteelFill, line 2522: UpdateForceWithRamp is expecting inputs for targetPowerW and maxPowerW but none of the implementations of UpdateForceWithRamp actually provide these inputs. This isn't causing a crash as the missing inputs are interpreted to be the optional rampDownWpS and rampZeroWpS inputs.

SteelFill (changes requested):
Not building on unstable for whatever reason but I did a local build to try things out and see how much I could break things.

- When using `ORTSDelayTimeBeforeUpdating`, the throttle value shown on the F5 and Alt + F5 HUDs show the throttle setting _with_ the delay. I think most people would expect the HUD to show the controls immediately without delay.
- Oddly enough, the same isn't true in dynamic braking, in that case only the F5 train driving HUD shows dynamic braking with a delay, while the Alt + F5 HUD is instant.
- This may be considered a nitpick, but the ramp parameters taking force/power units directly and assuming the change is per one second is inconsistent with other places in the program where the given units must explicitly be _rate of change units_. There's no provisions right now for power or force rate of change units but it could be useful.
- The force ramping doesn't work due to a mismatch in the inputs expected by the `UpdateForceWithRamp` function and the inputs given to the function.

That said, hackily changing the code to make force ramping work has promising results. I was worried about possible unexpected interactions where the ramp rate might interfere with situations that require forces to change more rapidly for reasons other than the control system (eg: AC locomotive dynamic brake effort curves tell dynamic brake force to rapidly drop to 0, which may require a more rapid drop than the ramp rates allow), but that seems accounted for.

- While doing this, I noticed that the diesel engine output power display would jump to 0 when reaching throttle 0, before the tractive effort ramped down all the way. See [DieselEngine.cs](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs#L1041), there is a check for `Locomotive.ThrottlePercent > 0` which is no longer accurate for showing the correct power when ramping down to 0.
- Similarly, I also saw ammeters drop to 0 instantly with idle throttle, ignoring the ramp down. Again, that's because there are checks for `ThrottlePercent > 0` which assume no force is developed at 0% throttle. See [AMMETER/AMMETER_ABS](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5291), [LOAD_METER](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5339), and [TRACTION_BRAKING](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5378).
- Have not looked thoroughly so I suspect there are more places in the code that assume "ThrottlePercent <= 0 means no tractive effort" and will cause problems since that's not true anymore.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
```diff
@@ -619,193 +634,90 @@ protected override void UpdateControllers(float elapsedClockSeconds)
...
-                }
-                else
-                {
-                    DynamicBrakeForceN = 0f;
-                }
+                float targetForceN = GetAvailableTractionForceN(t);
+                float limitForceN = GetAvailableTractionForceN(maxthrottle);
+                float maxForceN = float.MaxValue;
+                if (limitForceN >= targetForceN)
+                    maxForceN = limitForceN;
```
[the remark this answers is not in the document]
cesarBLG: That won't work, because AFAIK it is possible that a higher throttle setting gives a smaller tractive force. If I unconditionally set `maxForceN` to the one given by `maxthrottle`, the force would be incorrectly limited.

For example, people may have set tractive force curves such that Throttle=50% is used for Series and Throttle=100% for Parallel, such that at low speeds Force(Series) > Force(Parallel) and at high speeds Force(Series) < Force(Parallel)

It's likely that other parts of the code assume that lower throttle means lower force, but at this part of the code it must stay like this to keep existing stock working as intended.

cesarBLG:
>     * When using `ORTSDelayTimeBeforeUpdating`, the throttle value shown on the F5 and Alt + F5 HUDs show the throttle setting _with_ the delay. I think most people would expect the HUD to show the controls immediately without delay.
> 
>     * Oddly enough, the same isn't true in dynamic braking, in that case only the F5 train driving HUD shows dynamic braking with a delay, while the Alt + F5 HUD is instant.

This is outside the scope of this PR, but definitely something to check for consistency. The same happens with cruise control: some places show the % of dynamic brake applied by the electronics, and others show the demand by the driver (e.g. they show 0% even when cruise control is braking). Also, when TCS is cutting power, F5 shows 0% of throttle regardless of the throttle setting. It's something easy to change, but it should be agreed. I think this is worth a forum thread.

>     * This may be considered a nitpick, but the ramp parameters taking force/power units directly and assuming the change is per one second is inconsistent with other places in the program where the given units must explicitly be _rate of change units_. There's no provisions right now for power or force rate of change units but it could be useful.

I noticed that too, it makes sense to add the units for consistency with other variables.

>     * While doing this, I noticed that the diesel engine output power display would jump to 0 when reaching throttle 0, before the tractive effort ramped down all the way. See [DieselEngine.cs](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs#L1041), there is a check for `Locomotive.ThrottlePercent > 0` which is no longer accurate for showing the correct power when ramping down to 0.
> 
>     * Similarly, I also saw ammeters drop to 0 instantly with idle throttle, ignoring the ramp down. Again, that's because there are checks for `ThrottlePercent > 0` which assume no force is developed at 0% throttle. See [AMMETER/AMMETER_ABS](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5291), [LOAD_METER](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5339), and [TRACTION_BRAKING](https://github.com/cesarBLG/openrails/blob/da4b28b35eeb02c04178a7c22c3886701414cdc8/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L5378).
> 
>     * Have not looked thoroughly so I suspect there are more places in the code that assume "ThrottlePercent <= 0 means no tractive effort" and will cause problems since that's not true anymore.

I'll check that.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: I uploaded some changes addressing this. The logic is somewhat complex though.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSDieselLocomotive.cs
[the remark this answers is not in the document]
SteelFill: Hmm so in the case that the targetForce is greater than the limitForce because the throttle setting is less than the apparent throttle setting but the force curves produce a higher force at the lower setting...maxForce remains at infinity which would allow the ramp to smoothly change to the new (higher) force when it would otherwise be limited...

That is certainly a chain of events. If I wanted to account for the impacts of the diesel engine on force ramping, I'd consider checking the engine _power_ rather than reverse engineering a force from the lookup tables. (It would be very nice to drop the need to use `ReverseThrottleRPMTab` as that sets bothersome restrictions on RPM vs throttle settings.) Simply limiting the force ramping to not exceed the engine power would be clean and widely applicable. But that idea would sacrifice existing stock as I know there are things out there which rely on overloading the engine power output.

Perhaps some manipulation could be done on init to derive the power we should allow as a function of RPM (which may be lower than engine power due to auxiliary loads, or higher due to unrealistic configuration) but that's just making something else more complicated for this to be simpler.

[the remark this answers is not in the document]
cesarBLG: I believe there are also some electric locomotives where tractive force decreases when throttle increases (depending on the speed), trying to model rheostatic control. Ideally we would have an entirely different tractive force calculation for that case, while keeping the existing code for chopper and AC. But still, we have to use this for legacy stock.

Maybe we could entirely avoid ApparentThrottleSetting by correctly adjusting `DieselPowerSupply.AvailableTractionPowerW` using DieselPowerTab, but I'm not entirely familiar with that.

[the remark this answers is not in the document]
SteelFill: Maybe. The idea I have in mind is to give the user the option to enter a table of auxiliary load/loss power ("AuxiliaryPowerTab"?) as a function of RPM in the diesel engine (aux loads that vary or don't depend on RPM could be added by other systems in the simulation). This load gets subtracted from the gross power already entered in the DieselPowerTab to produce the traction power allowed, and the power is prevented from going above that. If the user doesn't enter aux loads then they are estimated such that the locomotive achieves the same performance before and after the code change (in the case of my slug, that would mean giving it a negative auxiliary load so its "engine" produces enough power for it to move).

That idea is definitely outside the scope of force ramping.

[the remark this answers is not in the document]
cesarBLG: This could work by setting `AvailableTractionPowerW = DieselEngineOutputPowerW` in the DieselPowerSupply class, but I'm not sure about what will happen when ThrottleRPMTab is defined but DieselPowerTab is not.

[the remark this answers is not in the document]
SteelFill: You say that but that's actually something I'd expect to be able to achieve. Maybe not get maximum tractive effort instantly by _default_ (though I'm pretty sure that's what legacy content did before this PR), which would require automatically entering an ORTSTractiveForceRampUpRate based on the engine parameters if that parameter wasn't included.
But there are definitely locomotives out there, like early EMD GP and SD-series locos, that will nearly immediately adjust starting tractive effort in response to throttle setting. While max tractive effort can't be achieved with the engine at idle, notch 3 or so can be. But if the engine rpm is limiting the tractive effort curve to notch 1 then you are out of luck for a few seconds, and it's not as if you can construct tractive effort curves that allow anything different.

The real solution, as ever, is to simulate traction with something other than lookup tables.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2498,119 +2522,233 @@ protected virtual void UpdateControllers(float elapsedClockSeconds)
...
-                    else
+                }
+                if (targetPowerW < powerW && (targetPowerW == 0 ? rampZeroWpS : rampDownWpS) > 0)
+                {
+                    float maxChangeW = (targetPowerW == 0 ? rampZeroWpS : rampDownWpS) * elapsedClockSeconds;
+                    if (powerW - maxChangeW > targetPowerW)
                     {
-                        AbsTractionSpeedMpS = AbsSpeedMpS;
+                        targetPowerW = powerW - maxChangeW;
+                        targetForceN = Math.Max(targetForceN, Math.Min(forceN, targetPowerW / AbsTractionSpeedMpS));
```
SteelFill, line 2548: Overwriting `targetForceN` when considering the impact of power rate of change is causing unintended effects, particularly when the target force is 0. If the target force is 0, the power ramp check will see that the force is 0 and correctly use `rampZeroWpS` as the ramp rate.

However, `targetForceN` will then be overwritten to something other than 0 (at least, if the power output is still high enough). This prevents the subsequent force-ramp check from realizing that it should use `zampZeroNpS`, as the `targetForceN == 0` check has been changed to fail when we clearly want it to succeed still.

The easiest way to fix that is probably adding a temporary bool `bool zeroTarget = targetForceN == 0` at the very beginning of the function call, then replace all later checks for `targetForceN == 0` with this "zeroTarget", that way the change in the target doesn't throw off the ramp to zero rate.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -5305,17 +5443,17 @@ public virtual float GetDataOf(CabViewControl cvc)
                         if (LocomotiveAxles.Count > 0)
                         {
                             data = 0.0f;
-                            if (ThrottlePercent > 0)
+                            if (TractionForceN > 0)
```
SteelFill, line 5446: Using the traction force as the check does work better, but I am noticing ammeters decrease smoothly toward 0 when setting idle power, but then suddenly drop to 0 just before reaching 0 normally, which is jarring compared to the relatively smooth motion otherwise.

It seems that's down to the ammeter showing the _filtered_ tractive force, so ammeters will still be trying to display some nonzero value the moment the actual traction force hits 0, resulting in that sudden change. I'm not sure why the filtered force is used here (especially as dynamic brake indications aren't driven by a filter) so perhaps that behavior could be dropped, or the check here and in other cab view controls changed to `if (TractionForceN > 0 || FilteredTractiveForceN > 0)` not sure what the best option is.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
```diff
@@ -258,6 +259,8 @@ public override void Update(float elapsedClockSeconds)
                 DieselEngineMinRpm = 0;
             }
 
+            AvailableTractionPowerW = Math.Max(DieselEngineOutputPowerW - ElectricTrainSupplyPowerW, 0);
```
SteelFill, line 262: We'll have to be careful with this one. While limiting the traction power does preserve the laws of thermodynamics, the MSOR community has definitely created content which relies on breaking the laws of thermodynamics.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: I would be really happy to eliminate the filtered force from the cab controls. I thought that this would happen but I didn't tested how noticeable is this. The only reason for using filtered force was to allow a smooth changing of the ammeter in cab, but now this can be controlled with the ramp up/down rates, so I think we can remove it and make things simpler.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
cesarBLG: Sure, I think I'll add a parameter that either uses ApparentThrottleSetting or limits the power through the power supply

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -423,6 +424,19 @@ public float OdometerM
...
+        protected float TractionForceRampDownToZeroNpS=-1;
+        protected float TractionPowerRampUpWpS;
+        protected float TractionPowerRampDownWpS;
+        protected float TractionPowerRampDownToZeroWpS = -1;
+        protected float DynamicBrakeForceRampUpNpS;
+        protected float DynamicBrakeForceRampDownNpS;
+        protected float DynamicBrakeForceRampDownToZeroNpS=-1;
+        protected float DynamicBrakePowerRampUpWpS;
+        protected float DynamicBrakePowerRampDownWpS;
+        protected float DynamicBrakePowerRampDownToZeroWpS = -1;
```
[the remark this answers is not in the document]
cesarBLG: I overlooked it when I changed the power ramp up, but definitely it wasn't intentional, I'll fix it.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2498,119 +2514,236 @@ protected virtual void UpdateControllers(float elapsedClockSeconds)
...
+            // For diesel locomotives, also take into account the throttle setting associated to the current engine RPM
+            if (IsPlayerTrain && this is MSTSDieselLocomotive diesel && !diesel.TractiveForcePowerLimited) maxthrottle = Math.Min(maxthrottle, diesel.DieselEngines.ApparentThrottleSetting / 100.0f);
+            if (t > maxthrottle) t = maxthrottle;
+            t = MathHelper.Clamp(t, 0, 1);
 
-            // Calculate the total tractive force for the locomotive - ie Traction + Dynamic Braking force.
-            // Note typically only one of the above will only ever be non-zero at the one time.
-            // For flipped locomotives the force is "flipped" elsewhere, whereas dynamic brake force is "flipped" below by the direction of the speed.
-            if (DynamicBrakePercent > 0 && DynamicBrake && DynamicBrakeForceCurves != null && AbsSpeedMpS > 0)
+            if (maxthrottle > 0 && Direction != Direction.N && LocomotivePowerSupply.MainPowerSupplyOn)
```
SteelFill, line 2613: Things are working well, but another quibble I noticed is that transitioning from dynamic brakes to power will result in any remaining dynamic brake force being killed instantly, rather than waiting for the dynamic brake force to ramp to 0. An additional condition to check `DynamicBrakeForce <= 0` could be added to this if statement, and the `ThrottlePercent <= 0` condition removed from the [dynamic brake check](https://github.com/cesarBLG/openrails/blob/277dacdeefb70d21d42c3c9464e12b3d9d316d48/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L2654) to resolve that (at least, works on my end).

Note that the same is not true when transitioning from traction to dynamic braking, as the dynamic brake check already includes a check for `TractionForceN <= 0`.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2498,120 +2514,230 @@ protected virtual void UpdateControllers(float elapsedClockSeconds)
...
-                    DynamicBrakeForceN = f * (1 - PowerReduction);
-                    if (LocomotivePowerSupply.MaximumDynamicBrakePowerW > 0)
-                    {
-                        float maxPowerW = LocomotivePowerSupply.MaximumDynamicBrakePowerW * DynamicBrakePercent / 100 * (1 - PowerReduction);
-                        if (DynamicBrakeForceN * AbsTractionSpeedMpS > maxPowerW)
-                            DynamicBrakeForceN = maxPowerW / AbsTractionSpeedMpS;
-                    }
-                    TractiveForceN -= (SpeedMpS > 0 ? 1 : SpeedMpS < 0 ? -1 : Direction == Direction.Reverse ? -1 : 1) * DynamicBrakeForceN;                 
+                    float maxPowerW = supply.AvailableTractionPowerW;
+                    if (targetForceN * AbsTractionSpeedMpS > maxPowerW) maxForceN = maxPowerW / AbsTractionSpeedMpS;
```
SteelFill, line 2629: Enforcing the power limit only when the target force exceeds the power limit fails to capture situations where the engine power is the limiting factor when power is being reduced.
As in, if the throttle is at idle, the target force will be 0 so this check fails and the engine power won't be considered a limiting factor. If the engine RPM is set to reduce quickly, but the ramp down rates defined are slower than this, the result is the engine overloading.

I'm not sure there's any situation where the engine power limit should be skipped, as long as ORTSTractiveForceIsPowerLimited is being used.

cesarBLG: I'm not sure why I used `targetForceN * AbsTractionSpeedMpS`, it makes more sense to use `maxForceN * AbsTractionSpeedMpS`

SteelFill:
Ok, starting to throw this in more locomotives and it feels pretty solid. It'll take some time to roll this out over all my collection, mostly because it's hard to know exactly what values to use when equivalent values are almost never published. But even vibes-based setting of values is a dramatic improvement in the control of locomotive performance and the driving experience.

I was even able to somewhat properly depict a locomotive running at "constant" RPM (diesel engine model still enforces a 1 rpm gap between each table entry) for power generation, which is an achievement by OR standards.

It would be good to get this PR into more hands, but it must be conflicting with another PR-not sure which. Depending on which one, it might be appropriate to make this PR higher priority as it's in a pretty good state apart from needing more exposure.

cesarBLG:
My guess is #1057, it's modifying some lines that are contiguous to the ones modified here.

cesarBLG:
Thanks to @cjakeman for approving the trello card!

 @SteelFill I temporarily added the for-unstable label so we can test it in Unstable, and then we can merge it if there are no issues

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2127,34 +2164,10 @@ public override void Update(float elapsedClockSeconds)
...
-                if (!DynamicBrake)
-                {
-                    if (DynamicBrakeCommandStartTime + DynamicBrakeDelayS < Simulator.ClockTime)
-                    {
-                        DynamicBrake = true; // Engage
-                        if (IsLeadLocomotive() && DynamicBrakeController?.CurrentValue > 0)
-                            Simulator.Confirmer.ConfirmWithPerCent(CabControl.DynamicBrake, DynamicBrakeController.CurrentValue * 100);
-                    }
-                    else if (IsLeadLocomotive() && DynamicBrakeController?.CurrentValue > 0)
-                        Simulator.Confirmer.Confirm(CabControl.DynamicBrake, CabSetting.On); // Keeping status string on screen so user knows what's happening
```
SteelFill, line 2145: Noticed dynamic brake control confirmations weren't working quite right-looks like these need to be added back in somewhere else.

cesarBLG: I'll look at it. IIRC my motivation was to fix some issues related to cruise control. But now that I look at it, it seems that this is related to DynamicBrakeController.CurrentValue and not DynamicBrakePercent, so I'll check what was trying to do here.

cesarBLG: I restored the previous behavior. My goal was to actually confirm the current value of the controller, as is done for the rest of controllers. With the restored behavior, it shows "setup" regardless of the dynamic brake setting, which makes it more difficult to adjust it, but keeps compatibility.

SteelFill: Some changes could be made so that setup is shown with the controller % but that doesn't matter too much for this PR.

Source/Orts.Parsers.Msts/STFReader.cs
```diff
@@ -1170,6 +1180,23 @@ internal double ParseUnitSuffix(ref string constant, UNITS validUnits)
...
+            if ((validUnits & UNITS.ForceRate) > 0)
+                switch (suffix)
+                {
+                    case "": return 1.0;
+                    case "n/s": return 1;
+                    case "kn/s": return 1e3;
+                    case "lbf/s": return 4.44822162;
+                    case "lb/s": return 4.44822162;
+                }
+            if ((validUnits & UNITS.PowerRate) > 0)
```
SteelFill, line 1192: I did more experimenting with this latest version and found I couldn't get power rates to work...Looks like the limitations of the units enum are starting to bite, as `(validUnits & UNITS.PowerRate)` returns `-2147483648` when validUnits is UNITS.PowerRate, seems that gets casted to a signed value before doing the comparison. I think for now you can get away with changing the comparison to != 0, or casting the result of the bitwise operation to a (uint).
```suggestion
            if ((validUnits & UNITS.PowerRate) != 0)
```

cesarBLG: Yes, I thought that 1<<31 was moving to the 31th bit but it's actually moving to the 32nd, as 1<<0 already uses the first bit. I changed the enum to be ulong-based instead of int-based to temporary increase the number of different units to 63. Hope that works. Meanwhile, I will think about how to change the enum so we can still use Compulsory and Any without having to use one bit per unit, but that will go in a separate PR.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2452,30 +2460,38 @@ protected virtual void UpdateControllers(float elapsedClockSeconds)
...
+            var throttleCurrentNotch = ThrottleController.CurrentNotch;
+            ThrottleController.Update(elapsedClockSeconds);
+            if (CruiseControl != null && CruiseControl.ThrottlePercent != null)
+            {
+                LocalThrottlePercent = CruiseControl.ThrottlePercent.Value;
+            }
+            else
+            {
+                LocalThrottlePercent = ThrottleController.SavedValue * 100;
             }
```
SteelFill, line 2482: I swear I will stop finding things eventually...but for now, I tried all the parameters for the first time on a locomotive with animated levers in the 2D cab (I regularly run Streamlines/TrainSimulations content which does not have animated levers in the cab) and I discovered that, like parts of the HUD, cab view displays are also waiting to show throttle/dynamic brake changes until after the delay timer has elapsed. It's confusing to give a throttle input, only to see the throttle not move for a moment until it moves on its own (ghost in the machine, perhaps?), so I don't think this issue can be avoided.
I did observe a difference in behavior between the throttle handle and dynamic brake handle, but that's explained by the differences between line 2470 and line 2481. Either way, both were being interfered with by the delay timer.

Clearly some processes need a throttle/dynamic brake percent after the delay, but others should have it before the delay. In my opinion, the user facing stuff like HUDs and cab views should be told to use `MSTSNotchController.CurrentValue`, while all the actual physics stuff should use `MSTSNotchController.SavedValue`.

cesarBLG: I tested with a CP_HANDLE locomotive, which gets the information from a different place and actually uses CurrentValue, so that explains why I didn't observe that.

For me, it is clear that THROTTLE and DYNAMIC_BRAKE should get the information from CurrentValue. However, it's not so clear what to do for HUDs and displays. I'm fine with the current behaviour for HUDs, which use the Throttle/DynamicBrakePercent, as it also displays the % used by cruise control and blending. Regarding displays, there are two legitimate usages:
- Make the display follow the lever value. But the user could just use a THROTTLE without MouseControl instead.
- Make the display follow the actual demanded percent (including blending, cruise control, etc.). We could use THROTTLE_DISPLAY for that.
 
I uploaded the proposal where THROTTLE uses CurrentValue and THROTTLE_DISPLAY uses ThrottlePercent. What do you think?

SteelFill: That thought did cross my mind, it makes sense to me that a _DISPLAY should show the control setting the locomotive 'thinks' is active (after the delay), though I was concerned about the case where someone could use a _DISPLAY cab view control for a display that's mechanically linked to the throttle rather than electronic...but yes in that case it could just be a THROTTLE with no mouse control. I'll try the effect on something with both levers and displays.

SteelFill: Ok, looking at the latest version the throttle and throttle displays seem to work as intended, though some new (maybe some existing too?) unusual behaviors exist in dynamic braking. Setting the dynamic brake handle off results in _immediate_ killing of dynamic brake effort, ignoring the delay. Re-applying the dynamic brake in this state waits for the dynamic brake setup timer to elapse before braking effort returns, _ignoring the controller delay_, but then the dynamic brake effort returns to the same setting it was at before turning it off (assuming the control delay is longer than the dynamic brake setup time), not the new setting. That happens even if the dynamic brake handle has been left in the off position for _longer than the control delay_. Seems like the control delay stops updating if the dynamic brake is off.

Also, another thing (sorry), it seems the dynamic brake control interlock considers the 'saved' throttle setting, rather than the 'current' throttle setting (ie: can't move the dynamic brake handle until moving the throttle to idle _and_ waiting for the control delay). Such interlocking is generally physical, rather than electrical, so the dynamic brake handle should be unlocked when the throttle is idle even if power is still applied. The more problematic consequence of this is that the dynamic brake handle can be _moved even if the throttle handle is above idle_, so long as the throttle control delay has not elapsed yet. When the delay is over and power is applied, things behave oddly as the situation where dynamic braking and throttle are above 0 simultaneously isn't supported.

SteelFill: Alright so trying again the dynamic brake interlock feels right now, but I'm still experiencing the problem that the dynamic brake control delay doesn't update when the dynamic brake is set all the way off.

The dynamic brake doesn't instantly turn off when moving the handle off, except now it _never turns off_, the 'saved' setting remaining at the last saved value before it was turned off. This allows for an interesting state where you can apply throttle and get the engine to rev up (at least when driving a diesel), but you still get dynamic brake force. This would be unrecoverable but traction power returns on its own when dynamic brake force drops to 0 due to low speed.

Anyway, I think this is happening because the `TimeSinceLastChange` variable inside the `DynamicBrakeController` doesn't update when the dynamic brakes are off. If you look at [line 2437 of MSTSLocomotive.cs](https://github.com/cesarBLG/openrails/blob/1a62721986dd45a284c4e0f2eeb7e4611e7d4766/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L2437), the check to update dynamic brakes ignores the saved value. That means `DynamicBrakeController.Update(elapsedClockSeconds)` on [line 2442](https://github.com/cesarBLG/openrails/blob/1a62721986dd45a284c4e0f2eeb7e4611e7d4766/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs#L2442) isn't called when the dynamic brakes are active, but the handle is set off. If the update function isn't called, the TimeSinceLast change doesn't update, so the saved value never updates, and the dynamic brakes get stuck on.

I think the solution there is to change the line 2437 if statement from 

```C#
if (DynamicBrakeController != null && (DynamicBrakeController.CurrentValue > 0 || DynamicBrakeController.UpdateValue > 0))
```

to 

```C#
if (DynamicBrakeController != null && (DynamicBrakeController.SavedValue > 0 || DynamicBrakeController.CurrentValue > 0 || DynamicBrakeController.UpdateValue > 0))
```

Add an additional check for the saved value, that way the dynamic brakes are updated if the lever is above 0, the active value is above 0, or the controller is being updated (all 3 of which seem like good situations to trigger dynamic brake updating).

cesarBLG: Probably it would make sense to always update the dynamic brake controller, which is what we do for the rest of controllers. Still, your suggested solution should work and has less risk of breaking something else.

SteelFill (approved):
I think I have scrutinized the traction/dynamic force/power ramping and control delay enough, I don't have any more odd behavior to report. This feature is great for customizing the driving behavior of locomotives and I'm looking forward to using this on more of my content, as well as investigating some of the future opportunities provided by the refactoring done here.

cesarBLG:
Thanks for the thorough review!

PR #1049, opened 2025-02-25
Re-organise document storage and access

cjakeman:
Changes the Menu > Documents to list only documents which are useful when running Open Rails.
Other documents are moved to the website where they remain available through a link in the drop-down. These include:
- installation
- translating
- notifications
- content creation
- code-related
- architecture

The intended use of the \Source\Documentation folders is also documented.

Marked DRAFT and not-for-unstable until website also ready (https://github.com/openrails/openrails.org/pull/32 needs approval and merging).

6-4-2025 Now taken out of draft

PR #1068, opened 2025-03-29
Build for online-only documentation files

twpol:
To go with #1049


cjakeman (approved):
Hi James,

Thanks for these changes. Approved and merged.

Chris

PR #1069, opened 2025-04-04
Mouse wheel control only with Alt – Followup for #1051

pzgulyas:
A followup for #1051 by the requests [in this thread](https://www.elvastower.com/forums/index.php?/topic/37779-handling-throttle-brakes-etc-with-mouse-scrollwheel/page__view__findpost__p__317234)
Roadmap: https://trello.com/c/V6pIZGjM/580-add-mousewheel-support

### DIFF
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2340,9 +2340,10 @@ namespace Orts.Viewer3D.RollingStock
         /// </summary>
         float NormalizedMouseMovement()
         {
+            var mouseWheelChange = UserInput.IsDown(UserCommand.GameSwitchWithMouse) ? UserInput.MouseWheelChange : 0;
             return (ControlDiscrete.Orientation > 0
-                ? (float)(UserInput.MouseMoveY + UserInput.MouseWheelChange) / (float)Control.Height
-                : (float)(UserInput.MouseMoveX + UserInput.MouseWheelChange) / (float)Control.Width)
+                ? (float)(UserInput.MouseMoveY + mouseWheelChange) / (float)Control.Height
+                : (float)(UserInput.MouseMoveX + mouseWheelChange) / (float)Control.Width)
                 * (ControlDiscrete.Direction > 0 ? -1 : 1);
         }
 
--- a/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
+++ b/Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
@@ -2340,7 +2340,7 @@ namespace Orts.Viewer3D.RollingStock
         /// </summary>
         float NormalizedMouseMovement()
         {
-            var mouseWheelChange = UserInput.IsDown(UserCommand.GameSwitchWithMouse) ? UserInput.MouseWheelChange : 0;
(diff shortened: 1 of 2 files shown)


SteelFill (changes requested):
While this successfully allows zooming in exterior cam without needing to mess around, this still fails to give the intended behavior of only interacting with 3D cab views. Right now, holding Alt will erroneously disable zoom in exterior views for no reason, and holding Alt is required to manipulate controls with the mouse wheel in 2D cabs even though you can't zoom in 2D cabs. Either an override for ZoomByMouseWheel specifically for 3D cabs is needed or the conditions in the existing ZoomByMouseWheel need to check if the camera is in 3D cab mode.

Also, I am suspecting something is wrong with the core mouse wheel control behavior as I am finding it effectively impossible to use. On everything I've tried, just one click of the mouse wheel throws the control all the way to the max/min value. (For example, brake is released, scroll the mouse up one click, brake is instantly moved to emergency. Not helpful at all.)

pzgulyas:
> Also, I am suspecting something is wrong with the core mouse wheel control behavior as I am finding it effectively impossible to use. On everything I've tried, just one click of the mouse wheel throws the control all the way to the max/min value. (For example, brake is released, scroll the mouse up one click, brake is instantly moved to emergency. Not helpful at all.)

I've added a division by 5 to the movement to slow it down. I'm not sure it would be good to alter the behavior regarding the Alt button between the 2D and 3D cabs, so I left that part like it is for now. But one may now zoom in other views together with Alt too.

Source/RunActivity/Viewer3D/RollingStock/MSTSLocomotiveViewer.cs
```diff
@@ -2340,9 +2340,10 @@ public virtual int GetDrawIndex()
         /// </summary>
         float NormalizedMouseMovement()
         {
+            var mouseWheelChange = (UserInput.IsDown(UserCommand.GameSwitchWithMouse) ? UserInput.MouseWheelChange : 0) / 5;
```
SteelFill, line 2343: This is better but I'm still finding controls hard to manipulate precisely. An 8-notch throttle tends to change by 2 notches per mouse wheel click. I'd expect any control to move by no more than 1 notch per mouse wheel click to be the most intuitive.

pzgulyas: So what divider would you like to see instead of 5 here? To me the main value in this PR (or rather its predecessor that already got merged) is not the added mouse wheel control, but the refactoring part. The eliminating of the direct shape animation and hierarchy calls, so the whole thing can play more clearly with gltf. The mouse wheel support is a surplus only, mainly for being able to avoid creating new trello, bug report, blueprint and who knows what else entries, I don't care too much about it. :-)

SteelFill: You should see the amount of things I crammed into #1029...it's not bad for a 'follow up' PR to be an extension of the original, even if that means many different topics are covered.

Anyway, mouse wheel control is something I wanted to look at when it was first tried out, I even did some experimenting with the original (closed) PR, and I could probably say I'd approach it differently right now...if I wasn't trying to do a dozen other things. Obviously that's not important now, people just want to be able to zoom in exterior camera views. My vibe is that dividing by 10 (so 10 mouse wheel clicks to go over a control's entire range?) would be comfortable without reworking the whole darn thing when it wasn't even the main objective in the first place.

SteelFill (approved):
That'll do for now, this functionality is much more in line with how folks expect things to work.

PR #1070, opened 2025-04-07
Fix a NullReferenceException in TCS scripts

pzgulyas:
[Forum thread](https://www.elvastower.com/forums/index.php?/topic/38716-sigscr-prooblem)

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
@@ -670,9 +670,10 @@ namespace Orts.Simulation.RollingStocks.SubSystems
                     }
 
                     var functionHead = default(SignalHead);
-                    foreach (var head in trainSignal.SignalObject.SignalHeads)
-                        if (head.Function == function)
-                            functionHead = head;
+                    if (trainSignal?.SignalObject?.SignalHeads != null)
+                        foreach (var head in trainSignal.SignalObject.SignalHeads)
+                            if (head.Function == function)
+                                functionHead = head;
                     signalTypeName = functionHead.SignalTypeName;
                     foreach (var key in functionHead.signalType.DrawStates.Keys)
                     {
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
@@ -670,17 +670,19 @@ namespace Orts.Simulation.RollingStocks.SubSystems
                     }
 
                     var functionHead = default(SignalHead);
-                    if (trainSignal?.SignalObject?.SignalHeads != null)
-                        foreach (var head in trainSignal.SignalObject.SignalHeads)
-                            if (head.Function == function)
-                                functionHead = head;
+                    foreach (var head in trainSignal.SignalObject.SignalHeads)
+                        if (head.Function == function)
+                            functionHead = head;
                     signalTypeName = functionHead.SignalTypeName;
-                    foreach (var key in functionHead.signalType.DrawStates.Keys)
+                    if (functionHead?.signalType?.DrawStates != null)
                     {
-                        if (functionHead.signalType.DrawStates[key].Index == functionHead.draw_state)
+                        foreach (var key in functionHead.signalType.DrawStates.Keys)
                         {
-                            drawStateName = functionHead.signalType.DrawStates[key].Name;
-                            break;
+                            if (functionHead.signalType.DrawStates[key].Index == functionHead.draw_state)
+                            {
+                                drawStateName = functionHead.signalType.DrawStates[key].Name;
+                                break;
+                            }
                         }
                     }
                     textAspect = functionHead?.TextSignalAspect ?? "";


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
```diff
@@ -670,9 +670,10 @@ SignalFeatures NextGenericSignalFeatures(string signalFunctionTypeName, int item
                     }
 
                     var functionHead = default(SignalHead);
-                    foreach (var head in trainSignal.SignalObject.SignalHeads)
-                        if (head.Function == function)
-                            functionHead = head;
+                    if (trainSignal?.SignalObject?.SignalHeads != null)
```
cesarBLG, line 673: Adding a null check is just hiding the real problem. Those structures should never be null, so there is no point in adding null pointer checks. If they are null, that's due to a corrupted signalling in the route. Maybe we should be able to recover from that instead of crashing, but that should be done by ensuring the consistency of the signalling at initialization.

pzgulyas: Yeah, you are right, but... They admitted that the signal scripts may be wrong, but they are unable to debug it, and the original creator passed away. And I think it is never a good approach to run the game onto a nullreferenceexception. Yes, an initial check could be added, and the problem logged, but still, if the script is wrong, we should not run on a nullreferenceexception. That's why all the sript functions are written in a way to return at least an empty string, not a null, to avoid this.

cesarBLG: I agree that it's wrong to throw the exception. The thing is that it makes no sense to me that those structures are null. I think this started with the refactoring in #537, previous code was more tolerant to badly configured SIGCFGs. We should ensure that they are not null instead of adding null pointer checks everywhere, which makes the code much less readable. Maybe it is enough to check for null at initialization, and then default-initialize SignalHeads, SignalType and so on.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
```diff
@@ -674,12 +674,15 @@ SignalFeatures NextGenericSignalFeatures(string signalFunctionTypeName, int item
...
                     {
-                        if (functionHead.signalType.DrawStates[key].Index == functionHead.draw_state)
+                        foreach (var key in functionHead.signalType.DrawStates.Keys)
                         {
-                            drawStateName = functionHead.signalType.DrawStates[key].Name;
-                            break;
+                            if (functionHead.signalType.DrawStates[key].Index == functionHead.draw_state)
+                            {
+                                drawStateName = functionHead.signalType.DrawStates[key].Name;
+                                break;
```
pzgulyas, line 684: Looks like we are out of sync with the guy reporting the error, he still run the yesterday morning version. This is his real line of error, 677. And here it is possible to be null, because if in the previous foreach no head was found with the function, the signalType will be default, which is null.

pzgulyas: I've reverted my first commit, and added this instead.

cesarBLG: But functionHead can't be null, because if it were null the SignalObject wouldn't be in playerTrainSignalList.

pzgulyas: No, that is not null, but var functionHead = default(SignalHead), and if (head.Function == function) not found, then functionHead.signalType stays default, which is null.

cesarBLG: Yes, but if (head.Function == function) should be always found, because otherwise the signal wouldn't be present in playerTrainSignalList

cesarBLG: But if it's not a valid signal type, then 
if (!Locomotive.Train.signalRef.SignalFunctions.ContainsKey(signalFunctionTypeName))
applies, which handles that case.

If the signal type is valid, then we get all signals which have a head of that type, which is
var playerTrainSignalList = Locomotive.Train.PlayerTrainSignals[dir][function];
Therefore, all signals in playerTrainSignalList always have at leas one matching signal type

pzgulyas: Then another idea: Still the signalType may be null. The SignalHead has 2 constructors, and only one of them set the signalType. The other one requires that to be set via the SetSignalType() method. But not all paths there set this. There is also a warning:

Trace.TraceWarning("SignalObject trItem={0}, trackNode={1} has SignalHead with undefined SignalType {2}.",
              mainSignal.trItem, mainSignal.trackNode, sigItem.SignalType);

This would needed to be searched for in the log file, which unfortunately we don't have. But if it is there, that means the signalType is null.

cesarBLG: Fair enough, signalType can be null if there is a mismatch between TDB/world and SIGCFG. I agree with the null pointer check for signalType then. What I don't understand is why a similar code works in 1.5, but it really doesn't matter.

Anyway, the TDB from the original user seems to be quite out of sync. I agree that not crashing is still nicer.

cjakeman:
Hi Peter,

Thanks for this fix. Is this fix complete now?

Chris

pzgulyas:
I think this can be merged. @cesarBLG Is it all right from your side?

cesarBLG:
Hi Peter,

IIRC the bug still exists, there was a logfile from the unstable release where a crash is still present at line 676

`signalTypeName = functionHead.SignalTypeName;`

Unfortunately I wasn't left enough time to reply before this PR was merged, so this must go in a new PR now. It would have been nice to wait a little bit more time before merging so I could have replied. It's not a big deal though.

PR #1072, opened 2025-04-21
Content Creation Shortcuts & Advanced Wagon Shape Interactions

SteelFill:
This PR covers some ideas I've had to simplify certain steps of content creation and installation, especially things that have bothered me in particular. The goal is to give the option to automate certain frequently repeated, yet predictable, steps of the creation process, add alternate ways to achieve a result that take less effort than the current state of art, and even outright add new features to give additional control over content. At the end, this should provide the opportunity for fewer tedious steps of content creation and installation, so that more time can be spent on the interesting parts of development and playing.

These features won't be useful everywhere, but run-of-the-mill content such as a reskin of a boxcar is the perfect opportunity to use the features here.

[Discussion thread](https://www.elvastower.com/forums/index.php?/topic/38223-potential-features-to-simplify-engwag-creation/) (yes this started last year)
[Related Trello card](https://trello.com/c/tSHSbR36/606-simplify-repetitive-steps-of-train-creation)

### Current features for train cars:
- **Automatic calculation of Davis resistance values**, with A and B calculated from the weight of the rolling stock, number of axles, and bearing type. Each calculation is only run if an `ORTSBearingType` is defined, but the Davis value itself was missing. Davis C, if missing, is automatically calculated from the `ORTSFrontalArea` and `ORTSDavisDragConstant` if given. The calculations used rely on the original Davis formula and the CN resistance formula which produces results similar to the FCalc utility (but, of course, happen automatically saving time and file space normally dedicated to calculating and storing those Davis values). This also works with freight animations.
- **Automatic setting of the Size ( ) parameter based on the 3D model**, as proper settings of `Size ( )` are often based on the 3D model. Rather than opening the shape file in an external program and checking the bounding box, `ORTSAutoSize ( dx, dy, dz )` can be used to tell OR to inspect the main shape and attached freight animations and determine the width, height, and length of the rolling stock based on the min and max dimensions of the shape files. The 3 inputs `ORTSAutoSize` can take are _offsets_ from the actual model dimensions, so `ORTSAutoSize ( 0m, 0m, -0.18m )` makes the simulated wagon have the same width and height as the 3D model, but makes the length 18 centimeters shorter than the 3D model. That's useful for getting proper coupler spacing.
- **Automatic centering of the 3D model** by automatically determining the `CentreOfGravity` Z value needed to center the model. While coupler alignment can be fixed by inspecting the 3D model bounding box and calculating exactly how off-center it is, then manually setting `CentreOfGravity`, this is tedious, easy to get wrong, and it's easy to simply miss. Using similar methods to auto size, it's quite easy to automate with `ORTSAutoCenter ( 1 )`. This will overwrite any existing Z value for the CoG (the X and Y values are left unchanged) and insert the value that exactly centers the 3D model (or 0, if the model is already centered).
- **More center of gravity control** with the ability to enter center/centre (both spellings now accepted) of gravity values component by component using `ORTSCentreOfGravity_X`, `ORTSCentreOfGravity_Y`, and `ORTSCentreOfGravity_Z`. Values entered here will overwrite the corresponding value(s) entered in `CentreOfGravity` so long as the new parameter is placed lower down in the file. A notable improvement here is that `ORTSCentreOfGravity_Z` is _not_ limited to less than 2 meters, whereas the Z value given in `CentreOfGravity` is (which was done to improve compatibility with some MSTS locomotives). This will help in some specific cases where a 3D model needs to be shifted substantially forwards/backwards, but auto centering isn't appropriate (note that ORTSAutoCenter also bypasses the 2 meter limit on CoG.Z).
- **Manual definition of articulation** to handle some of the cases the automatic system isn't built for. `ORTSFrontArticulation ( 1 )` forcibly enables articulation at the front of the rail vehicle, `ORTSRearArticulation ( 1 )` does the same for the rear. Entering a value of 0 forcibly disables articulation. -1 would give the default, automatic behavior.
- **Refactor of most engine/wagon shape definitions** to allow giving a custom .sd file, instead of always using the .sd file with the same name as the .s file. This is generally achieved by entering the shape definition as normal, then adding an extra input giving the path/name of the .sd file after all the normal inputs. The only exception is MSTS freight animations, which already have an optional input. Use ORTS freight animations if you want custom .sd files. So `WagonShape ( dash9.s CustomFile.sd )` would load the "CustomFile.sd" shape descriptor rather than the "dash9.sd" shape descriptor. Additionally, if you have another engine/wagon using the same shape, but with a different .sd (like `WagonShape ( dash9.s MyReskin.sd )`, it will be treated as a separate shape entirely, allowing one "actual" shape file on the disk to be used as if it were multiple different shapes, saving file space. Note that relative file pathing works here, so the .eng/.wag, .s, and .sd can all be in entirely different folders-organize files however you like.

### Current features for .sd files:
- **New option to override textures inside the .sd file**, allowing for a shape to be reskinned without editing the original shape file. Inside the .sd file, add `ESD_ORTSTextureReplacement ( OriginalTexture.ace ReplacementTexture.ace )` to replace the texture definition for "OriginalTexture" inside the shape file with "ReplacementTexture.ace". While .ace is given in the example, .dds works too. Relative file pathing works here as well, so textures can be in different folders from the shape. Multiple textures can be replaced at once by adding additional statements, or adding multiple pairs of textures to one statement like `ESD_ORTSTextureReplacement ( OldTex1.ace NewTex1.dds OldTex2.ace NewTex2.dds )`. If one of the 'old' textures can't be found in the shape file, a warning is recorded and no changes will be made to the shape. **This feature is for the reskinners out there** as, when combined with the previous feature, it allows for implementing **reskins without requiring any shape file copying/pasting/renaming/uncompressing/editing**. For example, the .eng of a reskin could use `WagonShape ( "..\\DASH9\\dash9.s" "dash9reskin.sd" )` to reference the original shape file in its original location, but then overwrite the textures using the `dash9reskin.sd` file inside the current folder, so long as that .sd file specifies something to specify the new textures like `ESD_ORTSTextureReplacement ( us2dash9.ace dash9reskinA.dds us2dash92.ace dash9reskinB.dds )`.
- **Ability to modify position of shape sub objects** from inside the .sd file, without actually editing the shape. To do this, add `ESD_ORTSMatrixTranslation ( MATRIX x y z )` to the .sd file, where `MATRIX` is the name of the shape matrix that should be modified (matrix names can be found in utilities like shape viewer). If the shape doesn't have a matrix named MATRIX, a warning is added to the log and the change isn't applied. `x y z` are the offsets from the original position of the sub object in +right/-left, +up/-down, and +forward/-backward direction, respectively. The offset uses units of length, default is meters. Note that this, and the other similar MATRIX changes can only be applied once per matrix, but can be applied multiple times as long as its to a different matrix each time. Also, applying any transform to one matrix will cause all sub-objects _below it in the hierarchy_ to be transformed as well (the hierarchy can also be seen in utilities like shape viewer).
- **Ability to modify the scale of shape sub objects** from inside the .sd file. Similar to changing position, use `ESD_ORTSMatrixScale ( MATRIX x y z )` but this time `x y z` are the widthwise, heightwise, and lengthwise scale factors, respectively. Scale factors greater than 1 increase size, between 0 and 1 reduce size (these are units), and negative scale can be used to mirror a sub object. This can be especially useful for shapes which were built for the wrong track gauge.
- **Ability to rotate shape sub objects** from inside the .sd file. Similar setup to the other two, use `ESD_ORTSMatrixRotation ( MATRIX y p r )` where this time `y p r` are the yaw (+left), pitch (+up), and roll (+counter-clockwise) angle to rotate the MATRIX by. Units are radians by default, but the `deg` suffix can be used to give angles measured in degrees. Note that _rotation is applied last_ out of the 3 matrix transformations, which should reduce risk of model warping. Be careful when changing shape transforms as this can have far-reaching effects as many simulator systems rely on the 3D model data to function correctly.
- **Also, shape matrices can be renamed** from the .sd file without changing the shape file. Add `ESD_ORTSMatrixReanme ( OLDNAME NEWNAME )` to rename the matrix called OLDNAME to a name of NEWNAME. If the shape doesn't contain a matrix with OLDNAME then a warning will be added to the log and nothing will change. The name of the matrix is important for a few systems, especially animations, but sometimes matrices could be named incorrectly, so this allows that to be fixed without needing to uncompress and edit the shape file.
- **LOD distances can be changed** from the .sd file without changing the shape file. Add `ESD_ORTSLODOverride ( LODindex LODdistance )` to change the LOD with an index of LODindex to have a viewing distance of LODdistance (measured in distance units, default meters). Note that the LODindex starts at 0! 0 is the closest LOD, 1 is the second closes LOD, etc...
- **Shape hierarchy can be changed** from the .sd file without changing the shape file. `ESD_ORTSMatrixParent ( MATRIXNAME PARENTNAME )` changes the hierarchy parent of the matrix called MATRIXNAME to be the matrix called PARENTNAME. Changing the hierarchy could cause things like infinite loops, but error checking is done on the resulting hierarchy to prevent setting a broken hierarchy, with a warning added to the log if the settings are invalid (or if either named matrix can't be found). Note that changing a matrix's hierarchy parent may require changing the position/rotation of the matrix for it to appear in the expected location, as the position/rotation are measured relative to the hierarchy parent, not relative to the 3D model as a whole.
- **Existing shaders can be replaced with a different type of shader** from the .sd file, using `ESD_ORTSShaderReplacement ( id ShaderName )` where id is the index of the shader in the `shader_names` list of the shape file (as an index, the first `named_shader` listed is actually shader 0, not shader 1) and ShaderName is the name of the shader, case-sensitive (valid shader names are `Tex`, `TexDiff`, `BlendATex`, `BlendATexDiff`, `AddATex`, and `AddATexDiff`, experienced creators will be familiar with all those shaders). The intended use for this is to enable/disable transparency on shapes without the need to decompress the shape. If the id or ShaderName is invalid for any reason, warnings will show up in the log.
- **Specific parts of a shape can be hidden from the GPU** using the .sd file and the parameter `ESD_ORTSObjectVisibility ( MATRIXNAME 0/1 )`. Like other parameters, MATRIXNAME refers to a named matrix in the shape hierarchy, but this time the second input will tell the GPU to render that matrix if the input is nonzero (this is the default behavior), and if the second input is 0 the GPU will not render anything directly attached to that matrix. The parent and children of that matrix will still be rendered, only the single specific matrix mentioned won't appear on the screen. The matrix will still be simulated though, so this can save some GPU power if shape objects don't need to be rendered, but it won't save any CPU.

### Bonus features (technically related to the main features, but aren't creation shortcuts):
- Refactor of most systems that attach to engine/wagon shapes to **allow attachment to arbitrary sub-objects with arbitrary offsets**. This includes lights (this behavior was already implemented in #917), particle emitters, freight animations, sounds, and interior views. This was done to properly handle the transformations that can now be applied to shapes in the .sd file.
- To attach these addons to different sub-objects use `ORTSShapeHierarchy ( MATRIXNAME )` in a `Light (` block (existing feature), `Stream (` block (in an .sms file), `ORTS3DCab (` `ORTSAlternate3DCabViewPoint (` `Inside (` `ORTSAlternatePassengerViewPoint (` blocks for 3D interiors, `FreightAnimStatic (` `FreightAnimContinuous (` blocks for ORTS freight animations (MSTS freight animations are not supported), `HeadOut (` block for head out views (note that ORTS specific parameter in the HeadOut ( block **must come AFTER the x y z position numbers**!), in any **particle FX** subblock (again, the ORTS specific parameters must be placed **AFTER the position, direction, and radius numbers**!), and in the `Tr_CabViewFile (` header of a .cvf file (for a .cvf file, `ORTSShapeHierarchy` must be included _once for each view point_, which also means each viewpoint can have a different attachment). As with previous implementations, this will attach the addon to the sub object with the matrix called "MATRIXNAME", resulting in the addon translating and rotating with the sub object instead of with the main shape. Any parameters relating to positional offset of the addon will be measured relative to the position of the sub object instead of the main object. If the named matrix doesn't exist in the wagon shape file, a warning appears in the log and the addon will attach to the main object instead. Similarly, if the shape hierarchy parameter is missing, the addon will attach to the main object as well.
- A lot of addons were also missing the ability to reposition the location of the addon relative to the sub object to which it is attached. For a `Stream (` in the .sms file, use `ORTSPosition ( x y z )` to define the position of the stream's sound emitter, `ORTS3DCab (` `ORTSAlternate3DCabViewPoint (` `Inside (` `ORTSAlternatePassengerViewPoint (` blocks all now accept an `ORTSShapeOffset ( x y z )` input to allow shifting the position of the interior shape file (note that **this will NOT shift the view point**, make sure to update the view point location if shape offset is used), `FreightAnimContinuous (` blocks now accept the `Offset ( x y z )` and `Flip ()` parameters, which function the same as the counterparts in a static freight animation
- In an `FreightAnimStatic (` or `FreightAnimContinuous (` block, `ReplaceObject ()` can be added to make the freight animation replace the main shape object it is attached to. This is best used with `ShapeHierarchy` to attach a freight animation to a specific sub object, then "delete" (in reality, the sub object is simply not sent to the GPU, but is otherwise still present and simulated) that sub object from the main shape and replace it with the freight animation. By default, freight animations will not replace objects and will simply add on to existing objects.

### Future ideas:

- Automatic configuration of dynamic cargo weight for locomotives, at least diesels, where the volume and density of consumable resources is obvious enough to not require tedious manual data entry. (This might come in a future PR, after #1057 is merged due to the changes made to dynamic weights there)
- ShapeHierarchy for animated couplers (better support for articulated trains) (turns out this is not easy due to the way animated couplers are implemented, this will be a separate PR)

Should be clear that many of these things are related to shape files specifically. Manipulating shapes has historically been one of the worst aspects of participating in content manipulation so there's a lot to gain there. I'd love if every end user could avoid uncompressing a shape file without losing any capabilities.


KS44HUN:
Hi SteelFill,

Thanks for the extensive work on this PR. It's clear a lot of thought went into automating and simplifying parts of content creation, and I agree with the spirit of trying to reduce repetitive, manual work. 

However, there is one fundamental problem that cannot be ignored:
**This whole approach is only tailored to US, British, and Australian routes and vehicles.**

And this is a serious problem and a trend, because Open Rails should not only be about the railways of the continents and countries mentioned above. There is a very good friction calculator that can give more accurate, realistic values. especially when it comes to non-US vehicles. Your insight here could be a blind spot for European rail vehicles and European common rail systems. 

For example, articulated trainsets with Jacobs bogies (such as FLIRT, KISS, or Desiro Classic DMU/EMU units) are likely to be incorrectly handled using your current logic. These require precise control over pivot points and coupling distances that cannot be generalized. Automation without proper override mechanisms can do more harm than good in these cases.

When it comes to rolling resistance, there’s already a trusted friction calculator in use, particularly among European developers. Automatically estimating Davis values risks overwriting carefully tuned numbers with generalized, potentially inaccurate results. That can lead to misleading behavior in simulation and inconsistent experiences across regions per railway car.

I assume not only I, but I think others as well, that you test not only this, but everything else, exclusively on US tracks, with US vehicles. `This is a fundamental problem.` I understand that the US section of Open Rails is also under constant development, taking into account several factors towards improvement, but it should be remembered that this is not only a US-specific program. 

------

That said, I do think some of these tools could be genuinely useful, especially considering the reality of poorly configured rolling stock in certain regions. It might help clean up a lot of the mess. But the key here is choice.

There absolutely needs to be a way to enable or disable these systems, ideally not only from the in-game settings menus (which is the "lazy" option), but also through a clear, developer-facing method such as entries in the `.sd` file. That would strike a balance between ease of use and proper control, and make these features more acceptable for a wide range of use cases. If this flexibility is introduced, I believe many developers (myself included) would be open to trying it out.

------

In summary:

This PR is driven by good intentions, but its implementation is currently too narrow in scope. These features can only be genuinely useful if they are optional, transparent, and adaptable to different rolling stock architectures across the world.

Without that, these "smart" features will not bring universal benefits. Instead, they’ll help some users save time, while forcing others to correct errors, misalignments, and inaccurate behavior, especially in non-American contexts.

This is not progress. Without proper flexibility, this is a step backward.

Best regards,
KS44HUN
_(Just call me Krisztián or Krisz_)

SteelFill:
> Hi SteelFill,
> 
> Snip
> 
> This is not progress. Without proper flexibility, this is a step backward.
> 
> Best regards, KS44HUN _(Just call me Krisztián or Krisz_)

Sorry, but you are simply wrong, the features **_are already optional_**. As the documentation states (please read the documentation), friction values are only automatically calculated when _nothing has been entered except the ORTSBearingType_, the size isn't automatically calculated _except if ORTSAutoSize is manually added in_, and that CoG offset isn't overwritten _unless ORTSAutoCenter has been added_. The current behaviors (manually defining friction, size, CoG offset) are all still available, **old content will remain entirely unchanged** unless manually tampered with, and new content **should** use the manual methods when that makes sense (even in my content, I still manually define the size and CoG for articulated intermodal equipment).
The only edge case where something might change on old content is when someone defined ORTSBearingType but didn't define some Davis values, in which case the missing Davis value(s) will be inserted, but that can be considered error handling as nobody should have made content with a bearing type but without Davis values. Even when I get around to the whole shape file data injection thing, the _status quo will still be maintained_. The size, offset, rotation, textures, LODs, transparency, etc of the shape file will remain entirely unchanged unless extra data is _explicitly **added**_ to control those aspects of the system. So please, don't imagine then complain about problems that don't exist.

And in any case, it is entirely unreasonable to even attempt to create a system that works universally for literally everything as the implementation of such a system would be so complex and would require so much additional input from the user that they might as well enter the information manually-it would be the same amount of work for the user and much more work for development. And, of course, even when you do try to account for everything someone comes along 3 weeks after the PR was merged (seemingly missing that the PR was in testing for 4 weeks) and is baffled "why doesn't this work for my one-off locomotive and rail line". Ask me how I know.

peternewell:
A few points points -

i) I think that it would be good to have a message indicating when a "default" value has been calculated, and advising the value calculated. This can be set as a verbose message so that the messages can be turned off by the user if desired.

ii) Steam locomotives had mostly disappeared by 1992, so the variations to the Davis formula from that point on are probably not relevant to them. The earlier formulas allowed for mechanical resistance of the rods, etc which wasn't required for diesels in 1992. So some allowance for mechanical resistance maybe required. Also the resistance for the locomotive and tender were calculated as a combined value, and then "pro-rated" across the two units.

iii) The 1926 formula had a variation for light stock, has this been included?

iv) In the manual notes it should be clearly explained that _ideally users should enter their own Davis values_, as that will provide the most accurate and flexible approach. And thus custom values from other railway systems will override the "default" values. These changes are to only provide a default value where there are no figures provided by the user.

SteelFill:
I'm not against logging the auto-calculated values when in verbose mode, that would actually help with my debugging as I test out the features on my collection and make sure that automatic calculations are within error of those I already did manually.

As for resistance, I would like to do as much as I reasonably can with the information available in a rolling stock file otherwise devoid of information. There will be some situations where the file doesn't quite give enough information, but I could add the steam loco rod resistance, so long as the weight on drivers has been correctly entered (at least we actually have a parameter for this), and I suppose the lightweight variant of Davis wouldn't be too difficult to add.

But for all those edge cases, I already put a cop-out in the manual that basically says to not trust the calculation for steam locomotives, multiple units, nor high speed trains. The intent is mostly for wagons since it gets very tedious to enter hundreds of davis values for hundreds of wagons. (Ask me how I know...)

peternewell:
> But for all those edge cases, I already put a cop-out in the manual that basically says to not trust the calculation for steam locomotives, multiple units, nor high speed trains. The intent is mostly for wagons since it gets very tedious to enter hundreds of davis values for hundreds of wagons. (Ask me how I know...)

Whilst I can see that you have made some comments in the manual, these are not always read and hence this information can be overlooked. The more visible the information, the more likely that it will be taken in.

The verbose message could perhaps reinforce the fact that they have only been applied as a default.

Perhaps the message could be something like this -

"CarID aaa - No user information has been entered for Davis Resistance. Default values of A = xx, B = yy, and C = zz have been applied. Please enter user specific values if not accurate."

cjakeman:
> [Discussion thread](https://www.elvastower.com/forums/index.php?/topic/38223-potential-features-to-simplify-engwag-creation/) (yes this started last year)

HI Phillip, would you mind creating a Trello card for this PR in line with "[Contributing to Open Rails](https://github.com/openrails/openrails/blob/master/Docs/Contributing.md)". 

Thanks.

SteelFill:
**TL;DR:** To make these features work, I had to do 80% of the work required for the commonly requested feature of attaching things like freight animations, particle effects, and sounds to sub-objects other than the main object (I already did this for lights, and folks wanted to see the same for other systems), so I'm going to expand the scope a bit and include those features here to save time.

For the details...over the last couple of weeks, I have been working on the .sd file overrides. I have successfully produced a way for most shape definitions in engines/wagons to accept a second input for an .sd file to use (if the input is missing, as will be the case on all existing equipment, it is assumed the .sd file should have the same name & folder location as the .s file, which has always been the case) and allowing this input to be given to the shared shape manager so that _one .s file can be used with multiple different .sd files_ with each different .s and .sd pairing producing a different SharedShape object. Then on initialization the shape sees the .sd file (custom or default) and modifies information as specified by the .sd file. Currently not sure if it will be feasible to allow the same for scenery objects, as that would involve messing with world files.

I have already implemented the texture swap, matrix shift/scale/rotate, and matrix renaming, and additional .sd features will be easier to add than these first few. But allowing people to change transform matrices completely annihilates one assumption made repeatedly throughout the program: _'the MAIN object of the train car shape has no transform'_. The result of this is that applying a transform to the MAIN object would correctly move/scale/rotate the entire train car, but _wouldn't move freight animations_, wouldn't move sounds, wouldn't move the cab view, things like that.

My work over the last week has been pretty far-reaching to remove that assumption, and one consequence of that is allowing everything to establish a positional offset relative to the train shape, and to change which sub-object it's attached to. This involves changing a lot of files, which grows the reach of this PR but was bound to happen. One of the most exciting possibilities in this situation is **3D audio**; the program can now (on my end, haven't finished yet so no commit) accept an offset for a sound source, so you can pick where on the 3D model a sound stream comes from (think of making the horn come from the exact x, y, z position of the 3D horn object) and can attach sound sources to any sub-object you like (so wheel noise could actually come from the wheels and move as the wheels move with the bogie).

Similar is now (or soon enough) possible for freight animations, interiors, and particle effects. While this is technically outside the intended scope of the PR (this straight up is new features, not a shortcut for an existing process), the only thing that needs to be done is adding user inputs to the parser, some error detection, and documentation. That would take such little time that making a separate PR would be a waste.

SteelFill:
This is now feature complete, at least given the ideas I have right now. However, I believe there are conflicts with both #1057 and #570. I will use the for-unstable label to get some testing in with this PR as it's quite substantial and closer to complete, but if other PRs need more testing before this is merged then for-unstable can be removed.

cesarBLG:
Hi Philip,

I have been told that this PR is causing a problem with the engine sounds. When exiting the cab and then re-entering, some sounds stop playing. I'll try to get more details about the specific conditions.

cesarBLG:
We have been able to narrow down the issue: it seems to affect all _mono_ sounds in the 2D cab, when the cab view is exited and entered again. My guess is that the sound source position is not properly restored when entering the 2D cab.

SteelFill:
> We have been able to narrow down the issue: it seems to affect all _mono_ sounds in the 2D cab, when the cab view is exited and entered again. My guess is that the sound source position is not properly restored when entering the 2D cab.

That explains why I didn't notice this, it seems best practice is to use stereo sounds internally and mono sounds externally. Frankly, that makes sense given that 2D cab sounds can't rely on 3D positions to drive left/right sound, and go figure anything related to positioning sounds would break that.

I did have to 'cheat' slightly by editing a cab sound .sms file to use exterior sounds as I didn't have the time to search for something that used mono sounds in the cab by default, but that's functionally equivalent. I figure I can diagnose this over the weekend.

SteelFill:
It seems that Car.SoundSourceIDs is not a list of all sound sources attached to a car, it's only the list of sound sources that need a 3D position. The previous implementation of sound positioning only located sounds which were in the list of Car.SoundSourceIDs, meaning that interior sounds would remain located at 0,0,0. This wasn't immediately obvious, and the new implementation didn't need to use Car. SoundSourceIDs, so the position of all sounds was being set, even for the ones that shouldn't use 3D positioning.

However the audio engine is configured, it seems that we expect all 2D cab sounds to have a position of 0,0,0, so providing a position relative to the tile causes the sound to be treated as if it's absurdly far away. The fix is quite trivial, every sound source has and `Ignore3D` variable to indicate if it should not consider 3D position, and this includes interior sounds.

As for why stereo sounds work and why mono sounds work before going into an exterior view, it looks like something just ignores the position of stereo sounds and the sound-space position of the locomotive isn't updated from 0,0,0 until entering an external view so all the sounds started with a position of 0,0,0 by coincidence.

cesarBLG:
Seems to work now, thanks!

twpol:
There's a lot of interesting and potentially useful features in this PR, however, it is very large and it is not clear to me that all the features have the same usefulness for being built-in features.

In particular, it is worth remembering that every built-in feature has to be continuously maintained and supported indefinitely. We also have a contributed area specifically for unofficial but included tools, which would be a great place to include tools implementing the features not built-in.

Additionally, some of the features seem to cross boundaries that should be preserved as much as possible, mainly that the simulator does not depend on graphical data files like shapes. I am aware we already have some exceptions to this but that does not negate the need to keep as much separation going as we can, not least because it avoids the double loading of content.

I would love to see nearly every feature listed to be a separate PR (some are obviously very tied together), so that we can independently discuss and decide on approach and inclusion.

Anyway, here is how I would rate each identifiable feature according to how useful it appears to me to be built-in at this time. You are welcome to discuss all of this, but please keep the above comments in mind.

## Definitely good to be built-in

  - Manual definition of articulation
  - Refactor of most engine/wagon shape definitions to allow giving a custom .sd file
  - New option to override textures inside the .sd file
  - Allow most systems that attach to engine/wagon shapes to attach to arbitrary sub-objects with arbitrary offsets

## Probably good to be built-in

  - Automatic calculation of Davis resistance values
  - More center of gravity control [note: the separate components seems like a bad idea to me]

## Probably bad to be built-in

  - Ability to modify position of shape sub objects
  - Ability to modify the scale of shape sub objects
  - Ability to rotate shape sub objects
  - LOD distances can be changed
  - Existing shaders can be replaced with a different type of shader
  - Specific parts of a shape can be hidden from the GPU

## Definitely bad to be built-in

  - Automatic setting of the Size ( ) parameter based on the 3D model
  - Automatic centering of the 3D model
  - Also, shape matrices can be renamed
  - Shape hierarchy can be changed

PR #1073, opened 2025-04-22
Allow depart early

Roeterdink:
Timetable update : 
- Allow depart early (with or without set departure time)
- Request Stop (with or without set departure time)
- Passing Time (for player train only)
See blueprint https://blueprints.launchpad.net/or/+spec/tt-mode-additions


Roeterdink:
PR withdrawn for now to sort out merge errors.

PR #1077, opened 2025-04-25
Fix SME brakes not working

cesarBLG:


cesarBLG:
@peternewell It looks that the fix for SME brakes was quite easy. I tested and it's working on my side.

peternewell (approved):
Cesar,

It seems to now be working.

Thanks for following this up.

PR #1079, opened 2025-04-26
Adds Polish transations

cjakeman:
Adds Polish as a new language for Open Rails.

image

Translations provided by Robert Otremba at [https://github.com/RobertOtremba/Polish-language-for-Open-Rails ](https://github.com/RobertOtremba/Polish-language-for-Open-Rails )


twpol (dismissed):
Looks like the POT files have conflicts; I am not sure if we've documented this anywhere, but I suspect things would work better if the POT files are kept separate from PO files in PRs (e.g. one PR to update just POTs and then one PR per language with just POs)

cjakeman:
> Looks like the POT files have conflicts; I am not sure if we've documented this anywhere, but I suspect things would work better if the POT files are kept separate from PO files in PRs (e.g. one PR to update just POTs and then one PR per language with just POs)

Thanks.
I've resolved these - purely datestamps clashing.
You were right - there was no need to include the POT files in the PR.

twpol (approved):

PR #1081, opened 2025-04-27
Brake cuts power unification

cesarBLG:
This is a proposed solution to overcome the original compatibility issues with #527

It moves the DoesBrakeCutPower handling to the brake system, but the ultimate responsible of cutting power is the TCS (e.g. to restore traction only after pressing a button, setting throttle to 0, etc.).

Compatibility is preserved by taking into account that scripts are using

`DoesBrakeCutPower() && BrakeCutsPowerAtBrakeCylinderPressureBar() < LocomotiveBrakeCylinderPressureBar()`

I deprecated BrakeCutsPowerAtBrakeCylinderPressureBar, and now it is returning either -infinity or +infinity depending on whether traction cut-off is requested by the brake system.

https://blueprints.launchpad.net/or/+spec/brake-cuts-power-parameters

Context:
https://www.elvastower.com/forums/index.php?/topic/35678-ortsdoesvacuumbrakecutpower-is-no-longer-working/
https://www.elvastower.com/forums/index.php?/topic/38757-brakepipetokens-not-working-anymore-twinpipe/

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -5249,12 +5249,16 @@ the tables below.
 .. index::
    single: DoesBrakeCutPower
    single: BrakeCutsPowerAtBrakeCylinderPressure
+   single: ORTSBrakeCutsPowerAtBrakePipePressure
+   single: ORTSBrakeRestoresPowerAtBrakePipePressure
 
-Two other parameters in the Engine section of the ENG file are used by the TCS:
+Other parameters in the Engine section of the ENG file are used by the braking system to request traction cut-off to the TCS:
 
-- ``DoesBrakeCutPower( x )`` sets whether applying brake on the locomotive cuts the traction (1 for enabled, 0 for disabled)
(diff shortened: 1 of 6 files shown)


twpol (commented):
Although we'll have to keep an eye out for any issues as always, the proposed compatibility looks good to me 👍 

[ORMT approved]

cesarBLG:
Using the default implementation is not mandatory. You can just ignore the BrakeSystemTractionAuthorization and implement it entirely on the TCS script. However, as you said, the implementation depends on the brake system, so it looks sensible to me to implement it in the brake system. This ensures that if we introduce a new brake system, the traction-braking interlocking works for old scripts (e.g. for pure EP brakes, BrakeLine4 must be checked instead of the brake pipe).

Of course, I could implement the BrakeSystemTractionAuthorization variable inside the ScriptedTCS class, but from the scripting point of view it wouldn't change anything. We would need to add checks such as `if Car.BrakeSystem is AirSinglePipe` to cope with all possibilities, but I find it cleaner if we implement it inside the brake system class.

PR #1084, opened 2025-05-03
Timetable: Calculate Delay While Stopped at a Station

TheGwyd:
### Issue
In timetable mode trains do not recalculate their delay when stopped at stations. Trains futher up the line may have $wait, $follow, or $connect commands referencing a train stopped at a signal in a platform with a /maxdelay qualifyer.
[Thread](https://www.elvastower.com/forums/index.php?/topic/38801-issue-with-delay-calculations/)

### Desired Behaviour
Trains stopped at stations have their delay recalculated if they are stopped past their departure time.

### Possible Problems
I have added a delay calculation for trains at stations, regardless of whether they are leaving or not. The previous code explicitly only calculated delay on departure. I cannot think of a reason why only calculating delay on departure would be neccessary, but I would welcome other people's thoughts on any issues this could bring up.


cjakeman:
Hi Gwyd,

Congratulations on submitting this first PR. I've given it the label "for-unstable" so that it will automatically be included in the Unstable Version of Open Rails. We'll leave it there for a couple of weeks to see if it causes anyone else problems.

I wasn't sure if this is a bug fix or an enhancement.

Thanks and best wishes,

Chris

cjakeman:
Hi Gwyd,

It's been in Unstable for a couple of weeks, so I am happy to approve it.

However the "master" branch has been tweaked since you committed this PR.
Please would you resolve the conflict (you can  use the option in GitHub to do that on-line), so that I can complete the approval process.

Thanks and best wishes,

Chris

cjakeman (approved):
All done.

Approved and merged with thanks,

Chris

PR #1085, opened 2025-05-05
updates key commands for Train Operations window and also Daylight Offset

cjakeman:
And also a typo


SteelFill (approved):
Good catches, I didn't even know about the Alt+Shift+=/- option, that one will be useful to know. The spelling mistake in the brake friction area has caused some trouble, hopefully this is the end of that (part of me knows people will find forum posts with the old spelling and will be surprised when it doesn't work). And I guess it was simply forgotten to change the manual when the new train car operations window became official, that was certainly out of date.

PR #1088, opened 2025-05-17
Preserve previous log file.

rwf-rr:
The case that triggered this was that I started a second instance of Open Rails, and it failed because of the log file already being open. But I have also accidentally run Open Rails again, overwriting the log that I wanted to look at.

This keeps one previous log file (or redirects the log of an instance already running). 

When the log file name contains unique elements (eg. date), then there is no matching existing log (to move or delete).

Also renamed a variable to reflect that it is the path.


rwf-rr:
> This adds yet another file to user's desktop by default. 

I was a bit overeager with this PR. I also do not like files on my desktop. I could add a user setting to control this, but there already is an option to create unique log file names (there is just not GUI yet).

I will **cancel** (close) this PR.

PR #1089, opened 2025-05-20
Fix Superelevation on Rigid Frame Rolling Stock Again

SteelFill:
Follow up to #1029

Bug report: https://www.elvastower.com/forums/index.php?/topic/38542-problem-with-superelevation-and-two-axle-rolling-stock/page__view__findpost__p__318520

I previously fixed superelevation on two axle rolling stock, but a slight mistake meant it was only a partial fix. Also updated the method used to position two axle rolling stock during superelevation to solve instances of wheels sinking into rails on superelevated curves.


cesarBLG (approved):
Seems to be fixed according to the forum, thanks!

PR #1090, opened 2025-05-21
Add null check to avoid crash on corrupt signal files

cesarBLG:
Follow-up for #1070

If the signalling is corrupted, it is possible to get a signal added to the NORMAL signal list without actually having a NORMAL head, causing a null pointer exception.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/TrainControlSystem.cs
@@ -673,6 +673,8 @@ namespace Orts.Simulation.RollingStocks.SubSystems
                     foreach (var head in trainSignal.SignalObject.SignalHeads)
                         if (head.Function == function)
                             functionHead = head;
+                    if (functionHead != null)
+                        goto Exit;
                     signalTypeName = functionHead.SignalTypeName;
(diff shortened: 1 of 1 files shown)


cesarBLG:
This was initially reported at https://www.elvastower.com/forums/index.php?/topic/38716-sigscr-prooblem but PR #1070 was merged before I had time to say that the bug hadn't been fixed yet.

Definitely logging a warning would be useful, but not here. The TCS is merely consuming a corrupted data structure (functionHead should never be null: all the signals in the list should have one SignalHead from the requested type). 

I believe the bug happens specifically for NORMAL signals, but only in Activity or Timetable mode (i.e. not in Explorer mode). This is because the information for NORMAL signals is extracted from a different place in such case: https://github.com/openrails/openrails/blob/58ff988ed30f3e626ec0ac692e182e46e7cad03e/Source/Orts.Simulation/Simulation/Physics/Train.cs#L14778

My guess is that there is a signal in `SignalObjectItems` that qualifies as a NORMAL signal even though it doesn't have a NORMAL head. Not sure if this is caused by a bug in OR, or due to mismatch between SIGCFG, World and TDB files.

In my opinion, the bug should be solved by ensuring the consistency of SignalObjectItems and/or PlayerTrainSignals, but as this bug seems to happen only for specific routes, it's hard to debug. This fix at least solves the crash. The information is not used in many more places anyway, so solving the consistency bug is low priority.

PR #1091, opened 2025-05-21
Automatic speed control

cesarBLG:
Several improvements to the cruise control subsystem, including:

- Interface with the TCS to implement Automatic Train Operation. Exposed parameters are `SetSpeedMpS` and `SetSpeedAccelerationMpSS` which determine the permitted speed and deceleration of the braking curve.
- Use a linear conversion from 'delta' speed (difference between set speed and current train speed) to deceleration. The legacy algorithm (which is kept as default to maintain compatibility) is non-linear and causes oscillations for small deltas (small changes in speed result in big changes in throttle demand).
- Simplified the code

https://blueprints.launchpad.net/or/+spec/cruise-control-extensions

https://www.elvastower.com/forums/index.php?/topic/33841-cruise-control/page__view__findpost__p__318664


cesarBLG:
Thanks for reporting Carlo. I'm travelling and I don't have access to my PC, I'll fix it as soon as I come back.

I think it's not wise to have non-symmetrical save/restore for the NotchController, which differs from the rest of the save/restore methods, so I'll take a look.

cesarBLG:
I uploaded a fix. However, It looks like all the controller save/restore process is unnecessarily complicated. I don't know why it was done in such a way, but I'll try to simplify it.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/CruiseControl.cs
```diff
@@ -421,8 +486,84 @@ public void Initialize()
...
+                var notches = new List<MSTSNotch>();
+                if (SpeedSelectorIsDiscrete)
+                {
+                    if (!DisableZeroSelectedSpeedStep) notches.Add(new MSTSNotch(0, false, 0));
+                    if (MinimumSpeedForCCEffectMpS > 0) notches.Add(new MSTSNotch(float.Epsilon, false, 0));
+                    float numNotches = (float)Math.Round((Locomotive.MaxSpeedMpS - MinimumSpeedForCCEffectMpS) / SpeedRegulatorNominalSpeedStepMpS);
+                    for (int i = 1; i <= numNotches; i++)
+                    {
+                        notches.Add(new MSTSNotch(i / numNotches, false, 0));
+                    }
```
SteelFill, line 549: While doing unrelated testing, I encountered a crash caused when SpeedRegulatorNominalSpeedStepMpS ends up set to 0, resulting in an infinite number of notches, which overfills the notches list.

The system I'm using intentionally has the NominalSpeedStep set to 0, as the system does not allow changing the set speed after being enabled. Previously, a NominalSpeedStep of 0 would change the set speed by 0 every time a speed increase or decrease was commanded, basically nullifying the command to change speed. But now it's trying to create a speed selector controller for every system even if that system isn't supposed to have a speed selector.

Cruise control system is as follows (no updates have been made to this to take advantage of PID control, but this did work fine previously):
```
	ORTSCruiseControl (
		MaxForceSteps ( 8 ) 									Comment ( Throttle controls cruise control force, and the throttle has 8 notches. )
		SpeedIsMPH ( True )
		NominalSpeedStep ( 0 ) 									Comment ( This system does not allow manual speed setting. Speed is set based on current speed at activation time. )
		SpeedSelectorStepTimeSeconds ( 1e9 ) 					Comment ( This will also prevent manual adjustments of speed setting. )
		UseThrottleAsForceSelector ( True ) 					Comment ( The throttle sets the maximum power the cruise control will use. )
		UseThrottleInCombinedControl ( True ) 					Comment ( Required for locomotives which have a combined throttle and dynamic brake. )
		ModeSwitchAllowedWithThrottleNotAtZero ( True ) 		Comment ( This system allows, and requires, throttle to be applied to activate. )
		ThrottleNeutralPosition ( True ) 						Comment ( Setting zero throttle disables this cruise control system entirely. )
		StartReducingSpeedDelta ( 0.05 ) 						Comment ( Set quite low to make acceleration more conservative. Should avoid overshoots. )
		StartReducingSpeedDeltaDownwards ( 0.001 )
		Options ( "RegulatorManual, RegulatorAuto, EngageForceOnNonZeroSpeed" )
		ThrottleFullRangeIncreaseTimeSeconds ( 30 )
		ThrottleFullRangeDecreaseTimeSeconds ( 30 )
		UseDynBrake ( False )									Comment ( Genesis locomotives do not use dynamic brakes in cruise control. )
		DisableCruiseControlOnThrottleAndZeroSpeed ( True ) 	Comment ( Resets cruise when train departs after a stop. )
		DisableCruiseControlOnThrottleAndZeroForce ( True ) 	Comment ( Forces cruise to disable when throttling up from idle, as real system is disabled by idle throttle. )
		DoComputeNumberOfAxles ( True ) 						Comment ( No manual entering of axle information. )
	)

PR #1094, opened 2025-05-26
Blended Braking Compatibility and Effectiveness Improvements

SteelFill:
I noticed that blended braking wasn't playing nice with the advanced brake cylinder simulation introduced in #912 nor the dynamic brake ramp up/down introduced in #1067, so a few small changes help to fix these issues.

Continuation of https://blueprints.launchpad.net/or/+spec/braking-enhancement

- Added new `Engine(ORTSDynamicBlendingMinimumSpeed` parameter to mirror the `Engine(ORTSDynamicBrakeReplacementWithEngineBrakeAtSpeed` parameter added in #996. Previously this was using `DynamicBrakeSpeed1MpS` and in 996 we decided that we probably [shouldn't use the MSTS dynamic brake parameters](https://github.com/openrails/openrails/pull/996#discussion_r1903343306) since most new content will have left those values undefined, giving potentially unpredictable behavior. If the train speed is below the specified speed (default units mph), blended braking will be disabled. If not given, the MSTS DynamicBrakesMinUsableSpeed parameter will still be used as a fallback, and if neither is specified the default is 5 kph, same as before.
- Added new `Engine(ORTSDynamicBlendingRetainedPressure` parameter. When `Engine(ORTSDynamicBrakesHasPartialBailOff` is used, the blended braking system will retain the specified pressure (default PSI) in the brake cylinders regardless of what the dynamic brake does, and the dynamic brake will only respond to pressures higher than this when using `Engine(ORTSDynamicBlendingForceMatch`. Default pressure is 0 psi, unless a brake cylinder spring pressure has been set, in which case the retained pressure is twice the spring pressure. Many real locomotives do this as it prevents excessive cycling of the brake rigging and keeps the brake shoes warm.
- Fix for blended braking taking forever to turn off when advanced brake cylinder simulation is used; when brake cylinder pressure drops below the retained pressure setting, blended braking turns off (as opposed to waiting for the brake cylinder pressure to drop below 0.1 psi, which takes forever on advanced brake cylinders).
- Revised brake cylinder pressure control to consider the retained pressure feature.
- Fixed an error in the handling of bail off on graduated release brake valves that was slowing down the rate of brake cylinder reduction, interfering with blended braking.
- Revised blended braking dynamic brake control to reduce overshooting of target brake force (especially obvious when using dynamic brake rampup) by slowing down the control system response. Now the system waits for the dynamics to set up before attempting to adjust the force, scales the response speed depending on the magnitude of the difference between the current and target forces (small differences are corrected slowly), and limits the speed of the system response to match the dynamic brake ramp up/down rate. I no longer notice any overshooting/hunting behavior that was present previously.

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -3344,6 +3344,7 @@ the following parameters will adjust the behaviour of air brakes:
    single: DynamicBrakeHasAutoBailOff
    single: ORTSDynamicBrakesHasPartialBailOff
    single: ORTSDynamicBlendingRetainedPressure
+   single: ORTSDynamicBlendingMinimumSpeed
    single: ORTSTrainDynamicBlendingTable
    single: ORTSDynamicBrakeReplacementWithEngineBrake 
    single: ORTSDynamicBrakeReplacementWithEngineBrakeAtSpeed
@@ -3358,6 +3359,9 @@ the following parameters will adjust the behaviour of air brakes:
   pressure which, when used in combination with ORTSDynamicBrakesHasPartialBailOff,
   will remain applied regardless of the blended dynamic brake force. This
   pressure is also the minimum pressure at which the blended braking system will activate.
+- ``Engine(ORTSDynamicBlendingMinimumSpeed`` -- Below the specified speed
+  (default units mph, default value 5 mph / 8 kph), local dynamic brake blending
+  will be disabled, allowing locomotive brakes to hold the train while stopped.
   
 Sometimes the train brake controller is capable to apply the dynamic
 brakes for the whole consist, usually as a first step before air brakes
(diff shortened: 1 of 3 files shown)


cesarBLG (changes requested):

SteelFill:
> Hi Philip,
> 
> Regarding the 'overshooting', maybe something like the `AccelerationController` that I just introduced in #1091 can be used. It works similarly to a PID controller with a slight variation:
> 
> * It does a first estimation using `requiredForceN / GetAvailableDynamicBrakeForceN(100%)`
> * Then, it fine-tunes the throttle percent using the difference between DynamicBrakeForceN (i.e. current output) and requiredForceN

Yeah a PID controller would be better overall, but the modified form of what I think is effectively an I controller does good enough for now. This can be revisited once your PR is merged and the PID controller class is officially part of the code, no need to duplicate that here (I want to stay with smaller PRs while 1.6 is on the horizon).

cesarBLG (approved):
Thanks Philip, looks good to me. I linked a blueprint that covers blending changes.

I agree to stay with the I coefficient until the (modified-P)ID controller in #1091 is merged. Still, it won't make a huge difference as brake force increases slowly, and the P term is intended for quick changes. It may be more useful for two-stage braking or EP braking.

Let's wait a couple of days before merging just to check if some unintended side-effect appears.

PR #1095, opened 2025-05-26
Make Control Confirmations Fade Out in Real Time

SteelFill:
A few users have complained that control confirmation fade out doesn't make sense when the simulation speed isn't 100%. At high simulation speed, confirmers disappear too quickly to be read, but at low simulation speed confirmers stick around uncomfortably long, sometimes [sticking around in screenshots](https://www.trainsim.com/forums/forum/open-rails/open-rails-route-projects-and-questions/2300839-trainsimulations-sand-patch-csx-mainline-subdivision-development-thread#post2300839) when the user uses slow motion to frame their shot better.

With some investigation, it seems that confirmation messages were simply using game time when real time would have made more sense, so this was simply replaced to use real time.

Also fixed a crash I have randomly gotten a few times over the years, seemingly caused by the confirmation window attempting to manipulate the LabelText and/or LabelShadow before they were fully initialized. A null check should prevent that crash by waiting to update the objects until they are initialized.

### DIFF
--- a/Source/RunActivity/Viewer3D/Popups/MessagesWindow.cs
+++ b/Source/RunActivity/Viewer3D/Popups/MessagesWindow.cs
@@ -139,7 +139,8 @@ namespace Orts.Viewer3D.Popups
             }
 
             foreach (var message in Messages)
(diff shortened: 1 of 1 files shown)


cjakeman (approved):
Hi Phillip,

Thanks for fixing the time and also checking for null.
I remember working on this and never considered running fast or slow.

Approved and merged,

Chris

PR #1097, opened 2025-05-27
Fix For Delayed Particle Emitter Spawning

SteelFill:
Split from #1072. This is the fix for a bug I've seen sent my way a few times (can't find any formal bug reports, but it's made its way to my PMs more than once), particle emitters had a tendency to spawn particles too late, particularly when the particles or rail vehicle was moving at a high speed, or when there was a substantial delay between individual particles spawning. This resulted in a visible gap between where particles should appear from and where they actually appeared from.

The problem turned out to be a very simple off-by-one error (or, off by four in this case). The particle shader, which is actually responsible for putting particles on the screen, was being given a list of particles that was one particle out of date. So instead of drawing the newest particle, the shader would always draw the _second_ newest particle. By this time this particle was drawn on screen, it would have already moved, creating the visible gap in particles players have seen.

The fix is to change how the `vertex` counter works in `ParticleEmitterPrimitive.Update`. This variable decides where in the array of vertices to insert the information about the newest particle. Previously, the method would take the index of the first unoccupied space, _then add 1 to it_. Note that this means an empty space is left behind; when the `FirstFreeParticle` is at index 0, data is inserted at index 1 (in reality, this is index 4 as each particle has 4 vertices, but that distinction doesn't matter), _leaving index 0 empty_. The shader does not know this, so the shader starts at index 0, one index behind where the data actually is. This repeats, with the shader always grabbing data that is _one particle out of date_. Rather than change the shader, we simply change the `vertex` counter to use the `FirstFreeParticle` directly, before it is incremented. That way the emitter and the shader are perfectly in-sync with which particle is the newest one.

Observe how the problem is _much_ more complicated than the solution, such is life.


cesarBLG:
This simple fix is a candidate to enter into 1.6. You have to set the base branch to release/1.6. I'm not sure if you have the right permissions to do so, or it has to be done by ORMT.

cesarBLG:
From your side you just have to do 'git rebase release/1.6' and then force push here. This is because you based your changes in a master branch which is more recent than release/1.6. Still, not sure about what will be the bug fix policy for 1.6.

cjakeman:
Hi Cesar,

> Only ORMT has the right to merge the changes to the 1.6 branch once the PR is approved.

That isn't a rule that I recognise. Any authorised developer who can approve a PR also has the permissions to merge into the "master" branch. Once the PR is approved, the author also has the ability to merge it once approved.

We did have a discussion on ET where some devs preferred to do their own merge and others are happy to have it done by the approver. In either case, please don't leave it to ORMT to tidy up.

Best wishes,

Chris

twpol (approved):
Approved for 1.6, thanks!

PR #1099, opened 2025-05-29
Fix TCS orders not being sent to pantographs

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/38854-tcs-feature-no-more-working-in-testing-1850/page__pid__318796#entry318796

This is a regression from 1.5.1, I'm targeting 1.6 branch


Csantucci (approved):
That works. Obviously  no formal problems.

twpol (approved):
Approved for 1.6, thanks!

PR #1101, opened 2025-05-30
Added help buttons on Simulations options in Menu

sweiland-openrails:
There were no help buttons on the Options Simulation tab in the Menu


cjakeman (approved):
Hi Siebren,

Many thanks for spotting this and fixing it.

Approved and merged.

Chris

PR #1102, opened 2025-06-03
Fix Stuck Locomotive Brakes After Initialization

SteelFill:
Another quick fix for the next version, a nagging issue left in the wake of #916 and #997 is that locomotives not connected to the lead locomotive would get their locomotive brakes stuck on during initialization if the lead locomotive was initialized with the locomotive brakes applied if the locomotive brake was not synchronized by the distributed power system. This is because the locomotive brake would be initialized as if all locomotives were synchronized with the lead locomotive, but once initialized the trailing locomotive would ignore locomotive brake commands, resulting in stuck brakes (until re-initializing with the locomotive brake released).

Now, the locomotive brake won't be applied during initialization on any locomotives not connected to the lead locomotive unless independent brake sync is active. The locomotive may still develop brake cylinder pressure if the train brake is applied, but the train brake won't be stuck applied unlike the locomotive brake.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
@@ -594,10 +594,38 @@ namespace Orts.Simulation.RollingStocks.SubSystems.Brakes.MSTS
 
         public override void Initialize(bool handbrakeOn, float maxPressurePSI, float fullServPressurePSI, bool immediateRelease)
         {
+            MSTSLocomotive loco = Car as MSTSLocomotive;
+
             BrakeLine1PressurePSI = Car.Train.EqualReservoirPressurePSIorInHg;
             BrakeLine2PressurePSI = Car.Train.BrakeLine2PressurePSI;
-            if (Car is MSTSLocomotive && Car.Train.LeadLocomotive is MSTSLocomotive lead)
-                lead.EngineBrakeController?.UpdateEngineBrakePressure(ref BrakeLine3PressurePSI, 1000);
+            // Initialize locomotive brakes
+            if (loco != null && Car.Train.LeadLocomotive is MSTSLocomotive lead)
+            {
+                bool brakeLine3Init = false;
+
+                if (loco == lead) // Always initialize loco brakes on lead loco
+                    brakeLine3Init = true;
+                else
+                {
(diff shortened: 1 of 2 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/Brakes/MSTS/AirSinglePipe.cs
```diff
@@ -594,10 +594,38 @@ public override void Restore(BinaryReader inf)
...
+                            else if (loco.DPSyncIndependent && lead.DPSyncIndependent)
+                                brakeLine3Init = true; // Otherwise, only initialize loco brakes if synchronized by DP system
+
+                            break;
+                        }
+                    }
+                }
+
+                if (brakeLine3Init) // Sync loco brakes with lead brake system
+                    lead.EngineBrakeController?.UpdateEngineBrakePressure(ref BrakeLine3PressurePSI, 1000);
```
cesarBLG, line 625: This is definitely an improvement, but there's still this weird line. It's using the lead locomotive controller to update the pressure in another locomotive? What if they have incompatible controllers? It would be better to ensure that the lead locomotive is initialized first (to initialize the controller, as well as the reference pressure), and then for the rest of locomotives only set BrakeLine3PressurePSI = Train.BrakeLine3PressurePSI (or whatever variable that follows the lead loco pressure)

SteelFill: Such an implementation would be functionally identical to this. It looks odd, but all this is really doing is repeatedly setting BrakeLine3 to the pressure that the lead locomotive wants, which is the same as updating the lead locomotive specifically, then copying the result to other locomotives, just without needing to do the update in a specific order. The engine brake controller doesn't care, it just sees a number that isn't the target value and moves it toward the target value before spitting it back out.

And it doesn't matter what the engine brake controller is on the other locomotives as those are cut-out and the pressure in the independent brake pipe would be determined by the lead locomotive anyway, this is just simpler than running the brake pipe propagation to 'naturally' set the independent pipe to the pressure the lead locomotive's system wants.

cesarBLG: That's right, it was a quite confusing way to do it but it's effective. I didn't stop to fully understand it, but it obviously works as you say.

cesarBLG (approved):

twpol (approved):
Approved for 1.6, thanks!

PR #1103, opened 2025-06-03
Fix diverging force when exceeding max speed

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/38866-systemarithmeticexception-in-axlecs-nan/

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
@@ -2575,14 +2575,14 @@ namespace Orts.Simulation.RollingStocks
             if (TractiveForceCurves == null)
             {
                 powerW = Math.Min(powerW, MaxPowerW * t * t * (1 - PowerReduction));
-
-                if (AbsTractionSpeedMpS > MaxSpeedMpS - 0.05f)
-                {
-                    forceN = 20 * (MaxSpeedMpS - AbsTractionSpeedMpS) * MaxForceN * (1 - PowerReduction);
-                }
-                else if (AbsTractionSpeedMpS > MaxSpeedMpS)
+                
+                if (AbsTractionSpeedMpS > MaxSpeedMpS)
                 {
                     forceN = 0;
+                }
+                else if (AbsTractionSpeedMpS > MaxSpeedMpS - 0.05f)
+                {
+                    forceN = 20 * (MaxSpeedMpS - AbsTractionSpeedMpS) * MaxForceN * (1 - PowerReduction);
                 }
                 else
                 {


Csantucci (approved):
Seems a quite logical bug fix

rwf-rr:
This fixes the exception I reported in the referenced thread.

However, it is not an ideal solution, as the tractive effort now toggles between 0 (zero) and the normal value. Ideally, the wheel-slip control should reduce the power, or the wheel speed should reach a reasonable physical limit. But that my be a future thing.

Furthermore, setting the force to 0 (zero) seems to have an adverse performance effect (in my case the sound thread starts to stutter). Setting the force to a small value seems to avoid this while still avoiding the excessive wheel speed.

Suggested:

```
                if (AbsTractionSpeedMpS > MaxSpeedMpS)
                {
                    // limit to a small force, as setting the force to zero seems to have adverse effects
                    forceN = 0.05f * MaxForceN * (1 - PowerReduction);
                }
```

cesarBLG:
Setting the force to something that is non-zero but small is a workaround. If there is a bug, it must be fixed.

I don't understand why you say that it toggles between 0 and the normal value. If the speed exceeds the maximum speed, traction is completely removed by a safety system. So it's normal to me that it abruptly changes to 0. This was also the behaviour in 1.5, and changing that is a breaking change (which also makes no sense). Therefore I'm not accepting your suggestion, but I'll try to investigate the issue if you give me more details.

Regarding the performance, the Polach integrator is CPU hungry, especially during wheelslip. If Pacha method is not selected automatically, use simple adhesion.

Can you be more specific on the specific circumstances during which you experience a faulty behaviour? The most useful information would be a video with the 'Force information HUD' activated.

rwf-rr:
Here is the screen recording with your fix. Once the train starts moving, I throttle up fast. Once the wheel-slip starts, the speedometer and the ampere-meters jump between two values. Eventually the sound starts stuttering.
[Video after your fix](https://drive.google.com/file/d/1v-yVy71LpIfBLxqFMvUrdWUxkjaErWHZ/view?usp=sharing)

I agree that zero is a safer value.

cesarBLG:
Thanks for the video. I downloaded the Re 6/6 and I was able to reproduce it. See my comment in Elvas Tower, the engine has unrealistic physics, what causes the wheelslip.

This is what is happening:
1. You increase throttle. Due to the unrealistic tractive force, the wheels start slipping
2. When you start slipping, the adhesion coefficient is reduced. This means that wheels start spinning like crazy due to the huge tractive force, which is used to accelerate the wheels even more (because they have low inertia) and not the full engine.
3. When wheels reach the maximum engine speed, which happens almost instantly due to the huge tractive force, the electric motor is switched off for safety
4. As there is no tractive force anymore, all the wheel inertia is transmitted to the wheels and they stop spinning. Wheelslip gets back to normal
5. As wheel speed is back to normal, motor starts developing force again, and wheels start slipping, going back to 2.

All this happens quickly because the wheels have low inertia. That's life. That's also the reason for which the adhesion integrator is very CPU expensive (it requires very short update times to process those quick changes). Maybe it would allow higher update times, but we need a proper error estimator for the integrator, which is an advanced topic. If this affects performance, the solution could be simple adhesion, or a limit in the number of integrator substeps depending on the CPU capabilities. I think this applies both to Pacha and Polach methods, the only different is that the error estimator for Polach is more conservative and demands higher number of substeps, but maybe it could tolerate a lower resolution.

I don't think there's a way to solve that. If you detect wheelslip, you should reduce the throttle. Maybe a delay to cut and restore power could be applied, but that's something for the advance motor simulation that is still work in progress.

Another solution is to add slip control to the locomotive.

rwf-rr:
Thank you, Cesar, for the detailed explanation.

twpol (approved):
Approved for 1.6, thanks!

PR #1104, opened 2025-06-03
Handle simple adhesion within the axle module

cesarBLG:
Unify the handling of simple and advanced adhesion in the axle module. This avoids duplication of some parts of the code and handles all adhesion-related stuff inside the Axle class. This also makes per-axle wheel speed available for simple adhesion, which reduces the number of checks for `Simulator.UseAdvancedAdhesion` in other parts of the code.

Some other changes:
- Removed the 'compensated' axle forces, instead splitting the total rail out force into 'motive', 'brake' and 'friction' forces.
- Use the 'brake' out force (accounting for adhesion) as the BrakeForceN used for wagon physics.
- Add initial support for non-driven axles, and include a default non-driven axle for steam locomotives. Implemented the physics (not just the visuals) for DriveWheelOnlyBraking.
- Reimplemented the legacy AntiSlip and WheelslipCausesThrottleDown slip control systems to use axle drive force instead of physically moving the throttle lever.
- Cleanup

Left to future PRs:
- Handle wagon adhesion from the axle module (but keep simple adhesion for wagons to save CPU time).
- Integrate axle friction with wagon friction, by moving all rolling related friction (Davis A and B?) to the axle module.

https://blueprints.launchpad.net/or/+spec/organize-traction-adhesion


SteelFill:
With all the steam parameters who knows what's driving what, so here's the engine for the main loco and the articulated second engine I was using.

[Cab Forward 4244.zip](https://github.com/user-attachments/files/20826699/Cab.Forward.4244.zip)

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2729,40 +2724,73 @@ protected virtual void UpdateTractiveForce(float elapsedClockSeconds)
...
+                        {
+                            // Only restore traction when throttle is set to 0
+                            SlipControlActive[i] = false;
+                        }
+                        // Disable traction in the axle if slipping
+                        if (SlipControlActive[i]) axle.DriveForceN = 0;
+                    }
+                    else if (SlipControlSystem == SlipControlType.ReduceForce)
+                    {
+                        if (!DynamicBrake && axle.DriveForceN != 0 && AdvancedAdhesionModel)
```
SteelFill: Do we really want these two modes of slip control to not work in dynamic brake? I know that the previous (questionable) implementation only worked in powering but I feel that's more a product of the old system being less than optimal vs. an intentional feature. Locomotives having slip control on both powering and dynamic braking seems much more typical, at least with these types of systems that reduce power (as opposed to ones that use locomotive braking).

Also, is there any particular reason the ReduceForce method requires advanced adhesion?

cesarBLG: I think it makes sense to add it also for dynamic brake. My initial goal was to get something equivalent to the previous implementation.

> Also, is there any particular reason the ReduceForce method requires advanced adhesion?

The AntiSlip() token activates the ReduceForce method, but it also prevents wheelslip when simple adhesion is set (it behaves more like a Full slip control system), so having this for simple adhesion could be problematic for badly configured stock. But the condition should be (AdvancedAdhesionModel || !AntiSlip) so if someone enters SlipControlSystem ( ReduceForce ) directly (not via AntiSlip), it still works for simple adhesion.

SteelFill: Did not realize anti-slip disabled wheel slip in simple adhesion. Though now you mention it, "AntiSlip()" won't actually enable anti slip, as AntiSlip's parser defaults to false, which surely can't be right?
```C#
                case "engine(antislip": AntiSlip = stf.ReadBoolBlock(false); break;
```
If someone puts in AntiSlip we should be assuming they want it to be turned on even if it's not explicitly set to true.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -1104,7 +1105,7 @@ public override void Parse(string lowercasetoken, STFReader stf)
                 case "engine(dynamicbrakesresistorcurrentlimit": DynamicBrakeMaxCurrentA = stf.ReadFloatBlock(STFReader.UNITS.Current, null); break;
                 case "engine(numwheels": MSTSLocoNumDrvWheels = stf.ReadFloatBlock(STFReader.UNITS.None, 4.0f); if (MSTSLocoNumDrvWheels < 1) STFException.TraceWarning(stf, "Engine:NumWheels is less than 1, parts of the simulation may not function correctly"); break;
                 case "engine(ortsnumberdriveaxles": LocoNumDrvAxles = stf.ReadIntBlock(null); if (LocoNumDrvAxles < 1) STFException.TraceWarning(stf, "Engine:ORTSNumberDriveAxles is less than 1, parts of the simulation may not function correctly"); break;
-                case "engine(antislip": AntiSlip = stf.ReadBoolBlock(false); break;
+                case "engine(antislip": AntiSlip = stf.ReadBoolBlock(true); break;
```
SteelFill, line 1108: Another comment on AntiSlip parsing, it turns out MSTS accepted AntiSlip() when placed in the Wagon( section (look at the Dash 9, for example). While strange, this should also be allowed for compatibility with MSTS. MSTS even implements AntiSlip() on some wagon files, which probably means it's intended to work as wheel slide protection as well? Since OR already has ORTSWheelBrakeSlideProtection, AntiSlip could be an alias for that?
```suggestion
                case "engine(antislip":
                case "wagon(antislip":
                    AntiSlip = stf.ReadBoolBlock(true);
                    break;
```
I did some searching and this inconsistency, as well as the fact that "AntiSlip()" should be interpreted as "AntiSlip(true)", was actually [noticed many years ago](https://www.elvastower.com/forums/index.php?/topic/28692-antislip-parameter-not-correctly-parsed/) but it seems neither Carlo nor Peter acted on that discussion and it went forgotten.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2729,40 +2724,73 @@ protected virtual void UpdateTractiveForce(float elapsedClockSeconds)
...
+                            SlipControlActive[i] = false;
+                        }
+                        // Disable traction in the axle if slipping
+                        if (SlipControlActive[i]) axle.DriveForceN = 0;
+                    }
+                    else if (SlipControlSystem == SlipControlType.ReduceForce)
+                    {
+                        if (axle.DriveForceN != 0 && (AdvancedAdhesionModel || !AntiSlip))
+                        {
+                            if (axle.SlipPercent > axle.SlipWarningTresholdPercent) SlipControlActive[i] = true;
```
SteelFill: Using the axle.SlipPercent directly causes the slip control system to respond to the stable micro-slips present in the Polach system, so traction is constantly cut down to insignificant levels.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2729,40 +2724,73 @@ protected virtual void UpdateTractiveForce(float elapsedClockSeconds)
...
+                            SlipControlActive[i] = false;
+                        }
+                        if (SlipControlActive[i])
+                        {
+                            float absForceN = Math.Abs(axle.DriveForceN);
+                            float newForceN;
+                            if (axle.SlipPercent < axle.SlipWarningTresholdPercent * 0.9f)
+                            {
+                                // If well below slip threshold, restore full power in 10 seconds
+                                newForceN = Math.Min(Math.Abs(prevForceN) + absForceN * elapsedClockSeconds / 10, absForceN);
```
SteelFill, line 2804: This seems to be interacting poorly with the integrator (at least with the Polach method) as the axle.DriveForceN is further modified by the integrator. I'm observing this line calculate an increase in drive force, but that increase ends up smaller than the decrease in drive force after running the axle integrator (now maybe the integrator shouldn't be doing that), so the force never actually increases.

I think you may need to change this to not increase force relative to the current force, but increase force relative to the time elapsed since wheel slip was resolved.

cesarBLG: The integrator shouldn't change the drive force if the Axle is ForceDriven, which is where this clause applies. The bug is indeed that the drive force is taking into account the transmissionEficiency, but it shouldn't.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: This means that the integrator substeps are not high enough, but I have already said this multiple times for the Polach method, so I guess we will have to calculate an average slip percent instead.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Yes it seems like that was causing trouble there, and it is less troublesome in the latest version.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: I don't think there's an added value in trying to match MSTS anymore. Legacy MSTS content will probably have AntiSlip() just because people copied that from other .eng or .wag file, not because the wagon had slip protection. Changing it now to accept the Wagon(AntiSlip()) will change the behavior of OR itself, which is a breaking change. If nobody complains about OR not accepting that, let's keep it as is.

SteelFill: While I won't agree with ignoring MSTS out of this perception that 'old content is configured wrong', especially in this case where high effort locomotives that should indeed have wheel slip protection like the Dash 9 or HHP-8 get excluded, I'm not in the mood to fight that battle when I personally prefer to modify MSTS content and this particular series of adhesion updates doesn't need to be held around for decade-old quibbles.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Resolution works well with AntiSlip, though with the Full slip protection system similarly problematic behavior remains (more on that in another comment.)

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2729,40 +2724,77 @@ protected virtual void UpdateTractiveForce(float elapsedClockSeconds)
...
                     {
                         // Simple slip control
                         // Motive force is reduced to the maximum adhesive force
                         // In wheelslip situations, motive force is set to zero
-                        axle.DriveForceN = Math.Sign(axle.DriveForceN) * Math.Min(axle.MaximumPolachWheelAdhesion * axle.AxleWeightN, Math.Abs(axle.DriveForceN));
-                        if (axle.IsWheelSlip) axle.DriveForceN = 0;
+                        float adhesionLimit;
+                        if (axle.SlipPercent > 115) adhesionLimit = 0;
+                        else if (axle.SlipPercent > 95) adhesionLimit = axle.MaximumWheelAdhesion * (115 - axle.SlipSpeedPercent) / 20;
+                        else adhesionLimit = axle.MaximumWheelAdhesion;
```
SteelFill: Similar to the ReduceForce method, micro-slips are causing trouble here as well. The fact that Full slip protection entirely cuts traction at 115% slip is especially troublesome when being used with the features from #1067 as power can take some time to restore just because a micro-slip happened.

I think a combination of the adhesion limit behavior here (except without the whole instant cut power thing) and the rapid, but controlled, reduction in power of the ReduceForce method could work well. The adhesion limit should prevent most actual wheel slip, and should wheel slip truly occur the ReduceForce method is an effective backup.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -2729,40 +2724,93 @@ protected virtual void UpdateTractiveForce(float elapsedClockSeconds)
...
+                        float newForceN;
+                        if (axle.HuDIsWheelSlip)
+                        {
+                            newForceN = 0;
+                        }
+                        else if (!axle.HuDIsWheelSlipWarning)
+                        {
+                            // If well below slip threshold, restore full power in 10 seconds
+                            newForceN = Math.Min(Math.Abs(prevForceN) + absForceN * elapsedClockSeconds / 10, absForceN);
+                        }
```
SteelFill: This is running all the time during normal operation, which is greatly slowing down the build-up of tractive effort. This should probably use an implementation of SlipControlActive like ReduceForce does so it doesn't intervene when slip has yet to occur.

cesarBLG: Hopefully it works now as expected. I'm still not sure if this should respect the ramp up and down rates, probably it should do it for simplicity.

SteelFill: Using the ramp rates would make sense but would need a totally different implementation and would probably require assigning ramp rates to everything so you don't get pre-1.6 locomotives toggling tractive effort instantly during wheel slip. Then there's the whole thing that the ramp rates are defined and controlled outside the axle model (so the axle might need to signal to the locomotive to reduce force?)

Things are drivable in the current version, which is the important milestone.

cesarBLG:
Currently rail output force is not accurate during wheelslip due to discrepancies between the friction calculated in the axle module and the one calculated at MSTSWagon.UpdateTrainBaseResistance(). The unification of both force calculations thus cannot be postponed as I initially intended to do.

SteelFill (dismissed):
As this has had some time to settle in and crashes have been found and diagnosed, I feel this is ready to go in and free up files for future changes. The refactor should be very helpful to simplify future changes, and the rework to wheel slip protection is helpful for the driving experience too. While even more can be done for wheel slip protection, such additional ideas make more sense for a different PR.

cesarBLG:
Thanks for the approval @SteelFill 

I found a last-minute bug regarding legacy steam locomotive braking, but it should be fixed now. Also the merge conflict is solved

cesarBLG:
Thanks, Philip

@cjakeman the problems with this PR have been fixed, could you review the blueprint?

PR #1107, opened 2025-06-08
Fix DPMode when last remote is moved to front.

rwf-rr:
When the last remote loco is moved to the front, the train's DPMode remains as is (eg. traction). When the lead consist is moved to the other mode (eg. traction -> dynamic braking), and then a loco is moved to the back, the train's DPMode is still in the previous mode (eg. traction), even though the remote now provides braking. This results in the wrong colour on the Train DPU Info window.

When the last remove is moved to the front, the DPMode needs to be set to idle.

I also added a second change: When the first loco is moved to the back, the DPMode is set based on the lead loco. DPThrottlePercent or DPDynamicBrakePercent is set from the lead, so it does not make sense to leave the DPMode at idle.

I did not like that I had to set traction mode (Ctrl-L) after moving the loco to the rear before I can use the more/less functions (Ctrl-U and Shift-Ctrl-U). I believe that extra step is not required in the real world.

When the lead loco is idle (no traction or dynamic braking), the DPMode remains at idle.

PS: There are additional issues around DPMode. It is used to show the remotes (traction or dynamic braking) in the Train Driving Info window. When the remotes are in idle mode, there is no value for the remotes, even though the fence is up and the remotes have different values from the lead consist.

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -3344,6 +3344,7 @@ the following parameters will adjust the behaviour of air brakes:
    single: DynamicBrakeHasAutoBailOff
    single: ORTSDynamicBrakesHasPartialBailOff
    single: ORTSDynamicBlendingRetainedPressure
+   single: ORTSDynamicBlendingMinimumSpeed
    single: ORTSTrainDynamicBlendingTable
    single: ORTSDynamicBrakeReplacementWithEngineBrake 
    single: ORTSDynamicBrakeReplacementWithEngineBrakeAtSpeed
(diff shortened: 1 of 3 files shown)


SteelFill (approved):
Simple change, and seems to do the trick. For something this small it could be put in 1.6 as DPU is a 1.6 feature and it wouldn't be great to have glitchy behavior present in a 'new' feature for a year. You could rebase to the 1.6 release candidate branch and change the PR to use 1.6 as the baseline.

rwf-rr:
New PR: https://github.com/openrails/openrails/pull/1118

PR #1112, opened 2025-06-23
update for V1.6

cjakeman:
Adds settings to Installer to work with 64-bit Windows and prompt for destination folder. Also updates instructions to users to match new mechanisms for updating OR and installing content.

### DIFF
--- a/Source/Installer/Using the installer scripts.txt
+++ b/Source/Installer/Using the installer scripts.txt
@@ -1,8 +1,11 @@
 Using the installer scripts.txt
-Last updated: 2022-08-21 by Chris Jakeman
+Last updated: 2025-06-21 by Chris Jakeman
 
-1. If you wish to update the version name from "1.5", then edit the line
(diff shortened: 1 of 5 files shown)


Source/Installer/Using the installer scripts.txt
```diff
@@ -1,19 +1,24 @@
 Using the installer scripts.txt
-Last updated: 2022-08-21 by Chris Jakeman
+Last updated: 2025-06-21 by Chris Jakeman
```
twpol, line 2: Love the day being the same :)

Source/Installer/Version.iss
```diff
@@ -1 +1 @@
-#define MyAppVersion "none"
+#define MyAppVersion "1.6"
```
twpol, line 1: I changed this to `none` because the `Build.cmd` code sets it automatically, but it's not a big deal

twpol (approved):

PR #1114, opened 2025-06-25
Fix color of DB in DrivingInfo when in setup with DPU fenced. (r1.6)

rwf-rr:
**This is a PR for release 1.6.** _Since my other PR (DPMode) is desired for 1.6, I though I also do this for 1.6._

This affects the Train Driving Info window and its web-server presentation.

When the lead consist is in dynamic brake setup and the back consist is already in dynamic braking, instead of the cyan colour, a string with "%%%" is presented. The cause is that the value for the back consist is appended after the colour code. The colour code needs to be at the end of the string.

Also add the back-consist throttle percentage to the web-server presentation. It was missing.

Lastly, I changed the DPU value separator, from parenthesis to a vertical bar. This is more readable, and better matches the presentation in the DPU window.

Before:
DynamicBrakeInfo-before

After:
DynamicBrakeInfo-after


SteelFill (approved):
This was definitely buggy, I didn't even realize the dynamic brake display was intended to show the DPU braking setting as most of the time it simply didn't do that. The fixed version works much more consistently and prevents the formatting error.

twpol (approved):

PR #1115, opened 2025-06-26
Do not activate ETS switch if no suitable cars are attached

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/36034-changing-ets-behaviour/page__view__findpost__p__319484

https://trello.com/c/p975Tma4/611-ets-only-works-on-startup-if-required-by-consist

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselPowerSupply.cs
@@ -311,13 +311,13 @@ namespace Orts.Simulation.RollingStocks.SubSystems.PowerSupplies
                     SignalEventToBatterySwitch(PowerSupplyEvent.QuickPowerOn);
                     SignalEventToMasterKey(PowerSupplyEvent.TurnOnMasterKey);
                     SignalEventToDieselEngines(PowerSupplyEvent.StartEngine);
-                    SignalEventToElectricTrainSupplySwitch(PowerSupplyEvent.SwitchOnElectricTrainSupply);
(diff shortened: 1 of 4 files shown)


Source/Orts.Simulation/Common/Scripting/PowerSupply/LocomotivePowerSupply.cs
```diff
@@ -207,6 +207,25 @@ protected float ThrottleReductionPercent
...
+        protected int NumberOfElectricTrainSupplyConnectedCars
+        {
+            get
+            {
+                int count = 0;
+                foreach (var car in Train.Cars)
+                {
+                    if (car == null) continue;
+                    if (!(car is MSTSWagon wagon)) continue;
+                    if (!(wagon.PassengerCarPowerSupply?.ElectricTrainSupplyConnectedLocomotives.Contains(Locomotive) ?? false)) continue;
```
SteelFill: I could not get this feature to work. In the debugger, it seems that `ElectricTrainSupplyConnectedLocomotives` has yet to be set when first starting a session. `ElectricTrainSupplyConnectedLocomotives` seems to be established by `ScriptedPassengerCarPowerSupply.Update()` but that appears not to be called for the first time until after `DefaultDieselPowerSupply.Update()` is called for the first time (in the case of using the default diesel power supply).

Also, `ScriptedPassengerCarPowerSupply.Update()` seems to run the process for seeing which locomotives are attached every single frame, which does not sound particularly efficient.

SteelFill (changes requested):
Due to order of operations, this isn't having any effect. If ORTSElectricTrainSupply ( Mode ( Switch ) ) is used, ETH always starts turned off and if ORTSElectricTrainSupply ( Mode ( Automatic ) ) is used, ETH always starts turned on, regardless of the connected consist's demand for ETH.

PR #1117, opened 2025-06-28
ORTS_PLAYER_DIESEL_ENGINE_STATE manual fix

albertosaurio65:
The states indicated in the manual don't correspond to what actually happens, but with the changes made now they do. The example has also been changed, as "display-only" is specified and a multistate mouse control was being used.

### DIFF
--- a/Source/Documentation/Manual/cabs.rst
+++ b/Source/Documentation/Manual/cabs.rst
@@ -289,7 +289,7 @@ Example::
    single: ORTS_PLAYER_DIESEL_ENGINE_STATE
 
 ORTS_PLAYER_DIESEL_ENGINE_STATE: this control respectively selects frames 0, 
-1, 2, 3 for the player locomotive engine states Stopped, Stopping, Starting and Running and. It is a display-only control.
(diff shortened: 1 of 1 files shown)


Source/Documentation/Manual/cabs.rst
```diff
@@ -289,21 +289,33 @@ Example::
    single: ORTS_PLAYER_DIESEL_ENGINE_STATE
 
 ORTS_PLAYER_DIESEL_ENGINE_STATE: this control respectively selects frames 0, 
-1, 2, 3 for the player locomotive engine states Stopped, Starting, Running and 
-Stopping. It is a display-only control.
+1, 2, 3 for the player locomotive engine states Stopped, Stopping, Starting and Running and. It is a display-only control.
```
cesarBLG:
```suggestion
1, 2, 3 for the player locomotive engine states Stopped, Stopping, Starting and Running. It is a display-only control.
```

cesarBLG (changes requested):

cesarBLG (approved):

PR #1118, opened 2025-06-28
Fix DPMode when last remote is moved to front. (r1.6)

rwf-rr:
This is [PR 1107 ](https://github.com/openrails/openrails/pull/1107)recreated for release 1.6.

When the last remote loco is moved to the front, the train's DPMode remains as is (eg. traction). When the lead consist is moved to the other mode (eg. traction -> dynamic braking), and then a loco is moved to the back, the train's DPMode is still in the previous mode (eg. traction), even though the remote now provides braking. This results in the wrong colour on the Train DPU Info window.

When the last remove is moved to the front, the DPMode needs to be set to idle.

I also added a second change: When the first loco is moved to the back, the DPMode is set based on the lead loco. DPThrottlePercent or DPDynamicBrakePercent is set from the lead, so it does not make sense to leave the DPMode at idle.

I did not like that I had to set traction mode (Ctrl-L) after moving the loco to the rear before I can use the more/less functions (Ctrl-U and Shift-Ctrl-U). I believe that extra step is not required in the real world.

When the lead loco is idle (no traction or dynamic braking), the DPMode remains at idle.

PS: There are additional issues around DPMode. It is used to show the remotes (traction or dynamic braking) in the Train Driving Info window. When the remotes are in idle mode, there is no value for the remotes, even though the fence is up and the remotes have different values from the lead consist.

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -1628,6 +1628,7 @@ namespace Orts.Simulation.Physics
             if (LeadLocomotive == null || (LeadLocomotive as MSTSLocomotive).DPDynamicBrakeController == null)
                 return;
             int idToMove = -1;
+            bool stillHasRemotes = false;
             for (var i = 0; i < Cars.Count; i++)
             {
                 if (!(Cars[i] is MSTSLocomotive))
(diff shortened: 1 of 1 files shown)


rwf-rr:
I had to create a new PR for release/1.6. Something went wrong when I rebased the original PR (branch) from master to release/1.6.

SteelFill (approved):
Approved, see #1107 for original review comments.

PR #1120, opened 2025-07-03
Automatically Calculate Friction Values if Missing

SteelFill:
Split from #1072 

[Related Trello card](https://trello.com/c/OlRrZBEE/621-automatically-calculate-friction-values-if-missing)

This PR adds the ability for OR to estimate the Davis resistance values of a rail vehicle if an ORTSBearingType has been defined, but some or all of the ORTSDavis values were left undefined. The original Davis formula is used for Grease and Friction bearing types, while the CN resistance formula is used for Roller and Low bearings. (Compare this to the current method, which may set the friction value to 0, or using MSTS friction.)

- Davis A value is calculated based on the type of bearings, weight of the rail vehicle, and total number of axles (includes support for the special case in the original Davis formula of vehicles weighing less than 5 US tons per axle, and considers extra resistance due to rods on steam locomotives)
- Davis B value is calculated from the type of bearings, weight of the rail vehicle, and the type of rail vehicle (again, the less than 5 ton per axle case is also considered here)
- Davis C value is calculated more simply, just using the frontal area and Davis drag constant (both of which have default values if not defined)

Any manually entered friction values are used first, so to tell if automatic calculation is being used due to missing data, enable verbose configuration messages are enabled. Every time something has a resistance component calculated automatically, a message is logged giving the calculated resistance value and the data used to calculate it, which should make it easy to tell when a value was forgotten or if the wagon data is sensible.

ORTS freight animations are supported, with resistance calculated for both the empty and loaded states.

And some general refactoring was done in the way bearings, resistance, and resistance units of measure are handled to make things a little cleaner.

This feature is not intended as a replacement to external friction calculators as it's not possible to account for every case of resistance from within OR, and there are dozens of friction formulas that cannot all be added to OR. If in doubt, do what has always been done and manually enter resistance values, while keeping an eye out for any missing values (which will be indicated in the verbose configuration log).

### DIFF
--- a/Source/Documentation/Manual/physics.rst
+++ b/Source/Documentation/Manual/physics.rst
@@ -40,8 +40,8 @@ Resistive Forces
 ----------------
 
 Open Rails physics calculates resistance based on real world principles:
-gravity, mass, rolling resistance, and curve resistance. This is calculated
-individually for each car in the train.
+gravity, mass, wind, rolling resistance, and curve resistance. This is
+calculated individually for each car in the train.
 
 The program supports a few methods for determining rolling resistance.
 The oldest method, intended only to support legacy content, uses the
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
@@ -710,7 +710,7 @@ namespace Orts.Simulation.RollingStocks
                 {
                     if (ortsFreightAnim.ShapeFileName != null && !File.Exists(wagonFolderSlash + ortsFreightAnim.ShapeFileName))
                     {
-                        Trace.TraceWarning("ORTS FreightAnim in trainset {0} references non-existent shape {1}", file, wagonFolderSlash + ortsFreightAnim.ShapeFileName);
+                        Trace.TraceWarning("ORTS FreightAnim in trainset {0} references non-existent shape {1}", shortPath, wagonFolderSlash + ortsFreightAnim.ShapeFileName);
                         ortsFreightAnim.ShapeFileName = null;
                     }
 


SteelFill:
@rwf-rr Roger, since you've looked at train forces previously, I think you are in a good position to look over this PR. I've been using the feature for a little while without trouble and there's nothing else I want to do with it, no reason for it to sit around even longer.

Source/Documentation/Manual/physics.rst
```diff
@@ -39,50 +39,75 @@ the WAG / ENG file is discussed.
...
-rolling resistance, or friction, based on the Friction parameters in the
-Wagon section of .wag/.eng file. Open Rails identifies whether the .wag
-file uses the *FCalc* utility or other friction data. If *FCalc* was used to
-determine the Friction variables within the .wag file, Open Rails compares
-that data to the Open Rails Davis equations to identify the closest match
-with the Open Rails Davis equation. If no-FCalc Friction parameters are
-used in the .wag file, Open Rails ignores those values, substituting its
-actual Davis equation values for the train car.
+Open Rails physics calculates resistance based on real world principles:
+gravity, mass, rolling resistance, and curve resistance. This is calculated
```
rwf-rr: For completeness, should "wind resistance" be added here.

Source/Orts.Common/Conversions.cs
```diff
@@ -260,8 +260,23 @@ public static class NpM
     /// <summary>
     /// Resistance conversions from and to Newtons/metre/sec
     /// </summary>
-    public static class NpMpS
+    public static class NSpM
```
rwf-rr, line 263: I don't understand the scheme for the new class names (this and below).

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -599,9 +640,21 @@ public virtual void LoadFromWagFile(string wagFilePath)
                 WagonFrontalAreaM2 = CarWidthM * CarHeightM;
             }
 
-            // Initialise key wagon parameters
-            MassKG = InitialMassKG;
+            // If Davis C value is not defined, determine it from the drag constant and area
+            if (DavisCNSSpMM <= 0)
```
rwf-rr, line 644: Should this also be qualified by BearingType? I know, strictly the bearing type does not matter for wind resistance.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -657,7 +710,7 @@ public virtual void LoadFromWagFile(string wagFilePath)
                 {
                     if (ortsFreightAnim.ShapeFileName != null && !File.Exists(wagonFolderSlash + ortsFreightAnim.ShapeFileName))
                     {
-                        Trace.TraceWarning("ORTS FreightAnim in trainset {0} references non-existent shape {1}", WagFilePath, wagonFolderSlash + ortsFreightAnim.ShapeFileName);
+                        Trace.TraceWarning("ORTS FreightAnim in trainset {0} references non-existent shape {1}", file, wagonFolderSlash + ortsFreightAnim.ShapeFileName);
```
rwf-rr: Is there a reason why using "file" here, and not "shortPath" (as in other changes)?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -720,6 +768,15 @@ public virtual void LoadFromWagFile(string wagFilePath)
                     LoadEmptyWagonFrontalAreaM2 = WagonFrontalAreaM2;
                 }
 
+                if (FreightAnimations.EmptyORTSDavis_C > 0)
+                {
+                    LoadEmptyORTSDavis_C = FreightAnimations.EmptyORTSDavis_C;
+                }
+                else
+                {
+                    LoadEmptyORTSDavis_C = NSSpMM.FromLbfpMpH2(Me2.ToFt2(LoadEmptyWagonFrontalAreaM2) * LoadEmptyDavisDragConstant);
```
rwf-rr, line 777: It seems that DavisCNSSpMM (Davis_C) is not used at all. Yet it is mentioned in the documentation.

rwf-rr (commented):
The code changes look fine to me, though I do not always fully understand the reason behind the changes.
Since I am not yet familiar with the details in the friction calculations, I would prefer if a second reviewer provided feedback before approving the PR.

Source/Documentation/Manual/physics.rst
SteelFill: Fair enough, added in next build.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: The reason the Davis A and B values require a bearing type is because the bearing type affects the calculation, since the bearing type doesn't change Davis C there's no point in requiring a bearing type.

Because OR needs **all** 3 A B C values to be calculated before it will use **any** of them `IsDavisFriction = DavisAN != 0 && DavisBNSpM != 0 && DavisCNSSpMM != 0;` adding extra constraints on the calculation just increases the number of ways to get broken friction values. Making the system less error tolerant is the opposite of the goal of this PR.

Hypothetically, someone could enter Davis A and B, but not define a bearing type and not define Davis C. That's pretty unlikely, but in that case there is enough information to calculate Davis C, and thus enough info to calculate friction in general. On the other hand, arbitrarily requiring a bearing type even for Davis C would throw away the user's work on the A and B values and produce 0 friction force for no good reason.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: Not sure, that was probably supposed to be shortPath

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: DavisCNSSpMM is used, just on anything without freight animations. LoadEmptyORTSDavis_C is specific to dynamic freight animations.

The duplicate calculations are needed to handle this system. It's a mess but it has to be supported.

PR #1121, opened 2025-07-03
Manually Override Articulation

SteelFill:
Split from #1072 

[Related Trello card](https://trello.com/c/tSHSbR36/606-simplify-repetitive-steps-of-train-creation)

This simple PR allows the OpenRails articulation system to be manually controlled. For the purposes of OR, an "articulated" rail vehicle is purely graphical, set up such that it appears to pivot about the end(s) of the body, as if pivoting around a pin attached to the adjacent rail vehicle, as opposed to pivoting about a bogie. Normally, articulation is entirely automatic, but with some restrictions to maintain compatibility with classic content (articulation doesn't apply to locomotives, articulation only works if the number of axles is more than 1, but less than 4, articulation is decided based on the position of axles relative to the center of the model). This may mean articulation is enabled/disabled in unexpected ways, resulting in rolling stock that doesn't connect or follow curves as it should.

In such cases, articulation can now be manually defined using `ORTSFrontArticulation` for the front of the rail vehicle, and `ORTSRearArticulation` for the rear of the rail vehicle. Entering ( 1 ) will enable articulation on that side of the rail vehicle (ignoring the default rules), while ( 0 ) will force disable articulation on that side. If, for whatever reason, it is desired to force the default behavior, then entering ( -1 ) will allow the default automatic process to occur.

### DIFF
--- a/Source/Documentation/Manual/features-rollingstock.rst
+++ b/Source/Documentation/Manual/features-rollingstock.rst
@@ -403,6 +403,35 @@ OR supports tilting trains. A train tilts when its .con file name contains the
 
 .. image:: images/features-tilting.png
 
+Advanced articulation control
+-----------------------------
+
+A wide variety of modern rolling stock uses articulation, in which multiple rail vehicles
+share a single "Jacobs Bogie". Open Rails offers partial support for such passenger and
+freight units by allowing one wagon to include a bogie in its 3D model while the next
+wagon removes the bogie from its 3D model. Ideally, OR will then add an invisible bogie
+to the end of the wagon without the bogie to emulate "sharing" the bogie with the previous
+wagon.
+
+However, this automatic system is limited. OR will check for wheels in the wagon's 3D
+model and will assume the wagon is articulated at one end if there are no wheels towards
+that end of the 3D model. This approach will only be used on 3D models with 3, 2, or 0 axles
+(the 1-axle case is excluded for compatibility reasons) and won't be used on locomotives.
+In some cases, this approach will result in false negative or false positive detection
+of articulation. Should the automatic articulation method not produce the expected track
+following behavior, it is now possible to manually define whether a wagon or engine
+should use the articulation behavior.
+
+.. index::
+   single: ORTSFrontArticulation
+   single: ORTSRearArticulation
+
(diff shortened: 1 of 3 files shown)


Source/Documentation/Manual/features-rollingstock.rst
```diff
@@ -403,6 +403,43 @@ OR supports tilting trains. A train tilts when its .con file name contains the
...
+.. index::
+   single: ORTSFrontArticulation
+   single: ORTSRearArticulation
+
+To forcibly enable the articulation behavior at the front of the rail vehicle, use
+``ORTSFrontArticulation ( 1 )`` and at the rear use ``ORTSRearArticulation ( 1 )``.
+Conversely, use ``ORTSFrontArticulation ( 0 )`` or ``ORTSRearArticulation ( 0 )`` to
+force disable articulation behavior. Entering a value of -1 provides the default
+automatic behavior.
+
```
rwf-rr, line 444: It is unclear to me if these parameters are required at both ends of an intermediate car, or if only the end that does not have a bogie in the 3D model needs the parameter.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -2731,34 +2733,36 @@ public void SetUpWheels()
...
-            //Certain locomotives are testing as articulated wagons for some reason.
-            if (WagonType != WagonTypes.Engine)
-                if (WheelAxles.Count != 1 && (articulatedFront || articulatedRear))
-                {
-                    WheelAxlesLoaded = true;
-                    SetUpWheelsArticulation(carIndex);
-                }
+
+            // Force articulation if stock is configured as such
+            // Otherwise, use default behavior which gives articulation if there are no axles forward/reareward on the mode,
```
rwf-rr: I think that should be "model" (missing "l").

Source/Documentation/Manual/features-rollingstock.rst
SteelFill: Yeah you only need the articulation behavior on the end that doesn't have a bogie, otherwise the bogie is sufficient to sensibly position the train car on the tracks.

PR #1122, opened 2025-07-03
Improve Center of Gravity Adjustability

SteelFill:
Split from #1072 

[Related Trello card](https://trello.com/c/w4dh0pLc/623-wagon-size-and-centering-controls)

This PR adds some extra options for managing wagon center of gravity settings. Center of gravity can now be manually defined one component at a time, allowing for the Y value (height, used for physics) and the Z value (length, used for re-centering) to be defined independently of eachother.

- **Full manual control of centering the 3D model** with the new `ORTSShapeNudge ( z )` parameter. Manual offset of the 3D model is already possible using the Z value of `CentreOfGravity ( x y z )`, but this is limited to 2 meters (for compatibility with some MSTS rolling stock). `ORTSShapeNudge` has the same effect, but without any limit at all, greatly improving flexibility and reducing confusion as this value has nothing to do with the CoG in physics. Note that `ORTSShapeNudge` should be placed AFTER `CentreOfGravity ( x y z )` in the wagon ( ) section of the engine or wagon, otherwise the result will be overridden,
- *Define the vertical (Y) component of CoG exclusively* with `CentreOfGravity ( y )`. Entering only one length value, instead of three, in `CentreOfGravity` will define only the Y (height) value of the CoG, while leaving the X and Z values unchanged. The X value is not used, and the Z value can be defined by `ORTSShapeNudge`, without any risk of one parameter overriding the data given by another. I recommend using `CentreOfGravity ( y )` to avoid worrying about the parameter parse order.
- Slight refactoring to allow both spellings of "centre" and "center" to be used in CoG parameters.


SteelFill:
> > ... I know it's still dubious to load the 3D model for the train car simulation, ...
> 
> Was it ever considered to use a tool to update the .wag file? The consist editor already loads the shape, could it modify the .wag?

Using one of/adding a new contributed program is the most viable alternative to doing it in real time, suggested by the management team. I don't know of any precedent for editing files (especially *automatically* editing files) in any of the "official" programs, and I feel changing things is asking for trouble. I'm not sure that's any less taboo than loading the 'wrong type' of file.

It's certainly possible, but if I made a lone program to just determine the size of things I'd do that on my own since it doesn't seem substantial/user-friendly enough to be a formal addition. It feels that should only be part of a hypothetical larger 'engine updater' program that does multiple things (particularly, remove old MSTS data and introduce ORTS parameters where reasonable), since such a device is a commonly requested feature.

And the least offensive (but maybe least useful) approach for now is to add some parameter to define the bounding limits of the shape in the engine/wagon, then use these precomputed bounding boxes to do the size and centering as requested. Actually determining the bounding limits is then someone else's problem (whether automated or manual).

SteelFill:
Hey Roger @rwf-rr, after this feature set being held up so long by GLTF, I decided to do what I should have done originally and set up an automated process to determine wagon sizes and edit the `Size` parameter in my wagon files accordingly. Considering that was quite easy and avoids many of the questionable aspects of the `ORTSAutoSize` feature, I have elected to remove `ORTSAutoSize`, `ORTSAutoCenter`, and `ORTSShapeBounds`. Those just turned out to be bad features that were more trouble than they were worth.

All that remains is the much smaller and simpler set of adjustments to CentreOfGravity, which should be much less troublesome. I wouldn't mind getting this one off my plate after a year.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
```diff
@@ -109,6 +109,7 @@ public FreightAnimations(STFReader stf, MSTSWagon wagon)
                 new STFReader.TokenProcessor("emptymaxbrakeforce", ()=>{ EmptyMaxBrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
                 new STFReader.TokenProcessor("emptymaxbrakeshoeforce", ()=>{ EmptyMaxBrakeShoeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
                 new STFReader.TokenProcessor("emptymaxhandbrakeforce", ()=>{ EmptyMaxHandbrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
+                new STFReader.TokenProcessor("emptycenterofgravity_y", ()=>{ EmptyCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
                 new STFReader.TokenProcessor("emptycentreofgravity_y", ()=>{ EmptyCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
```
rwf-rr, line 113: Duplicate line. emptycenterofgravity_y is parsed twice.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
```diff
@@ -1021,6 +1022,7 @@ public FreightAnimationContinuous(STFReader stf, MSTSWagon wagon)
                 new STFReader.TokenProcessor("fullmaxbrakeforce", ()=>{ FullMaxBrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
                 new STFReader.TokenProcessor("fullmaxbrakeshoeforce", ()=>{ FullMaxBrakeShoeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
                 new STFReader.TokenProcessor("fullmaxhandbrakeforce", ()=>{ FullMaxHandbrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
+                new STFReader.TokenProcessor("fullcenterofgravity_y", ()=>{ FullCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
                 new STFReader.TokenProcessor("fullcentreofgravity_y", ()=>{ FullCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
```
rwf-rr, line 1026: Duplicate line. fullcentreofgravity_y is parsed twice.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
```diff
@@ -1147,6 +1149,7 @@ public FreightAnimationStatic(STFReader stf)
             new STFReader.TokenProcessor("fullmaxbrakeforce", ()=>{ FullStaticMaxBrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
             new STFReader.TokenProcessor("fullmaxbrakeshoeforce", ()=>{ FullStaticMaxBrakeShoeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
             new STFReader.TokenProcessor("fullmaxhandbrakeforce", ()=>{ FullStaticMaxHandbrakeForceN = stf.ReadFloatBlock(STFReader.UNITS.Force, -1); }),
+            new STFReader.TokenProcessor("fullcenterofgravity_y", ()=>{ FullStaticCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
             new STFReader.TokenProcessor("fullcentreofgravity_y", ()=>{ FullStaticCentreOfGravityM_Y = stf.ReadFloatBlock(STFReader.UNITS.Distance, -1); }),
```
rwf-rr, line 1153: Duplicate line. fullcentreofgravity_y is parsed twice.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1313,18 +1313,30 @@ public virtual void Parse(string lowercasetoken, STFReader stf)
                         stf.SkipRestOfBlock();
                     }
                     break;
+                case "wagon(centerofgravity":
                 case "wagon(centreofgravity":
```
rwf-rr, line 1321: Duplicate line. Same case statement twice.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1313,18 +1313,30 @@ public virtual void Parse(string lowercasetoken, STFReader stf)
...
+
+                        stf.SkipRestOfBlock();
+                    }
+                    else // User has entered a single value, only set the Y component to this value, leave other components unchanged
+                    {
+                        InitialCentreOfGravityM.Y = initialValue;
                     }
-                    stf.SkipRestOfBlock();
                     break;
+                case "wagon(ortsshapenudge": InitialCentreOfGravityM.Z = stf.ReadFloatBlock(STFReader.UNITS.Distance, null); break;
```
rwf-rr: This depends on the order of _wagon(centreofgravity_ and _wagon(ortsshapenudge_. Is that ok? Does it work when the new parameter is added in the "original" file and when it is added via the OpenRails directory?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1719,7 +1731,6 @@ public virtual void Copy(MSTSWagon copy)
             RearArticulation = copy.RearArticulation;
             TrackGaugeM = copy.TrackGaugeM;
             CentreOfGravityM = copy.CentreOfGravityM;
```
rwf-rr, line 1737: _CentreOfGravityM_ is copied here. But I can't find _CentreOfGravityM_ in any Save() or Restore() function. I looked at the MSTSWagon and TrainCar classes.

Should it not be saved and restored?

Source/Documentation/Manual/features-rollingstock.rst
```diff
@@ -411,6 +411,32 @@ some steps of the content creation process or allow more control over content th
...
+
+.. index::
+   single: CentreOfGravity
+
+However, this does not entirely replace ``CentreOfGravity``. The Y (height) value of the CoG is
+still used by the physics system and should still be defined. In this case, simply use
+``CentreOfGravity ( y )`` where y is the CoG height in meters (or other units, as desired).
+Unlike entering all 3 values for the CoG, entering only the Y value will NOT affect the alignment
+of the 3D model, allowing the "physical" CoG to be entered separately from the "visual" CoG.
+
```
rwf-rr, line 440: I am not quite understanding this.

When an existing wagon has a problem, is it sufficient to add the _ORTSShapeNudge_ parameter (preferably in the OpenRails folder)? Or does the existing _CentreOfGravity_ parameter need to be modified to only have one (the Y) value? If so, is the X value irrelevant?

A new wagon would only need _ORTSShapeNudge_ if the offset is more than 2 meters, and in that case _CentreOfGravity_ would only need the Y value. Correct?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
SteelFill: Not duplicate! This is subtle, but it's the difference between the American spelling "cent**ER**" and the British spelling "cent**RE**". I have done the same for all CoG parameters so creators can use the spelling they are comfortable with.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: The CoG values are not manipulated in any way that can't be re-initialized from other information that is saved. As far as I see, the CoG is only affected in real time by the freight animation system, where the CoG is just re-calculated from scratch every frame. Any saved value would immediately be overwritten by the calulcated value.

It does need to be copied though, for wagons without freight animations, in which case the value doesn't change so it still doesn't need to be saved.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: So right now it would require `ORTSShapeNudge` to be placed after `CentreOfGravity` to have the expected result. Considering there probably aren't any situations where someone would enter a ShapeNudge value but wouldn't want it to be used, it should be more intuitive for that to take priority no matter what. That is not a difficult change.

Source/Documentation/Manual/features-rollingstock.rst
SteelFill: There shouldn't be any modification required for `CentreOfGravity`, `ORTSShapeNudge` can override the CoG even if all 3 X, Y, Z values are defined.

The _preferred_ solution is to just define the Y (height) with `CentreOfGravity` and the Z (longitudinal) with `ORTSShapeNudge`, which is why the documentation suggests only using `CentreOfGravity` for the Y value and nothing else.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
peternewell: I think that this needs a broader discussion (potentially with the ORMT) as having slightly different spellings will cause overall parameter bloat and possibly cause some confusion if users swap between them. 

Interestingly we have only used English parameter names previously (what about foreign users?). If we are forcing some users to use a particular language, why don't we do that for all users, regardless of their native language (why are we accommodating two different English languages, but no others?)?

If your suggested approach is the way to go, then it should be an ORTS development architectural standard, and adopted by all users/developers.

rwf-rr: And I stared so hard at the two lines. Maybe a comment helps.

SteelFill: Haha, sure I can add a comment noting the cases are for different spellings.

rwf-rr (approved):

PR #1126, opened 2025-07-05
ShapeHierarchy Attachment for More Wagon Addons

SteelFill:
Split from #1072 

[Related Trello card](https://trello.com/c/R4YwPVMN/624-shape-hierarchy-attachment-for-more-wagon-addons)

This PR adds the "ShapeHierarchy" feature currently only present in lights #917 to particle emitters, freight animations, sounds, camera views, and interior views. This allows said addons to be attached to parts of a 3D model other than the MAIN body, so a freight animation attached to a bogie (using `ShapeHierarchy( BOGIE1 )`) will move as the bogie moves. This can be used on any sub object, even silly things like attaching a particle emitter to a wheel (in which case, the particles will spin around wildly). If the given matrix name doesn't exist in the model, a warning will be added to the log and the addon will attach to the MAIN sub object (use any sort of shape viewer to see the names of sub objects).

- To attach these addons to different sub-objects use `ORTSShapeHierarchy ( MATRIXNAME )` in a `Light (` block (existing feature), `Stream (` block (in an .sms file), `ORTS3DCab (` `ORTSAlternate3DCabViewPoint (` `Inside (` `ORTSAlternatePassengerViewPoint (` blocks for 3D interiors, `FreightAnimStatic (` `FreightAnimContinuous (` blocks for ORTS freight animations (MSTS freight animations are not supported), `HeadOut (` block for head out views (note that ORTS specific parameter in the HeadOut ( block **must come AFTER the x y z position numbers**!), in any **particle FX** subblock (again, the ORTS specific parameters must be placed **AFTER the position, direction, and radius numbers**!), and in the `Tr_CabViewFile (` header of a .cvf file (for a .cvf file, `ORTSShapeHierarchy` must be included _once for each view point_, which also means each viewpoint can have a different attachment).
- A lot of addons were also missing the ability to reposition the location of the addon relative to the sub object to which it is attached. For a `Stream (` in the .sms file, use `ORTSPosition ( x y z )` to define the position of the stream's sound emitter, `ORTS3DCab (` `ORTSAlternate3DCabViewPoint (` `Inside (` `ORTSAlternatePassengerViewPoint (` blocks all now accept an `ORTSShapeOffset ( x y z )` input to allow shifting the position of the interior shape file (note that **this will NOT shift the view point**, make sure to update the view point location if shape offset is used), `FreightAnimContinuous (` blocks now accept the `Offset ( x y z )` and `Flip ()` parameters, which function the same as the counterparts in a static freight animation

Also includes some refactoring needed for this to work as intended. Rolling stock animations are now processed before updating attached object positions to ensure nothing gets one frame out of sync. Each poseable shape now keeps track of its resultant matrices each frame so that doesn't need to be calculated separately by each system that needs a resultant matrix (if you need the position of a sub-object, please use PoseableShape.ResultMatrices[MAT_IDX] from here on out). Audio management reorganized slightly in a way that seems simpler and avoids some multithread nonsense.

To-Do
- [ ] Update implementation to work with GLTF when that is ready


SteelFill:
Removing from unstable so #1160 can be added to unstable builds.

This is still ready for review and I have no additional features planned. It would be helpful to free up the changes here to give room for everything else to change these associated files.

twpol:
I have split the Trello card for this PR into parts to match the separated implementations and approved it with one condition (see below): https://trello.com/c/R4YwPVMN/624-shape-hierarchy-attachment-for-more-wagon-addons

ORMT's condition for this work is that any parameters which are **overriding/replacing** those from other files include "Override" in the name, either directly or via a parent block.

I am not sure anything in this PR qualifies, perhaps `ORTSShapeOffset`, but please consider names with regards the above. Thanks!

SteelFill:
> I have split the Trello card for this PR into parts to match the separated implementations and approved it with one condition (see below): https://trello.com/c/R4YwPVMN/624-shape-hierarchy-attachment-for-more-wagon-addons
> 
> ORMT's condition for this work is that any parameters which are **overriding/replacing** those from other files include "Override" in the name, either directly or via a parent block.
> 
> I am not sure anything in this PR qualifies, perhaps `ORTSShapeOffset`, but please consider names with regards the above. Thanks!

From what I think you are going for, shape offset doesn't count as an override. It doesn't replace any data from other parameters (in this case, the position of the interior shape) as there is no other way to change this data. Before now, this position was always set to 0, 0, 0 and could not be changed, but with ShapeHierarchy I found situations where 0, 0, 0 isn't suitable, hence the new parameter.

PR #635, opened 2022-04-03
Minor steam locomotive issues.

peternewell:
Fix an issue with auxiliary water tender water level not being restored correctly.

https://bugs.launchpad.net/or/+bug/1967660

PR #1128, opened 2025-07-05
Particle Emitter Overhaul

SteelFill:
Discuss on [Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38982-particle-emitter-overhaul/)
Relevant [Trello Card #635](https://trello.com/c/1FDUw5Kg)

This PR makes many changes to particle emitters and adds loads of new customization options which can be applied to each individual emitter for greatly increased control over emitter behavior and appearance. The goal is to make particles look better out of the box, and to give basically endless options for custom particles to fit the unique character of each locomotive.

Difference between old and new particle emitter implementation

Hopefully it's quite clear that the changes have made steam effects more voluminous and forceful (without any changes required to the .eng file). There are fewer gaps between particles, particles expand much more when there is lots of exhaust (the main stack and cylinder cocks), and particles are actually being forced away from the emitter location instead of lazily floating upward. It makes things feel much more dynamic and powerful. The changes are much less visible for diesel locomotives, where less extreme volumes of exhaust are involved.

### The universal changes, applicable without any user input required, include:

- Replacing the time-delta system used to determine particle spawning time with an accumulating **'bucket' of particles** to more accurately determine when particles should be emitted, reducing some jitter that was present before
- **Improving particle limit behavior** so that old particles are deleted as the limit is hit, instead of preventing new particles from spawning, and limiting the spawning rate of particles to try and avoid hitting the particle limit in the first place
- Slightly **randomizing particle spawn time** to avoid particles from different emitters suspiciously being emitted at the exact same time
- Tweaking the other RNG used to make particle behavior a **bit less periodic**
- Calculating the spawning **position of a particle 'in between' frames** when more than one particle is emitted per frame, which greatly improves particle behavior when particles are emitted very quickly on a train moving at high speed
- Using the **train car velocity** (also used by the audio system) to determine the emitter velocity, rather than re-calculating the velocity for each individual emitter, which is more efficient and probably contributes to the reduced jitter
- Expanding the size of particles in two phases: instead of particles just expanding over time, they now additionally **expand rapidly between the emission time and the settling time**, with the amount of expansion depending on the difference between the emission speed and final speed of the particles, to better replicate the look of real exhaust plumes
- Nudging particles **upward as they expand** in size so they don't clip through the ground so obviously
- **Increasing settling time** by default, but decreasing it as particles are emitted at a higher speed, which gives particles more momentum without introducing unexpected behavior on high-speed particles
- Swapping the linear decay of particle velocity with a **quadratic decay**, which makes particles follow a more natural path and improves handling of high-speed particles
- **Rotating particles randomly** when they spawn to add a little bit more visual interest
- **Removing of a lot of arbitrary fudge factors**, thanks to the better handling of high-speed particles and more appropriate values for the density of steam
- **Refactoring the emitter rate functions** to be a bit more self-explanatory; they are now divided between setting rate in particles per second, volumetric flow rate, or particle initial velocity, with the un-specified properties being specified automatically
- Adjusting existing particle emitter calls to use the changed system, which should improve various existing effects, let me know if certain particle effects appear to be too strong/too weak as a lot of this is set by magic numbers not attached to the simulation (no major changes made to the calculations driving emitter speeds)
- Added default smoke and steam textures to the Content folder, which will be used if the textures can't be found in the GLOBAL folder (this will be useful for folks without a copy of MSTS or its files, but otherwise won't do anything)
- Added support for overriding particle opacity in the particle emitter and shader

### However, even more is possible when changing emitters with new parameters:

```
	Effects (
		FX_TYPE (
			FX_NAME (
				Comment ( Required parameters. )
				ORTSPosition ( 0m 4.75m 2.0m )
				ORTSInitialVelocity ( 0 1 0 )
				ORTSParticleDiameter ( 0.15m )
				
				Comment ( Optional parameters. )
				ORTSPositionVariation ( 0m 0m 0m )
				ORTSInitialVelocityVariation ( 0.1 )
				ORTSFinalVelocity ( 0 1m/s 0 )
				ORTSFinalVelocityVariation ( 0.75m/s )

				ORTSLifespanMultiplier ( 1.0 )
				ORTSLifespanVariation ( 0.5 )
				ORTSMomentumMultiplier ( 1.0 )
				ORTSMomentumVariation ( 0.1 )

				ORTSInititalExpansion ( 1.0 )
				ORTSExpansionSpeed ( 4.0 )
				ORTSRotationVariation ( 0.25 )
				ORTSWindMultiplier ( 1.0 )

				ORTSParticleOpacity ( 1.0 )
				ORTSPipeArea ( 0.018m^2 )
				ORTSRateMultiplier ( 1.0 )
				ORTSUseChaoticRandomization ( false )

				Comment ( Advanced optional parameters. )
				ORTSMaxParticles ( 2500 )
				ORTSGraphic ( "smokemain.ace" )
				ORTSGraphicAtlasLayout ( 4 4 )
			)
		)
	)
```

The exact functionality of each is described in the manual and would take too much space here. Hopefully it is mostly self-explanatory how each option can control the behavior and randomization of a particle emitter to produce new and unique effects. It is worth mentioning that most of these parameters are _unitless_ and act as multipliers for the thing they control, rather than directly setting a value. The is no "correct" or "physically accurate" value to put in most places, this is all down to tweaking things until it looks good.

Special mention should be made to `ORTSPositionVariation ( x y z )`, which sets the +/- deviation (as in, `ORTSPositionVariation ( 0.10m 0m 0m )` would allow particles to spawn 0.10m to the right _or_ to the left from the original position, for a total variation of 0.20m) in emitter position for each particle and is perfect for _rectangular exhaust ports_,  though be sure to also set the size of the exhaust port with `ORTSPipeArea ( a )` as the automatic calculation assumes a circular port. `ORTSUseChaoticRandomization ( 0/1 )` is a bit unique as it changes the randomness of each particle's properties to be completely random, with no regard to the random value generated for the previous particle. The default randomness algorithm is "smooth" in comparison, where each particle will randomly deviate from the previous particle only slightly. Which to use is a matter of taste.

And for custom textures, `ORTSGraphic ( texname )` will first search from the location of the eng/wag file, and if the texture isn't found there it'll search GLOBAL\TEXTURES, and if it still isn't there then the Content folder of the OR installation will be checked last. `ORTSGraphicAtlasLayout ( w h )` should be set depending on how the texture is set up. For MSTS smoke textures, the texture is actually a 4x4 grid of 16 total sprites. Each particle randomly selects one of these sprites. That grid would be represented by `ORTSGraphicAtlasLayout ( 4 4 )` (this is the default value, you do not need to enter this). If the texture used has only one sprite, you would use `ORTSGraphicAtlasLayout ( 1 1 )`, if it's a grid of 8 sprites that is 2 wide and 4 tall, you'd use `ORTSGraphicAtlasLayout ( 2 4 )`, and so on. Incorrect settings will be very obvious as you will see artifacting in the displayed particle. Note that particles are squares, so each sprite should ideally be a square to avoid squash/stretch effects.

And it should be noted that `ORTSPosition ( x y z )`, `ORTSInitialVelocity ( i j k )`, and `ORTSParticleDiameter ( d )` are replacements for the default MSTS particle data, but the MSTS formatting for position, velocity, and size can still be used so long as **all ORTS parameters come after the MSTS data**.

```
	Effects (
		FX_TYPE (
			FX_NAME (
				0 4.75 2.0
				0 1 0
				0.15
				
				Comment ( Now that MSTS numbers are given, ORTS parameters can follow: )
				ORTSabc ( xyz )
				...etc...
			)
		)
	)
```

I recommend just dropping the MSTS parameters entirely as it's not as clear what each number means. And the placement/order of the ORTS parameters doesn't matter, while the order of the MSTS numbers matters a lot.

It's also worth noting this PR is _not_ about any of the simulated calculations driving particle emitters. Those behaviors will be similar to before, even when that means something isn't realistic. Particle effects that weren't driven by simulation data before are still not driven by simulation data now. However, ones that are driven by simulation data should behave more accurately as fudge factors have been removed and calculations adjusted to consider basic fluid flow mechanics.


twpol:
> It is my intention to add a patch for steam locomotives similar to the one attached (I am not sure whether a couple of the "fudge factors" will be deleted from the patch).
> 
> I think that it will make it simpler for users, rather then having to add a large stack of parameters that may not be easy to identify from prototypical stock.

Thanks for working on that Peter, I also agree that we should have fewer of these "fudge factor" parameters (I called them "code tuning" parameters in my post: https://www.elvastower.com/forums/index.php?/topic/38982-particle-emitter-overhaul/page__view__findpost__p__320508)

Csantucci:
Phillip, you could consider attached patch which solves this https://www.elvastower.com/forums/index.php?/topic/32640-or-newyear-mg/page__view__findpost__p__322663 .
[ddsvsacepatch.diff.txt](https://github.com/user-attachments/files/23567811/ddsvsacepatch.diff.txt)

SteelFill:
Removing from unstable to give room for hot reloading, which for some silly reason seems to be conflicting.

SteelFill:
Merging into unstable is fixed, but I want to have some builds with #1263 in so that the "harmless" parts of this can be merged and the scope of this PR reduced to just the new features. Can't merge both PRs simultaneously, and I don't quite trust myself to not irreversibly screw up everything by rebasing this branch onto my other branch.

PR #1132, opened 2025-07-08
Fixes For Correct Questionable Braking Parameters

SteelFill:
Since `TrainPipeLeakRate` was fixed back in #912, some users have noticed their locomotives are unable to keep up with the leak rate defined in the .eng file because the leak rate was set too high (presumably in an attempt to counteract the bug preventing the leak from working as expected) and/or because the locomotive compressor was unreasonably weak.

The simplest approach that does not interfere with creator's choice (I cannot judge someone for intentionally wanting a high leakage for an activity or something) nor require excessive effort from the user is to use the Correct Questionable Braking Parameters setting.

- `TrainPipeLeakRate` is limited to 2.5 psi/min if CQBP option is enabled
- Fixed calculation of main res charging rate for MSTS locomotives, so MSTS locomotives should have more appropriate brake pipe charging performance
- Cleaned up `CorrectBrakingParams` by moving the electric locomotive dynamic brake setup timer to only be adjusted if CQBP is enabled instead of all the time (I'm not sure why this feature is there, but it's not correct to enforce arbitrary limits like that all the time)
- Adjusted documentation for CQBP to reflect what the option actually does, as the documentation was missing many pieces

A note on MSTS main res charging rate: The previous code would never function as it was checking for `MainResChargingRatePSIpS <= 0` but MainResChargingRatePSIpS defaulted to 0.4 psi/s, so that calculation never ran. The charging rate now defaults to -1 to indicate it has not been set, which allows the correction code to run as expected on MSTS locomotives (but not on ORTS locomotives, where the charging rate should be defined already). MSTS locomotives should no longer get stuck with the pitifully slow 0.4 psi/s charge rate, and the minimum charge rate has been changed to 0.025 m^3/s so main res volume doesn't impact the 'actual' minimum rate of air pumping.

### DIFF
--- a/Source/Documentation/Manual/options.rst
+++ b/Source/Documentation/Manual/options.rst
@@ -863,21 +863,23 @@ are a problem for OR, which has a more sophisticated braking model. The
 problem usually is that the train brakes require a long time to release,
 and in some times do not release at all.
 
-.. index::
-   single: AirBrakesAirCompressorPowerRating
-
 The following checks and corrections are performed if the option is
-checked (only for single-pipe brake system):
+checked:
 
 - if the compressor restart pressure is smaller or very near to the max
   system pressure, the compressor restart pressure and if necessary the max
-  main reservoir pressure are increased;
+  main reservoir pressure are increased (single pipe air brakes only)
 - if the main reservoir volume is smaller than 0.3 m\ :sup:`3` and the
   engine mass is higher than 20 tons, the reservoir volume is raised to 0.78
-  m\ :sup:`3`;
-- the charging rate of the reservoir is derived from the .eng parameter
-  ``AirBrakesAirCompressorPowerRating`` (if this generates a value greater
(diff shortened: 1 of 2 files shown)


SteelFill:
> I agree on all points in your comment except the first one. If CQBP is enabled I'd not disable TrainPipeLeakRate; instead I'd cap it if it is higher than a default, reasonable value (which might eventually also depend from the locomotive compressor charging rate) . This is also more coherent to what CQBP does with other parameters.

I suppose a couple psi/minute is unlikely to completely break things, much less than a few psi per second, but it won't be foolproof. If that's still too much there's probably something seriously wrong with the setup beyond just leaking, though disabling leaks entirely is still the safer option at minimal gameplay cost (the current leak system doesn't change a lot about train handling, really just keeps the compressor running).

twpol (approved):

PR #1133, opened 2025-07-08
Minor Fix for Brake Pipe Charging

SteelFill:
Back with #912 I implemented some changes to add smoothing to brake pipe charging. One of these smoothing factors was supposed to reduce the rate of charging as the _brake pipe pressure_ approached the main res pressure (as the pressure drop across the feed valve decreases, the flow rate decreases due to fluid dynamics), but it was instead mistakenly implemented as a reduction when the _equalizing res pressure_ approached the main res pressure.
Obviously, that doesn't make as much sense, and we can even see the if statement uses the difference between MR and BP pressure, so I do not know how that MR vs ER pressure calculation made it to the final version of the air brake PR.


cesarBLG (approved):

SteelFill:
@twpol James, I saw you going through the bug fixes from the last couple of weeks and saw you missed this one...but that was my fault, I forgot to set this one to target 1.6. Now it should be good to go.

twpol (approved):

PR #1137, opened 2025-07-14
Apply brakes at startup on minimal reduction

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/38763-demo-model-1-update/page__view__findpost__p__319969


SteelFill (approved):
Fix works great. I encountered this problem previously when setting up some manually lapped braking systems, now I don't have to worry as much about the train rolling away when starting the session.

PR #1139, opened 2025-07-20
Fix for bug https://bugs.launchpad.net/or/+bug/2117357. 

sweiland-openrails:
https://bugs.launchpad.net/or/+bug/2117357

Exception index of of range when opening Train Car Operations window:

reproduction:

- after starting OR open http://localhost:2150/TrainCarOperations/index.html: Train Car Operations
- select last car (click twice on it) on above mentioned page
- decouple last car
- open Train Car Operations with F9
- Exception in:

                    if (LastCarIDSelected == null)
                    {
                        LastCarIDSelected = PlayerTrain.Cars[SelectedCarPosition].CarID;
                    }

  SelectedCarPosition has value of last car before uncoupling --> Index was out of range


sweiland-openrails:
hi Mauricio,

Do you have time to have a look at this small fix? Not that serious but I hate bugs which cause exceptions which make OR end. Just when you are in the middle of an activity.

thanks, Siebren.

cjakeman (approved):
Hi Siebren,

Thanks for this simple fix.

Approved and merged,

Chris

PR #1143, opened 2025-07-25
Status in Work Orders popup set too fast

sweiland-openrails:
Status in Work Orders popup set too fast, see https://bugs.launchpad.net/or/+bug/2118776

### DIFF
--- a/Source/Orts.Simulation/Simulation/Activity.cs
+++ b/Source/Orts.Simulation/Simulation/Activity.cs
@@ -1398,11 +1398,7 @@ namespace Orts.Simulation
                     }
                     if (nWagonListCars == wagonIdList.Count)
                     {
-                        if (((wagonIdList[0] == trainItem.Cars[0].CarID) && (wagonIdList[wagonIdList.Count - 1] == trainItem.Cars[trainItem.Cars.Count - 1].CarID)) || 
-                            ((wagonIdList[0] == trainItem.Cars[trainItem.Cars.Count - 1].CarID) && (wagonIdList[wagonIdList.Count - 1] == trainItem.Cars[0].CarID)))
-                        {
-                            return trainItem;
-                        }
+                        return trainItem;
                     }
                 }
             }


sweiland-openrails:
This fix is not that important, so just leave it out of v1.6.

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1376,33 +1376,28 @@ private Train matchesConsist(List<string> wagonIdList)
...
         /// </summary>
         /// <param name="wagonIdList"></param>
         /// <returns>train or null</returns>
         private Train matchesConsistNoOrder(List<string> wagonIdList)
         {
             foreach (var trainItem in Simulator.Trains)
             {
-                int nCars = 0;//all cars other than WagonIdList.
-                int nWagonListCars = 0;//individual wagon drop.
+                int nWagonListCars = 0;
```
SteelFill, line 1392: We need an additional check here to verify that the train `trainItem` is actually located on the siding that the drop off needs to be left at. Right now, this will check **every** train, even ones not on the siding, which can and does result in the work order being completed when the set of cars has been uncoupled somewhere other than the siding it should be at (the only check for being at the correct siding is done on the _player_ train).

SteelFill (changes requested):
Unfortunately, this fix reveals some new problems. While the code for matchesConsistNoOrder is cleaner and makes sense, the implementation of EventWrapper.Triggered means that Drop Off work orders will be marked "Done" as long as the _player train_ is at the drop off siding and the wagons to be dropped off isn't coupled to anything else. Note that this means the drop off wagons can be literally anywhere for the work order to be completed, even if that means they aren't at the proper siding.

While the old implementation does mark drop off work orders as "Done" while the drop off wagons are still attached, it also explicitly excludes static trains from the drop off check, so it's impossible to get the work order marked as "Done" if the wagons have been uncoupled in the wrong place.

Neither approach is ideal, we want something that _both_ only completes the work order after the wagons have been uncoupled _and_ only completes the work order when the wagons are where they are supposed to be, so some additional behavior is needed to ensure that the uncoupled wagons are actually where they need to be.

sweiland-openrails:
hi Phillip, you are right. I did not have a look at the siding check. In the added fix I changed that. Hope the total fix now covers all situations. Regards, Siebren.

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1376,33 +1376,28 @@ private Train matchesConsist(List<string> wagonIdList)
...
-                //support individual wagonIdList drop
-                if (trainItem.Cars.Count - nCars == (wagonIdList.Count == nWagonListCars ? wagonIdList.Count : nWagonListCars))
+                if (nWagonListCars == trainItem.Cars.Count)
                 {
-                    if (excludesWagons(trainItem, wagonIdList)) listsMatch = false;//all wagons dropped
-                    
-                    if (listsMatch) return trainItem;
-                    
+                    return trainItem;
                 }
```
SteelFill: Unfortunately, you can't have this check set to match the train to the wagon list exactly. There are activities out there where a drop-off work order gives only the _first and last_ wagon IDs to drop off. The intent is that the work order is easier to read than listing all 30 wagons but still completes when you drop all of them off (because you drop off the two listed ones as requested, and all the 'extra' wagons are ignored).

image

We should only be checking if the train _contains all_ the required wagons `nWagonListCars == wagonIdList.Count`, not checking if the train is exclusively composed of the required wagons.

sweiland-openrails: Oeps, I really do not know enough about activities. Thanks for your comments. Next time I'll start a discussion first.

I've undone the previous fix in matchesConsistNoOrder. So that the function returns when the cars in the list are found.

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1376,33 +1376,30 @@ private Train matchesConsist(List<string> wagonIdList)
...
-                    if (listsMatch) return trainItem;
-                    
+                    if (wagonIdList.Contains(item.CarID))
+                    {
+                        nWagonListCars++;
+                    }
+                    if (nWagonListCars == trainItem.Cars.Count)
+                    {
+                        return trainItem;
+                    }
```
SteelFill: This is still trying to find a train that matches the work order list exactly. In the case of a work order giving only the first and last wagons, nWagonListCars will be 2, but trainItem.Cars.Count is going to be much bigger, so the two will not be equal.

`nWagonListCars == wagonIdList.Count` would check if all the required wagons are contained in the train, while allowing for extra wagons to still be attached. (And would still fix the original problem where dropping off just one of the desired wagons would be enough to complete the whole work order.)

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1376,33 +1376,30 @@ private Train matchesConsist(List<string> wagonIdList)
...
         /// </summary>
         /// <param name="wagonIdList"></param>
         /// <returns>train or null</returns>
         private Train matchesConsistNoOrder(List<string> wagonIdList)
         {
             foreach (var trainItem in Simulator.Trains)
             {
-                int nCars = 0;//all cars other than WagonIdList.
-                int nWagonListCars = 0;//individual wagon drop.
+                int nWagonListCars = 0;
```
SteelFill, line 1392: I think we want a check here at the beginning of the loop to verify that the trainItem is _not the player train_, to guarantee that the work order won't get marked as done until uncoupling has actually happened.

sweiland-openrails:
hi Phillip, see commit https://github.com/openrails/openrails/pull/1143/commits/36d87c32be8be66bfc955738f76c46a82be3c42a.

- function matchesConsistNoOrder now returns the wagon when "nWagonListCars == wagonIdList.Count" as advised. So it returns when the tested train contains all the wagons in the activity list. 
- however if more wagons are dropped as mentioned in the activity list the above condition is not sufficient, it will return the train also. Therefore I added a test for the first and last wagon in the activity list. They should match the first and last wagon of the the tested train. I guess in 99% of the drop off's the sequence of the wagons in the activity list will match those in the drop off train. This will also take care of the fact that for long lists in the activity only the first and last wagon is mentioned.
- I also added the condition "consistTrain.TrainType == Train.TRAINTYPE.STATIC". Not that necessary, the test on the first and last wagon should be sufficient. But anyway in some situations it might help. Or should I test the traintype for not being Train.TRAINTYPE.PLAYER and Train.TRAINTYPE.AI_PLAYERDRIVEN?

Thanks for looking into this.

Source/Orts.Simulation/Simulation/Activity.cs
```diff
@@ -1376,33 +1379,34 @@ private Train matchesConsist(List<string> wagonIdList)
...
+                    {
+                        nWagonListCars++;
+                    }
+                    if (nWagonListCars == wagonIdList.Count)
+                    {
+                        if (((wagonIdList[0] == trainItem.Cars[0].CarID) && (wagonIdList[wagonIdList.Count - 1] == trainItem.Cars[trainItem.Cars.Count - 1].CarID)) || 
+                            ((wagonIdList[0] == trainItem.Cars[trainItem.Cars.Count - 1].CarID) && (wagonIdList[wagonIdList.Count - 1] == trainItem.Cars[0].CarID)))
+                        {
+                            return trainItem;
+                        }
```
SteelFill: I can not agree with adding this condition. While this technically accounts for the specific case of a truncated work order, this adds a soft requirement to _every_ work order for the consist to be in a similar order to the order listed. That requirement for order did not exist before and shouldn't be added.

For example, I have one activity with a drop off order like this:
```
32779 - 0
32779 - 1
32778 - 0
32778 - 1
32778 - 2
32778 - 3
32777 - 0
32777 - 1
```

This is a drop off, not an assemble train at location, implying the order of the 3 sets of wagons from consists 32777, 32778, and 32779 is not intended to matter. But with this condition, someone who casually puts consist 32779 in between 32778 and 32777 would be unable to complete the work order because 32779 - 0 wasn't at either end of the drop off.

A potentially worse case is to have individual wagons in the wrong order because two of the consists weren't facing the same direction, such as:
```
32779 - 1 ( - 1 at front instead of - 0)
32779 - 0
32778 - 0
32778 - 1
32778 - 2
32778 - 3
32777 - 0
32777 - 1
```
This also wouldn't work because 32779 - 0 isn't at either end but would be even more confusing to fix. I suspect most players would see they have all the wagons in the intended spot, then give up as it makes no sense why it wouldn't complete.

Simply checking if the static consist contains all the required cars with no additional checks is going to give the behavior closest to what people expect. Yes, hypothetically this allows a work order to complete with something extra like a caboose attached to the train, but that's always been allowed and is one of the least problematic ways to go off-script.

sweiland-openrails: No problem. I've removed this extra check on the first and last wagon.  You've got more experience in this area. Thanks.

PR #1152, opened 2025-08-12
fix: Clean up multiple issues with data logger

twpol:
- Add explicit setting for steam power curve data
- Label and make steam data options properly exclusive
- Correct mismatched columns (multiple)
- Fix memory column


cjakeman (approved):
Hi James,

Much neater with the #DEBUG.

Approved and merged.

Thanks,

Chris

PR #1153, opened 2025-08-17
Fix TrItems in track viewer (events were over-writing mileposts)

rwf-rr:
The track viewer also displays activity events. It uses the TrItemTable to capture the events. The bug causes the events to over-write some existing TrItem, instead of appending them to the TrItemTable.

TrItemTable is an array, making it more complex to append.

An alternative fix would be to change the TrItemTable into a List. But that affects a lot of main code.

The bug was discovered because the path chart (and track view) was missing some mileposts.

**Update**: I added another, unrelated commit. It adds the path name to the chart window title. It has been tested with multiple languages.


Source/Contrib/TrackViewer/Editing/Charts/DrawPathChart.cs
rwf-rr, line 1: This is unrelated to the fix. It adds the path name to the chart title. I had this done and tested in another branch (that I am abandoning).

sweiland-openrails (approved):

PR #1155, opened 2025-08-20
fix for illegal characters in refeerence to Shape file

cjakeman:
Where a World File contains a bad reference to a Shape File (e.g. a path containing an illegal ">" character, this led to a fatal error without identifying the file.

This fix changes that to a Warning in the log, so Open Rails can continue, and provides location information.

It solves a problem reported by Chris Gerlach.


Csantucci (approved):
That's OK to me, thanks

PR #1156, opened 2025-08-22
Fix incorrectly disabled options in train operations window

cesarBLG:
https://bugs.launchpad.net/or/+bug/2121205

The new Train Operations window disables many options that are required for standard handling of trains in Europe, which were available in the old implementation:

- Opening/closing of angle cocks in the first and last car of the consist, to empty the brake pipe. This is a mandatory operation that has to be done after a train is formed, to check brake pipe continuity.
- MU command for non-diesel locomotives. Someone decided that electric locomotives must always be MUed, and only allowed de-MUing of diesel locomotives. This is not correct, if you disconnect the MU wire of an electric locomotive, it will not be remote controlled.

### DIFF
--- a/Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
+++ b/Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
@@ -1189,6 +1189,7 @@ namespace Orts.Viewer3D.Popups
                     if (wagon.PowerSupply.BatterySwitch.Mode == BatterySwitch.ModeType.AlwaysOn)
                     {
                         Texture = BattAlwaysOn32;
+                        TrainCarViewer.BatterySwitchOn = wagon.PowerSupply.BatterySwitch.On;
                     }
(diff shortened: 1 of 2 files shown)


cesarBLG:
I'm not familiar with the DPU system, so I decided to provide a simpler variant of GetMultipleUnitsConfiguration() to be used for the Train Car Operations window, but preserving the old behavior for diesels. In MSTSDieselLocomotive, GetMultipleUnitsConfiguration() also checks the DPU Id, which I don't know about.

If GetMultipleUnitsConfiguration() was only intended to be used for DP locomotives, then it may be better to just avoid checking it in the Train Car Operations window. If it should be valid for all MU variants, then it has to be implemented for electric (and probably steam) locomotives, as I did. Both alternatives are fine for me.

mbm-OR:
Hi César,
Thank you for providing a way to avoid using localised strings for core operations.
This fix was expected a long time ago.
I will try it out as soon as possible.

Regards,
Mauricio

PR #1158, opened 2025-09-05
Temporary fix for bug 2121985: F9 TCO out-of-range after resume (release/1.6)

rwf-rr:
Bug [2121985 ](https://bugs.launchpad.net/or/+bug/2121985)- it is present in **1.6-rc8**

This is a temporary fix. It prevents writing an out of range index into the save file. As a comment it contains a better fix for the Window, but I could not come up with a simple fix for the Webpage.

The problem with the Web-Page happens when the Window is closed. Thus the web page cannot easily change state in the window.

For master, we will need a better fix. There are issues with the interaction between the Window and the Web-Page. Maybe have a separate state object and use the window and web-page objects for the UI parts only.

### DIFF
--- a/Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
+++ b/Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
@@ -29,6 +29,7 @@ using ORTS.Common;
 using ORTS.Common.Input;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 
@@ -167,6 +168,12 @@ namespace Orts.Viewer3D.Popups
             outf.Write(Location.Width);
             outf.Write(Location.Height);
 
+            // rwf-rr: temporary fix for bug 2121985
+            if (CarPosition >= Owner.Viewer.PlayerTrain.Cars.Count)
+            {
+                Trace.TraceWarning("TrainCarOperationsViewerWindow.CarPosition {0} out of range [0..{1}]", CarPosition, Owner.Viewer.PlayerTrain.Cars.Count - 1);
+                CarPosition = Owner.Viewer.PlayerTrain.Cars.Count - 1;
+            }
             outf.Write(CarPosition);
             outf.Write(ResetAllSymbols);
         }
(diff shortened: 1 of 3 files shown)


Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
```diff
@@ -574,6 +581,9 @@ public override void PrepareFrame(ElapsedTime elapsedTime, bool updateFull)
 
                 //required by traincarwindow to ModifyWindowSize()
                 windowHeight = Vbox != null ? Vbox.Position.Height : 0;
+
+                // rwf-rr: part of debugging bug 2121985
+                // System.Diagnostics.Debug.Assert(CarPosition < PlayerTrain.Cars.Count, "Viewer SelectedCar (index) out of range");
```
twpol: I'm happy with the temporary fix including the comment saying as much, but please could we remove the commented-out debugging code

twpol (changes requested):

Source/RunActivity/Viewer3D/Popups/TrainCarOperationsViewerWindow.cs
rwf-rr: Done.

Source/RunActivity/Viewer3D/Popups/TrainCarOperationsWindow.cs
```diff
@@ -642,6 +649,13 @@ public override void PrepareFrame(ElapsedTime elapsedTime, bool updateFull)
                     SupplyStatusChanged = false;
                     Layout();
                     updateLayoutSize();
+
+                    // rwf-rr: potential partial fix for bug 2121985
+                    // if (trainCarViewer.CouplerChanged && CarPosition >= Owner.Viewer.PlayerTrain.Cars.Count)
+                    // {
+                    //     SelectedCarPosition = CarPosition = Owner.Viewer.PlayerTrain.Cars.Count - 1;
+                    //     LastCarIDSelected = PlayerTrain.Cars[SelectedCarPosition].CarID;
+                    // }
```
twpol, line 637: Looks like we still have this block.

rwf-rr: I left this one because it is a permanent fix for the F9 window. I wanted to keep a record of this option.

rwf-rr: @twpol I would like to leave this last comment in there, as it will guide a future permanent fix. Please let me know if you agree.

PR #1159, opened 2025-09-16
Skip OR warnings about TSRE-specific token Ruler

Csantucci:
Feature request: https://www.elvastower.com/forums/index.php?/topic/39102-warning-skipped-unknown-token-ruler-in/

### DIFF
--- a/Source/Orts.Parsers.Msts/TokenID.cs
+++ b/Source/Orts.Parsers.Msts/TokenID.cs
@@ -1718,6 +1718,9 @@ namespace Orts.Parsers.Msts
         MaxStackedContainers,
         Length,
         Flipped,
+
+        // TSRE specific
+        Ruler
     }
 
 }
--- a/Source/Orts.Formats.Msts/WorldFile.cs
+++ b/Source/Orts.Formats.Msts/WorldFile.cs
@@ -82,6 +82,7 @@ namespace Orts.Formats.Msts
         {
             TokenID.VDbIdCount,
             TokenID.ViewDbSphere,
+            TokenID.Ruler,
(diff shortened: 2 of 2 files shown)


Source/Orts.Parsers.Msts/TokenID.cs
```diff
@@ -1718,6 +1718,9 @@ public enum TokenID : uint
         MaxStackedContainers,
         Length,
         Flipped,
+
+        // TSRE specific
+        Ruler
```
cesarBLG, line 1723: It's a simple fix, thanks. My only question is whether ruler appears in compressed .w files and as such needs to be assigned a specific number. I think TSRE does not compress .w files, so it's not required, right?

rwf-rr: I looked at TSRE, https://github.com/GokuMK/TSRE5/blob/6c2b51123fa2b36252ee7984f4d79a5d87ae99b1/WorldObj.h#L49:

Ruler has the number 22 assigned. But the rest of the numbers do not match with ORTS. For example, platform is 11 in TSRE, (3)60 in ORTS. So I (not the author) think the PR is ok as is.

cesarBLG: I don't think the numbers that you pointed from the TSRE sources have anything to do with compressed .w files. From my understanding of the compressed SBR format, each token is assigned a value, but I'm not familiar with that either. Anyway, .w files produced by TSRE are not compressed, so this is not a problem.

cesarBLG (approved):

PR #1160, opened 2025-09-21
Route Based TTrack Sounds

peternewell:
This PR adds the ability to define track based sounds "within" the route, and thus removes need for individual stock to have track sounds associated with it.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
@@ -1890,7 +1890,7 @@ namespace Orts.Simulation.RollingStocks
 
             outf.Write(WheelBrakeSlideProtectionActive);
             outf.Write(WheelBrakeSlideProtectionTimerS);
-            outf.Write(AngleOfAttackRad);
+            outf.Write(AngleOfAttackmRad);
             outf.Write(DerailClimbDistanceM);
             outf.Write(DerailPossible);
             outf.Write(DerailExpected);
@@ -1947,7 +1947,7 @@ namespace Orts.Simulation.RollingStocks
 
             WheelBrakeSlideProtectionActive = inf.ReadBoolean();
             WheelBrakeSlideProtectionTimerS = inf.ReadInt32();
-            AngleOfAttackRad = inf.ReadSingle();
+            AngleOfAttackmRad = inf.ReadSingle();
             DerailClimbDistanceM = inf.ReadSingle();
             DerailPossible = inf.ReadBoolean();
             DerailExpected = inf.ReadBoolean();
--- a/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
@@ -33,10 +33,16 @@
 // Debug Brake Slide Calculations
 //#define DEBUG_BRAKE_SLIDE
 
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
 using Microsoft.Xna.Framework;
 using Orts.Common;
 using Orts.Formats.Msts;
 using Orts.Parsers.Msts;
+using Orts.Simulation;
 using Orts.Simulation.Physics;
 using Orts.Simulation.RollingStocks.Coupling;
 using Orts.Simulation.RollingStocks.SubSystems;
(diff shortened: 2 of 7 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1890,7 +1890,7 @@ public override void Save(BinaryWriter outf)
 
             outf.Write(WheelBrakeSlideProtectionActive);
             outf.Write(WheelBrakeSlideProtectionTimerS);
-            outf.Write(AngleOfAttackRad);
```
cjakeman, line 1893: What happens to content that has used the old parameter "AngleOfAttackRad"?
Won't that content no longer work?

Source/Documentation/Manual/sound.rst
```diff
@@ -446,6 +446,62 @@ Trigger       Function
 =========     =====================================
 
 
+Track Based Sounds for Routes
+-----------------------------
+
+As track based sounds (such as track joints, curve squeal, etc) are common to all rolling stock these types of sounds can be added at a route level, rather then 
+at an individual rolling stock level. This has the advantage of ensuring a level of consistency for all individual stock rather then requiring each piece of stock 
+to have appropriate sounds associated with it. It means that all rolling stock will be "automaticvally" have track sounds given to it.
```
cjakeman:
```suggestion
to have appropriate sounds associated with it. It means that all rolling stock will automatically have track sounds given to it.
```

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
peternewell: That is an internal parameter change made to highlight the UoM, see line underneath. Hence shouldn't impact any user content.

Source/Documentation/Manual/sound.rst
peternewell: I have corrected the spelling mistake, thanks for picking it up.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -2709,7 +2903,7 @@ public void SetUpWheels()
                 }
             }
             // Using WheelAxles.Count test to control WheelAxlesLoaded flag.
-            if (WheelAxles.Count > 2)
+            if (WheelAxles.Count >= 2) // Some cars only have two axles.
             {
                 WheelAxles.Sort(WheelAxles[0]);
                 WheelAxlesLoaded = true;
```
SteelFill, line 2909: Peter,

I was chasing down bugs for PR #1169, particularly train cars going invisible, thinking I messed something up in #1121, but further digging has revealed that this specific change is what causes things to go invisible.

`WheelAxlesLoaded` should _only_ be set to true if the train car has enough information to determine its position with the linear regression method (otherwise a fallback method is used, the two methods should probably be unified). Something with 2 axles does not necessarily have enough information to do this, but something with 3 axles almost certainly has enough information, which is why this check was previously requiring 3 or more axles.

The problem opened up here is the case of a train car with 1 bogie, and that bogie having 2 axles. 1 bogie is not sufficient to determine the position of a train car, at least 2 bogies or 1 bogie and 1 rigid mounted axle are required to determine the position of the train car. However, this code is telling OR that these train cars with 1 bogie which has 2 axles _do_ have enough information for that calculation even though they do _not_. This results in the calculation failing, and those train cars appearing invisible.

I will attempt a fix with #1169 but I don't know enough about the implementation of track sounds to know if there will be an impact. Do track sounds require the wheel axles to be loaded? Many bits of rolling stock simply won't have the right setup for the wheel axles to be loaded, so that should be accounted for in the track sound system.

peternewell: Thanks for identifying this.

As you suggest, WheelAxlesLoaded is used as a test to find the position of the car (car.ComputePosition(traveller, true, elapsedTime, distance, SpeedMpS);) in Train.cs.

With it set to > 2, the position of 2 axle (prototypical 4 wheel) wagons were not being correctly calculated, and hence the sound trigger for 2 axle wagons was not being set correctly. Thus Crossover sounds, etc for 2 axle wagons were not being correctly played. 

Hence the change to recognise (>=) 2 axles being a valid scenario was required to facilitate the car position for 2 axle wagons being correctly identified. This change then appeared to give correct positioning of 2 axle wagons.

It appeared that the fallback method (when WheelAxlesLoade was not set)  is not providing a similar position indication as when the WheelAxlesLoaded is set.

Provided that the location of the 2 axle wagon can be correctly identifed by OR in either approach, then there shouldn't be any impacts on the track sounds for these potential code changes.

SteelFill: I see. I will attempt some alternate ways to detect if the 'wheel axles are loaded', hopefully without making it too complicated, that still allow "actual" 4 wheel wagons to work without breaking "false" 4 wheel wagons (like articulated EMUs or intermodal cars, which is where the problems were presenting).

I am on a work trip this week, so I won't have anything to show until next weekend, but that will be part of #1169 so keep an eye out for that.

peternewell: Ok, thanks.

PR #1161, opened 2025-09-21
Fix string trim in .ini files

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/38055-seranas-tcs-scripts-not-working-in-ornymg-after-rev-140/page__view__findpost__p__321570

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/PowerSupply/LocomotivePowerSupply.cs
+++ b/Source/Orts.Simulation/Common/Scripting/PowerSupply/LocomotivePowerSupply.cs
@@ -606,7 +606,7 @@ namespace ORTS.Scripting.Api
 
                 if (length > 0)
                 {
-                    buffer.Trim();
+                    buffer = buffer.Trim('\0').Trim();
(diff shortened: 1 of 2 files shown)


SteelFill (approved):
Simple change, original poster reports problem is fixed and nothing new has been noted for a week. Looks good to merge.

Source/Orts.Simulation/Common/Scripting/PowerSupply/LocomotivePowerSupply.cs
```diff
@@ -606,7 +606,7 @@ protected T LoadParameter<T>(string sectionName, string keyName, T defaultValue)
 
                 if (length > 0)
                 {
-                    buffer.Trim();
+                    buffer = buffer.Trim('\0').Trim();
```
twpol, line 609: This is not really what we should have done here; the Windows API returns the number of characters written in to the buffer and we should be using that

cesarBLG: You're right, I blindly copied the fix that I used inside the TCS script, where the string length is not available. I'll submit a new PR.

PR #1163, opened 2025-09-29
Fix: Crash when using Camera 8 and F9.

mbm-OR:
Crash when camera 8 is active, F9 window visible and any wagon clicked.
[Bug #2125924](https://bugs.launchpad.net/or/+bug/2125924)


Csantucci (commented):
Let's wait some days for feedbacks, and then I can approve.

rwf-rr (approved):
Looks fine to me.

PR #1169, opened 2025-10-19
Revise TrainCar.SetUpWheels to Better Handle Unusual Rolling Stock

SteelFill:
Semi follow up to #1121 and #1160. A seemingly innocent change made in #1160 allowed for train cars with insufficient information to compute their position to use `TrainCar.ComputePosition()` anyway. This would result in said train cars appearing invisible, as noted in [this bug](https://www.elvastower.com/forums/index.php?/topic/39187-stock-missing-in-consists/) and [this bug](https://www.elvastower.com/forums/index.php?/topic/39269-missing-loco-shapes-in-rmd-east-route/page__pid__322641#entry322641).

This is because `TrainCar.ComputePosition` requires at least two points in 3D space to position a train car, as linear regression is used and linear regression needs at least two points. These two reference points can be the position of an axle on the train car, or a bogie on the train car (though if a bogie is used, the _bogie itself needs an additional two or more reference points_ to determine its position). Whether or not `TrainCar.ComputePosition` is used at all depends on the state of `TrainCar.WheelAxlesLoaded`, where a true value indicates `TrainCar.ComputePosition` can be used, and a false value indicates it cannot be used. This means that `TrainCar.WheelAxlesLoaded` should only be true if the train car has at least two reference points for its position.

However, the process to determine `TrainCar.WheelAxlesLoaded` in `TrainCar.SetUpWheels` hasn't really been touched in 10 years and was suboptimal. Previously, `TrainCar.WheelAxlesLoaded` was set to true if the train car had **3 or more simulated axles**. This was merely a rough approximation of the actual check to see "are there two or more reference points for position?" as models with 3 simulated axles _usually_ have enough reference points for `TrainCar.ComputePosition` to do its thing. However, #1160 revealed how problematic this strategy was by changing the condition to be **_2_ or more simulated axles**. Unlike checking for **3** or more simulated axles, many models with **2** simulated axles do _not_ have two reference points for `TrainCar.ComputePosition`. The most obvious example is a model with one bogie, and that bogie having two axles. Because the model has only one bogie, it has only one positional reference point, and thus `TrainCar.WheelAxlesLoaded` should be set to false for this model. But with the changes made, `TrainCar.WheelAxlesLoaded` would be set to true for this model and calculating its position with `TrainCar.ComputePosition` would fail.

While the change made in #1160 could be reverted, going back to checking for 3 or more simulated axles, that would break track sounds on models with 2 axles. Instead, `TrainCar.SetUpWheels` has been greatly refactored to achieve the actual intended result, instead of some approximation of it. With this PR, `TrainCar.SetUpWheels` will inspect the configuration of simulated axles and bogies and will only set `TrainCar.WheelAxlesLoaded` to true if the train car has 2 or more axles attached to it, 2 or more bogies (which themselves have 2 or more axles attached) attached to it, or some combination of axles and bogies that adds up to at least 2 positional references.

This won't necessarily provide the same resulting behavior as previous versions, but any changes should be imperceptible. I think this might even help content creators, by removing some of the jank that was previously present (you should no longer need to add fake axles or fake bogies to non-bogie rolling stock).

Additionally, should `TrainCar.WheelAxlesLoaded` somehow be set to true incorrectly, a fallback calculation has been added to `TrainCar.ComputePosition` that will still produce a reasonable position instead of causing the model to become invisible.

While refactoring `TrainCar.SetUpWheels`, some related problems that have been observed were also fixed. `ORTSFrontArticulation` and `ORTSRearArticulation` were not being copied to subsequent instances of a train car, so they have been added to `MSTSWagon.Copy`. Fake axles were not being added to all bogies that were missing axles, resulting in nonsense position calculations that would [cause said bogies to spin](https://www.elvastower.com/forums/index.php?/topic/39191-rapidly-rotating-bogies/), so that was rebuilt to work reliably. And when fake axles were added, a few systems were treating these fake axles as if they were real axles, so the distinction has been made more clear.

Bug report: https://www.elvastower.com/forums/index.php?/topic/39191-rapidly-rotating-bogies/

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/CruiseControl.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/CruiseControl.cs
@@ -432,7 +432,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems
                 SelectedNumberOfAxles = 0;
                 foreach (TrainCar tc in Locomotive.Train.Cars)
                 {
-                    SelectedNumberOfAxles += tc.WheelAxles.Count;
+                    SelectedNumberOfAxles += tc.WheelAxles.Sum(w => w.Fake ? 0 : 1);
                 }
             }
         }
--- a/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
@@ -911,9 +911,11 @@ namespace Orts.Simulation.RollingStocks
             // Initialize RigidWheelBaseM in first loop if not defined in ENG file, then ignore
             if (RigidWheelBaseM == 0 && !RigidWheelBaseInitialised)   // Calculate default values if no value in Wag File
             {
-                float Axles = WheelAxles.Count;
-                float Bogies = Parts.Count - 1;
-                float BogieSize = Axles / Bogies;
+                int Axles = WheelAxles.Sum(w => w.Fake ? 0 : 1); // Only consider real axles
+                int Bogies = Parts.Sum(p => p.Bogie ? 1 : 0);
+                int BogieSize = Axles;
+                if (Bogies > 0)
+                    BogieSize = (int)(WheelAxles.Sum(w => !w.Fake && w.Part.Bogie ? 1 : 0) / Bogies); // Only consider axles attached to bogies
 
                 RigidWheelBaseM = 1.6764f;       // Set a default in case no option is found - assume a standard 4 wheel (2 axle) bogie - wheel base - 5' 6" (1.6764m)
 
@@ -2804,16 +2806,20 @@ public string GetCurveDirection()
             // No parts means no bogies (always?), so make sure we've got Parts[0] for the car itself.
             if (Parts.Count == 0)
                 Parts.Add(new TrainCarPart(Vector3.Zero, 0));
-            // No axles but we have bogies.
-            if (WheelAxles.Count == 0 && Parts.Count > 1)
+            // Determine how many parts are considered bogies (used to calculate position of the car)
+            int bogieCount = 0;
+            foreach (TrainCarPart p in Parts)
+                if (p.Bogie == true)
+                    bogieCount++;
+            // No axles but we have bogies. Each bogie needs axles to function correctly
+            if (WheelAxles.Count == 0 && bogieCount > 0)
             {
-                // Fake the axles by pretending each bogie has 1 axle.
+                // Add a fake axle to each bogie, a second fake axle will be added later if needed
                 for (int i = 1; i < Parts.Count; i++)
-                    WheelAxles.Add(new WheelAxle(Parts[i].OffsetM, i, Parts[i].iMatrix));
-                Trace.TraceInformation("Wheel axle data faked based on {1} bogies for {0}", WagFilePath, Parts.Count - 1);
+                    if (Parts[i].Bogie == true)
+                        WheelAxles.Add(new WheelAxle(Parts[i].OffsetM, i, Parts[i].iMatrix, true));
+                Trace.TraceInformation("Wheel axle data faked based on {1} bogies for {0}", WagFilePath, bogieCount);
             }
-            bool articFront = !WheelAxles.Any(a => a.OffsetM.Z < 0);
-            bool articRear = !WheelAxles.Any(a => a.OffsetM.Z > 0);
             // Validate the axles' assigned bogies and count up the axles on each bogie.
             if (WheelAxles.Count > 0)
             {
@@ -2833,10 +2839,6 @@ public string GetCurveDirection()
                     w.Part = Parts[w.BogieIndex];
                     w.Part.SumWgt++;
                 }
-
-                // Make sure the axles are sorted by OffsetM along the car.
-                // Attempting to sort car w/o WheelAxles will resort to an error.
-                WheelAxles.Sort(WheelAxles[0]);
             }
 
             //fix bogies with only one wheel set:
@@ -2856,7 +2858,7 @@ public string GetCurveDirection()
                         {
                             if (w.OffsetM.Z.AlmostEqual(Parts[i].OffsetM.Z, 0.6f))
                             {
-                                var w1 = new WheelAxle(new Vector3(w.OffsetM.X, w.OffsetM.Y, w.OffsetM.Z - 0.5f), w.BogieIndex, i);
+                                var w1 = new WheelAxle(new Vector3(w.OffsetM.X, w.OffsetM.Y, w.OffsetM.Z - 0.5f), w.BogieIndex, i, true);
                                 w1.Part = Parts[w1.BogieIndex]; //create virtual wheel
                                 w1.Part.SumWgt++;
                                 WheelAxles.Add(w1);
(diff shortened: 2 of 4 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -3020,8 +3020,18 @@ public void ComputePosition(Traveller traveler, bool backToFront, float elapsedT
                 if (p.SumWgt > 1.5f)
                     p0.AddPartLocation(1, p);
             }
-            // Determine facing direction and position of train car
-            p0.FindCenterLine();
+            if (Parts.Count == 2)
```
rwf-rr: Is this correct when a car has more than two bogies? Code around lines 2740 supports more than two bogies.
The Swiss Re460 electric locomotive has three bogies of type B.

Actually I am confused by the Parts list. Why does a car with a single Bogie have a parts count of 2?

rwf-rr (commented):
The changes look fine. 

But the whole code is not very intuitive to me. Like in the original code, loop from index 1 and then explicitly call for index 0.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
SteelFill: Parts[0] is the train car itself. Setting the position and orientation of Parts[0] will ultimately set the position and orientation of the train car. Parts[1] is the first bogie, then Parts[2] is the second bogie, and so on. The algorithm is designed specifically for 2 bogies or more, but 1 bogie is an edge case. "0" bogie equipment is handled totally differently.

rwf-rr (dismissed):

rwf-rr (dismissed):
Additional changes look good. Also did some testing with a freight car with two fixed axles.

SteelFill:
Determined that I was going on a wild goose chase, the issues I've been chasing down here came from a change I didn't even notice was introduced with #1160. I suspect the change is needed for the track sounds to work so I will look for a more capable workaround than either undoing the change or continuing to target the symptoms as I have done this week.

SteelFill:
Root cause has been located and addressed. Some testing will be needed to verify the new methodology employed will correctly compute positions for all types of rolling stock

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -911,9 +911,11 @@ public virtual void Update(float elapsedClockSeconds)
...
             if (RigidWheelBaseM == 0 && !RigidWheelBaseInitialised)   // Calculate default values if no value in Wag File
             {
-                float Axles = WheelAxles.Count;
-                float Bogies = Parts.Count - 1;
-                float BogieSize = Axles / Bogies;
+                int Axles = WheelAxles.Sum(w => w.Fake ? 0 : 1); // Only consider real axles
+                int Bogies = Parts.Sum(p => p.Bogie ? 1 : 0);
+                int BogieSize = Axles;
+                if (Bogies > 0)
+                    BogieSize = (int)(WheelAxles.Sum(w => !w.Fake && w.Part.Bogie ? 1 : 0) / Bogies); // Only consider axles attached to bogies
```
rwf-rr, line 918: Here the check for axles includes a check for the part being a bogie. This is not present in CruiseControl.cs.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/CruiseControl.cs
```diff
@@ -432,7 +432,7 @@ private void ComputeNumberOfAxles()
                 SelectedNumberOfAxles = 0;
                 foreach (TrainCar tc in Locomotive.Train.Cars)
                 {
-                    SelectedNumberOfAxles += tc.WheelAxles.Count;
+                    SelectedNumberOfAxles += tc.WheelAxles.Sum(w => w.Fake ? 0 : 1);
```
rwf-rr, line 435: In CruiseControl.cs, the check also includes a check for the part to be a boogie.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -2804,16 +2806,20 @@ public void SetUpWheels()
...
+            // Determine how many parts are considered bogies (used to calculate position of the car)
+            int bogieCount = 0;
+            foreach (TrainCarPart p in Parts)
+                if (p.Bogie == true)
+                    bogieCount++;
+            // No axles but we have bogies. Each bogie needs axles to function correctly
+            if (WheelAxles.Count == 0 && bogieCount > 0)
+            {
+                // Add a fake axle to each bogie, a second fake axle will be added later if needed
+                for (int i = 1; i < Parts.Count; i++)
```
rwf-rr, line 2818: Nit: Since we now check that the part is a bogie (inside the loop), the loop could start at 0. That would be more intuitive when one does not know that part 0 never is a bogie.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -2907,43 +2904,50 @@ public void SetUpWheels()
...
+            // expected motion, a fake axle is added at the articulated end(s)
+            // of the car, attached to the car itself. This will drive the positioning
+            // of the car as expected.
+            if (front)
+                WheelAxles.Add(new WheelAxle(new Vector3(0.0f, BogiePivotHeightM, -CarLengthM / 2.0f), 0, 0, true) { Part = Parts[0] });
 
-                if (rear)
-                    WheelAxles.Add(new WheelAxle(new Vector3(0.0f, BogiePivotHeightM, CarLengthM / 2.0f), 0, 0) { Part = Parts[0] });
-
-                WheelAxles.Sort(WheelAxles[0]);
```
rwf-rr, line 2945: Since we could add either a front or rear axle, would it not be necessary to sort the axles again?

rwf-rr (approved):
The code changes look good, with only minor comments.

I can't say I fully understand the handling of bogies and axles, so a second review (or some extra testing with unusual wagons) would be good.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
SteelFill: The axle list is sorted on line 2919. There is no point in sorting the list repeatedly, so it is just sorted once after all axles are added.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
SteelFill: By convention, Parts[0] is always the main body of the rail vehicle. Axles are added to the main body as needed in other steps, so the main body should always be skipped here.

Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
SteelFill: This is doing something completely different from what cruise control is doing. This process is explicitly determining the number of axles per bogie as that affects the wheelbase of the bogie, which is what RigidWheelBaseM represents, so we only count axles that are attached to bogies.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/CruiseControl.cs
SteelFill: Cruise control is interested in the total number of axles in the train, which means it does not care if any of those axles are on a bogie or not, just that they aren't fake axles.

PR #617, opened 2022-03-13
02m: Removed menu option Debrief Evaluation. Now always available.

cjakeman:
See https://blueprints.launchpad.net/or/+spec/update-menu-options
and http://www.elvastower.com/forums/index.php?/topic/35301-menu-options/

Renamed in the Manual as Activity Evaluation.

Evaluation data saved in a file has been merged with the data aready saved by Game > Save (F2).

The text report is now saved in the same folder as OpenRailsLog.txt (by default the Windows Desktop) and named OpenRailsEvaluation.txt.

It is copied by Game > Save (F2) to the same folder as the *.save file and named *.evaluation.txt so will be deleted and undeleted with the *.save files.

PR #1175, opened 2025-10-29
Make Data Logger interval configurable

rwf-rr:
The Data Logger currently logs at the frame rate. This PR makes the interval configurable.

A new input box on the data logger tab specifies the interval in milliseconds. The interval is saved in the settings.

An interval of 0 (or less than 1/FPS) will log for each frame.

Unrelated: Add confirmations that the Data Logger was started or stopped (by F12 key).

See [forum discussion](https://www.elvastower.com/forums/index.php?/topic/39186-who-is-using-the-data-logger-and-for-what)

Also see Trello card [#617](https://trello.com/c/kGbRsMeL/617-data-logger-configurable-interval-configurable-units-more-train-data).


cjakeman (changes requested):
Hi Roger,

This is a good step forward.

Would you add a short note in the [6.7. Data Logger Options](https://open-rails.readthedocs.io/en/unstable/options.html#data-logger-options) to explain what happens when the new Logging Interval (msec) is set to 0?

Thanks,

Chris

cjakeman (approved):
Hi Roger,

Very happy with this.

Approved and merged with thanks,

Chris

PR #1186, opened 2025-12-30
Hot Reloading of Trains, Scenery, and Terrain

SteelFill:
This PR adds the often-requested ability to reload changes made to content without restarting the entire program. Changes made to select file types will cause in-game assets referencing that file to be reloaded automatically, reflecting the changes made to the file. These changes can be seen in just a couple of seconds, compared to the perhaps minute it might take to close and reload the simulation, which should save a whole lot of time in the long run.

[Discuss on Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38030-models-reload-while-games-running/)
[Relevant Trello card](https://trello.com/c/pZla7I3P)

The rough architecture is as follows:
- There is a new option to "Enable hot reloading of simulator files" in the experimental tab of the options menu, which should be enabled for the following behaviors to occur (_most of what's described here is automatic but can be triggered manually_).
- If hot reloading is enabled, `FileSystemWatcher` objects are generated in the `Simulator` object, one each for the GLOBAL, ROUTES, SOUND, and TRAINS folder in the currently selected MSTS install. Having separate watchers for each folder helps with organization and better handles some symbolic link setups (FileSystemWatcher does not handle symbolic links well).
- Event handlers have been added to the `Simulator` and `Viewer` objects that respond to file changes by making lists of files that have changed. These lists are sorted based on file type for organization and optimization purposes (if only an .eng file is changed, there is no need to check for any updates to sounds or textures).
- Objects in code with properties affected by the contents of some file on disk have a `bool StaleData` variable. When this is true, it indicates that some changes have been made on the disk that are not currently reflected by the data in the object, so the file should be reloaded.
- In the case of objects that are handled through a shared manager, if the shared manager is requested to serve an entry from the dictionary with `StaleData` set to true, it will load a new instance of that object and insert that into the dictionary, instead of serving the object that was previously present. This means re-loading any files used, capturing any new changes made.
- After 1 second has passed since the last file change, the lists of changed files are processed to figure out which objects need to have `StaleData` set to true. Setting only the affected objects to stale reduces the amount of time spent reloading unnecessary files, speeding up the whole process. And that 1-second delay allows for multiple file changes to be processed at once, as one change made by the user can throw multiple events in rapid succession. While the timer check is run every frame, the processing only happens once per set of changed files, reducing performance impact.
- "Stale-ness" is also propagated upward until it reaches an object that can trigger a reload. For example, a texture going stale will then mark any materials using the texture as stale, then any shapes using those materials as stale, finally marking any viewer using that shape as stale. Then the viewer can reload itself, which the texture could not do.
- For comparison, if the keybind to do a _manual reload_ is pressed, the Viewer intercepts this input and commands every "staleable" graphical/audio object to go stale, regardless of any file changes, then commands the Simulator to do the same for simulation objects. This results in everything being reloaded, which takes longer but is more reliable at capturing changes.
- Every frame in the loader thread, each train car viewer, scenery tile, and terrain tile is checked to see if it is stale. For each stale object, an updated version will be reloaded, then the existing object will be seamlessly replaced with the updated one. During reloading, any stale assets will be reloaded from scratch, capturing changes made.
- Also every frame, each train car checks to see if it is stale, using its own `StaleData` variable. If it is stale, `StaleViewer` is also set to true to force reloading the visual model, and the train car is reloaded from the wag file and re-initialized. Unlike reloading graphical things like the viewer or scenery, the reloading of the simulation overwrites data in the existing object rather than destroying and recreating the object.

The main features are reloading of trains, scenery, and terrain. For trains, any change to a texture (.ace/.dds/.png/.jpg), shape (.s/.sd), sound (.sms/.wav, or any .inc used by the .sms), or cabview (.cvf, .inc used by the .cvf) an engine/wagon uses will cause the model of that engine/wagon to be reloaded to reflect those changes. Meanwhile, any change to the .eng/.wag or any .inc used by the .eng/.wag will cause the simulation of that engine/wagon to be reloaded (as well as the model) so additional changes can be reflected. For scenery, any change to a world file (.w/.ws) or to any track profile (.stf/.xml), texture, shape, or sound used by the world file will cause that tile of scenery to be reloaded. And for terrain, any change to a tile file (.raw, .t) or to a terrain/water texture used by the tile will cause that tile of terrain to be reloaded.

By only reloading the specific engines, wagons, and/or tiles affected by a file change, fewer files are processed at once so the loader thread completes its work very quickly, though this is not foolproof. If hot reloading doesn't capture a change, a **manual reload** can be executed using `Ctrl + Alt + F5`. Instead of reloading specific assets, a manual reload **reloads everything**, which is slower but is more reliable. The automatic option is preferred.

While many systems in Open Rails are built to allow unloading and reloading in real time, other systems are more fragile and could break with even slight changes. I would not anticipate ever seeing support for hot reloading of signal scripts, the track database, activities, paths, services, things like that. But for the systems that can be supported, I hope this proves to be useful to speed things up.

This is not perfect; there are probably some file changes that won't cause a reload, even if theoretically the file could be reloaded, as these systems have to be set up on a case-by-case basis. Any deletion/creation/moving/renaming of files might not work as the system is happiest working with changes to/overwriting files that were already there when the sim started. And there's no sort of preemptive error checking, so saving a file with even minor errors can cause OR to reload that file and crash when seeing the error. All that means is, be careful and don't use this feature unless you are doing content development and can cope with the occasional glitch or crash.

Some refactoring was required to facilitate hot reloading:
- As mentioned, `StaleData` has been added as a variable to many classes.
- A new `TextureTag` class was created, which is a modification of the previous `AceInfo` class but with an added `bool StaleData` field, intended to be used as the object for the `Texture2D.Tag` parameter. `SharedTextureManager` can then use the tag on each texture to tell when a texture has gone out of date.
- Some `Initialize()` functions have been given a `bool reinitialize = false` parameter. `reinitialize` is to be set to true when the object has been initialized before, but is being initialized again anyway (eg: due to a hot reload), which allows for different conditional logic to occur on the first initialization versus subsequent initializations (eg: don't reset the throttle to idle on reinitialization).
- A few simulation parameters have had parsing adjusted to reduce errors and glitches during hot reloading, but there should be no functional changes when doing an initial load.
- Each `STFReader` object now keeps track of a list (actually a HashSet, we don't need duplicate entries) of `FileNames` it has encountered, recursively adding any files encountered by the `includeReader`. This basically means the STF reader contains a list of the original file as well as any .inc files referenced by the original file, very useful for dependency tracking.
- Some objects thus have a HashSet of `FileReferences` to keep track of which .inc files are referenced by the object, though these lists are only created if hot reloading is enabled, as they are a waste of memory otherwise.
- Many variables storing strings of file paths were changed to modify the user-entered path with `Path.GetFullPath(myFilePathVariable).ToLowerInvariant()` before storing it to a variable. This is to ensure the path stored in memory can be compared to the path reported by the `FileSystemWatcher`. Users can define relative paths with modifiers like "..\\\\" to go up in the folder tree, but a file system event will always give the absolute path to a file which does not contain "..\\\\". `Path.GetFullPath()` disambiguates the path entered by the user so it can be matched to the path reported in a file system event. And then the use of uppercase and lowercase may be different between the path in memory and the path in the file system, but case shouldn't matter, so both file paths are set to lowercase. The default string equality check can then be used to successfully match a file system path to a path in memory.
- Shared assets with marking (indicating which assets are in use and thus can't be removed from memory) were updated to make marking more efficient. Instead of searching through a dictionary repeatedly for every single used asset to indicate each is in use, a hash set is built up indicating which assets are in use, and this set is then searched _once_ at the end of loading. This reduces the amount of lookups needed to mark everything, which gives a measurable reduction to load times.
- `TerrainTile` objects were given local variables to store a reference to the `Tile` and `TileManager` used to create the terrain tile. This additional linking of objects makes a few things easier as there are situations where it's nice to know which tile or tile manager is used by a terrain tile.

This means there are many changes spread all across the program, particularly in graphical systems. While none of the changes are so deep to break every other graphical change, hot reloading does require additional work for larger scale changes like .glTF, as these add new file types that hot reloading has not been set up for.

Plans
- [x] Support for reloading scenery and terrain tiles automatically
- [x] Keybind that forces everything that can be reloaded to be reloaded, in case some changes are not detected properly by the automatic system
- [ ] Wait for GLTF to be ready, then add support for hot reloading of GLTF related assets
- [ ] Support changes to additional obscure filetypes, I'm considering car spawner configs, others(?)
- [ ] Continued checking to verify that file changes cause reloads when expected, and ensure reloads reflect the expected changes with minimal side effects


Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -563,11 +664,47 @@ internal void UpdateShaders()
...
 
+    public class SharedTexture
+    {
+        public bool StaleData;
+        private Texture2D _texture;
+        public Texture2D Texture
+        {
+            get
+            {
+                return _texture;
```
pzgulyas: I see you wrap the Texture2D class into an own class for the sake of assigning an additional bool property. This seems a bit too intrusive to me, because it entails modifying all the Texture2D variables used in the whole project. Wouldn't it be possible to utilize the existing Texture2D.Tag property instead? You can see the usage of it in this file at around line 910. An Orts.Formats.Msts.AceInfo object is assigned to the Texture2D.Tag, which has an AlphaBits property. You could add there an additional StaleData one for your purposes.

Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -768,8 +907,8 @@ public SceneryMaterial(Viewer viewer, string texturePath, SceneryMaterialOptions
             var texture = SharedMaterialManager.MissingTexture;
             if (Texture != SharedMaterialManager.MissingTexture && Texture != null) texture = Texture;
             else if (NightTexture != SharedMaterialManager.MissingTexture && NightTexture != null) texture = NightTexture;
-            if (texture.Tag != null && texture.Tag.GetType() == typeof(Orts.Formats.Msts.AceInfo))
-                AceAlphaBits = ((Orts.Formats.Msts.AceInfo)texture.Tag).AlphaBits;
+            if (texture.Texture.Tag != null && texture.Texture.Tag.GetType() == typeof(Orts.Formats.Msts.AceInfo))
+                AceAlphaBits = ((Orts.Formats.Msts.AceInfo)texture.Texture.Tag).AlphaBits;
```
pzgulyas: See here the usage of the Texture2D.Tag. I think it would be better to reuse this object instead of a project-wide Texture2D-wrap. The Orts.Formats.Msts.AceInfo class could be moved out to a more common place somewhere, and renamed to something more generic of course, since it would be no longer msts-related.

pzgulyas (changes requested):

Source/RunActivity/Viewer3D/Lights.cs
```diff
@@ -23,6 +23,12 @@
 //#define DEBUG_LIGHT_CONE
 //#define DEBUG_LIGHT_CONE_FULL
 
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using LibGit2Sharp;
```
pzgulyas:
```suggestion
```

SteelFill: I do love how visual studio has a mind of its own on which namespaces to use.

Source/RunActivity/Viewer3D/Materials.cs
SteelFill: It seems that should be flexible enough to do what I need, I'll give that a try.

Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -73,16 +74,19 @@ public Texture2D Get(string path, bool required = false)
         /// <summary>
         /// Loads a game texture file; DO NOT use with internal data, use <see cref="LoadInternal(GraphicsDevice, string)"/> instead.
         /// </summary>
-        /// <returns>The <see cref="Texture2D"/> created from the given <paramref name="path"/> or a missing placeholder.</returns>
+        /// <returns>The <see cref="SharedTexture"/> created from the given <paramref name="path"/> or a missing placeholder.</returns>
```
pzgulyas: Not yet reverted.

Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -204,30 +209,58 @@ public static Texture2D LoadInternal(GraphicsDevice graphicsDevice, string path,
...
+        /// <summary>
+        /// Sets the stale data flag for ALL textures to the given bool
+        /// (default true)
+        /// </summary>
+        public void SetAllStale (bool stale = true)
         {
-            foreach (var path in TextureMarks.Where(kvp => !kvp.Value).Select(kvp => kvp.Key))
+            foreach (Texture2D texture in Textures.Values)
+                if (texture.Tag is TextureTag tag)
+                    tag.StaleData = stale;
```
pzgulyas, line 256: Didn't you want to write something like this one instead?
```suggestion
            foreach (Texture2D texture in Textures.Values)
            {
                if (!(texture.Tag is TextureTag tag))
                    texture.Tag = tag = new TextureTag();
                tag.StaleData = stale;
            }
```
Otherwise the textures with null .Tag would not get the flag assigned.

Source/RunActivity/Viewer3D/Materials.cs
```diff
@@ -204,30 +209,58 @@ public static Texture2D LoadInternal(GraphicsDevice graphicsDevice, string path,
...
             {
-                Textures[path].Dispose();
-                Textures.Remove(path);
+                if (Textures.ContainsKey(texPath) && Textures[texPath].Tag is TextureTag tag && !tag.StaleData)
+                {
+                    tag.StaleData = true;
+                    found = true;
+
+                    Trace.TraceInformation("Texture file {0} was updated on disk and will be reloaded.", texPath);
+                }
```
pzgulyas, line 285: The same here:
```suggestion
                if (Textures.ContainsKey(texPath))
                {
                    if (!(Textures[texPath].Tag is TextureTag tag))
                        Textures[texPath].Tag = tag = new TextureTag();
                    if (!tag.StaleData)
                    {
                        tag.StaleData = true;
                        found = true;

                        Trace.TraceInformation("Texture file {0} was updated on disk and will be reloaded.", texPath);
                    }
                }
```

pzgulyas (changes requested):

Source/RunActivity/Viewer3D/Lights.cs
pzgulyas: Yeah, it happend to me a numerous time too...

Source/RunActivity/Viewer3D/Materials.cs
SteelFill: This _shouldn't_ be needed as every texture is given a TextureTag when loaded, though I suppose it would not hurt to ensure a tag is added to guarantee the texture gets reloaded even if something goes wrong.

pzgulyas: It just came into my mind because for gltf not all textures are loaded via the Viewer.TextureManager.Get(), thus the preventive assignment there would not work for that case. When the image is packed together into a glb for instance, it is loaded via a stream provided by the glTFLoader.Interface.OpenImageFile() function, thus will not get the Tag. So such a "lazy assigning" approach would be preferable from this point of view.

PR #1190, opened 2026-01-09
Fix Map Window position saved when minimized.

rwf-rr:
When the Map (Dispatcher) window (Ctrl-9) is minimized, its position was saved as "-32000,-32000,199,34". This is how Windows sizes and hides a minimized window. Subsequently, when the Map window is opened again, it is not visible (far off the screen), and the user may think that the Map is no longer working.

See [discussion on ElvasTower](https://www.elvastower.com/forums/index.php?/topic/37165-reworking-the-dispatcher-window/page__view__findpost__p__322764).

When the window is minimized, use the _restore_ bounds (instead of the current bounds) to save in the settings. This affects both the position and the size.

Common practice seems to be to also save the _restore_ bounds when the window is maximized. So I included this.

### DIFF
--- a/Source/RunActivity/Viewer3D/Map/MapForm.cs
+++ b/Source/RunActivity/Viewer3D/Map/MapForm.cs
@@ -297,9 +297,12 @@ namespace Orts.Viewer3D.Debugging
             {
                 string name = "Map_" + mapViewer.Name;
 
-                int X = this.Bounds.X;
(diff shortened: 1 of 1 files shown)


sweiland-openrails (approved):
I checked and tested this code change and it's ok.

Did not merge as some developers like to do this themself.

regards, Siebren.

PR #1195, opened 2026-02-14
fix: Allow telemetry form to be completely localised

twpol:
### Changes

- fix: Allow telemetry form to be completely localised
- chore: Update all locale templates and pseudo-localisation


cjakeman (approved):
Hi James,

No issues with this.

I see that you found a number of other updates needed, which you confined to a separate commit. Good catch.

Approved and merged.

Chris

PR #1197, opened 2026-02-15
Implementation of Riggenbach brake

peternewell:
Adds Riggenbach braking to steam locomotives.

https://blueprints.launchpad.net/or/+spec/riggenbach-braking

ET thread:

https://www.elvastower.com/forums/index.php?/topic/39476-riggenbach-counter-pressure-brake/

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
@@ -3971,11 +3971,16 @@ public readonly SmoothedData StackSteamVelocityMpS = new SmoothedData(2);
             CylinderSteamExhaustSteamVelocityMpS = 100.0f;
             CylinderSteamExhaustParticleDurationS = 1.0f;
 
-            CounterPressureBrake1SteamVolumeM3pS = CounterPressureBrakeOn ? Math.Abs(cutoff * 5.0f * SteamEffectsFactor) : 0.0f;
+            float WatertoSteamVolumeLbstoM3 = 0.759f; // 1lb of water = 0.759M of water @ 100 degC
+            // Amount of water used has to be divieded between steam cylinders, and converted to steam volume
+            float CounterPressureBrakeWaterSteamExhaustM3pS = (CounterPressureBrakeWaterUsedLBpS / 2) * WatertoSteamVolumeLbstoM3;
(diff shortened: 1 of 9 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -1541,6 +1559,13 @@ public override void Initialize()
             if (ZeroError(BoilerVolumeFT3, "BoilerVolume"))
                 BoilerVolumeFT3 = 1;
 
+            // Only set counter pressure braking fitted to the steam engine if it is specified in the ENG file and there is at
+            // least one steam engine block.  Otherwise, leave it as false (not fitted) which is the default value.
+            if (CounterPressureBrakingFitted && SteamEngines.Count < 1)
```
cesarBLG:
```suggestion
            if (CounterPressureBrakingFitted && SteamEngines.Count == 1)
```
Probably you meant this, if count is 0 the following line would be out of range

cesarBLG (changes requested):
There's just a minor issue that I spotted

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
peternewell: Thanks.

I have adjusted it.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -69,6 +69,7 @@
 using System;
 using System.Diagnostics;
 using System.IO;
+using System.Security.Cryptography;
```
cesarBLG: Was this intentional?

peternewell: No it wasn't intentional. I have removed it.

Visual Studio seems to add these "automatically", perhaps the AI at work?

cesarBLG: I suspect it's the Autocomplete feature. It sometimes happen to me. Although in the end AI is essentially an advanced autocomplete

cesarBLG (approved):

PR #1205, opened 2026-03-01
Various controls by mouse enhancements

sweiland-openrails:
- sorted mouse controls by rectangle size so that lager ones do not overlap smaller ones


sweiland-openrails:
I've closed this PR whithout merging. I think it's not that usefull...

PR #1206, opened 2026-03-01
Set missing texture color depending on shape warning option (was: build mode)

rwf-rr:
Gray (grey) ~~for the release (and testing) build~~ by default.
Magenta (highly visible) ~~for the debug build~~ when "Show shape warnings" is checked (enabled) in the experimental options, so that missing textures stand out.

This was triggered by [this thread](https://www.elvastower.com/forums/index.php?/topic/39453-please-explain-what-is-going-on-here/).

The solution is not entirely clean, as GetInternalMissingTexture() is a static, but checking options requires a SharedTextureManager object.

However, the only textures that are loaded before the Viewer is initialized are BMP and PNG images for the splash (loading) screen. If those were missing, even grey would make it obvious.

All simulation (global, route, trains) textures are loaded (SharedTextureManager initialize) well after the settings are read.

Roadmap: https://trello.com/c/chR6f9OT/629-set-missing-texture-color-depending-on-build-mode

### DIFF
--- a/Source/Documentation/Manual/compatibility.rst
+++ b/Source/Documentation/Manual/compatibility.rst
@@ -47,7 +47,8 @@ file and run properly. Open Rails software logs these errors in a log file on
 the user's desktop. This log file may be used to correct problems identified 
 by the Open Rails software. The parser will also correct some of the problems 
 that stumped MSTS.  For example, if a texture is missing Open Rails will 
-substitute a neutral gray texture and continue.
+substitute a neutral gray texture (or high-vis magenta texture when compiled 
+in DEBUG mode) and continue.
 
 MSTS Files Used in Whole or Part by Open Rails
 ==============================================
--- a/Source/Documentation/Manual/compatibility.rst
+++ b/Source/Documentation/Manual/compatibility.rst
@@ -47,8 +47,8 @@ file and run properly. Open Rails software logs these errors in a log file on
 the user's desktop. This log file may be used to correct problems identified 
 by the Open Rails software. The parser will also correct some of the problems 
 that stumped MSTS.  For example, if a texture is missing Open Rails will 
-substitute a neutral gray texture (or high-vis magenta texture when compiled 
-in DEBUG mode) and continue.
+substitute a neutral gray texture (or high-vis magenta texture when the 
+experimental option "Show shape warnings" is checked) and continue.
 
 MSTS Files Used in Whole or Part by Open Rails
 ==============================================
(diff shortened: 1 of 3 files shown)


SteelFill:
I think hard coding different graphical behavior on stable vs unstable could be confusing. An alternative thought I had surrounding this issue was to make the missing texture customizable, such as using an actual image in the content folder (think of Train Sim Classic, where the missing texture is a texture that has "missing texture" written on it), which users could replace as they need. And in that case, (I assume?) we could still ship a different texture with different builds to provide different default behavior depending on the build.

rwf-rr:
> ... using an actual image in the content folder (think of Train Sim Classic, where the missing texture is a texture that has "missing texture" written on it), ...

The original solution loaded a file, but oddly named "blank.bmp". This file was loaded every time a missing texture was encountered. As part of the texture loading optimization, the file was replaced with a hard-coded color-based texture.

The challenge is that the missing texture class is static. It would need significant refactoring to properly load from a file (or use a color based on the "Show shape warnings" setting).

James wanted a simple solution. The other options I was considering was to use magenta only when compiled with the DEBUG flag. But that is impractical for content creators.

twpol:
> I think hard coding different graphical behavior on stable vs unstable could be confusing. [...] And in that case, (I assume?) we could still ship a different texture with different builds to provide different default behavior depending on the build.

That would still result in different behaviour for different update channels, no? This is a way simpler implementation of that

With this PR, it is only Unstable that gets the magenta, which seems like the best and simplest compromise between highlighting a new graphical problem (desired) and highlighting oversights in existing content (not desired)

cjakeman:
Hi Roger,

I think this is a good idea and a simple solution.
It seems to be an enhancement rather than a bugfix, so would you mind creating a Trello card for it which I can approve.

Thanks,
Chris

rwf-rr:
> ... so would you mind creating a Trello card ...

Trello card created. https://trello.com/c/chR6f9OT/629-set-missing-texture-color-depending-on-build-mode

rwf-rr:
There is some discussion to make unstable not use the DEBUG flag. That would make the current solution pointless.

That brings us back to the option of using the experimental setting "Show shape warnings". The challenge with this is that the missing texture class is a static variable. It is initialized before the settings are read.

rwf-rr:
A different solution, using the **Show shape warnings** experimental option, is ready for review.

twpol (approved):

PR #1207, opened 2026-03-01
Allow Zero and Negative Davis Values

SteelFill:
Follow up to #1120 +
Bug report: https://www.elvastower.com/forums/index.php?/topic/39475-friction-b-value-zero-override-check/

The previous PR considered friction values "missing" if the davis value was less than or equal to 0, as the default value of the unassigned variables was 0. However, it seems there are some situations where a 0 or negative value should be entered intentionally, which would look like missing data and get overridden anyway.

To allow that, every davis parameter has been made a nullable float, so now the distinction between no value being entered (which gives a `null` result) and 0/negative values intentionally entered is obvious. 0 and negative values will not be overridden, only null values. Be careful using negative coefficients!


peternewell (dismissed):
This looks fine, thanks for addressing it.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
rwf-rr: Is the default still correct? Should it be null or 0f?
Similar for the _B and _C variables below.

rwf-rr: Ditto for defaults.

rwf-rr: Is "fullortsdavis_a" correct here, or should it be "fullstaticortsdavis_a" (static added)?
Also below for the additional 3.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
```diff
@@ -2148,12 +2148,12 @@ public override void Initialize()
                                     Trace.TraceWarning("SteamGearRatioHigh not found in ENG file, or doesn't appear to be a valid figure, and has been set to default value");
                             }
                             // Adjust resistance for neutral gearing
-                            GearedRetainedDavisAN = DavisAN; // remember davis a value for later
-                            NeutralGearedDavisAN = DavisAN; // Initialise neutral gear value
+                            GearedRetainedDavisAN = DavisAN.Value; // remember davis a value for later
+                            NeutralGearedDavisAN = DavisAN.Value; // Initialise neutral gear value
                             float TempDavisAAmount = N.FromLbf((DavisMechanicalResistanceFactor * Kg.ToTUS(SteamEngines[0].AttachedAxle.WheelWeightKg))); // Based upon the Davis formula for steam locomotive resistance
                             if (TempDavisAAmount > 0.5 * DavisAN)
```
rwf-rr, line 2154: Should that not be "DavisAN.Value"?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
rwf-rr, line 1114: My understanding is that if any of the right-side variable is NULL, the left side will be NULL. Is that ok in this case (and below)?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
```diff
@@ -1942,9 +1962,9 @@ public override void Save(BinaryWriter outf)
             outf.Write(Variable1_3);
             outf.Write(Variable1_4);
             outf.Write(Friction0N);
-            outf.Write(DavisAN);
-            outf.Write(DavisBNSpM);
-            outf.Write(DavisCNSSpMM);
+            outf.Write(DavisAN.Value);
```
rwf-rr, line 1965: Is it save to assume that there always is a value at this stage? Or should saving also support NULL?

rwf-rr (commented):
In general looks good. I am a bit concerned about some references to .Value that are not checked for NULL. I am not very familiar with how C# handles nullable floats.
Maybe there is a location (during loading or initializing) to do an assertion that all variables have a value. 

There also is a case where the original code seems to be reading the wrong value from the config file (tag and variable names do not match).

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
SteelFill: Hmm, previously the default values for the parser didn't matter much, but now that would make a difference. I think 0 is the best default for these purposes.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
SteelFill: It is DavisAN.Value?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/FreightAnimations.cs
SteelFill: "FullORTSDavis_A" is the correct token syntax. The same token is used for both static and continuous freight animations, but in the code a distinction needs to be made, hence the different variable name.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: LoadFromWagFile guarantees that any Davis coefficient with a missing value has its value set to 0.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSWagon.cs
SteelFill: LoadFromWagFile guarantees that any Davis coefficient, including the freight animation ones, with a missing value has its value set to 0.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
rwf-rr: I meant line 2154, which multiplies with "DavisAN". If DavisAN were null, then the condition will be false.
But I gather from other feedback that at this stage DavisAN would never be null.

rwf-rr (approved):

Source/Orts.Simulation/Simulation/RollingStocks/MSTSSteamLocomotive.cs
SteelFill: Ah, that shouldn't cause any trouble.

PR #1208, opened 2026-03-02
Fix URL shown during installation

cjakeman:
The Readme.txt shown during installation has an out-of-date URL pointing to the content catalogue.


rwf-rr (approved):
Looks fine. URL works.

PR #1211, opened 2026-03-19
Shows icon to indicate that a notification is available

cjakeman:
If Internet access to the Notifications server fails, then this PR shows the Notification Icon, prompting the user to view the Notifications, which will show just the page for "Retry after error".


sweiland-openrails (approved):
fix ok, thanks

PR #1213, opened 2026-03-26
feat/adds Refresh button for Unstable version

cjakeman:
Also refreshes the form after Retry.

This feature requires an update to the openrails/notifications using [PR 7](https://github.com/openrails/notifications/pull/7)

This feature request was part of a related [bug report](https://bugs.launchpad.net/or/+bug/2142392)

[Forum thread](https://www.elvastower.com/forums/index.php?/topic/39561-refresh-button-for-menu-updates-for-open-rails/#entry325191)

[Trello card](https://trello.com/c/foqyIvMd/628-add-refresh-button-to-detect-new-version-available)


Source/Menu/MainForm.cs
```diff
@@ -371,6 +377,16 @@ void CheckForTelemetry()
             });
         }
 
+        public virtual void OnRefreshUpdates(EventArgs e)
+        {
+            // Force a Refresh of the notifications, which will check for updates and show the notification pages if there are any new notifications.
+            UpdateManager.SetChannel("Unstable"); // The 'Refresh' button is only available for the 'Unstable' mode.
```
twpol: I wonder if it would be better to either provide the channel in the button, or use the user's current channel?

twpol (dismissed):

Source/Menu/MainForm.cs
cjakeman: Good point. I was being a bit lazy, I guess.

twpol (approved):

PR #1215, opened 2026-03-30
Don't Assume MSTS ENG and WAG Files

SteelFill:
Bug report: https://www.elvastower.com/forums/index.php?/topic/37448-ammeter-configuration-steeringcarriages-emudmu/page__view__findpost__p__325239

Some cab cars do not work as expected when overriding a vehicle implemented in MSTS as a locomotive, then overwritten as a control car inside an `OpenRails` folder. Related to #1176 

It appears that some parts of the program were not scanning the `OpenRails` folder for an OR replacement of an MSTS engine or wagon, leading to an assortment of inconsistent results. This PR changes the cases I noticed to ensure that engine and wagon parsing will prefer to load the OR version of an engine or wagon file, as long as it is available.

To make all these cases cleaner, some helper methods were added in the new class `ORFileHelper` to give a shorthand way to replace a path to an MSTS file like `MSTS\TRAINS\TRAINSET\DASH9\dash9.eng` with the OR-specific path `MSTS\TRAINS\TRAINSET\DASH9\OpenRails\dash9.eng` without duplicating code.

For other devs: `ORFileHelper.GetORTSFilePath(string path)` would take in `MSTS\TRAINS\TRAINSET\DASH9\dash9.eng` as input, and return `MSTS\TRAINS\TRAINSET\DASH9\OpenRails\dash9.eng` _even if that file doesn't exist_.
On the other hand, `ORFileHelper.FindORTSFile(string path)` would take in `MSTS\TRAINS\TRAINSET\DASH9\dash9.eng` as input, and would return `MSTS\TRAINS\TRAINSET\DASH9\OpenRails\dash9.eng` _only if that file actually exists_. If that OpenRails file doesn't exist, then the original path `MSTS\TRAINS\TRAINSET\DASH9\dash9.eng` would be returned instead (note: it does not check if the original path exists, only if the OR-specific path exists).
In my inspection of existing processes, I found that there were cases where we want to replace the MSTS path with the OR-specific path (if it exists), but also cases where both the MSTS path and the OR-specific path were needed, so the two methods will cover both these use cases.

The cases where I noticed loading OR-specific files wasn't working are:

- `GenericWAGFile` was ignoring OR-specific engines and wagons, this resolves the control car vs electric locomotive problem in the original bug thread. Because some people have already made content that relies on this bug to function, if the engine type is missing from the `OpenRails` engine but not the MSTS one, the MSTS engine type will be used and a warning will be logged about the missing engine type.
- `Orts.Formats.Msts` `EngineFile` and `WagonFile` were ignoring OR-specific engines and wagons, which meant that non-simulator parts of the program (such as the main menu) would only load the MSTS versions of engines and wagons for displaying data. That data will now correctly reflect any changes/corrections made to the OR-specific files.

### DIFF
--- a/Source/Orts.Common/ORFileHelper.cs
+++ b/Source/Orts.Common/ORFileHelper.cs
@@ -24,13 +24,20 @@ namespace Orts.Common
     {
         /// <summary>
         /// Given any file path, returns a path to the same file type and name but contained inside an
-        /// "OpenRails" subfolder inside the same folder as the original file. Usually used to specify
-        /// an "Open Rails only" file to be loaded instead of an original MSTS file.
+        /// "OpenRails" subfolder inside the same folder as the original file. Usually used to determine
+        /// the location of an "Open Rails specific" version of a file to load alongside the original
+        /// "MSTS specific" file. <br />
+        /// (See <see cref="FindORTSFile(string)"/> if the ORTS-specific file should not be returned
+        /// unless it exists.)
         /// 
-        /// eg: Given "MSTS\TRAINS\TRAINSET\DASH9\dash9.eng", returns "MSTS\TRAINS\TRAINSET\DASH9\OpenRails\dash9.eng"
+        /// <para>
+        /// eg: Given <c>"MSTS\TRAINS\TRAINSET\DASH9\dash9.eng"</c>, returns <c>"MSTS\TRAINS\TRAINSET\DASH9\OpenRails\dash9.eng"</c>
+        /// </para>
+        /// 
+        /// See pull request #1215 for additional context.
         /// </summary>
-        /// <param name="path">File path to the "MSTS" file to be replaced with an "OpenRails" file.</param>
-        /// <returns>File path string pointing to the same file given in the 'path' parameter,
+        /// <param name="path">File path to a file that an OR-specific equivalent is desired for.</param>
+        /// <returns>File path string pointing to the same file given in the <paramref name="path"/> parameter,
         /// but contained inside an "OpenRails" subfolder.</returns>
         public static string GetORTSFilePath(string path)
         {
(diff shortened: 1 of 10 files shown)


Source/Orts.Formats.Msts/EngineFile.cs
```diff
@@ -41,6 +41,12 @@ public class EngineFile
 
         public EngineFile(string filePath)
         {
+            string dir = Path.GetDirectoryName(filePath);
+            string file = Path.GetFileName(filePath);
+            string orFile = dir + @"\openrails\" + file;
+            if (File.Exists(orFile))
+                filePath = orFile;
```
twpol: This pattern exists in a couple of places already and we're adding a couple more here, so it should probably be made into a utility somewhere

SteelFill: I can add something to one of our helper files to get the 'Open Rails file' given a regular 'MSTS' file. There are definitely a lot of cases out there where an equivalent of this is used.

Source/Orts.Common/ORFileHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with Open Rails.  If not, see <http://www.gnu.org/licenses/>.
+
+using System;
+using System.IO;
+
+namespace Orts.Common
+{
+    public class ORFileHelper
```
twpol: If all members of the class are `static`, we can mark the class itself as `static` too

Source/Orts.Simulation/Simulation/Simulator.cs
```diff
@@ -338,19 +338,19 @@ public Simulator(UserSettings settings, string activityPath, bool useOpenRailsDi
                 RDB = new RoadDatabaseFile(rdbFile);
             }
 
-            var carSpawnFile = RoutePath + @"\carspawn.dat";
+            string carSpawnFile = RoutePath + @"\carspawn.dat";
```
twpol, line 341: Slightly unsure why this change is being made, when we have the style rules configured opposite to it: https://github.com/openrails/openrails/blob/a7801848225ad43ba85e79349d958f4d48269788/Source/.editorconfig#L74-L77

twpol (dismissed):

Source/Orts.Simulation/Simulation/Simulator.cs
SteelFill: Personal preference really, I feel better with strong types and when I change a bunch of the uses of a variable, I end up checking the declaration of it and adjusting the type to keep track of what I'm doing.

Source/Orts.Common/ORFileHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (String.IsNullOrEmpty(path))
+                return "";
+
+            return Path.GetDirectoryName(path) + @"\OpenRails\" + Path.GetFileName(path);
+        }
+
+        /// <summary>
+        /// Given any file path, returns a path to the same file type and name but contained inside an
+        /// "OpenRails" subfolder if such a file exists. Otherwise, returns the original file path.
```
cjakeman: Unlike the summary for GetORTSFilePath(), this method lacks the "Usually used . . . " advice.

The opening comment in the PR explains very clearly why this Helper is useful and it would be a shame to lose that insight, so could we also have a reference back to the PR?

SteelFill: I looked into the documentation for comments and made the documentation here prettier and more legible as well, as it was becoming a wall of text otherwise.

cjakeman (approved):
Terrific analysis, especially "Because some people have already made content that relies on this bug to function". That makes it tricky, so thanks for this clean solution.

Approved and merged.

PR #1216, opened 2026-04-02
Add missing sounds on control cars

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/37448-ammeter-configuration-steeringcarriages-emudmu/page__view__findpost__p__325343

https://bugs.launchpad.net/or/+bug/2147612

### DIFF
--- a/Source/Orts.Simulation/Common/Scripting/PowerSupply/ControlCarPowerSupply.cs
+++ b/Source/Orts.Simulation/Common/Scripting/PowerSupply/ControlCarPowerSupply.cs
@@ -46,10 +46,9 @@ namespace ORTS.Scripting.Api
             }
         }
         protected override void SetCurrentMainPowerSupplyState(PowerSupplyState state) {}
-        protected override void SetCurrentAuxiliaryPowerSupplyState(PowerSupplyState state) {}
         protected override void SetCurrentElectricTrainSupplyState(PowerSupplyState state) {}
         protected override void SetCurrentDynamicBrakeAvailability(bool avail) {}
-        public override PowerSupplyState GetPowerStatus() => PowerSupplyState.Unavailable;
+        public override PowerSupplyState GetPowerStatus() => LpsHost.AuxiliaryPowerSupplyState;
         public void SignalEventToControlActiveLocomotive(PowerSupplyEvent evt)
         {
             ControlActiveLocomotive?.LocomotivePowerSupply.HandleEvent(evt);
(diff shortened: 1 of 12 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/ElectricTrainSupplySwitch.cs
```diff
@@ -18,6 +18,7 @@
 using Orts.Common;
 using Orts.Parsers.Msts;
 using ORTS.Scripting.Api;
+using SharpDX;
```
SteelFill: Is this intended? I find there are quite a few ways to get VS to think you need SharpDX when you actually wanted something else.

Source/Orts.Simulation/Common/Scripting/PowerSupply/ControlCarPowerSupply.cs
```diff
@@ -46,17 +46,16 @@ public int IndexOfControlActiveLocomotive
             }
         }
         protected override void SetCurrentMainPowerSupplyState(PowerSupplyState state) {}
-        protected override void SetCurrentAuxiliaryPowerSupplyState(PowerSupplyState state) {}
```
SteelFill, line 49: Is this one supposed to be removed entirely?

SteelFill (changes requested):
Looks like the relevant issue has been fixed without adverse consequences, just saw a couple things in my scan of the code that seemed off.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/ElectricTrainSupplySwitch.cs
cesarBLG: It's not, good catch! I guess it was automatically added who knows why.

Source/Orts.Simulation/Common/Scripting/PowerSupply/ControlCarPowerSupply.cs
cesarBLG: Yes, now control cars also have auxiliary supply, which means that the default behavior (i.e. activate the supply) is required instead of a no-op.

SteelFill (approved):
Ok, looks good to go.

PR #1219, opened 2026-04-20
Fix for https://bugs.launchpad.net/or/+bug/2148824 Null crash in timetable mode

Csantucci:


Roeterdink:
This patch is incorrect.
This is indeed the location where the crash occurs, but this is not the error.
When a train has a valid stop, it must have both valid arrival and departure times. That allows these values to be used elsewhere (e.g. for display, delay processing etc.) without the need for additional checks.
Setting the patch in this way may lead to crashes elsewhere.

This bit of code has always been like this. Care was taken that always both arrival and departure times were calculated for every stop. This rule has been broken somewhere else, and that problem has to be solved.

If I knew how I can set a review I would reject this patch - but I can't work out how I can set a review.

Regards,
Rob Roeterdink

Roeterdink (changes requested):
As said in comment, this patch is not correct.

Csantucci:
What if, waiting for the whole solution, the temporary patch would be as follows, to allow people to run their timetable:
                   if (arrivalDT.HasValue)
                    {
                        thisStation.arrivalDT = arrivalDT.Value;
                        if (departureDT.HasValue)
                        {
                            thisStation.departureDT = departureDT.Value;
                        }
                        else
                        {
                            thisStation.departureDT = thisStation.arrivalDT;
                        }
                    }

PR #1220, opened 2026-04-22
Fix documentation of Web Server API.

rwf-rr:
The Track Monitor request was incorrect, and the Driving and DPU requests were missing.

I am not set up to build the documentation. I would appreciate if someone could verify that my changes render correctly.


cjakeman (approved):
Hi Roger,

Thanks for sorting this out.

Approved and merged,

Chris

PR #630, opened 2022-03-28
Sky Color Fix (Addresses Trello Roadmap Card #367 for More accurate sunrise and sunset)

DR-Aeronautics:
**I updated SkyShader.fx to make the sunset and sunrise more realistic.**
**Link to Road Map Card**: https://trello.com/c/cQi6FPz5/367-more-accurate-sunrise-and-sunset
**Link to forum**: http://www.elvastower.com/forums/index.php?/topic/36123-discussion-code-fixing-the-openrails-sky/
Please forgive me if I'm uploading the code/doing pull requests wrong- this GitHub stuff is new to me.
The old pull request wasn't correct so it was closed and this one replaces it.
Old pull request: https://github.com/openrails/openrails/pull/628

**Improvements**
1.	Blue Sky remains until dusk (and darkens slowly)
2.	Stars no longer visible when the sun is above the horizon
3.	Sky color is now citrus, as opposed to full yellow or full red
4.	No rapid sky color changes (removed the yellow flash effect)
5.	Sunrise coloring now supported (it's a little more yellow than sunset)
Fixes were tested in 4 seasons on PRRV2 (Pennsylvania, USA), VBSR5 (Norway), and Blue Mountains (Australia)

**Videos of the results**:
Timelapse real vs old code vs new code: https://www.youtube.com/watch?v=6NHYx8R1vWo
Realtime test on PRRV2 Route: https://www.youtube.com/watch?v=OvW9w8UGZaI&t=259s

Attached is a memo on more information about the changes, including why I made them
[OpenRails.Sky.Recoding.Memo.pdf](https://github.com/openrails/openrails/files/8366374/OpenRails.Sky.Recoding.Memo.pdf)
:

PR #1221, opened 2026-04-22
New Sounds for Diesel Locomotives (+ Refactor of Sound Debug Window)

SteelFill:
Discuss on [ElvasTower](https://www.elvastower.com/forums/index.php?/topic/39618-new-sound-capabilities-especially-for-diesel-locomotives/)
Trello card [#630](https://trello.com/c/FH5txnc4)

To start a series of diesel engine improvements I would like to implement concurrently, I am beginning with sounds. After assorted discussions and perspective of comparing OR to other train sims, there were some notable limitations that this PR aims to resolve:
- Variable2 is too limiting, it only works between the engine idle RPM and max RPM (can't be used for any sounds below idle, like engine starting), it only responds to what the first engine is doing (multi-engine locos are basically incompatible with the MSTS sound system), and it's just confusing as the values produced by Variable2 are not in meaningful units
- There is no way to make sounds vary with engine load, which is severely limiting when there are many locomotives with engines that hold a constant RPM yet need to produce very different sounds depending on how hard the engine works.
- Diesel locomotives have no sound controls for tractive effort, which would be useful for motor or transmission sounds, as electric locomotives use Variable2 for tractive effort, but that's already reserved for engine RPM on diesels.

So, some new capabilities have been added to the sound system, give them a try:

New sound variable! ``EngineRPMControlled``
- Allows for programming sounds to respond based on actual engine RPM values, instead of a unitless value from 0 to 1.
- Works for all RPM values, even below idle RPM or above max RPM.
- Special setup in the parser allows for an arbitrary number of engines to be supported. ``EngineXRPMControlled`` will respond to the RPM of the Xth engine, so ``Engine2RPMControlled`` will use the RPM of the second engine. (If the engine number specified is invalid, or the locomotive has no diesel engines, the control will always return 0 RPM. If the value "X" is not present, the 1st engine is used.)
- For triggers, ``EngineRPM_Inc_Past`` and ``EngineRPM_Dec_Past`` can be used, and this also supports multiple engines ``EngineXRPM_Inc_Past`` and ``EngineXRPM_Dec_Past``

New sound variable! ``EnginePowerControlled``
- Controls sounds in response to actual instantaneous engine power output, with all values assumed to be in **watts** unless other power units are specified. Very useful for making sounds that change depending on how hard the engine is working, instead of how fast it is spinning.
- Also supports multiple engines, following the syntax ``EngineXPowerControlled``, eg ``Engine2PowerControlled``. As before, if there is no engine "X", the control will return 0 W, and if the value is not present then the power output of the 1st engine will be used.
- For triggers, use ``EnginePower_Inc_Past`` or ``EnginePower_Dec_Past``, and for multiple engines ``EngineXPower_Inc_Past`` and ``EngineXPower_Dec_Past``.

New sound variable! ``EngineTorqueControlled``
- Similar to engine power, but for personal preference some creators may find torque more appropriate to measure instantaneous power output. All torque values specified in the .sms file are assumed to be **newton-meters**.
- For multiple engines, use ``EngineXTorqueControlled``. Same rules as usual, if engine number "X" doesn't exist, a value of 0 is used, and if "X" isn't included the 1st engine is assumed.
- On triggers, use ``EngineTorque_Inc_Past`` and ``EngineTorque_Dec_Past`` for single engine locomotives, and ``EngineXTorque_Inc_Past`` and ``EngineXTorque_Dec_Past`` on multi engine locomotives.

New sound variable (for all locomotives)! ``TractiveEffortControlled``
- Control sounds depending on the tractive effort applied by the locomotive, regardless of type, using values in **newtons** by default, or other force units if entered, instead of arbitrary 0-1 values.
- Positive tractive effort values indicate tractive effort in the same direction as motion, but negative values will be produced as well to indicate effort opposite the direction of travel (such as dynamic braking).
- This also has triggers, ``TractiveEffort_Inc_Past`` and ``TractiveEffort_Dec_Past``.

New sound variable (for all locomotives)! ``TractivePowerControlled``
- Influence sounds using the power, default in **watts**, other power units allowed too, applied at the rail, regardless of locomotive type. Note that this is after all transmission losses, set values accordingly.
- Like tractive effort, positive power is in the same direction as the direction of travel (like when accelerating), while negative power will be returned if power is in the opposite direction to the direction of travel (like dynamic braking).
- Triggers are ``TractivePower_Inc_Past`` and ``TractivePower_Dec_Past``.

While testing this, I was hoping to use the sound debug window (Alt + S) to help experiment, but I found this menu lacking in detail and incredibly performance inefficient. So, I rewrote a lot of it:
- The selected sound source now lists **all** variables used by the sound system, including things like wheel RPM and angle of attack (in addition, of course, to the new variables for diesel locomotives), so it is more obvious what the sound system is thinking. This is very helpful for finding values the sound system is using that are not normally shown in the HUD, like engine torque.
- The units used to display variables are now consistent with the values that should be entered into the .sms file (eg: Variable2 will now show "0.867" instead of "86.7%" as "0.867" is the value used by the .sms file, .sms files do not speak in percentages.) and the actual units are given in the label for clarity (unless the quantity is unitless).
- Variables with multiple values, such as the RPM of each engine on a locomotive with 2 engines, are all shown in the text box separated by commas. This should help multi-engine sound tuning as it is clear what values each individual engine produces.
- The display for "Volume" now displays the volume of the specific sound stream that has been selected, instead of displaying an arbitrary stream's volume. A "Frequency (Hz)" display has also been added to verify that sound streams are running with the expected volume and frequency values.
- Track sounds are now shown in the sound debug window, expand the list entry for a train car to see the list of sound sources (both track sources and car sources) attached to that particular train car.
- Sound sources attached to train cars are now marked as [EX] for external sounds or [IN] for interior sounds, and [TRK] for track sounds or [CAR] for car sounds. This should make it more obvious which sound file is responsible for the sound source, in addition to listing the file name.
- Sound sources in the world are now marked as "World" sounds, with additional text indicating which world tile the sound is located in to help differentiate between instances of the same sound.
- Since the values of variables shown in the text boxes can get quite long, it is now possible to click and drag the split between the variable list and the list of sound sources to expand or contract the size of either as is needed. The list of variables itself is now scrollable so more values can be inspected without increasing the window size.
- To reduce performance impact, the sound list shown is only updated 1/20th at a time (except for the selected sound source, which is always updated). However, every time the sound debug window is opened the entire list is updated once before this kicks in.
- To reduce performance impact, the list of sound streams under each sound source is no longer destroyed and recreated on every update. Instead, the list of streams is created once, then the existing entries are updated on each subsequent iteration.
- To reduce performance impact, the list of sound streams under each sound source is only updated once, unless the sound source has been expanded to show the list of sound streams. There is no reason to update this list when it is not visible. (Note: the number of active sound streams is still tracked even with the sound source collapsed to hide the list of sound streams.)

Other bits of refactoring include:
- Updating ``Variable1_X_Controlled``, ``Variable1_X_Inc_Past``, and ``Variable1_X_Dec_Past`` to accept arbitrary values of X, instead of just 2, 3, or 4, to allow for arbitrary numbers of steam engine sounds to be supported. Same rules as diesel engines apply (if the value of "X" specifies a steam engine that doesn't exist, the control will always use a value of 0, if X is not present then it's assumed the first steam engine is used).
- Refactored the implementation of sound variables for both curves and triggers. Now curves and triggers use the same list of options, there is no need to manually add new sound variables to both curves and triggers, and the parsing for "_Inc_Past" and "_Dec_Past" triggers is handled procedurally, so these do not need to be defined individually. This should make it dramatically easier to add new sound variables.
- As a consequence, without any extra effort I have added missing handling for the triggers ``AngleofAttack_Dec_Past``, ``WheelRPM_Dec_Past``, ``ConcreteSleepers_Dec_Past``, and ``CarInTunnel_Dec_Past``. For some reason, all 4 of these had their appropriate "_Inc_Past" trigger handling, but the "_Dec_Past" triggers were ignored by the switch statement. Very strange, but now that these are interpreted automatically this should not happen anymore.
- Added a quick calculation for wheel speeds on wagons (assumes wheel speed is equal to actual speed, unless brake skid is happening, in which case wheel speed is 0) to 
- Variable sound curves and triggers now accept units of measure like anywhere else in the OpenRails program. Instead of using whatever arbitrary units are set as default, you can use units you are comfortable with.
- Implemented a fix for disposing unneeded wave files (previously, wave files would never be removed from memory even if they were not used anymore) that was originally implemented for hot reloading. This PR is a better place for this fix than hot reloading is.


Csantucci:
As you are reworking the sound debug window, you could add therein also the track sound sources.

SteelFill:
> As you are reworking the sound debug window, you could add therein also the track sound sources.

I can check, could be easy to set up, but integrating it _well_ might be a different story.

SteelFill:
Hey @peternewell, could you review this feature set? You were the last one to touch the sound system so you'll be more familiar with what's going on here (don't worry, Erick and I did some testing, and your track sounds still work after refactoring).

peternewell (approved):
Seems ok. Please keep an eye on it as it moves into Testing.

Thanks

PR #1226, opened 2026-04-29
Patch for Timetable Crash Bug 2148824

Roeterdink:
Bug: https://bugs.launchpad.net/or/+bug/2148824
Discussion: https://www.elvastower.com/forums/index.php?/topic/39610-timetable-loading-and-resuming-and-brakes/

### DIFF
--- a/Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
+++ b/Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
@@ -2503,7 +2503,6 @@ namespace Orts.Simulation.Timetables
             /// <returns> StopInfo structure</returns>
             public StopInfo ProcessStopInfo(string stationInfo, StationInfo stationDetails)
             {
-                string[] arr_dep = new string[2] { String.Empty, String.Empty };
                 string fullCommandString = String.Empty;
 
                 if (stationInfo.Contains('$'))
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
rwf-rr, line 3382: Should _validPassTime_ also be considered here?

Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
```diff
@@ -3323,12 +3348,34 @@ public StopInfo(string name, string arrTime, string depTime, TimetableInfo ttinf
...
+                        departureDT = arrivalDT;
+                    }
+                    else if (!validArrTime && validDepTime)
                     {
-                        departureTime = Convert.ToInt32(atime.TotalSeconds);
-                        departureDT = new DateTime(atime.Ticks);
+                        arrivalTime = departureTime;
+                        arrivalDT = departureDT;
                     }
                 }
```
rwf-rr, line 3380: Should there be a log when _arrSet_ is false? Or are there cases where there is no valid arrival time?

Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
```diff
@@ -3280,6 +3289,10 @@ public StopInfo(string name, string arrTime, string depTime, TimetableInfo ttinf
                         passTime = Convert.ToInt32(atime.TotalSeconds);
                         passDT = new DateTime(atime.Ticks);
                     }
+                    else
+                    {
+                        Trace.TraceInformation("Invalid passing time {0} at station {1}", arrTime, name);
```
rwf-rr, line 3294: Should these and subsequent traces be _Trace.TraceWarning_?

I believe TraceInformation is used when there is a suitable default behaviour, TraceWarning when there could be problems later.

Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
Roeterdink: No. The name is a little misleading, but the variable is only used for processing of actual stop, not passing.
Passing times are processed separately.

Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
Roeterdink: If no (valid) arrival time is defined but a valid departure has been defined, no specific arrival time is required - it does not really matter.

Source/Orts.Simulation/Simulation/Timetables/ProcessTimetable.cs
Roeterdink: No, Information is enough here. If no valid times are defined, the stop is simply removed. This will, in general, not seriously affect the timetable. If this is an AI train, most likely it is not even noticed. If it is the player train, the stop will not show up in the Station Stop window, but even that will not affect the running of the train.
Only if the missed stop had additional commands like attach, detach or transfer could this cause incorrect handling, but that is too complex to work out here.
Also, in general, this kind of incorrect time definition is rare. The crash for which this pull request has been set up, is the first of this kind since the creation of timetable mode more than 10 years ago.

rwf-rr (approved):

rwf-rr:
I am going to merge this PR, since it has been in approved state for a while, and with no additional comments in the PR, Bug or Discussion.

PR #1227, opened 2026-05-01
Track based sounds trigger error

peternewell:
Correct an error where sound axle count may not align with the sound triggers.

[Forum thread](https://www.elvastower.com/forums/index.php?/topic/39128-route-based-track-sounds)

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
@@ -918,12 +918,8 @@ namespace Orts.Simulation.RollingStocks
             realTimeTrackJointDistanceM = (float)Simulator.TRK.Tr_RouteFile.DistanceBetweenTrackJointsM; // Initialise track joint distance
             SoundAxleCount = (LocoNumDrvAxles + WagonNumAxles);
 
-            // make sure that sound axle count does not align with or exceed maximum possible sound triggers
-            if (SoundAxleCount < 2)
-            {
-                SoundAxleCount = 2;
-            }
-            else if (SoundAxleCount == 5) // No trigger for 5 axles, so increase to 6
+            // Where sound axle count does not align with sound triggers, then set to next highest value, also limit to 8 axles
+            if (SoundAxleCount == 5) // No trigger for 5 axles, so increase to 6
             {
                 SoundAxleCount = 6;
             }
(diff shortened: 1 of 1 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/TrainCar.cs
```diff
@@ -918,10 +918,22 @@ public virtual void Initialize()
...
             SoundAxleCount = (LocoNumDrvAxles + WagonNumAxles);
 
-            // make sure that axle count does not exceed maximum possible trigger
-            if (SoundAxleCount > 8)
+            // make sure that sound axle count does not align with or exceed maximum possible sound triggers
+            if (SoundAxleCount < 2)
             {
-                SoundAxleCount = 8f;
+                SoundAxleCount = 2;
+            }
```
SteelFill: A case where this change might have unexpected consequences is if the number of axles is 1 (I believe 0 axles is disallowed), which might be the case for a dummy wagon or a segment of an articulated unit. In that case we might not want that to make axle sounds at all, like would happen before, instead of triggering the 2-axle sounds.

A better solution would be to not hardcode the triggers for certain numbers of axles, though I suspect the 1, 5, and 7-axle edge cases now accounted for are fairly uncommon.

peternewell: Thanks for pointing this out, I will limit it to a minimum of 2 axles.

peternewell: > A case where this change might have unexpected consequences is if the number of axles is 1 (I believe 0 axles is disallowed), which might be the case for a dummy wagon or a segment of an articulated unit. In that case we might not want that to make axle sounds at all, like would happen before, instead of triggering the 2-axle sounds.
>
I have now disabled sounds if there is less the 2 axles. 

If you are happy with it, can you approve it?

 
> A better solution would be to not hardcode the triggers for certain numbers of axles, though I suspect the 1, 5, and 7-axle edge cases now accounted for are fairly uncommon.
>
The odd axle counts would predominantly cover steam locomotives. 5 axles = 4-6-0, or 2-6-2 locomotive, and 7 axles = 4-6-4, and etc.

SteelFill (approved):
Looks good, this will help handle those edge cases in the track sound system.

peternewell:
Thanks for that.

PR #1228, opened 2026-05-12
Trigger the watchdog at load time

pzgulyas:
Trigger the watchdog after each loading phase.

Bug report: https://www.elvastower.com/forums/index.php?/topic/39652-threadhangexception-thread-loader-process-has-hung/

### DIFF
--- a/Source/RunActivity/Viewer3D/Materials.cs
+++ b/Source/RunActivity/Viewer3D/Materials.cs
@@ -444,20 +444,16 @@ namespace Orts.Viewer3D
         public void Mark()
         {
             MaterialMarks.Clear();
-            foreach (var path in Materials.Keys)
-                MaterialMarks.Add(path, false);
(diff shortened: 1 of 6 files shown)


Source/RunActivity/Viewer3D/Materials.cs
rwf-rr, line 1: Just to confirm, the changes in this file are unrelated to the watchdog, correct?

rwf-rr (commented):
Changes look fine to me. I poked around a bit and did not find any other locations where a ping of the watchdog would be necessary.

Source/RunActivity/Viewer3D/Materials.cs
pzgulyas: Yes, in my case this was the function where the watchdog got triggered in more than one occasion, so I tried to implement a little speed-up that never hurts.

rwf-rr (approved):

PR #1231, opened 2026-06-09
Log consists with duplicate UiDs

Csantucci:
Feature request: https://www.elvastower.com/forums/index.php?/topic/39687-camera-3-does-not-go-to-the-end-of-train/


SteelFill (approved):
Looks good, and I see it has already helped diagnose some malformed consists. Good catch on checking both stationary and moving consists, this problem can show up anywhere.

Csantucci:
> Looks good, and I see it has already helped diagnose some malformed consists. Good catch on checking both stationary and moving consists, this problem can show up anywhere.

Thanks, Phillip

PR #1232, opened 2026-06-16
Fix Settings Exporter description to match all other applications

twpol:
Just a tiny bug I spotted


Csantucci (approved):
Just a tiny positive review :)

PR #1233, opened 2026-06-18
adds Log Motion Data to DataLogger

cjakeman:
This is a temporary addition to the DataLogger in order to track down an anomaly in which the steam loco running gear (i.e. coupling rods and cranks) drift out of sync with the driving wheels.

DRAFT as not intended to be merged into the Testing Version.

Instructions: 
On the Options > Data Logger, select "Log motion data" and set the "Logging interval" to 2,000 msecs.
Start the loco moving and press F12 to start the logger. Run the loco for 15 mins to provoke drift and press F12 to stop the logger.
The results will be in file OpenRailsDump.csv on the Windows Desktop.


SteelFill:
Hey Chris, this was related to the issue solved by #1230, right? If this experiment is no longer needed, we could close this PR, at least to de-clutter the PR list a little.

PR #1234, opened 2026-07-08
Use system settings for Data Logger Units, for train physics only.

rwf-rr:
Previously, the data logger had a setting for the speed unit, but used internal units for everything else.

This PR changes the data logger to use the system settings to determine the units. The speed, force, distance and pressure units are then converted from the internal unit to the system-configured units.

There is an override in the data logger options to use internal (native) units, i.e do not do any conversion.

These changes apply only to the train physics section of the data logger (the previous speed-unit setting also applied to the train physics only). The steam locomotive data seems rather specific, and I am not familiar with steam locomotives. So I left them unchanged. Maybe someone else can add unit conversion to them. There is a proposal for steam in [this thread](https://www.elvastower.com/forums/index.php?/topic/39448-units-for-steam-locos-in-the-data-logger/page__view__findpost__p__324380), but it lacks feedback.

See [forum discussion](https://www.elvastower.com/forums/index.php?/topic/39186-who-is-using-the-data-logger-and-for-what) and Trello card [#617](https://trello.com/c/kGbRsMeL/617-data-logger-configurable-interval-configurable-units-more-train-data).

A **significant change** is that the unit is now included in the header line, and the subsequent cells only have the value. This makes graphing the values simpler. Originally the parameters with conversion had the unit with the data cells.

This is how the data looked before:

```
... | Player Throttle [%] | Player Motive Force [N] | ... | Player Speed [route] | ... | Distance [m] | ...
... |                  13 |                   66304 | ... |               1.1mph | ... |          402 | ...
```
When using internal units, then it will look like this:

```
... | Player Throttle [%] | Player Motive Force [N] | ... | Player Speed [m/s] | ... | Distance [m] | ...
... |                  13 |                   66304 | ... |                0.5 | ... |          402 | ...
```
When using imperial units (eg. system settings is _route_, and the route uses miles):

```
... | Player Throttle [%] | Player Motive Force [klbf] | ... | Player Speed [mph] | ... | Distance [mi] | ...
... |                  13 |                     14.905 | ... |                1.1 | ... |         0.076 | ...
```

When using metric units (eg. system settings is _route_, and the route uses km):
```
... | Player Throttle [%] | Player Motive Force [kN] | ... | Player Speed [km/h] | ... | Distance [km] | ...
... |                  13 |                   66.304 | ... |                 1.7 | ... |         0.402 | ...
```

Data Logger Options form:

DataLoggerOptions-Form

A sample output file, with different settings (look for header lines).

[OpenRailsDump.csv](https://github.com/user-attachments/files/29777762/OpenRailsDump.csv)


SteelFill:
Looks like we got unlucky and I have added some documentation related to logging in my PR #1235 that conflicts with the documentation added here. I have marked #1235 as for-unstable so it can be tested too. Feel free to remove for-unstable from that PR if you want to test additional changes here, I will be out of town and unable to progress on anything code related for a couple weeks anyway.

SteelFill (approved):
Looks good to me, I have tested with a variety of units and with the system units and it worked as it should. This is definitely easier to use than the previous implementation.

PR #1235, opened 2026-07-14
Expanded Engine RPM Behaviors

SteelFill:
Discuss on [Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/39781-expanded-engine-rpm-behaviors/)
Some related Trello card(s): [Engines for ETS](https://trello.com/c/Ie6a79WQ), [Dynamic Brake RPMs](https://trello.com/c/WT2Nof6m)

The main goal of this PR is to allow diesel engines (at least, on diesel electric and diesel hydraulic locomotives) to operate at any arbitrary RPM and provide sensible performance while doing so, as the current limitations and assumptions placed on engine RPM behavior are entirely inadequate for simulating much beyond the most basic diesel locomotives. This has required and motivated some substantial refactoring of diesel engines, but I have also added some long-desired features too.

## Disconnect Tractive Effort from Engine RPM (for diesel-electric and diesel-hydraulic)
The current diesel engine implementation in Open Rails only allows the throttle setting to change engine RPM, and as throttle increases the engine RPM must also strictly increase. In reality, there are very few locomotives where both of those limitations are true (locomotives with dynamic brakes usually change engine RPM with dynamic braking, and a lot of locomotives don't have strictly increasing engine RPM as idle RPM is the same as the RPM for the first power setting), so this has greatly hindered the maximum accuracy possible when modeling diesel engines.
The reason OR has implemented these rules is that the engine RPM has been used to determine the amount and rate of change of locomotive tractive effort. If engine RPM could be influenced by something other than throttle, then tractive effort would also be affected, which wouldn't make sense for some of the systems being requested. And if engine RPM wasn't strictly increasing as throttle increases, it would become ambiguous which tractive effort curve should be applied at which RPM. Given that, it's clear more flexible and accurate engine RPM behavior requires splitting tractive effort from engine RPM.

- **Removed** the `ReverseThrottleRPMTab` and disconnected all behaviors related to this. The target tractive effort is now determined exclusively by the throttle setting (actual tractive effort will still respect engine performance, read below).
- Removed the check for strictly increasing RPM values in `ThrottleRPMTab`. RPM values can be **held the same** at multiple throttle settings **or even decrease** as throttle increases.
- To still allow for **smooth changes in tractive effort** without using engine RPM, PR #1067 is used. Default values (only if user hasn't entered any) for `ORTSTractivePowerRampUpRate`, `ORTSTractivePowerRampDownRate`, and `ORTSTractivePowerRampDownToZeroRate` are calculated based on the min/max engine RPM and rates of change of RPM. This will provide similar power delivery to the previous engine RPM based system, but without the limitations that entailed.
- `ORTSTractiveForceIsPowerLimited` if not set is set automatically based on the tractive power and engine power the locomotive uses, to prevent overloading the engine in cases where the engine power has been set realistically (fake 'diesel' locomotives with lower engine power than tractive power will not be power limited, as has been the case)
- `ORTSTractiveForceRamp` and `ORTSDynamicBrakeForceRamp` values are now calculated automatically if they were missing, using the MSTS parameter `RunUpTimeToMaxForce`. If the run up time isn't specified, 10 seconds is assumed for diesels. This will prevent instant jumps in tractive effort at low speed.
- The ["dynamically evolving tractive force"](https://open-rails.readthedocs.io/en/unstable/physics.html#dynamically-evolving-tractive-force) feature was not compatible with tractive force ramp rates and was causing other undesired effects. The calculation has been updated to integrate better and (I think?) work closer to the original intent: tractive effort will be limited to the continuous tractive effort after some time running at or above the continuous tractive effort. Below the continuous tractive effort, nothing changes.
- Force ramp rates also weren't playing nice with activities that start the train in motion, initialization of locomotives was changed to also run a few methods to initialize axle and adhesion data so that such activities begin with the proper amount of tractive effort applied.

Note that for mechanical transmissions everything's going to be pretty much the same, these changes are only meaningful for diesel electric and diesel hydraulic styles.

## Disconnect Fuel Consumption from Engine RPM
For a long while now, fuel consumption of ORTS diesel engines has been based entirely on the current engine RPM. This is a passable assumption when locomotive power output is also directly tied to RPM (though it still gives inaccurate fuel consumption when at low load, like with the reverser in neutral), but with engine RPM no longer the sole factor in determining tractive effort that assumption falls apart and there are now many more situations where fuel consumption would make no sense.

- **Fuel consumption now scales with the actual engine power output.** An engine running below its rated power output (like in neutral) will consume less than the rated amount of fuel. An engine running above its rated power output will consume more than the rated amount of fuel. No additional user input needed, but there are some things users can do to make fuel use more accurate.
- To prevent excessively low fuel consumption at low load, like what happened [here](https://www.elvastower.com/forums/index.php?/topic/23758-diesel-engine-fuel-consumption/) on a previous implementation, **engine friction is now considered**. Based on some research, an estimate of friction vs RPM is considered, using the fuel consumption at idle (when the engine is working against friction only) to determine the exact amount. Higher RPM leads to higher friction, requiring more fuel to work against it even in neutral. The user cannot enter friction values directly, as these are not given in spec sheets.
- **Auxiliary power draw has also been added**, which combines with friction to establish the fuel use at low load. Unlike friction, the user can actually enter auxiliary power values since these often are documented, using the new `AuxiliaryPowerTab` (like `DieselPowerTab`, this is a table of RPM values and power values), and auxiliary power draw adds to engine power output. For now, the only variable auxiliary load is ETS, and the other auxiliary loads are always present. but in the future, this can be expanded to consider intermittent loads like fans and compressors. If an aux power table is not defined, it will be estimated based on the difference between engine gross power (as defined by `DieselPowerTab`) and the engine traction power (as defined by `ORTSMaxTractiveForceCurves`).
- Added new display to the Alt + F5 locomotive debug HUD. The **power in/out graph now has an orange segment**, indicating power the engine is making that isn't going to the transmission (like auxiliary loads or ETS).
- New locomotive engine section parameter `ORTSDieselTransmissionEfficiency` (provide a value from 0 to 1) can be used to determine how much load should be applied to the engine relative to the locomotive's drive power. The efficiency here is the efficiency between the engine shaft and the wheels, not just the efficiency of the motor/gearbox/torque converter. A correct efficiency setting will result in the correct amount of load on the engine, and thus accurate fuel consumption. Power before efficiency loss is also used in some menus where traction power is more appropriate than rail power (such as horsepower per ton calculations).
- New diesel engine parameter `IdleDieselConsumption` can be used to define the fuel used when the engine is at idle RPM and is only working against the constant auxiliary loads and friction. Giving this value helps OR **estimate friction** and makes fuel consumption at low load more accurate at other RPMs. If `IdleDieselConsumption` is not given, OR uses the `DieselConsumptionTab` value at the idle RPM, but as this value should be set for a loaded engine, not an unloaded engine, that won't be quite as accurate for determining low-load fuel use.
- For accurate fuel use when RPM is changing (additional fuel should be required to speed the engine up, less fuel to slow it down), **engine rotational inertia is now considered**. While inertia can be entered in the engine with `ShaftInertia` (in units of kilogram * meters^2), this usually isn't going to be in spec sheets. If the parameter is missing, inertia will be estimated using the previously mentioned friction values and `ChangeDownRPMpS`, assuming the engine slows down entirely under friction. This isn't strictly correct for all cases but can provide a good starting point for manual data entry.
- Changes in engine RPM also affect the simulated power output, with engines speeding up producing more power and engines slowing down producing less power. This can be used by things like the sound system to reflect the work the engine is doing just to change speed.
- Revised the handling of **changing engine RPM to have more smoothness**, should avoid sudden changes in fuel use caused by changing engine RPM.
- **Reworked the built-in cooling system behavior**, which was quite odd and probably never worked as intended, to use the newly redone fuel calculations as a proxy for heat generation, and to consider ambient temperature, when determining temperature increase and decrease. You can now expect the engine to get hot running hard on a summer day, and stay cool at idle in the winter, instead of whatever it was doing before.

This set of changes in particular will be very important for the future idea of actually simulating engine aspiration and forced induction. Better fuel modeling certainly helps right now, but that's also half the battle for better air modeling too.

## General Refactoring
Alongside all the other updates, I did some serious cleanup, particularly for `DieselEngine.cs` to make things work a bit better and cleaner where I could. Not all of this is strictly useful for engine RPM behavior but has been done to keep things clean for future diesel engine features.

- Added parsing for missing parameter `IdleExhaustMagnitude`. This is the lifespan of each exhaust particle at idle. Previously, the only way to enter this was with the MSTS parameter `DieselSmokeEffectInitialMagnitude` outside the ORTS diesel definition. A complete diesel engine should not require any parameters outside the diesel definition. If this parameter is missing, the MSTS parameter will still be used as a fallback.
- Similarly, added parsing for missing parameter `MaxExhaustMagnitude`. This is the lifespan of each exhaust particle at max RPM, used to require entering `DieselSmokeEffectMaxMagnitude` outside the ORTS diesel definition.
- And also added `ExhaustDecelColor` as a diesel engine parameter. For some reason, this color was hardcoded, while the other variations on exhaust color were not hardcoded. Now you can set whatever color you want.
- Refactored how power is split between engines on locomotives with more than one engine to better handle engines operating at different power levels.
- Removed the `SettingsFlags`/`initLevel` behavior. This was far too clunky and non-expandable. Now, to check if something is initialized, we just check if it's not `null` or nonzero instead of adding all this extra bespoke behavior.
- Cleaned up initialization of diesel engines and diesel locomotives, including adding a couple helper functions to automatically determine if a value is initialized and construct a log message for verbose logging, greatly cutting down on duplicated code in this part of the process.
- Removed the separate behavior for basic/advanced engine initialization, there was no need to treat these differently. The log will still specify if the engine has been created from MSTS or ORTS parameters, but otherwise the process is the same.
- Added additional conversions to the `pS` conversion class for to/from radians per second, as there are many situations with engines and other rotating things where radians per second are needed for a calculation.
- Added the `RPM` conversion class. Behaves exactly like the `pS` class but considers things per minute instead of per second. Also very handy for rotating things where we are using rotations per minute but often need radians per second too.
- General cleanup of variable names, spelling, comments, indentation, spacing, etc.

## New Diesel Engine Features
Finally, all that work allows for some interesting new capabilities for arbitrarily setting engine RPM that people have been asking for since long ago, without causing any strange behaviors with tractive effort or fuel consumption. There is more that I want to do than just the current feature set, but considering the sheer scale of this PR as imposed by the refactoring needed just to get to this point, I will probably wait to do more later.

### Discrete Speed Control
Due to the way ORTS engines are built out of lookup tables in the form of linear interpolators, it is possible for input values in-between those defined to result in output values in-between what's defined. That is just how interpolation works, of course, but for certain systems in some locomotives that just isn't accurate; only discrete output values can be produced even if the input is continuous. One such case is the engine RPM demanded by a governor. Electro-hydraulic governors can only produce a few **predefined RPMs, and not anything in between** (hence the 8-notch throttle seen on North American equipment, as opposed to a fully analog throttle handle).

To replicate that, rounding has been added as a feature to interpolators (which technically means it's not an interpolator anymore) which can be implemented in a diesel engine by adding the `SpeedControl` parameter to the diesel engine block. The accepted values for `SpeedControl` are `SpeedControl ( "RoundDown" )`, `SpeedControl ( "RoundUp" )`, `SpeedControl ( "RoundNearest" )`, and `SpeedControl ( "Continuous" )` (this is the default mode). If any of the "Round" modes are selected, then any table that produces an RPM value (eg: `ThrottleRPMTab`) will round the input (eg: throttle %) from the actual value to one of the values defined in the table, then return the RPM for that. This prevents producing an "in-between" RPM value, much like an electro-hydraulic governor would. The direction of rounding is specified by the specific mode ("Up" "Down" or "Nearest"), though if the input throttle % is already an exact value then no rounding happens.

For example, let's say we have this `ThrottleRPMTab`:
```
Engine (
    ORTSDieselEngines ( 1
        Diesel (
            ThrottleRPMTab (
                0.0         400
                25.0        600
                50.0        800
                75.0        1000
                100.0       1000
            )
        )
    )
)
```
If the throttle % is currently 40%, then the default behavior (`SpeedControl ( "Continuous" )`), then the resulting engine RPM will interpolate between 600 RPM (the 25% throttle setting) and 800 RPM (the 50% throttle setting), giving 720 RPM. If the governor is electronic, it wouldn't actually be able to set 720 RPM. Instead, if `SpeedControl ( "RoundDown" )` is added to the engine, the 40% throttle value will be "rounded down" to 25% (the next lowest or equal throttle setting in the table) and the RPM will be set to 600 RPM. Conversely, `SpeedControl ( "RoundUp" )` would round the 40% throttle to 50%, giving 800 RPM. And `SpeedControl ( "RoundNearest" )` would also round 40% up to 50%, as 40% is closer to that than to 25%.

Hopefully that makes sense. This isn't accurate for every locomotive, but it is accurate for some and can allow you to make smaller lookup tables instead of defining individual RPM values for every possible throttle setting.

### Engine RPM During Dynamic Braking
Many diesel locomotives with dynamic braking will use engine RPM other than idle to provide extra cooling to the various parts of the transmission that are likely to get hot during dynamic braking. People have known this should be a thing for a long time, but as mentioned earlier ORTS previously considered engine RPM a function of throttle setting only. However, now with tractive effort disconnected from engine RPM, it is viable to change the engine RPM without causing trouble for tractive effort, so we can finally give engine speed during dynamic braking.

This is done with the `DynamicsRPMTab` in the engine definition, which behaves just like the `ThrottleRPMTab` but instead of using the throttle handle to set RPM, uses the dynamic brake handle. If a locomotive doesn't have a dynamic brake controller, then `DynamicsRPMTab` will be ignored. And, of course, if the dynamic brakes are disabled then these RPMs will be ignored. `DynamicsRPMTab` works with `SpeedControl` too, which can be used to do rounding of the dynamic brake setting.
```
Engine (
    ORTSDieselEngines ( 1
        Diesel (
            SpeedControl ( "RoundDown" )
            ThrottleRPMTab (
                0.0         400
                25.0        600
                50.0        800
                75.0        1000
            )
            DynamicsRPMTab (
                0.0         400
                75.0        800
            )
        )
    )
)
```
In this example, dynamic braking keeps the engine at idle RPM of 400 until 75% dynamic braking is reached, at which point RPM will increase to 800 RPM and stay there up to 100% dynamic braking. Many other arrangements of engine RPM are possible, enter in as many dynamic brake and RPM values as needed to replicate whatever behavior your locomotive employs.

Fuel use will increase as RPM increases, but not to the same level as when powering thanks to the previously mentioned fuel consumption rework.

### Engine RPM With ETS Enabled
Another commonly considered situation where something other than throttle affects engine RPM is locomotives with electric train supply. Many diesel passenger locomotives use a diesel engine to provide ETS power, but not without changing the RPM behavior of that engine (perhaps because the engine can't make sufficient power at its idle RPM, or a specific RPM is needed to produce a specific frequency of power). Previously, this would require a different engine file with a different diesel engine definition, but now `ETSThrottleRPMTab` and `ETSDynamicsRPMTab` can be added to a diesel engine to give correct behavior in both normal power and with ETS enabled.
```
Engine (
    ORTSDieselEngines ( 1
        Diesel (
            SpeedControl ( "RoundDown" )
            ThrottleRPMTab (
                0.0         400
                25.0        600
                50.0        800
                75.0        1000
            )
            DynamicsRPMTab (
                0.0         400
                75.0        800
            )
            ETSThrottleRPMTab (
                0.0         600
                50.0        800
                75.0        1000
            )
            ETSDynamicsRPMTab (
                0.0         600
                75.0        800
            )
        )
    )
)
```
In this example, with ETS enabled the engine RPM increases from a minimum of 400 to 600, which requires changes to both the throttle and dynamic brake RPM behavior. Obviously, `ETSThrottleRPMTab` and `ETSDynamicsRPMTab` will only be used if ETS is enabled for the locomotive. Note that if `ETSDynamicsRPMTab` is not defined, then the 0% throttle setting of `ETSThrottleRPMTab` will be used, even if `DynamicsRPMTab` is present.

Fuel use with ETS enabled will vary with the ETS load, which is added on to the auxiliary power draw and subtracted from the traction power.

### Engines for ETS Only
Related to giving the proper RPM for ETS, some locomotives (or non-locomotive power cars) employ engines that only supply power for ETS and not for tractive effort (and vice versa). This can be controlled using the diesel engine parameters `ProvidesTractionPower ( 0/1 )` and `ProvidesETSPower ( 0/1 )` (both of these are set to 1 [enabled] by default). Disabling traction power means the engine performance will be unaffected by the tractive effort the locomotive applies, while disabling ETS power means the engine performance will be unaffected by the ETS draw from the coaches. ETS will not actually enable unless at least 1 currently running engine `ProvidesETSPower`.

This can be combined with multiple engines to produce locomotives with a dedicated engine for traction and a small engine for ETS. In this case, the power draw from traction and ETS will not interfere with eachother, giving more accurate locomotive performance and fuel consumption.
```
Engine (
    ORTSDieselEngines ( 2
        Diesel (
            ProvidesTractionPower ( 1 )
            ProvidesETSPower ( 0 )
            ThrottleRPMTab (
                0.0         400
                25.0        600
                50.0        800
                75.0        1000
                100.0       1000
            )
        )
        Diesel (
            ProvidesTractionPower ( 0 )
            ProvidesETSPower ( 1 )
            ETSThrottleRPMTab (
                0.0         1800
                100.0       1800
            )
        )
    )
)
```

There are many other non-throttle engine RPM responses I would like to tackle, such as compressor speed-up and low idle, but dynamic brake and ETS RPM are the easiest to implement and will be most familiar to those who already set up engines with throttle RPM, so they are the best first example to show off the new capabilities while I work on the more complicated features.


SteelFill:
(Now resolved) Unfortunately, #1234 conflicts with this due to adding additional documentation at the exact same spot in Options.rst. Marking as for-unstable so this gets added to at least one unstable build.

Source/Documentation/Manual/physics.rst
```diff
@@ -579,195 +579,515 @@ It is also possible to operate a locomotive with the own engine off and
...
 
-Diesel Exhaust
-..............
+Types of Governors
+^^^^^^^^^^^^^^^^^^
+
+By default, OpenRails assumes engine speed control is fully
+analog and the engine can maintain any arbitrary speed, including
+speeds in-between the values defined in ``ThrottleRPMTab``. This
+is appropriate for locomotives with penumatic controlled governors,
```
cesarBLG:
```suggestion
is appropriate for locomotives with pneumatic controlled governors,
```

Source/Documentation/Manual/physics.rst
```diff
@@ -579,195 +579,515 @@ It is also possible to operate a locomotive with the own engine off and
...
-
-In ORTS, single or multiple engines can be set for one locomotive. In case
+In ORTS, one locomotive may have any number of diesel engines, as opposed
+to MSTS which only considered one engine per locomotive. In case
 there is more than one engine, other engines act like *helper* engines
 (start/stop control for helpers is ``<Ctrl+Y>`` by default). The power of
 each active engine is added to the locomotive power. The number of such
-diesel engines is not limited.
+diesel engines is not limited. Depending on the settings of ``ProvidesTractionPower``
+and ``ProvidesETSPower``, various multi-engine locomotives can be constructed,
```
cesarBLG, line 617: What do you think about using ProvidesAuxiliaryPower instead? E.g. for electrical compressors, or DMUs with heating/air conditioning.

Source/Documentation/Manual/physics.rst
```diff
@@ -789,8 +1110,13 @@ diesel engine rated power.
 
 In ORTS, the diesel-electric locomotive can use
 ``ORTSTractionCharacteristics`` or tables of ``ORTSMaxTractiveForceCurves``
-to provide a better approximation to real world performance. If a table is
-not used, the tractive force is limited by MaxForce, MaxPower and
+to provide a better approximation to real world performance.
+``ORTSTransmissionEfficiency`` can also be added to consider the
+efficiency of all transmission components between the engine and
+wheels (generator, cabling, motors, gearboxes, etc.) and increase the
```
cesarBLG: There's already a TransmissionEfficiency parameter in the Axle module, so the sentence 'all transmission components up to the wheels" is probably not correct. We should probably use a per-diesel-engine transmission efficiency (instead of your global ORTSTransmissionEfficiencyParameter) between engine and gearbox/generator, and the Axle one between gearbox/electric motor and wheels

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -611,6 +613,38 @@ public override void LoadFromWagFile(string wagFilePath)
...
+                TractionForceRampDownToZeroNpS = 10.0f * TractionForceRampUpNpS;
+            }
+            if (MaxDynamicBrakeForceN > 0 && RunUpTimeToMaxForceS > 0 &&
+                DynamicBrakeForceRampUpNpS <= 0 && DynamicBrakeForceRampDownNpS <= 0 && DynamicBrakeForceRampDownToZeroNpS <= 0)
+            {
+                DynamicBrakeForceRampUpNpS = MaxDynamicBrakeForceN / RunUpTimeToMaxForceS;
+                DynamicBrakeForceRampDownNpS = 2.0f * DynamicBrakeForceRampUpNpS;
+                DynamicBrakeForceRampDownToZeroNpS = 10.0f * DynamicBrakeForceRampUpNpS;
+            }
+
```
cesarBLG, line 641: What is the intended use of RunUpTimeToMaxForceS? I think it only makes sense for diesel-electric locomotives where it represents the rate of change of the inverter frequency (and maybe for diesel-hydraulic locomotives where it represents something related to the torque converter). For diesel-mechanical locomotives, the force should be directly provided by the engine, right?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -648,145 +612,266 @@ public bool IsInitialized
...
         /// <summary>
         /// Engine load table - Max output power vs. RPM
         /// </summary>
         public Interpolator DieselPowerTab;
-
-         /// <summary>
+        /// <summary>
+        /// Auxiliary power table - Auxiliary power draw vs. RPM
+        /// </summary>
+        public Interpolator AuxPowerTab;
```
cesarBLG, line 780: Electrical auxiliary loads should be handled by the Power Supply, also for electric locomotives. It's fine for mechanical loads, but maybe this is a simplification (e.g is a mechanical compressor always active?)

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -966,51 +1043,58 @@ public void Copy(DieselEngine other)
...
         }
 
         public void Initialize()
         {
             if (!Simulator.Settings.NoDieselEngineStart)
             {
                 RealRPM = IdleRPM;
+                RawRPM = RealRPM;
                 State = DieselEngineState.Running;
             }
```
cesarBLG, line 1089: Could you look at the bug that prevents diesel-mechanical locomotives with some gearboxes from starting/stopping? This block shouldn't be needed once that is fixed

cesarBLG (changes requested):

Source/Documentation/Manual/physics.rst
SteelFill: I don't want to lump in all auxiliaries together as there are configurations out there where a second engine does have a generator for ETS but is disconnected from other auxiliary systems like air compressors. Right now, the only things that can dynamically affect engine load are traction and ETS, hence why those two things specifically can be turned on or off. As more variable loads are considered, those should have additional, _individual_, ways to determine which engine should take the load. Combining multiple load types into one parameter would arbitrarily disallow certain configurations that are realistic.

cesarBLG: Agreed, then I think that power supply code has to be adjusted so the ETS is only connected when the relevant engine is running?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: This is just a simplification and abstraction of many different factors. Actual auxiliary power draw comes from cooling fans, water pumps, oil pumps, battery charging, blowers, running a compressor/exhauster, and so on. In an ideal world, an auxiliary power table would not be needed, as the engine would simply check every single subsystem to see which ones require power...but a lot of these things aren't simulated, so a basic constant load vs RPM is a decent simplification (this is also the way auxiliary power draw is depicted in some documents, allowing for direct data entry).

Even an estimate for power used by fans and pumps is important as this sets the minimum possible fuel consumption, but in the future we should actually simulate the power used by fans, and users who choose to use that simulation should then remove the estimated contribution of the fans from the auxiliary power table. Repeat as features are added until the auxiliary power table just represents the various tiny loads that don't justify dedicated simulations.

cesarBLG: Agreed, but I think we should point in the manual that this only affects mechanical loads. Electrical loads (e.g. electrical compressor, battery chargers) should be handled by the power supply, as they also appear in electrical locomotives. Maybe adding an AuxiliaryPowerDraw parameter that is accepted for any locomotive type is the best approach for them.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: There is a lot going on with diesel mechanical locomotives that is questionable and I have had to restrain myself from changing everything there, this PR is big enough. I think a lot of the issues with diesel mechanical come down to trying to calculate the engine RPM directly, resulting in a lot of 'faked' behaviors instead of simulating torques directly and allowing RPM change to result from torque imbalance. Trying to tackle that any other way just means more rules and workarounds with more odd behaviors, so something for another time.

Source/Documentation/Manual/physics.rst
SteelFill: I believe after changing the engine load to be calculated based on `Locomotive.LocomotiveAxles.DrivePowerW` that does mean the transmission efficiency is technically everything up to, but _not_ including, the axle model. There just isn't a way to define the axle efficiency, which I hesitated to add because that could interact oddly with tractive force curves as axle force is multiplied by the axle efficiency, but users expect the force they enter to be the at rail force, which should already be accounting for this efficiency loss, so some extra calculating would be needed to un-account for efficiency. I'll have to see the consequences of that.

I definitely don't think it's a good idea to attach efficiency per engine as it's not a property of the engine. The same engine with the same settings can (and should, especially for the sake of reusing data with include files) be compatible with a diesel hydraulic, DC diesel electric, and AC diesel electric transmission. Each of those should have different efficiency, despite using the same engine.

Source/Documentation/Manual/physics.rst
SteelFill: Yeah it shouldn't be much trouble to add a check for at least one running engines that `ProvidesETSPower`.

Source/Documentation/Manual/physics.rst
cesarBLG: The wheel-rail interface has no 'transmission efficiency' by itself, as the heat losses are taken into account by the friction parameters. I guess the parameter should be removed from the axle model and moved somewhere else (it is currently unused). But, for diesel-electric locomotives, we still need to distinguish between the transmission efficiency between diesel engine and generator, and between electric motor and axle (the latter also affecting electric locomotives). For electric motors, this means that the force produced by the motor is greater than what is specified by the tractive force curves, which represent the rail force. Having two separate parameters would be helpful for dual locomotives.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: Everything I've seen about the MSTS `RunUpTimeToMaxForce` suggests that it does exactly what it says; sets the time required for the locomotive to go from 0 force to `MaxForce`. This could be a characteristic of the control system, load regulator, mechanical linkages, torque converter filling, it's a pretty generic parameter that should only show up in pretty old content, kind of the MSTS equivalent for traction force ramp. This is here to make MSTS locomotives work better with the changes to the system.

As for diesel mechanicals, what I've seen in testing suggests they already ignore ramp rates, which makes sense, the rate of change of power there should solely depend on what the engine can do (though the engine can change power output instantly, which is something for a different time).

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: So I understand what you are going for; loads that are simulated elsewhere, like in the power supply system, should definitely _not_ be part of the auxiliary power table. ETS should not be in the auxiliary power table since there is a different system for that. But that consideration is not unique to "electrical" loads, "mechanical" loads should also be removed from the auxiliary power table if they are being calculated elsewhere (right now, I don't know of any examples for this). Should there be a system for calculating the load of "mechanical" compressors (in reality, such a system would be shared for both electrical and mechanical compressors) then that load should not be included in the auxiliary power table, even though it's a "mechanical" component.

The important thing for users to know is not that the auxiliary power table needs to consider mechanical vs electrical, it's that they need to be sure they don't double-count auxiliary loads; if cooling fan power draw is simulated on your engine (soon, TM) then the cooling fan power should not be included with the auxiliary power table. Doesn't matter if the cooling fan is electrical or mechanical, either way it still takes power from the engine so it must still be accounted for, but only once.

A single parameter in the engine might work for electric locomotives (though electric locomotives don't have something that is influenced by total energy use, while diesel locomotives have fuel consumption to worry about), but that just isn't adequate for diesel locomotives where it's well known that average auxiliary power draw varies with RPM. I did not dream up the idea of auxiliary power vs RPM, that relationship shows up in many places, such as from [NASA](https://ntrs.nasa.gov/citations/19820018174) (where they termed it "accessory horsepower").
image

cesarBLG: > So I understand what you are going for; loads that are simulated elsewhere, like in the power supply system, should definitely not be part of the auxiliary power table. 

Not exactly. My point is that there are some "tiny loads that don't justify dedicated simulations" which come from electrical devices. Since those can appear both on electric and diesel locomotives, they should be handled by the power supply. That's why I'm suggesting a static parameter that represents those subsystems, e.g. "engine(ortsauxpowersupplycontinuouspower", equivalent to "wagon(ortspowersupplycontinuouspower" for passenger cars. So to sum up:
- Subsystems that are properly simulated are not included elsewhere (e.g. battery chargers)
- Other electrical charges are included in the power supply
- Mechanical or RPM-dependent charges are included in AuxPowerTable

> though electric locomotives don't have something that is influenced by total energy use

For example, high speed trains have current consumption limits in conventional lines that reduce their maximum speed and acceleration

> I did not dream up the idea of auxiliary power vs RPM, that relationship shows up in many places

Yes, it's reasonable that some loads, even electrical ones (e.g. the fuel pump) depend on RPM. I'm in favor of keeping AuxPowerTable, it's just that there should be another parameter that tells the diesel engine how much power is demanded by the power supply subsystem.

cesarBLG: But this can be done in a separate PR since it's additional things and not related to diesel engines

Source/Documentation/Manual/physics.rst
cesarBLG: #1238 removes the transmission-related parameters from the axle module (and sets efficiency and ratio to 1 for electric motors)

Source/Orts.Simulation/Common/Scripting/PowerSupply/DieselPowerSupply.cs
```diff
@@ -51,6 +51,7 @@ public DieselEngineState CurrentDieselEngineState(int id)
             }
         }
         protected float DieselEngineOutputPowerW => DieselEngines.MaxOutputPowerW;
+        protected float DieselEngineAvailablePowerW => DieselEngines.AvailablePowerW;
```
cesarBLG: Maybe naming it "AvailableTractionPowerW"?

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
```diff
@@ -611,6 +613,38 @@ public override void LoadFromWagFile(string wagFilePath)
...
+            if (DynamicBrakePowerRampDownToZeroWpS < 0) DynamicBrakePowerRampDownToZeroWpS = DynamicBrakePowerRampDownWpS;
+
+            // Estimate tractive and braking ramp rates if none were given
+            if (RunUpTimeToMaxForceS < 0 && this is MSTSDieselLocomotive)
+            {
+                if (this is MSTSDieselLocomotive)
+                    RunUpTimeToMaxForceS = 10.0f; // Use 10 second ramp up as default for diesels
+                else
+                    RunUpTimeToMaxForceS = 5.0f; // Use 5 second ramp up as default for electrics
+            }
```
cesarBLG: What was the intended behavior here? There are two checks for MSTSDieselLocomotive. If it's intended to work for electric locomotives, wouldn't it break existing content by inserting unrealistic delays?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -648,145 +612,266 @@ public bool IsInitialized
...
+        /// Engine dynamic brakes settings table - RPM vs. dynamic brakes settings
         /// </summary>
-        public Interpolator ReverseThrottleRPMTab;
+        public Interpolator DynamicsRPMTab;
         /// <summary>
-        /// Throttle setting as calculated from real RpM
+        /// Engine throttle settings table w/ ETS - RPM vs. throttle settings when ETS is active
         /// </summary>
-        public float ApparentThrottleSetting;
+        public Interpolator ETSThrottleRPMTab;
```
cesarBLG, line 800: This deprecates DieselEngineMinRpmForElectricTrainSupply, I guess we should remove it

Source/Documentation/Manual/physics.rst
SteelFill: When that's merged I'll make sure it still works here. Though I think right now the only time a motor is ever created is specifically for locomotives set to use the AC traction motor type, everything else ends up with force driven axles. A generic "DC" motor implementation that accepts forces from tractive effort curves (as opposed to considering voltages and currents, which we are not ready for) would be needed for an `ORTSMotorEfficiency` value to have an effect in all (diesel)-electric locomotives.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: I didn't realize I accidentally removed the line of code that was handling `DieselEngineMinRPM` when removing references to `ReverseThrottleRPMTab`, so now it does nothing. I think that parameter should still function, but it does behave slightly redundantly to `ETSThrottleRPMTab`.

Do you think `DieselEngineMinRpmForElectricTrainSupply` should be used to generate an `ETSThrottleRPMTab` that just holds the engine at a constant RPM equal to `DieselEngineMinRpmForElectricTrainSupply` (unless an `ETSThrottleRPMTab` has been defined manually, or the engine does not provide ETS, of course)? I think that would roughly do the same thing as before.

Source/Orts.Simulation/Common/Scripting/PowerSupply/DieselPowerSupply.cs
SteelFill: Still refining naming, this metric is specifically for power free to be used by the transmission so that would be a bit more descriptive.

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
SteelFill: I was experimenting a bit with this, and the current setup intentionally suppresses the behavior for electric locomotives since I wasn't too sure on that. Everyone's used to the diesels taking time to load, but not the electrics so it seemed out of place. This will probably be left to diesel locomotives only, if people want to slow down the electric locomotives they can do that manually.

Source/Documentation/Manual/physics.rst
cesarBLG: Maybe the parameter can be renamed to "ORTSDieselTransmissionEfficiency" to show that it's for diesel locomotives

Source/Orts.Simulation/Simulation/RollingStocks/MSTSLocomotive.cs
cesarBLG: Yes, it's better to keep it only for diesel locomotives. 5 seconds is too high for metro railways

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
cesarBLG: Yes, creating a table with a single value would work.

Then we have the DieselEngineMinRpm variable handled by the power supply. That one should also be removed as it's not compatible with multiple engines. The idea behind that variable was to implement a governor that increases RPM depending on demanded power, not on throttle/ETS status, but it's better to implement that inside the DieselEngine class

Source/Documentation/Manual/physics.rst
SteelFill: New behavior checks for at least one engine that is running, provides ETS, and is at or above the minimum RPM for ETS. The ETS switch can still be turned on or off, but ETS won't actually turn on until the engine(s) are ready.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: Removed `DieselEngineMinRpm` and added automatic creation of `ETSThrottleRPMTab` based on the value of `DieselEngineMinRpmForElectricTrainSupply`.

Source/Orts.Simulation/Common/Scripting/PowerSupply/DieselPowerSupply.cs
```diff
@@ -50,22 +50,47 @@ public DieselEngineState CurrentDieselEngineState(int id)
...
-        protected float DieselEngineOutputPowerW => DieselEngines.MaxOutputPowerW;
-        protected float DieselEngineAvailablePowerW => DieselEngines.AvailablePowerW;
-
-        public float DieselEngineMinRpmForElectricTrainSupply => DpsHost.DieselEngineMinRpmForElectricTrainSupply;
-
-        public float DieselEngineMinRpm
+        /// <summary>
+        /// True if at least 1 engine can provide ETS power
+        /// </summary>
+        public bool ElectricTrainSupplyEquipped()
```
cesarBLG: This variable name could be confused with "ElectricTrainSupplyUnfitted()". I'd replace both methods with a single one CurrentElectricTrainSupplyGeneratorStatus() that returns a PowerSupplyState (Unavailable, PowerOff, PowerOn).

SteelFill: `ElectricTrainSupplyUnfitted()` has to be retained for non-diesel power supplies, but `CurrentElectricTrainSupplyGeneratorState()` replaces that for the diesel power supply, giving `PowerSupplyState.Unavailable` if no diesel engine `ProvidesETSPower`, `PowerSupplyState.PowerOff` if at least one diesel engine does provide ETS but none that provide ETS are running/are not at sufficient RPM, and `PowerSupplyState.PowerOn` if at least one diesel engine does provide ETS, is running, and is at the proper RPM.

cesarBLG: Sorry, I was suggesting to combine ElectricTrainSupplyAvailable and ElectricTrainSupplyEquipped into CurrentElectricTrainSupplyGeneratorState, which is what you did, so it's fine.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -2050,8 +2045,10 @@ public void InitRailPower(float totalTractionPower, float totalETSPower)
                 }
             }
 
-            // Set rail power vs RPM from calculated rail powers
-            if (RailPowerTab == null)
+            // Determine rail power vs RPM, used for other calculations
+            Interpolator railPowerTab = null;
+
+            if (railPowerTab == null)
```
cesarBLG: Is there a reason to keep the if clause? It always evaluate to true since railPowerTab is always null at this point

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
```diff
@@ -638,155 +602,272 @@ public bool IsInitialized
...
+        public bool EngineBrakingLockout = false;
 
         /// <summary>
         /// The RPM controller tries to reach this value
         /// </summary>
-        public float DemandedRPM;           
-        float demandedThrottlePercent;
-        float throttleAcclerationFactor = 1.0f;
+        public float DemandedRPM;
+        float ThrottleAcclerationFactor = 1.0f;
```
cesarBLG:
```suggestion
        float ThrottleAccelerationFactor = 1.0f;
```
maybe rename the variable to correct the typo?

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/Axle.cs
```diff
@@ -1093,7 +1093,9 @@ public void Initialize()
         }
         public void InitializeMoving()
         {
-            AxleSpeedMpS = TrainSpeedMpS;
+            // Small amount of slip is required for adhesion, set axle speed just above true speed
+            AxleSpeedMpS = TrainSpeedMpS * 1.01f;
```
cesarBLG: I don't think this is correct. The 101% fixed ratio can be above the slip threshold, causing the simulation to start with a wheel slip. Axle speed can be above (accelerating), below (braking) or equal (coasting) train speed, so I don't see a justification to set it above true speed. Setting axle speed equal to train speed ensures that wheel slip/skid does not appear at startup if the applied force is below the adhesion limit.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: Fixing spelling in this codebase could be a full time job, and I am not helping!

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerSupplies/DieselEngine.cs
SteelFill: This keeps the scope of some temporary variables in check. Since I have no intent to ever make this a user input, the slightly simpler approach is to remove the `if` and leave the calculation inside `{ }` which isn't pretty but it works.

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/Axle.cs
SteelFill: If the axle speed is set equal to the actual train speed, then the resulting lack of slip will cause the axle to report maximum adhesion happens at 0 slip per [line 903](https://github.com/openrails/openrails/blob/69f105bb5bb82b44fc861ac485a316469c7056cb/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/Axle.cs#L903), which results in calculating the axle has 0 adhesion per [line 1379](https://github.com/openrails/openrails/blob/69f105bb5bb82b44fc861ac485a316469c7056cb/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/Axle.cs#L1379), which tells the locomotive it can not output any tractive effort (at least, with the "Full" slip control system), so the activity begins with the locomotive giving no tractive effort even if throttle is applied...which is the entire problem I was encountering and trying to fix.

I have my doubts about `ComputeWheelSlipThresholdMpS()` returning 0 m/s as the slip threshold when there is currently exactly 0 m/s of slip as this simply doesn't make sense (no part of the calculation relies on the slip speed, it's not like an error happens if the slip speed is 0), so I can try removing that arbitrary behavior and seeing what happens.

cesarBLG: Yes, line 903 makes no sense, I don't know why it was added. Maybe it was intended for the case where TrainSpeedMpS==0 as then the maximum adhesion happens at 0 slip.

cesarBLG (approved):
Thanks for the work and for addressing all my comments, it looks good to me

PR #1236, opened 2026-07-18
adds MissingItem type to handle new item types in notifications/menu.json

cjakeman:
With this bug fix, notifications/menu.json can add new item types without breaking the Notifications in older versions of Open Rails.

OR v1.6 and v1.6.1 do not have this fix. Before a new item type (e.g. Refresh button) can be added, these versions will need to fetch notifications/menu.json from a different source.

### DIFF
--- a/Source/Menu/Notifications/NotificationManager.cs
+++ b/Source/Menu/Notifications/NotificationManager.cs
@@ -32,6 +32,7 @@ using ORTS.Updater;
 using static ORTS.Common.SystemInfo;
 using static Menu.Notifications.NotificationPage;
 using Newtonsoft.Json.Serialization;
+using System.Diagnostics;
 
 // Behaviour
 // Notifications are read only once as a background task at start into Notifications.
@@ -250,7 +251,7 @@ namespace Menu.Notifications
                 // Check criteria for each item and add the successful items to the current page
                 foreach (var item in n.ItemList)
                 {
-                    if (AreItemChecksMet(item)) AddItemToPage(Page, item);
+                    if (AreItemChecksMet(item) && !(item is MissingItem)) AddItemToPage(Page, item);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


Source/Menu/Notifications/NotificationManager.cs
```diff
@@ -671,11 +672,24 @@ public Type BindToType(string assemblyName, string typeName)
...
+                    var result = typeof(Notifications).Assembly.GetType($"{@namespace}.{name}");
+
+                    // Any errors are silenced, so write to the debug output instead.
+                    if (result == null)
+                    {
+                        Debug.WriteLine($"WARNING: {@namespace}.{name} not recognised item type");
+                    }
+
+                    return (result is null)
+                        ? typeof(Notifications).Assembly.GetType("Menu.Notifications.MissingItem")
```
twpol:
```suggestion
                        ? typeof(MissingItem)
```

twpol (dismissed):

Source/Menu/Notifications/NotificationManager.cs
cjakeman: Thanks. Adopted in a new commit.

Source/Menu/Notifications/NotificationManager.cs
```diff
@@ -671,11 +672,24 @@ public Type BindToType(string assemblyName, string typeName)
...
+
+                    // Any errors are silenced so, for debugging, write a message instead.
+                    if (result is null)
+                    {
+                        Debug.WriteLine($"WARNING: {@namespace}.{name} not recognised item type");
+                    }
+
+                    return (result is null)
+                        ? typeof(MissingItem)
+                        : typeof(Notifications).Assembly.GetType($"{@namespace}.{name}");
```
rwf-rr: Should that not be ": result"? It is the same _typeof_ expression as on line 677.

cjakeman: Hi Roger,
Yes, of course.
Fixed and thanks.

rwf-rr (approved):

PR #1237, opened 2026-07-21
added Control Rectangles (Ctrl+F5) for 3d locomotives

sweiland-openrails:
Added Control Rectangles (Ctrl+F5) for 3d locomotives. They are not rectangles like for 2d locomotives, but yellow points instead. Rectangles are to difficult to draw for this 3d locomotives. But they give a nice overview where the controls are to be found.


cjakeman:
Nice idea, Siebren.

Could we have a Trello card please and a forum thread?

Thanks,

Chris

SteelFill:
I had a try with this feature and I noticed that sometimes the position of the yellow dots would lag behind the actual 3D position of the control. This deviation would get worse as speed increases. That kind of thing happens when the 3D position used for one calculation is out of sync with what's on the screen, though it seemed to be alternating between the correct and incorrect position, so there might be a misalignment in update rate between the dots and the 3D world.

PR #1238, opened 2026-07-23
Remove transmission from axle module

cesarBLG:
For diesel-mechanical locomotives, it should depend on the gearbox. For electric locomotives, moved to ElectricMotor class. Since it is always set to 1, it does not change anything.

### DIFF
--- a/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/ElectricMotor.cs
+++ b/Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/ElectricMotor.cs
@@ -36,7 +36,7 @@ namespace Orts.Simulation.RollingStocks.SubSystems.PowerTransmissions
         /// <summary>
         /// Positive nonzero transmission ratio, given by n1:n2 ratio
         /// </summary>
-        public float TransmissionRatio { get; protected set; }
+        public float TransmissionRatio { get; protected set; } = 1.0f;
 
         /// <summary>
         /// Transmission efficiency, relative to 1.0, within range of 0.0 to 1.0 (1.0 means 100%, 0.5 means 50%)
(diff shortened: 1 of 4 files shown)


Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/ElectricMotor.cs
```diff
@@ -33,6 +33,16 @@ public class ElectricMotor : ISubSystem<ElectricMotor>
         public float SurfaceM;
         public float WeightKg;
 
+        /// <summary>
+        /// Positive nonzero transmission ratio, given by n1:n2 ratio
+        /// </summary>
+        public float TransmissionRatio { get; protected set; }
```
SteelFill: I'm still out of town, so I cannot check for sure, but I think `TransmissionRatio` needs to be set to 1.0 by default, as if left at the default value of 0 (which will always happen without a way to define the gear ratio) division by 0 will occur in other calculations. I think that could be the source of [this crash](https://www.elvastower.com/forums/index.php?/topic/39800-error-systemarithmeticexception-function-does-not-accept-floating-point-not-a-number-values) but it is hard to tell where exactly things went wrong with NaN errors since the error isn't thrown until after the calculation that produced NaN.

```suggestion
        public float TransmissionRatio { get; protected set; } = 1.0f;
```

SteelFill (changes requested):

Source/Orts.Simulation/Simulation/RollingStocks/SubSystems/PowerTransmissions/ElectricMotor.cs
cesarBLG: You're right, I don't know how I forgot that

SteelFill (approved):
No new trouble reported in the last week, and I haven't noticed any change in behavior in my driving. Looks good to go.

PR #1240, opened 2026-08-03
Fix bug with auxiliary tender

peternewell:
https://www.elvastower.com/forums/index.php?/topic/39811-aux-water-tender-problem/page__pid__327586#entry327586

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -199,7 +199,7 @@ namespace Orts.Simulation.Physics
         public float MaxAuxTenderWaterMassKG;
         public bool IsAuxTenderCoupled = false;
         bool AuxTenderFound = false;
-        string PrevWagonType;
+        TrainCar PrevWagonType;
 
         public bool HasControlCarWithGear = false;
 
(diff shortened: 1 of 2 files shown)


rwf-rr:
The changes in code look fine and safe. As I am not a steamer, I hope someone else, that experimented with the fix, can provide the approval.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2349,52 +2354,55 @@ public void UpdateAuxTender()
...
-                                    mstsSteamLocomotive.CurrentAuxTenderWaterMassKG = Cars[i].AuxTenderWaterMassKG * (Simulator.Activity.Tr_Activity.Tr_Activity_Header.FuelWater / 100.0f); // 
-                                    IsAuxTenderCoupled = true;      // Flag to advise MSTSSteamLovcomotive that tender is set.
-                                    AuxTenderFound = true;      // Auxililary tender found in consist.
+                                    if (mstsSteamLocomotive.AuxTenderMoveFlag == false)  // If locomotive hasn't moved and Auxtender connected use fuel presets on aux tender
+                                    {
+                                        MaxAuxTenderWaterMassKG = Cars[i].AuxTenderWaterMassKG;
+                                        mstsSteamLocomotive.CurrentAuxTenderWaterMassKG = Cars[i].AuxTenderWaterMassKG * (Simulator.Activity.Tr_Activity.Tr_Activity_Header.FuelWater / 100.0f); // 
+                                        IsAuxTenderCoupled = true;      // Flag to advise MSTSSteamLovcomotive that tender is set.
+                                        AuxTenderFound = true;      // Auxililary tender found in consist.
```
SteelFill, line 2370: Is there not a reason to `return` from this method after finding the first functional auxiliary tender? As far as I can tell, there is no reason to continue scanning through the entire train once one functional auxiliary tender is found. It even looks to me like having a second auxiliary tender in the consist could cause `MaxAuxTenderWaterMassKG = 0` and `IsAuxTenderCoupled = false` even if a functional aux tender was found earlier.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -199,7 +199,7 @@ public int ActivityDurationS
         public float MaxAuxTenderWaterMassKG;
         public bool IsAuxTenderCoupled = false;
         bool AuxTenderFound = false;
-        string PrevWagonType;
+        TrainCar PrevWagonType;
```
SteelFill: I don't see any reason this variable needs to be stored for the entire train, it is only used temporarily in `UpdateAuxTender`, and the value doesn't appear to require persistence between frames as it is re-calculated every time.

Source/Orts.Simulation/Simulation/Physics/Train.cs
```diff
@@ -2334,6 +2334,11 @@ public void UpdateWindComponents()
         //================================================================================================//
         /// <summary>
         /// Update Auxiliary Tenders added to train
+        /// NOTE TODO:  It has been reported that setting the WagonType to "Tender for an AuxTender will cause a crash.
+        /// To overcome this in MSTSWagon.cs in FindAuxTendersSteamLocomotive() set the following line.
+        /// if (TendersSteamLocomotive == null) ignore, and return (break out of the loop) with an appropriate error message.
+        /// Note, ideally this AuxTender function should be expanded to allow more then one AuxTender to be added to a consist and,  
+        /// also multiple steam locomotives in the smae consist, each with an AuxTendershould also be catered for.
```
SteelFill: Spelling
```suggestion
        /// also multiple steam locomotives in the same consist, each with an AuxTender should also be catered for.
```

SteelFill (changes requested):
Fix was reported as working, but there might be some situations where an aux tender fails to be detected. See comments.

Source/Orts.Simulation/Simulation/Physics/Train.cs
peternewell: Fair, enough, though it will be fixed when multiple aux tenders are allowed for.

Source/Orts.Simulation/Simulation/Physics/Train.cs
peternewell: Fair enough

Source/Orts.Simulation/Simulation/Physics/Train.cs
peternewell: Thanks.

SteelFill (approved):
Looks good to go now, thanks!

peternewell:
Thanks

PR #1241, opened 2026-08-03
Add Support for Dynamic Road Sections

SteelFill:
Extension of #954 
Code originally written by goku

Discuss on [Elvas Tower](https://www.elvastower.com/forums/index.php?/topic/38114-multiple-track-profiles/)
[Trello card](https://trello.com/c/AwthpV6y)

This PR adds support for dynamic road sections and custom track profiles for dynamic track. Previously, only tracks could have dynamic sections (with arbitrary length and curve radius) generated, and these sections would only use the default track profile.

With the addition of a `ShapeTemplate` parameter to world objects, regular track and dynamic track can now define a specific track profile to use (the given value is interpreted as the file name of the track profile to use, or the `Name` of the track profile if none of the file names match, `ShapeTemplate ( "DEFAULT" )` uses the default track profile while `ShapeTemplate ( "DISABLED" )` forces the static shape to be used instead of a track profile) instead of the previous behavior that automatically selected a track profile. Using this feature will require editing a route, so the previous behavior has been retained (as long as `ShapeTemplate` is not given) so track profiles still work without editing a route.

Dynamic track objects with the 8th bit of `StaticFlags` set (hex flag 0x00000100) will be interpreted as dynamic *road* objects instead (this flag only applies to dynamic track objects). This allows for different behavior when using a "track" profile to procedurally generate these shapes. Namely, no superelevation will be applied (superelevation isn't calculated for the road database), no overhead wires will be generated, and different track profiles will be used (ideally, these "track" profiles should be built to have the appearance of a road). Otherwise, the capabilites and functionality are the same as for dynamic track, as much of the existing code is reused.
Unlike for track profiles, there is no hardcoded default road profile, so if dynamic roads are to be used then custom profiles must be provided (else dynamic roads will have the appearance of dynamic track). The track profile named `TrProfileRoad` (name pending) will be used as the default profile for roads, unless a specific profile is given by `ShapeTemplate`.

This system is probably expandable to things other than roads, additional behavior may be added for generic scenery objects (think of fences, walls) or the existing terminology might be rephrased to be less road-centric.

Todo:

- [ ] Decide on file naming convention
- [ ] Add some way to differentiate between profiles meant for track and profiles meant for road, to prevent the two being mixed
- [ ] Add support for multiple track profiles in a singular track profile file to reduce clutter in the TRACKPROFILES folder
- [ ] Documentation

### DIFF
--- a/Source/Orts.Formats.Msts/WorldFile.cs
+++ b/Source/Orts.Formats.Msts/WorldFile.cs
@@ -508,6 +508,7 @@ namespace Orts.Formats.Msts
         public float Elevation;
         public uint CollideFlags;
         public JNodePosn JNodePosn;
+        public string ShapeTemplate;
 
         public TrackObj(SBR block, int detailLevel)
         {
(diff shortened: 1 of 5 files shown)


cjakeman:
Sounds really good, Phillip. Is there a Trello card that I can approve?

SteelFill:
> Sounds really good, Phillip. Is there a Trello card that I can approve?

Trello [card #633](https://trello.com/c/AwthpV6y) added.

PR #1242, opened 2026-08-04
Upgrade to .NET 6

twpol:
**Support lost for**

- Microsoft Windows 7
- Microsoft Windows 8.1
- Microsoft Visual Studio 2019

**Support remains for**

- Microsoft Windows 10 (1607+)
- Microsoft Windows 11 (22H2+)
- Microsoft Visual Studio 2022
- Microsoft Visual Studio 2026

**Note:** MonoGame upgraded from 3.8.0 to 3.8.1 so that it matches with .NET 6

[Forum thread 1](https://www.elvastower.com/forums/index.php?/topic/39706-moving-on-from-windows-7-sp1)
[Forum thread 2](https://www.elvastower.com/forums/index.php?/topic/39789-macos-port-of-open-rails/)

### DIFF
--- a/Source/Contrib/DataValidator/DataValidator.csproj
+++ b/Source/Contrib/DataValidator/DataValidator.csproj
@@ -1,7 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework Condition="'$(BuildDotNet)' == 'true'">net6-windows</TargetFramework>
-    <TargetFramework Condition="'$(TargetFramework)' == ''">net472</TargetFramework>
+    <TargetFramework>net6-windows</TargetFramework>
     <OutputType>Exe</OutputType>
     <AssemblyName>Contrib.DataValidator</AssemblyName>
     <IsPublishable>False</IsPublishable>
--- a/Source/Contrib/DataValidator/DataValidator.csproj
+++ b/Source/Contrib/DataValidator/DataValidator.csproj
@@ -17,7 +17,6 @@
     <ProjectReference Include="..\..\Orts.Simulation\Orts.Simulation.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="Microsoft.CSharp" Version="4.7.0" />
     <PackageReference Include="Microsoft.DotNet.UpgradeAssistant.Extensions.Default.Analyzers" Version="0.4.355802">
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
--- a/Source/Launcher/Program.cs
+++ b/Source/Launcher/Program.cs
@@ -20,7 +20,7 @@
  * This is the program which users launch. Its purpose is to check for the
  * required dependencies and Open Rails files before launching the menu.
  *
- * .NET 6 checks for itself on launch
+ * .NET checks for itself on launch
  */
 
 using System;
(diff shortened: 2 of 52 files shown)


pzgulyas:
There are references left to .NET 4.7.2 here:
https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/Launcher/Program.cs#L98-L111

twpol:
> There are references left to .NET 4.7.2 here:
> 
> https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/Launcher/Program.cs#L98-L111

Removed the .NET Framework check in the launcher entirely, as .NET Core/.NET 5+ does this check itself

pzgulyas (commented):
These app.config files still hold a reference:

https://github.com/openrails/openrails/blob/master/Source/Menu/app.config#L4

Should the files be deleted entirely?

Source/Launcher/Program.cs
```diff
@@ -18,30 +18,19 @@
 /* Open Rails Launcher
  *
  * This is the program which users launch. Its purpose is to check for the
- * required dependencies before launching the menu, so it is designed to
- * run on clean versions of Windows 7 and up
+ * required dependencies and Open Rails files before launching the menu.
  *
- * Dependencies checked for:
- *   - Microsoft .NET Framework 4.7.2
+ * .NET 6 checks for itself on launch
```
pzgulyas, line 23: Might make the comment version independent.

```suggestion
 * .NET checks for itself on launch
```

twpol:
> These app.config files still hold a reference:
> 
> https://github.com/openrails/openrails/blob/master/Source/Menu/app.config#L4
> 
> Should the files be deleted entirely?

I've removed all of the `<startup>` sections (removing most but not all files - some had other unrelated items in)

pzgulyas (commented):
There is a win7 reference here:
https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/Launcher/app.manifest#L8-L9

and here:
https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/RunActivity/Prerequisites.txt#L9-L10

Source/Contrib/DataValidator/DataValidator.csproj
pzgulyas: This line can be deleted, I think.

pzgulyas (dismissed):

twpol:
> https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/Launcher/app.manifest#L8-L9

I left this for now as it seems harmless

> https://github.com/openrails/openrails/blob/b5109a8ec55ffcd5f155e49a0b3114815e166480/Source/RunActivity/Prerequisites.txt#L9-L10

I fixed up and tested the installer, which now downloads .NET 6 and installs it - not sure it is worth making it fully offline but if we want, I believe it is easy enough

Source/Contrib/DataValidator/DataValidator.csproj
twpol: I can't actually work out what this package _does_ but removing it doesn't seem to break the build, so it's gone `¯\_(ツ)_/¯`

pzgulyas (approved):

PR #1243, opened 2026-08-07
Fix for system name not extracted by Notifications

cjakeman:
On moving to .Net 6, Open Rails will no longer be compatible with Windows 7 and 8.1.
A Notification Check will be needed to block the installation of an Open Rails update on such systems.

This fix corrects the code which was returning not a Name like "Microsoft Windows 10 Home Edition" but the object containing property Name.

### DIFF
--- a/Source/Menu/Notifications/NotificationManager.cs
+++ b/Source/Menu/Notifications/NotificationManager.cs
@@ -448,11 +448,21 @@ namespace Menu.Notifications
             return url;
         }
 
+        /// <summary>
+        /// Given a function that replaces a parameter with its value, this method replaces all parameters in the Notifications object.
+        /// </summary>
         void ReplaceParameters()
         {
             Notifications.ReplaceParameters(ReplaceParameterValues);
         }
 
+        /// <summary>
+        /// Given a string, replaces any parameters in the form {{parameter}} with their value from the ParameterDictionary.
+        /// If the ParameterDictionary is already loaded with override values, these are used first. Otherwise the program is queried for the value.
+        /// If a parameter is not recognised, it is not replaced.
+        /// </summary>
+        /// <param name="value"></param>
+        /// <returns></returns>
         string ReplaceParameterValues(string value)
         {
             if (value == null) return value;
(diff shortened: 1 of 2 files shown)


Source/Menu/Notifications/NotificationManager.cs
```diff
@@ -507,7 +517,7 @@ string GetParameterValue(string parameter)
                         replacement = Runtime.ToString();
```
rwf-rr: I think this should also be fixed. Runtime is of the same type as OperatingSystem (type Platform). The Name and Version fields are set in SystemInfo.cs.

rwf-rr (commented):
The change looks fine. However, I think I found two additional places where ToString() does not result in what is needed.

Source/Menu/Notifications/NotificationManager.cs
```diff
@@ -507,7 +517,7 @@ string GetParameterValue(string parameter)
...
                         break;
                     case "system":
-                        replacement = SystemInfo.OperatingSystem.ToString();
+                        replacement = SystemInfo.OperatingSystem.Name;
                         break;
                     case "memory":
                         replacement = Direct3DFeatureLevels.ToString();
```
rwf-rr: Similar here. Direct3DFeatureLevels is a LIst. I believe the default implementation of ToSTring() for a List yields the type of the list, not the list's content.

Source/Menu/Notifications/NotificationManager.cs
twpol: It also seems like we're missing a way to check the version with these `Platform` types, doesn't it? :(

twpol (commented):
Seems like some more items have problems, but also I noticed none of them provide versions

Source/Menu/Notifications/NotificationManager.cs
cjakeman: Good catch, Roger. 
I failed to test some items when Notifications were first added.

Source/Menu/Notifications/NotificationManager.cs
cjakeman: Sorted

cjakeman:
> Seems like some more items have problems, but also I noticed none of them provide versions

Tested them all and added versions for system and runtime.
Thanks

rwf-rr (approved):
Additional changes look good.

PR #1244, opened 2026-08-08
Fix Speed Limit Warning Signs in Track Monitor

SteelFill:
This PR fixes an inconsistency I recently noticed; the track monitor would display speed limit warning signs (with the speed text in yellow, instead of white), but only in explorer mode. With some minor changes, the track monitor can also display speed limit warning signs in automatic mode (the normal mode during activities) too, giving a bit more information to help with driving.

- Changed `UpdatePlayerTainData` in AUTO mode to consider warning speed limit signs. Previously, only speed limit signs with `actual_speed > 0` would be considered, but for warning signs the `actual_speed` is set to -1 (thus, warning signs were not considered). Now, the freight/passenger speed limit (depending on the train type) and the train max speed are considered when determining if a speed limit sign is valid, so warning signs will be included in the AUTO mode list.
- Changed speed limit signs in `UpdatePlayerTrainData` in MANUAL/EXPLORER mode to consider the max speed of the train, so the track monitor speed limits shown in this mode should reflect the actual train speed.
- Added support for speed limit warning signs on the web interface track monitor. Warning speeds are displayed in `Yellow` (same behavior as the in-game track monitor).
- Changed the color of the beginning of a restricted speed limit zone on the track monitor from `Red` to `OrangeRed` for consistency between the track monitor window and the web interface track monitor (both use `OrangeRed` now).

### DIFF
--- a/Source/Orts.Simulation/Simulation/Physics/Train.cs
+++ b/Source/Orts.Simulation/Simulation/Physics/Train.cs
@@ -14873,14 +14873,20 @@ namespace Orts.Simulation.Physics
                             PlayerTrainSignals[dir][SignalFunction.NORMAL].Add(thisItem);
                             signalProcessed = true;
                         }
-                        else if (signalObjectItem.ObjectType == ObjectItemInfo.ObjectItemType.Speedlimit && signalObjectItem.actual_speed > 0)
+                        else if (signalObjectItem.ObjectType == ObjectItemInfo.ObjectItemType.Speedlimit)
                         {
-                            thisItem = new TrainObjectItem(thisSpeedMpS: signalObjectItem.actual_speed,
-                                isWarning: signalObjectItem.speed_isWarning,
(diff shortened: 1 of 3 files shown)


rwf-rr:
I am not sure if the track monitor should show the speed warning signs at all. With the speed limit showing in the track monitor, the player has ample warning for the point where the limit will apply.

SteelFill:
> I am not sure if the track monitor should show the speed warning signs at all. With the speed limit showing in the track monitor, the player has ample warning for the point where the limit will apply.

Considering there was already explicit behavior for showing speed limit warnings in yellow, and this worked in manual mode, the design intent does seem to be that speed limit warning signs _should_ be shown. Is it strictly _necessary_ for gameplay? No, but this is meant to be a bugfix and not a serious re-evaluation of how the track monitor should be designed.

PR #1245, opened 2026-08-11
Allow Defining Multiple SMS Files Simultaneously

SteelFill:
This is one I've [seen suggested](https://www.elvastower.com/forums/index.php?/topic/36488-cab-sounds/page__view__findpost__p__288154) a few times
New [Trello Card](https://trello.com/c/s6jzwXvZ) generated.
Discuss in the [diesel sounds thread](https://www.elvastower.com/forums/index.php?/topic/39618-new-sound-capabilities-especially-for-diesel-locomotives/), this is not strictly related to diesels, hence the separate PR, but it comes out of the same desire to do more here and there with the sound system

Historically, every parameter that accepts a .sms file as an input has assumed that only a single .sms file will be given as input. While this is a practical assumption, it does limit how sounds can be implemented in a way that can make things awkward and inefficient. 

The commonly cited example is horn sounds; many locomotives can have the same engine but different horns. The same horn is often used on different engines. An .sms file ends up being required for every combination of engine and horn, leading to a lot of file clutter and a lot of duplicated data (for example, having a selection of 20 horns and 8 engines, _these numbers are not made up_, would require 20 * 8 = **160 .sms files** to account for every combination! And then _double_ that for the in-cab sounds.) It would be much simpler, less error prone, and more file-efficient, to define the horn separately from the engine sound so that the horns and engines can be mixed and matched as defined by the .eng file (this would reduce our 20 horns and 8 engines down to just 20 + 8 = **28 .sms files**, much better!)

This PR adds the ability to simultaneously define multiple .sms files in locomotives/wagons and `Soundsource` blocks of world sounds in .ws files. Simply add more paths for .sms files to a `Sound` or `FileName` parameter, with each sound file separated by a space, like this:
```
    Sound ( "..\\..\\Common.Snd\\Engine_16-645E3.sms" "..\\..\\Common.Snd\\Horn_K3LA.sms" ) [in .eng/.wag files]
    FileName ( "ForestSurroundings.sms" "WideRagingRiver.sms" ) [in .ws files]
```
Under the hood, the additional .sms files specified are added as **additional sound sources** to the sound-emitting object. This is functionally identical to how cab sounds or track sounds work, those are just extra sound sources that are added to the list. Note that the list of sounds is present in a single parameter, they should not be split between multiple `Sound`/`FileName` parameters, as only the final parameter will have any effect (much like how OR handles other parameters).

And as much as I have used horns as an example, this could be used to split .sms files however you like to make files more convenient to manage. However, it is technically more efficient to have all streams defined as a single sound source as each sound source has a bit of computational overhead. It appears to not be common knowledge, but **you can use include files inside .sms files**. Include files can be more complicated to implement and only makes the issue of file count worse but are nicer on the CPU. The choice of approach should be whatever the creator is comfortable with.

Note that versions of OR without this feature will only read the very first .sms listed, so the first .sms should be the most important.

### DIFF
--- a/Source/Documentation/Manual/sound.rst
+++ b/Source/Documentation/Manual/sound.rst
@@ -49,6 +49,33 @@ sound is called, MSTS will play the sound at double speed. In order to have it
 play at the correct speed, a frequency curve halving the speed has to be 
 inserted. OR behaves the same as MSTS in this case.
 
+Additional .sms Files
+---------------------
+
+OR also offers some additional audio features beyond what was possible in MSTS.
+One of the simplest to use is the ability to define more than one .sms file
+in a few locations where MSTS only allowed one. Each extra .sms file will be added
+as a separate sound source and behaves independently of any other .sms files.
+This can be used to make sounds in a modular manner, such as using one .sms file
+for a locomotive's engine sounds and a second .sms file for the horn, allowing
+locomotives with different engine sounds to reuse the same horn sound implementation.
+There is a slight increase in computational load for each additional sound source,
+but minimizing the number of simulatenous active streams is more impactful than this.
+
+Currently, additional .sms files are allowed in train external sounds,
+cab/passenger view sounds, (TODO: more places?).
+
+To use this, simply list off multiple .sms files in a `Sound` parameter, each
+separated by a space::
+
+    Sound ( "..\\..\\Common.Snd\\Engine_16-645E3.sms" "..\\..\\Common.Snd\\Horn_K3LA.sms" )
+
+It should also be noted that :ref:`"include" files <physics-inclusions>` can be
+used inside .sms files to faciliate modular sounds, such as having the horn sound
+stream inside a dedicated include file referenced by multiple sound sets. This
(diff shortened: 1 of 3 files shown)


SteelFill:
Also think you are the most prepared to check this one as well @peternewell. Nothing dramatic changing here, hopefully it's obvious how this works.

peternewell:
> Also think you are the most prepared to check this one as well @peternewell. Nothing dramatic changing here, hopefully it's obvious how this works.

Does this PR have a Trello card or Approved Blueprint linked to it?

SteelFill:
> > Also think you are the most prepared to check this one as well @peternewell. Nothing dramatic changing here, hopefully it's obvious how this works.
> 
> Does this PR have a Trello card or Approved Blueprint linked to it?

This came from the same discussion that led to [card #630](https://trello.com/c/FH5txnc4) but unlike the original feature request, this is obviously not specific to diesel engines. I've added card [#634](https://trello.com/c/s6jzwXvZ).

peternewell (approved):
Thanks for the Trello card.

PR #1247, opened 2026-08-21
Allow cross-compilation on non-Windows systems

cesarBLG:


SteelFill (approved):
Haven't noticed any trouble caused by this, not that it could cause much trouble. Feel free to merge.

PR #1248, opened 2026-08-23
Fix scripts not working in .NET 6

cesarBLG:
C# scripts stopped working due to missing referenced assemblies during compilation

https://www.elvastower.com/forums/index.php?/topic/39858-latest-testing-418-and-latest-unstable-u20260822-1838-dont-load-scripts/page__pid__328049#entry328049


Csantucci (approved):
This worked here, so I do approve

PR #1249, opened 2026-08-25
Fix longer loading times in .NET 6

cesarBLG:
https://www.elvastower.com/forums/index.php?/topic/39857-with-t161-418-loading-times-have-grown-2-10-times/


SteelFill (approved):
Nice simple change, normal loading times have been restored in mine and others' testing.

PR #1250, opened 2026-08-26
Fix Diesel Starting Behavior

SteelFill:
Fix for [this bug](https://www.elvastower.com/forums/index.php?/topic/39866-is-the-diesel-engine-start-function-broken-mg174/page__gopid__328146#entry328146).

The diesel engine demanded RPM should not be reset to 0 if no demanded RPM can be calculated, it should instead remain at whatever the previous demanded RPM was. Resetting the demanded RPM to 0 overrides the demanded RPM that's set once when engine start-up is requested.


SteelFill:
@cesarBLG As is tradition, someone found an issue immediately after merging the previous PR despite the issue being present for over a month. This is a one-line fix that brings #1235 back in line with the intended engine starting behavior.

cesarBLG (approved):

PR #1251, opened 2026-08-26
MSTS legacy shaders

vitro-mod:
## Summary

This PR improves MSTS shape/material compatibility in the 3D viewer, primarily to support older TML Studios add-ons such as U-Bahn and Linie 51. Until now, much of the distinctive tunnel and station lighting in these routes was lost in Open Rails because several MSTS rendering features they relied on were not implemented.

Some of these features, especially `GlossMap`, were also used more broadly in MSTS content and may be useful for future content as well.

- Added support for `uv_op_uniformscale`, `uv_op_nonuniformscale`, and `uv_op_reflectmapfull`.
- Added scenery shader support for `DetailMod2X`, `GlossMap`, `AlphRefMap`, and `nightlight` / `NightLight`.
- Added support for baked per-vertex lighting, vertex colors, and vertex alpha in MSTS shapes. For vertex colors to be used, `lightMatIdx` must be set to `-1`; this is the third parameter of the corresponding `vtx_state` entry in the shape file.
- Added second texture-stage handling for detail/environment maps, including independent UV scale, address mode, and mipmap bias.
- Updated shadow rendering to respect UV scaling.

| Shader | Description |
| --- | --- |
| **DetailMod2X** | Uses two texture slots. The textures are multiplied together and the result is multiplied by two. Works similarly to terrain rendering. |
| **GlossMap** | Uses two texture slots: one for the base texture and one for an additive reflection layer. See: https://web.archive.org/web/20231210194836/https://msts.steam4me.net/tutorials/supershine.html |
| **AlphRefMap** | Similar to `GlossMap`, but with alpha support. Discovered in TML's `app.fbk`. |
| **nightlight**, **NightLight** | Uses additive blending modulated by inverse daylight, making the material visible primarily at night. TML used it for glowing windows. Discovered in their `.fbk` files. TML uses the lowercase `nightlight` spelling; support for both that spelling and the conventional PascalCase `NightLight` variant was added. |

## Result

Older add-ons relying on MSTS-specific UV operations, detail/gloss/reflection shaders, night lighting, and baked vertex colors should now render much closer to their original MSTS appearance. These capabilities also become available for newly created content.

### DIFF
--- a/Source/Orts.Formats.Msts/AceFile.cs
+++ b/Source/Orts.Formats.Msts/AceFile.cs
@@ -159,21 +159,24 @@ namespace Orts.Formats.Msts
                         scanlineOffsets[imageIndex][y] = reader.ReadInt32();
                 }
 
-                var scanlineDataEnd = reader.BaseStream.Position;
-                for (var imageIndex = 0; imageIndex < imageCount; imageIndex++)
+                if (reader.BaseStream is BufferedInMemoryStream)
                 {
-                    var imageWidth = width / (int)Math.Pow(2, imageIndex);
-                    var imageHeight = height / (int)Math.Pow(2, imageIndex);
-                    var scanlineLength = channels.Sum(channel => (channel.Size * imageWidth + 7) / 8);
-                    for (var y = 0; y < imageHeight; y++)
-                        scanlineDataEnd = Math.Max(scanlineDataEnd, offsetBase + scanlineOffsets[imageIndex][y] + scanlineLength);
-                }
+                    var scanlineDataEnd = reader.BaseStream.Position;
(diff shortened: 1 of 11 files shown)


pzgulyas (changes requested):
Hi Victor,

This PR will not be able to merge, because the 3D shaders and related cs files got extensively modified in #570 and this PR conflicts with that. That PR is not yet merged, that's why you did not notice it when targeted the current 'master'. Instead you should rework this PR to target #570. This PR will still not be able to merge afterwards because of the AppVeyor though, so for this to finally appear in the 'unstable' it must wait until #570 gets merged. Sorry for the inconvenience,but the review process takes time for that one.

PR #1252, opened 2026-08-26
Fix for exception on "What is new" in main menu Notifications.

rwf-rr:
With .NET 6, _UseShellExecute_ needs to be explicitly set to true.

See https://www.elvastower.com/forums/index.php?/topic/39862-unhandled-exception-when-clicking-what-was-new-button/.

Update 2026-09-03:
* Chris found a few more locations where the same fix (using  UseShellExecute = true) is required.
* Reviewed all Process.Start() calls. Changed those where the reference is a URL or a non-exe type file (eg: pdf, log, etc).
* In addition, discovered a few **URL that were no longer correct**. Fixed these too.


cjakeman (commented):
Hi Roger,

Thanks for finding and fixing this.

I did a quick search of the solution and found 3 more occurrences of ProcessStartInfo(), all in the Menu project. Would you mind checking these as well?

Thanks,

Chris

rwf-rr:
> I did a quick search of the solution and found 3 more occurrences of ProcessStartInfo(), all in the Menu project. Would you mind checking these as well?

Thanks, Chris. I found a few more outside of the main Menu. I believe we got all of them now.

cjakeman (approved):
All looks good.

Approved and merged with thanks.

PR #1259, opened 2026-09-09
Fix exception when opening Car Ops (F9) for the first time while Free Cam (8) is active.

rwf-rr:
This fixes a NullReferenceException when opening the Train Car Operations window (F9) the first time while the Free Roam Camera (8) is active. 

The bug does not exist for the Web page version.

The cause is that on the first open (with free roam camera), `Owner.Viewer.Camera.AttachedCar` is null.

PS: The code in the affected block would benefit from some simplifications. But that is deferred to later.


mbm-OR:
Thank you for pointing it out.
It looks fine to me. 
Let's wait a week for comments from the users.

Regards,
Mauricio

PR #1262, opened 2026-09-17
Fix bugs in data logger: steam interval, steam data.

rwf-rr:
Data for Steam locomotives is logged at speed intervals. However, new rows were always written at time intervals, resulting in empty rows for steam loco data.

Some of the steam locomotive data was logged from the loco object. However, the data was never copied from the SteamEngine object (where it is maintained).

SteamPerformanceTimeS was not updated for steam locomotives using the simple adhesion model.

A few points to note:

- The Steam Performance data is logged at 5 mph intervals (and only when increasing). In contrast, the Steam Power Curve is logged at 1 mph intervals (but the comment still said 5 mph). Does that make sense?
- The previous log speed is initialized at -1 mph. This makes sense for the Steam Power Curve, but the Steam Performance will be logged at 4, 9, 14, ... mph. Does that make sense?
- The tracked speed is truncated to an integer. What could be the reason for that?
- With the upcoming change to configurable units, should the speed interval be based on the configured unit?

I am not familiar with steam locomotives, so I don't have the background to judge some of the code decisions. I also don't have rolling stock to test more complex steam locomotives (eg. compound cylinders) (or geared diesel locomotives). I tested with MSTS routes and the Scenic Sub starter route.


peternewell:
The main reason for these features was to allow a "direct" comparison of performance data with test report data. See page 46 of this test report - http://users.fini.net/~bersano/english-anglais/BR-tests/BR_P&E_No15_BR_Std8.pdf  . It relied on the locomotive being able to run without interruption so that the results could be compared against the test report. I hope that this is of some help.

You can look at some of my CTN rolling stock for testing provided ORTS features have not been changed since they were built.

